PR #1, opened 2017-02-14
Build on Appveyor

jobevers:


lbrynet/lbry.onefile.spec
```diff
@@ -3,6 +3,7 @@ import platform
 import os
 
 import lbryum
+import requests
```
jobevers: can remove this, pyinstaller does properly bundle the cacert on its own

lbrynet/lbry.onefile.spec
```diff
@@ -12,26 +13,40 @@ repo_base = os.path.abspath(os.path.join(cwd, '..'))
...
+        os.path.join(os.path.dirname(lbryum.__file__), 'wordlist', language),
+        'lbryum/wordlist'
+    )
+    for language in languages
+]
+# + [
+#    (requests.certs.where(), '')
+#]
+
+print datas
```
jobevers: can remote this too

lbrynet/lbry.py
```diff
@@ -1,4 +1,8 @@
 from lbrynet.lbrynet_daemon import DaemonControl
 
+import os
+print 'Bundle:', os.environ.get('REQUESTS_CA_BUNDLE')
+print
+
```
jobevers: these lines should go

lbrynet/lbry.spec
```diff
@@ -1,61 +1,28 @@
...
+exe = EXE(pyz,
+          a.scripts,
+          a.binaries,
+          a.zipfiles,
+          a.datas,
+          name='lbry',
+          debug=False,
+          strip=False,
+          upx=True,
+          console=False , icon='..\\build\\icon.ico')
```
jobevers: this whole file can go

package.json
```diff
@@ -34,6 +34,9 @@
     },
     "linux": {
       "target": "deb"
+    },
+    "win": {
+      "target": "zip"
```
jobevers: nsis was giving an error for me, unfortunately - should dig into that soonish, but zip works so lets get it out.

PR #5, opened 2017-02-21
Add release script

jobevers:


jobevers:
Lets fix this on master.

PR #7, opened 2017-03-02
add required package

kauffj:
pyinstaller is required 

(also instructions should indicate you need a virtual environment)


lyoshenka:
pyinstaller shouldn't be in requirements, because its only a requirement of the build process, not of lbry. it does get installed by build.sh, but only when its running on teamcity (for some reason). im looking at this stuff right now, will rewrite. also, no idea what `posix.txt` is

PR #9, opened 2017-03-08
Add basic menubar

alexliebowitz:
This includes a Developer menu that can be enabled or disabled through the UI.

This Web UI PR is needed to enable and disable the menu: https://github.com/lbryio/lbry-web-ui/pull/193

### DIFF
--- a/app/menu/main-menu.js
+++ b/app/menu/main-menu.js
@@ -2,75 +2,106 @@ const {Menu} = require('electron');
 const electron = require('electron');
 const app = electron.app;
 
-const template = [
+const baseTemplate = [
   {
     label: 'Edit',
     submenu: [
       {
-        role: 'undo'
+        role: 'undo',
       },
       {
-        role: 'redo'
+        role: 'redo',
       },
       {
-        type: 'separator'
+        type: 'separator',
       },
       {
-        role: 'cut'
+        role: 'cut',
       },
       {
-        role: 'copy'
+        role: 'copy',
       },
       {
-        role: 'paste'
+        role: 'paste',
       },
-      {
-        role: 'delete'
-      },
-      {
-        role: 'selectall'
-      }
     ]
   },
   {
     label: 'Help',
     submenu: [
       {
-        label: 'FAQ',
-        click () { require('electron').shell.openExternal('https://lbry.io/faq') }
+        label: 'Help',
+        click(item, focusedWindow) {
+          if (focusedWindow) {
+            focusedWindow.loadURL(`file://${__dirname}/../dist/index.html?help`);
+          }
+        }
       }
     ]
   }
 ];
 
-module.exports = {
-  showNormalMenubar: () => {
-    const menu = Menu.buildFromTemplate(template);
-    Menu.setApplicationMenu(menu);
-  },
-  showDeveloperMenubar: () => {
-    const devTemplate = template.slice();
-    devTemplate.push({
-      label: 'Developer',
-      submenu: [
-        {
-          label: 'Reload',
-          accelerator: 'CmdOrCtrl+R',
-          click (item, focusedWindow) {
-            if (focusedWindow) focusedWindow.reload()
-          }
-        },
-        {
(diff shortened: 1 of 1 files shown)


app/menu/main-menu.js
```diff
@@ -0,0 +1,76 @@
+const {Menu} = require('electron');
+const electron = require('electron');
+const app = electron.app;
+
+const template = [
+  {
+    label: 'Edit',
+    submenu: [
+      {
+        role: 'undo'
```
kauffj: What do "undo" and "redo" do in context of this app?

app/menu/main-menu.js
```diff
@@ -0,0 +1,76 @@
...
+        type: 'separator'
+      },
+      {
+        role: 'cut'
+      },
+      {
+        role: 'copy'
+      },
+      {
+        role: 'paste'
```
kauffj: I also have a "paste and match style" option, which I don't see listed here? Is there a way to just add paste without "paste and match style"?

app/menu/main-menu.js
```diff
@@ -0,0 +1,76 @@
...
+      },
+      {
+        role: 'delete'
```
kauffj: What does this do?

app/menu/main-menu.js
```diff
@@ -0,0 +1,76 @@
...
+      },
+      {
+        role: 'selectall'
```
kauffj: Same as above questions... why do we need/want this?

app/menu/main-menu.js
```diff
@@ -0,0 +1,76 @@
...
+      }
+    ]
+  },
+  {
+    label: 'Help',
+    submenu: [
+      {
+        label: 'FAQ',
+        click () { require('electron').shell.openExternal('https://lbry.io/faq') }
```
kauffj: We have a help option in the UI. If we add help to the toolbar, it should go to the same place. If we want to encourage FAQ usage, the main "Help" section should do that.

app/menu/main-menu.js
alexliebowitz: Undo and redo typing, just like in a browser. They're not going to be heavily used, but I feel like people expect that to be in any app with an Edit menu.

app/menu/main-menu.js
alexliebowitz: I did remove the Paste and Match Style option, so it sounds like you're getting the default Electron menu (the one that appears in developer mode). It looks similar because I based this on the example from the Electron docs which shows you how to make the same menu as the default developer menu. I'll take a look at the office today.

app/menu/main-menu.js
alexliebowitz: I feel like these options are supposed to be there for text editing in every app that has text boxes.

I checked the apple HIG [here](https://developer.apple.com/library/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/MenuBarMenus.html#//apple_ref/doc/uid/20000957-CH29-SW7) and it confirms:

> Even if your app doesn’t handle text editing within its documents, the expected Edit commands should be available for use in dialogs and wherever users can edit text.

I haven't checked the Windows or GNOME/KDE guidelines, though. Might be different there.

app/menu/main-menu.js
alexliebowitz: True. I wasn't giving this part a lot of thought because I knew it would be at least today before it was merged. Just trying to get the general framework of the menu up for review. We should probably decide today on the full set of menu items and what they should do.

app/menu/main-menu.js
kauffj: Chrome does not have undo and redo in it's menu bar. I do not think that we should either.

app/menu/main-menu.js
kauffj: Confirmed, I no longer see this.

app/menu/main-menu.js
kauffj: This should be removed probably.

PR #14, opened 2017-03-15
Make publishes appear immediately in My Files

alexliebowitz:
Uses a client side cache to simulate entries in the file manager and  claims list.

Also adds new utility functions for using Local Storage.

### DIFF
--- a/ui/js/utils.js
+++ b/ui/js/utils.js
@@ -1,0 +1,15 @@
+/**
+ * Thin wrapper around localStorage.getItem(). Parses JSON and returns undefined if the value
+ * is not set yet.
+ */
+export function getLocal(key) {
+  const itemRaw = localStorage.getItem(key);
+  return itemRaw === null ? undefined : JSON.parse(itemRaw);
+}
+
+/**
+ * Thin wrapper around localStorage.setItem(). Converts value to JSON.
+ */
+export function setLocal(key, value) {
+  localStorage.setItem(key, JSON.stringify(value));
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
   * You can now make API calls directly on the lbry module, e.g. lbry.peer_list()
   * New-style API calls return promises instead of using callbacks
   * Wherever possible, use outpoints for unique IDs instead of names or SD hashes
+  * New publishes now display immediately in My Files, even before they hit the lbrynet file manager.
 
 ### Changed
   * Update process now easier and more reliable
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Web UI version numbers should always match the corresponding version of LBRY App
 
 ## [Unreleased]
 ### Added
+  * The app is much more responsive switching pages. It no longer reloads the entire page and all assets on each page change.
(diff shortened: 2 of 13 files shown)


ui/js/utils.js
```diff
@@ -0,0 +1,100 @@
+/**
```
kauffj: Should all these functions actually be in the global namespace?

ui/js/utils.js
```diff
@@ -0,0 +1,100 @@
...
+ */
+export function setLocal(key, value) {
+  localStorage.setItem(key, JSON.stringify(value));
+}
+
+/**
+ * Records a publish attempt in local storage. Returns a dictionary with all the data needed to
+ * needed to make a dummy claim or file info object.
+ */
+export function savePendingPublish(name) {
```
alexliebowitz: Possibly break these out into their own module?

kauffj: These functions should probably be moved/kept with the LBRY API logic?

ui/js/utils.js
```diff
@@ -0,0 +1,100 @@
...
+  const pendingPublishes = getLocal('pendingPublishes') || [];
```
alexliebowitz: I don't love "pending publish" as the name for this data structure, kinda wordy. But what else to call it? Publish stub? Dummy publish?

ui/js/utils.js
alexliebowitz: Before, we were talking about aiming to make the lbry API only have API wrapper functions, not random utility functions and such. We could still put helper functions like this as private functions that are not exported, though.

The main reason I didn't do that is that I felt like lbry.js is already way too big. But it's going to shrink over time, so I guess it's OK.

ui/js/utils.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+  setLocal('pendingPublishes', newPendingPublishes);
+  return newPendingPublishes
+}
+
+/**
+ * Gets a pending publish attempt by its name or (fake) outpoint. If none is found (or one is found
+ * but it has timed out), returns null.
+ */
+export function getPendingPublish({name, outpoint}) {
```
alexliebowitz: I know it's awkward to use Local Storage for queryable, list-like data like this. The main alternatives would be IndexedDB (too complicated without a wrapper), Sqlite (possibly too complicated and would require doing stuff in the main process and IPC), or a light JS key/value database like nedb (adds a dependency and stuff to learn).

If we don't like local storage, it would be fairly quick to switch to something like nedb.

ui/js/utils.js
alexliebowitz: In ES6 it defaults to module namespace. That might be an issue if we don't like the calling style (`getLocal()` vs `utils.getLocal()`) but it's not going to pollute scope anywhere we don't want it to.

ui/js/utils.js
kauffj: The name is fine, probably more important to make it some kind of const if referenced in multiple places.

ui/js/utils.js
kauffj: I think lbry is the right place for this. They change the behavior of API calls provided by lbry.js.

ui/js/utils.js
kauffj: Local storage is fine.

kauffj:
@alexliebowitz you can merge this when ready

kauffj:
@alexliebowitz let's get this out!

PR #16, opened 2017-03-17
Real update system

alexliebowitz:
- Now asks the daemon to close, even if it wasn't started by the same app
 - Improved UX during upgrade process (cancel buttons, final dialog where you approve the update, etc.)
 - Saves updates in temp directory, closes app and launches the installer

Todo (none of these are blocking merge):
 - Force kill the existing daemon if it doesn't stop
 - Improve visuals
 - Don't upgrade to RCs

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -115,16 +115,19 @@ app.on('activate', () => {
 })
 
 
-function shutdownDaemon() {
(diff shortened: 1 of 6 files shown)


lyoshenka:
Don't worry about not upgrading to RCs. That's on me to mark them as prereleases

PR #20, opened 2017-03-22
Upgrade improvements and refactor shutdown process

alexliebowitz:
- Manually call `xdg-open` instead of using `shell.openItem()`, which doesn't reliably work from the main process
 - If there's a connection error or timeout when asking the daemon to close, fall back on trying to force kill any copies of the daemon that are running

Outstanding issues before ready to merge:
 - Finish testing
 - Try on Windows (all of the known issues have to do with the logic of killing the daemon and restarting, so there shouldn't be anything huge that comes up here)
      - Will need to have Jack help get a dev install going on a Windows VM. Haven't done that since we switched to Electron
 - `xdg-open` doesn't work reliably if you're not the same user as the package manager. `shell.openItem()` would have the same problem as it ultimately just calls xdg-open. We should probably at least detect this scenario and do something different, e.g. reveal the file instead of open it, and display a notice that you'll have to click the file to start the upgrade.
    - This is an independent thing and isn't blocking merge.

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -15,8 +15,8 @@ let win;
 // Also keep the daemon subprocess alive
 let daemonSubprocess;
 
-// This is set to true right before we try to kill the daemon subprocess --
-// if it dies when we haven't made a request, we want to alert the user.
+// This is set to true right before we try to shut the daemon subprocess --
+// if it dies when we didn't ask it to shut down, we want to alert the user.
 let daemonSubprocessKillRequested = false;
 
 // When a quit is attempted, we cancel the quit, do some preparations, then
@@ -61,15 +61,18 @@ function handleDaemonSubprocessExited() {
   console.log('The daemon has exited.');
   daemonSubprocess = null;
   if (!daemonSubprocessKillRequested) {
-    // We didn't stop down the daemon subprocess on purpose, so display a
+    // We didn't stop the daemon subprocess on purpose, so display a
     // warning and schedule a quit.
     //
     // TODO: maybe it would be better to restart the daemon?
-    console.log('Did not display, so scheduling quit');
     if (win) {
+      console.log('Did not request daemon stop, so quitting in 5 seconds.');
       win.loadURL(`file://${__dirname}/dist/warning.html`);
+      setTimeout(quitNow, 5000);
+    } else {
+      console.log('Did not request daemon stop, so quitting.');
+      quitNow();
     }
-    setTimeout(quitNow, 5000);
   }
 }
 
@@ -82,7 +85,7 @@ function launchDaemon() {
   } else {
     executable = path.join(__dirname, 'dist', 'lbrynet-daemon');
   }
-  console.log('Launching daemon: ' + executable)
+  console.log('Launching daemon:', executable)
   daemonSubprocess = child_process.spawn(executable)
   // Need to handle the data event instead of attaching to
   // process.stdout because the latter doesn't work. I believe on
@@ -94,8 +97,9 @@ function launchDaemon() {
 }
 
 /*
- * Quits without any preparation (when a quit is requested, we abort the quit, try to shut down
- * the daemon, and then call this to quit for real).
+ * Quits without any preparation. When a quit is requested (either through the
+ * interface or through app.quit()), we abort the quit, try to shut down the daemon,
+ * and then call this to quit for real.
  */
 function quitNow() {
   readyToQuit = true;
@@ -169,7 +173,7 @@ app.on('window-all-closed', () => {
 
 app.on('before-quit', (event) => {
   if (!readyToQuit) {
-    // We need to shutdown the daemons before we're ready to actually quit. This
+    // We need to shutdown the daemon before we're ready to actually quit. This
     // event will be triggered re-entrantly once preparation is done.
     event.preventDefault();
     shutdownDaemonAndQuit();
@@ -187,9 +191,9 @@ app.on('activate', () => {
   }
 });
 
-// When a quit is attempted, this is called, it attempts to shutdown the daemon,
-// and then calls app.quit() to quit for real.
-function shutdownDaemonAndQuit(shutdownEvenIfNotStartedByApp = false) {
+// When a quit is attempted, this is called. It attempts to shutdown the daemon,
+// then calls quitNow() to quit for real.
+function shutdownDaemonAndQuit(evenIfNotStartedByApp = false) {
   if (daemonSubprocess) {
     console.log('Killing lbrynet-daemon process');
     kill(daemonSubprocess.pid, undefined, (err) => {
@@ -205,7 +209,7 @@ function shutdownDaemonAndQuit(shutdownEvenIfNotStartedByApp = false) {
         // or because it's running but not responding properly (bad).
         // So try to force kill any daemons that are still running.
 
-        console.log('received error when stopping lbrynet-daemon. Error message: {err.message}');
+        console.log(`received error when stopping lbrynet-daemon. Error message: ${err.message}`);
         forceKillAllDaemonsAndQuit();
       } else {
         console.log('Successfully stopped daemon via RPC call.')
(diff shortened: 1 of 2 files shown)


app/main.js
```diff
@@ -122,10 +148,17 @@ function shutdownDaemon(evenIfNotStartedByApp = false) {
...
     });
   } else if (evenIfNotStartedByApp) {
-    console.log('Killing lbrynet-daemon, even though app did not start it');
-    client.request('daemon_stop', []);
-    // TODO: If the daemon errors or times out when we make this request, find
-    // the process and force quit it.
+    console.log('Stopping lbrynet-daemon, even though app did not start it');
+    client.request('daemon_stop', [], (err, res) => {
+      if (err) {
+        // We could get an error because the daemon is already stopped (good)
```
lyoshenka: wait, so this will to force-kill all daemons no matter what, even if the appropriate daemon was stopped successfully?

app/main.js
```diff
@@ -144,7 +177,10 @@ function shutdown() {
 
 function upgrade(event, installerPath) {
   app.on('quit', () => {
-    shell.openItem(installerPath);
+    // shell.openItem doesn't reliably work from the app process, so run xdg-open directly
+    child_process.spawn('xdg-open', [installerPath], {
```
lyoshenka: does this work on osx and windows? or is this only called on linux?

lyoshenka: also, i find it hard to believe that we cant tell whether this worked or failed

alexliebowitz: It's called everywhere, but yesterday I factored this out and made it choose the command based on OS. That'll go up with my next set of changes.

Re: checking if it worked or failed, we can detect _some_ failure conditions (where the installer or package manager completely fails to launch, arguably the least likely scenario) by seeing if the process returns with an error status code. But we can't detect anything that happens after that, like getting that weird dialog box because of a user mismatch. But you're right, I'll add a check for an error code.

app/main.js
alexliebowitz: The problem is that we don't know which one is the appropriate one, unless we launched it, which we didn't necessarily.

We talked about finding the process listening on 5279 instead, but we don't want to risk killing someone else's process.

Hmm... what if we combined both: we keep this code to get a list of pids for all running lbrynet-daemons, then find the process listening on 5279, confirm it's in the list of daemon pids, then kill just that one? That should handle every scenario except the case where someone made their own app that happens be called "lbrynet-daemon" and also runs on 5279, in which case it's kind of their problem :P

lyoshenka: im still confused. we have `client`, which is the connection to the daemon that we use for everything. we call `daemon_stop` on it. your comment says `we could get an error because the daemon is already stopped`. if the daemon is already stopped, then we're good, right? why do we then try to kill all the running daemons?

app/main.js
lyoshenka: gotcha

app/main.js
alexliebowitz: Sorry, should have written "frozen," not "stopped." The daemon process could be open but still it's refusing connections, or timing out, or giving 500 errors. Even if the error is "connection refused," it could be frozen but still have the port bound.

app/main.js
```diff
@@ -86,6 +125,37 @@ function launchDaemonIfNotRunning() {
...
+ * tries to force kill them.
+ */
+function forceKillAllDaemonsAndQuit() {
+  console.log('Attempting to force kill any running lbrynet-daemon instances...');
+
+  const fgrepOut = child_process.spawnSync('pgrep', ['-x', 'lbrynet-daemon'], {encoding: 'utf8'}).stdout;
+  const daemonPids = fgrepOut.match(/\d+/g);
+  if (!daemonPids) {
+    console.log('No lbrynet-daemon found running.');
+    quitNow();
```
alexliebowitz: We could probably use Promises to simplify the shutdown process so we're not calling quitNow in so many places, but for now my goal is to make the logic extremely explicit and easy to reason about.

app/main.js
lyoshenka: You sure about that last part? I'm pretty sure that if the daemon crashes, the app will close. See https://github.com/lbryio/lbry-app/blob/master/app/main.js#L46

app/main.js
```diff
@@ -26,49 +57,57 @@ function createWindow () {
...
 
+function handleDaemonSubprocessExited() {
+  console.log('The daemon has exited.');
+  daemonSubprocess = null;
+  if (!daemonSubprocessKillRequested) {
+    // We didn't stop down the daemon subprocess on purpose, so display a
+    // warning and schedule a quit.
+    //
+    // TODO: maybe it would be better to restart the daemon?
+    console.log('Did not display, so scheduling quit');
```
lyoshenka: What does this mean? `Did not display`?

app/main.js
```diff
@@ -86,6 +125,37 @@ function launchDaemonIfNotRunning() {
...
 
+/*
+ * Last resort for killing unresponsive daemon instances.
+ * Looks for any processes called "lbrynet-daemon" and
+ * tries to force kill them.
+ */
+function forceKillAllDaemonsAndQuit() {
+  console.log('Attempting to force kill any running lbrynet-daemon instances...');
+
+  const fgrepOut = child_process.spawnSync('pgrep', ['-x', 'lbrynet-daemon'], {encoding: 'utf8'}).stdout;
```
lyoshenka: is this cross-platform?

app/main.js
```diff
@@ -112,52 +185,62 @@ app.on('activate', () => {
...
-function shutdown() {
-  if (win) {
-    win.loadURL(`file://${__dirname}/dist/quit.html`);
-  }
-  quitting = true;
-  shutdownDaemon();
-}
-
 function upgrade(event, installerPath) {
+  console.log('top of upgrade()')
```
lyoshenka: reminder to remove this before merging

app/main.js
```diff
@@ -112,52 +185,62 @@ app.on('activate', () => {
...
   // what to do if no shutdown in a long time?
-  console.log('Update downloaded to ', installerPath);
+  console.log('Update downloaded to', installerPath);
   console.log('The app will close, and you will be prompted to install the latest version of LBRY.');
   console.log('After the install is complete, please reopen the app.');
 }
 
 ipcMain.on('upgrade', upgrade);
-
-ipcMain.on('shutdown', shutdown);
```
lyoshenka: we dont need this anymore?

ui/js/app.js
```diff
@@ -42,14 +42,26 @@ var App = React.createClass({
...
 
-  // Temporary workaround since electron-dl throws errors when you try to get the filename
+  getUpdateUrl: function() {
+    switch (process.platform) {
+      case 'darwin':
+        return 'https://lbry.io/get/lbry.dmg';
+      case 'linux':
+        return 'https://lbry.io/get/lbry.deb';
+      case 'win32':
+        return 'https://lbry.io/get/lbry.exe'; // should now be msi
```
lyoshenka: should now be msi?

app/main.js
alexliebowitz: Good point. Scratch that part. (Not sure if that actually means anything needs to change with the logic here, though.)

app/main.js
alexliebowitz: No idea where that came from. Maybe just a copy-and-paste error. I meant for it to say something like "did not intentionally quit daemon, so scheduling quit."

ui/js/app.js
alexliebowitz: I was going on Job's old comment about how the installer is now an MSI (deleted in this PR). But I guess that was written before we switched to Electron, and we're back to an exe.

app/main.js
alexliebowitz: It'll work on Mac and Linux. Still thinking about Windows.

app/main.js
alexliebowitz: Nope. The UI used to request the quit after an upgrade, but it was moved into upgrade() here in the main process.

lyoshenka:
merged

PR #24, opened 2017-03-26
More improvements to shutdown process

alexliebowitz:
- Gets Windows support working
 - Logic fixes

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -159,13 +159,15 @@ function forceKillAllDaemonsAndQuit() {
     console.log(`Found ${daemonPids.length} running daemon instances. Attempting to force kill...`);
 
     for (const pid of daemonPids) {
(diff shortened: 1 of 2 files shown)


alexliebowitz:
At this point all the deep issues are dealt with (getting it working with the sys calls for all 3 platforms, having a way to kill old/frozen daemon instances, etc.) and it's in a place where we can fiddle with the shutdown sequence however we want.

For example, we could do the trick we discussed earlier--i.e., when the daemon was started externally, instead of killing everything called "lbrynet-daemon," it only kills the one that opened the port. Or, if there were some way to detect which of the running daemons are frozen or left over from other versions, we could do that on _every_ upgrade (or even every shutdown!)

lyoshenka:
merged

PR #25, opened 2017-03-26
make page changes fast

kauffj:
This started as simply wanting to catch click events and re-draw the app inside of app.js without a full reload of index.html and all assets. However, it turned out to require adding a subscription feature for balances and fixing a few other areas where we were not unsubscribing/unmounting properly.


kauffj:
@alexliebowitz Make sure you see this. Also make sure to omit index.html in links moving forward. Just start with `?`.

alexliebowitz:
LGTM

PR #53, opened 2017-04-24
Redux

6ea86b96:
Very much a WIP still. Took a bit longer than expected to get some stuff working. However, here is my progress so far from the weekend. Hoping can get everything into Redux by the end of next weekend.

There are already some noticeable improvements. Discover page and show page now reuse the same claim data and load pretty much instantly.


kauffj:
Awesome progress.

Whenever you think this is can go live without BC breaks, I'm fine to put it in even if we're not 100% doing things the Redux way. App development is hot and broad right now, so the more we can keep branches in check the fewer head aches down the road.

6ea86b96:
Will have this in a state where it can be merged by tomorrow morning @kauffj. Need to get published files and rewards working. Merging with master isn't going to be easy though :/

kauffj:
@6ea86b96 we'll own the merge. Can't wait!

6ea86b96:
Ok, so I think this is good now @kauffj 

Bugs:
- published files are showing up in downloaded files
- reward for purchasing won't get claimed if the user refreshes the client before the download finished
- Data doesn't get refreshed right now after initial fetch. Need to dispatch actions to fetch latest data in some places. Then perhaps when data is first being fetched when can display a loading message and on subsequent fetches we display the data along with a spinner saying it's updating.
- Data not refreshing means content priced in dollars is going to get out of date
- some reward callbacks might be missing
- the URI we build to store fileinfo is different to the uri returned from featured content (which has a claim sequence number), so we start to see duplicates as we view files.
- probably lots of other things :)

Todo:
- refactor redux state to store files by `sd_hash`. Now I understand the data model better I think storing everything by URI is not right.
- get rid of the content actions/reducer/selector
- I think we can completely remove all the `lbry` file info subscribe events and rely on components being updated by changes to the global state.
- refactor rewards out of components and into actions.
- lots of loading/fetching state improvements. For example we know that the claim and or cost info is being fetched when loading featured content file tiles so we could display some nice loading message instead of "this file is pending confirmation".

Other things I came across:

- `lbry.sendToAddress` seems to return true even when an address is invalid
- would be nice if `lbry.sendToAddress` returned a transaction
- `lbry.removeFile(null)` removes all every single downloaded file ever. Bug or feature? :)
- Removing a paid file (including from fs) means you need to purchase it again to watch it

kauffj:
@6ea86b96 merged but I think I made at least a few mistakes. Coming back to this a bit later today or tomorrow depending on how things go.

6ea86b96:
Woah, that was insanely fast. Expected this to take much longer :) I will get a coffee now and take a look.

PR #91, opened 2017-04-27
Move version checking into Electron

alexliebowitz:


package.json
```diff
@@ -42,5 +42,8 @@
   "devDependencies": {
     "electron": "^1.4.15",
     "electron-builder": "^11.7.0"
+  },
+  "dependencies": {
+    "semver": "^5.3.0"
```
lyoshenka: is this the correct package.json to be putting this in? i thought app requirements go into app/package.json?

lyoshenka: is this the correct package.json to be putting this in? i thought app requirements go into app/package.json?

lyoshenka:
Another question: how would I test this? Is it possible to write tests for this?

kauffj:
@alexliebowitz please merge yourself after you review what @lyoshenka said

package.json
alexliebowitz: Yup, just did npm install in the wrong directory. Fixed.

alexliebowitz:
Re: testing, this is well suited to unit tests, but we don't have a framework on the Electron side yet. We ought to get something like Mocha or Spectron going.

Meanwhile, a quick and dirty (but still highly accurate) way to test the upgrade behavior when a new version is available would be to flip the local and remote versions on [this line](https://github.com/lbryio/lbry-app/pull/91/files#diff-3c56f1dfc265d93d754c52ea6397c36aR58). That's how I did it.

PR #92, opened 2017-04-27
Better effects for search input

Plasmmer:


kauffj:
Really appreciate this and it's a good idea, but this is about to be dropped for a new bar entirely. So I'm going to reject this.

Also, we try very hard to avoid `!important`, and it should really never be necessary.

PR #98, opened 2017-04-27
prefer wallet_unused_address over wallet_new_address

lyoshenka:
did i do this right?

### DIFF
--- a/ui/js/page/wallet.js
+++ b/ui/js/page/wallet.js
@@ -16,7 +16,7 @@ var AddressSection = React.createClass({
       this.setState({
         address: address,
       });
-    });
+    }.bind(this));
   },
 
   _getNewAddress: function(event) {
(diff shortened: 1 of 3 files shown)


ui/js/page/wallet.js
```diff
@@ -16,20 +16,20 @@ var AddressSection = React.createClass({
       this.setState({
         address: address,
       });
-    });
+    }.bind(this));
```
lyoshenka: why did the one below have `bind(this)` but this one didnt?

kauffj: Because it doesn't use an arrow function and thus loses the scope of `this` otherwise. Wherever you see `.bind(this)` it can probably be re-written as an arrow function.

lyoshenka: I did not realized that arrow functions work differently from normal functions. I assumed that was just the new way of writing them.

PR #99, opened 2017-04-27
.pkg format for macos (bounty, 20000 LBC)

ChristianBeach:
Changes the electron-build script to automatically build to .pkg if being compiled on a mac.
Introduces a script for creating a .pkg if on linux.


build/PackageInfo.xml
```diff
@@ -0,0 +1,6 @@
+<pkg-info format-version="2" identifier="io.lbry.LBRY.pkg" version="0.10.0" install-location="/" auth="root">
```
jackrobison: what does `auth="root"` do?

build/PackageInfo.xml
```diff
@@ -0,0 +1,6 @@
...
+  <payload installKBytes="173006" numberOfFiles="3608"/>
```
jackrobison: Can these be moved to env variables?

build/PackageInfo.xml
```diff
@@ -0,0 +1,6 @@
...
+<bundle-version>
+    <bundle id="io.lbry.LBRY" CFBundleIdentifier="io.lbry.LBRY" path="./Applications/LBRY 0.10.0.app" CFBundleVersion="0.10.0"/>
```
jackrobison: The release script will have to be updated to bump the version here

build/PackageInfo.xml
ChristianBeach: Defines if the user needs to authenticate as an admin user to install the package

build/PackageInfo.xml
ChristianBeach: I don't think so but I'm sure I can figure something out with sed.

lyoshenka: how did you calculate the number of files and the installation size? we can't have this hardcoded, since it will need to change anytime we make code changes

build/PackageInfo.xml
lyoshenka: is it possible to install this without asking for root?

package.json
```diff
@@ -8,29 +8,7 @@
     "appId": "io.lbry.LBRY",
     "mac": {
       "category": "public.app-category.utilities",
-      "target": "dmg"
-    },
-    "dmg": {
-      "iconSize": 128,
```
lyoshenka: is the icon still included in this build? or do .pkg builds not have an icon?

build/PackageInfo.xml
ChristianBeach: I believe the /Applications/ directory always requires root access, so no.

build/PackageInfo.xml
ChristianBeach: find root | wc -l < if executed in the .pkg directory will show the number of lines and

du -b -s root < will give the number of bytes (needs to be converted to kbytes)

ChristianBeach:
Dependencies:

- [https://github.com/hogliux/bomutils](bomutils): Creates apples's BOM file (bill of materials) which gets turned into a receipt that you can use to uninstall the app. This file contains a manifest of everything the package will install and where to

- [http://xmlstar.sourceforge.net/](xmlstarlet): Used to modify .xml files. When I was trying to use sed, it would break .xml files for some reason.

- [https://github.com/mackyle/xar](eXtensible ARchiver): .pkg files are actually .xar archives, this utility is needed in the final step to create the .pkg

lyoshenka:
I know I said I'd look at this today, but I didn't get to it. Sorry. It's the first thing on my list tomorrow.

ChristianBeach:
Don't worry about it.

package.json
ChristianBeach: As far as I can tell, without using xcode there is no way to use an icon for the .pkg

jackrobison:
A couple small things:

- build dependancies should be installed by `build/build.sh`

- `build/PackageInfo.xml` and `build/Distribution.xml` (versions, file counts, size of built app) should be updated by `build/build.sh` so it doesn't have to be done manually

- please remove `flat/base.pkg/Payload`, `build/daemon.zip`, `flat/Distrubution`, `flat/base.pkg/Bom`, and `flat/base.pkg/PackageInfo`

ChristianBeach:
I have work so I won't be able to work on these until later tonight but I'll get it done.

lyoshenka:
More discussion about dmg vs pkg here: https://github.com/lbryio/lbry/issues/589

lyoshenka:
@ChristianBeach any chance you can finish this?

lyoshenka:
Closing this for now. @ChristianBeach please reopen when you make the changes Jack mentioned.

PR #103, opened 2017-04-29
Removed .bind() from lambda

MaxiBoether:
wallet.js did not compile with .bind after the lambda function, and after talking in #dev in Gitter we came to the conclusion to just remove the bind as it's not necessary


alexliebowitz:
Thanks! I removed all the other bind() calls too; there's really no need for it in our codebase now that we're fully onto ES6.

PR #107, opened 2017-05-02
Add missing DAEMON_READY reducer

6ea86b96:
The missing `DAEMON_READY` reducer was breaking it @kauffj.


6ea86b96:
I added a bunch of other little fixes here. Not sure if the last commit should be rebased out and committed in a separate PR. It switches out the video player for https://github.com/Selz/plyr which seems to be much better. It works for more video types and can be used for audio later on.

PR #113, opened 2017-05-05
Make error modals display full screen properly

alexliebowitz:
Was not passing in the generic modal-overlay class. (ExpandableModal doesn't provide it automatically so you can omit it and make a totally custom overlay if you want.)

### DIFF
--- a/ui/js/component/modal.js
+++ b/ui/js/component/modal.js
@@ -28,7 +28,7 @@ export const Modal = React.createClass({
     return (
       <ReactModal onCloseRequested={this.props.onAborted || this.props.onConfirmed} {...this.props}
                   className={(this.props.className || '') + ' modal'}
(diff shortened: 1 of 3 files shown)


ui/js/component/modal.js
```diff
@@ -28,7 +28,7 @@ export const Modal = React.createClass({
     return (
       <ReactModal onCloseRequested={this.props.onAborted || this.props.onConfirmed} {...this.props}
                   className={(this.props.className || '') + ' modal'}
-                  overlayClassName={[null, undefined, ""].indexOf(this.props.overlayClassName) === -1 ? this.props.overlayClassName : 'modal-overlay'}>
+                  overlayClassName={![null, undefined, ""].includes(this.props.overlayClassName) ? this.props.overlayClassName : 'modal-overlay'}>
```
alexliebowitz: No logic problem here, just updating to use the ES6 function

PR #116, opened 2017-05-05
URI handling (WIP)

alexliebowitz:
Implements opening LBRY URIs using Electron's built-in support.

Electron only has very basic support for this: on Mac it works well, but on Windows it can only pass the URI in as a command line argument (i.e. you can't launch URIs in an app that's already running), and there's no Linux support at all yet.

I'm thinking we'd be best off to use the built-in functionality like this, and then eventually hack on improvements for Win and Linux. [Here are some ideas](https://github.com/electron/electron/issues/4857#issuecomment-201124281) about how to do it in Linux.

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -303,3 +303,12 @@ function upgrade(event, installerPath) {
 }
 
 ipcMain.on('upgrade', upgrade);
+
+if (process.platform == 'darwin') {
+  app.on('open-url', (event, uri) => {
+    win.webContents.send('open-uri-requested', url);
+  });
+} else if (process.argv.length >= 3) {
+  // No open-url event on Win, but we can still handle URIs provided at launch time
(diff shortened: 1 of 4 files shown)


lyoshenka:
Should this validate that the requested URI is a valid URI? what if they request `x&sendcreditstohacker=100000`

kauffj:
Oops, possibly shouldn't have merged this, was discussing with Alex on Slack and missed @lyoshenka's latest comment.

alexliebowitz:
This doesn't directly alter window.location. It calls App.showUri() which updates the UI to display the Show page. That function does alter the current URI stored in state and creates a history item, but even there it runs the URI through encodeURIComponent so I don't think there's any potential for injection attacks.

We should probably add a separate error page for invalid URIs, though. Right now it just silently does nothing (at least on master; might have changed on Redux).

PR #122, opened 2017-05-11
Linux URI handling (WIP)

alexliebowitz:
Still left to do: get single-instance app mode working on Linux.

In theory, it should work the same as Windows without any changes. The problem is that `app.makeSingleInstance()` is always returning true on Linux, meaning "this is the not the original process and the app should close." Probably an Electron bug, but there might be a workaround, and the worst case is that we implement a similar system from scratch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,11 @@
       "schemes": ["lbry"]
     }],
     "linux": {
-      "target": "deb"
+      "target": "deb",
+      "desktop": {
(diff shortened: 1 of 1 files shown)


kauffj:
Even in this state, it seems better than not having it at all.

6ea86b96:
This is definitely the cause. LBRY uses a HTML5 `video` tag under the hood, which does not support this file format.

PR #125, opened 2017-05-17
Convert all remaining old-style components to ES6 classes (WIP)

alexliebowitz:
Close to done. There are probably still a few callbacks that need to be wrapped in arrow functions.

### DIFF
--- a/ui/js/component/auth.js
+++ b/ui/js/component/auth.js
@@ -9,23 +9,28 @@ import {CreditAmount, Address} from "../component/common.js";
 import {getLocal, getSession, setSession, setLocal} from '../utils.js';
 
 
-const SubmitEmailStage = React.createClass({
-  getInitialState: function() {
-    return {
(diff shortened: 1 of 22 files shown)


6ea86b96:
@alexliebowitz there's a huge refactoring already in https://github.com/lbryio/lbry-app/tree/redux where, among other things, everything has been switched to ES6 classes.

alexliebowitz:
I'm in the loop about the Redux branch and I've been coordinating with Jeremy to make sure we don't duplicate effort. This PR is actually based on the Redux branch, which does have a lot of components that hadn't been converted yet (e.g. [Publish page](https://github.com/lbryio/lbry-app/blob/redux/ui/js/page/publish/view.jsx).

6ea86b96:
ah, ok, all good then :)

kauffj:
Even though this is marked WIP, I'm going to merge this now since it'll only diverge further.

PR #127, opened 2017-05-21
Redux

kauffj:
@lyoshenka said it's tricky to make a sharable release on a non-master branch, so I'm going to merge this despite known and likely unknown issues. Known issues:

* Broken snack bar links
* Broken help menu item
* Title is not properly set on navigating back
* URI handling not fully working (at least on Linux)
* New user flow should be tested from start to finish, including all reward collection.


kauffj:
I'm merging this so I can make an RC, but please take a look at this @alexliebowitz @lyoshenka @14bd4beb. See comments above re: known issues.

6ea86b96:
Awesome!

@kauffj I'm not sure that storing files by outpoint works for all files? I've had a couple of cases in the past where files have had no outpoint because they are not yet uploaded I think. The only thing that seems to always exist is the `sd_hash`.

PR #129, opened 2017-05-22
Some small fixes

6ea86b96:
- [x] Fix snackbar links
- [x] Fix title when navigating back
- [x] Always log in development
- [x] Fix playing videos that were purchased/downloaded but then deleted or moved on file system
- [x] Make navigating to discover page faster by already displaying content if it's loaded.
- [x] Fix navigating back through search results
- [x] Make search results appear immediately if they are already in the state


6ea86b96:
@kauffj a bunch of fixes here.

PR #132, opened 2017-05-23
Messing around with an audio player

6ea86b96:
@kauffj thoughts on adding audio playback? Here's an MVP showcasing my super awesome design skills.

2017-05-23 at 22 14


6ea86b96:
Closing this. I think we can get a lot working with the `render-media` lib.

kauffj:
:+1: on going native or as close to native when feasible

PR #142, opened 2017-05-26
Pre-fill metadata on Publish page

alexliebowitz:
Also includes some general refactoring and improvements to the state management

### DIFF
--- a/ui/js/component/form.js
+++ b/ui/js/component/form.js
@@ -90,6 +90,10 @@ export class FormField extends React.PureComponent {
     return this.refs.field.options[this.refs.field.selectedIndex];
   }
 
+  getOptions() {
+    return this.refs.field.options;
+  }
+
   render() {
     // Pass all unhandled props to the field element
     const otherProps = Object.assign({}, this.props),
@@ -218,6 +222,10 @@ export class FormRow extends React.PureComponent {
 
   getSelectedElement() {
     return this.refs.field.getSelectedElement();
+  }
+
+  getOptions() {
+    return this.refs.field.getOptions();
   }
 
   focus() {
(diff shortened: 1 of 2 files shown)


kauffj:
@alexliebowitz what needs to happen to merge this?

kauffj:
Related? https://github.com/lbryio/lbry-app/issues/186

alexliebowitz:
Just need to finish writing. The biggest part is done (refactoring so the metadata fields are controlled). I got stuck because I've been receiving a blank set in the "myClaims" prop and had trouble debugging it; there's a lot of Redux-y indirection that was giving me a headache. Good way to learn the new codebase, though. Over the weekend I'm going to plow through and finish this. IIRC that issue started happening suddenly for no apparent reason, which makes me think it's a daemon bug. If it is, I'll make a simple mock for the correct behavior and file an issue in the lbry repo.

alexliebowitz:
#186 is related in that the updates won't actually work properly until we fix that. But this is purely form filling logic within the Publish page, doesn't touch Redux at all.

alexliebowitz:
Whew. Figured out the issue with the myClaims prop being empty.

The Publish page is populating the prop with a value from the Redux store: https://github.com/lbryio/lbry-app/blob/master/ui/js/page/publish/index.js#L15

... but it never actually dispatches an action to call `file_list_mine` and get the value, so it defaults to an empty set:
https://github.com/lbryio/lbry-app/blob/master/ui/js/selectors/claims.js#L84

But the My Files page does request the info: https://github.com/lbryio/lbry-app/blob/master/ui/js/page/fileListDownloaded/index.js#L24

So the prop is only populated if you have already visited the My Files page. If you haven't visited My Files, the Publish page will act like you have no claims, and if you enter the name of something you already published, it doesn't indicate that you're updating an old claim. (This is happening on master too.)

Now that I finally figured that out, tomorrow I will fix this bug, finish the pre-fill stuff, and everything should be ready for review by late tomorrow.

@kauffj

kauffj:
You might want to check out #192, seems related.

6ea86b96:
Can I help out at all on this @alexliebowitz?

PR #149, opened 2017-05-28
Try to only keep one copy of each claim in the store

6ea86b96:
I think it would be cool to try to only keep one copy of any object in the store, to avoid weird problems later on with the same thing having different properties.


ui/js/reducers/claims.js
```diff
@@ -40,20 +40,33 @@ reducers[types.RESOLVE_URI_CANCELED] = function(state, action) {
...
+    claims,
+  } = action.data
+  const myClaims = new Set(state.myClaims)
+  const byUri = Object.assign({}, state.claimsByUri)
+
+  claims.forEach(claim => {
+    const uri = lbryuri.build({
+      contentName: claim.name,
+      claimId: claim.claim_id,
+      claimSequence: claim.nout,
```
kauffj: I suspect this will cause misses, since they are not typically built with claimSequence and claimId.

6ea86b96: I think rather than misses, this is going to cause duplicates. So rather than having dupes separated into `claimsByUri` and `myClaims` now we've just moved the duplication problem into the `claimsByUri` object, but a step in the right direction perhaps for removing dupes?

I think perhaps long term it might be better to store claims in `claims.byId` and then map `byUri`, `myClaims`, etc to claim ids.

ui/js/reducers/file_info.js
```diff
@@ -18,7 +18,9 @@ reducers[types.FILE_LIST_COMPLETED] = function(state, action) {
 
   const newFileInfos = Object.assign({}, state.fileInfos)
   fileInfos.forEach((fileInfo) => {
-    newFileInfos[fileInfo.outpoint] = fileInfo
+    const { outpoint } = fileInfo
+
+    if (outpoint) newFileInfos[fileInfo.outpoint] = fileInfo
```
6ea86b96: I have a file with no outpoint locally. It's one that I tried to upload but it didn't work for some reason.

ui/js/reducers/claims.js
kauffj: We were saying the same thing here.

PR #151, opened 2017-05-28
Use BusyMessage for price lookup

6ea86b96:
Maybe better than the current "???"

Before
2017-05-28 at 17 56

After
2017-05-28 at 17 53

### DIFF
--- a/ui/js/component/filePrice/view.jsx
+++ b/ui/js/component/filePrice/view.jsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import {
+  BusyMessage,
   CreditAmount,
 } from 'component/common'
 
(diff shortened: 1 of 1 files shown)


kauffj:
This was a conscious choice. I didn't like 40 loading indicators popping up all at once on the landing page.

I do think we can something better than ???.

kauffj:
I think in general it would be useful to develop a UI indicator for "this information is being refreshed/queried in the background" that is NOT an active indicator. For example, as we persist claim and other information more, this icon could be used when rendering tiles off of "stale" claim info, and likely in other places.

kauffj:
Possibly `icon-refresh`?

PR #157, opened 2017-05-29
Fix naming:  Publishes > Published for address bar

tzarebczan:
When going to Published files, the address bar would show "Publishes". Changed this to "Published".


kauffj:
This is Publishes on purpose, to match the tense of "Downloads & Purchases".

PR #158, opened 2017-05-29
Downloads & Purchases / Publishes to past tense

tzarebczan:
Change text "Downloads & Purchases" to "Downloaded & Purchased"
Change text "Publishes" to "Published"


kauffj:
I know I said to go ahead and change both, but I changed my mind again lol. I don't like that "Downloaded" will refer to currently downloading items. Just going to leave it as is.

PR #160, opened 2017-05-30
i18n Support

MaxiBoether:
Please excuse some double commits, but it seems like a rebase did go wrong. We doublechecked to make sure every String is translateable, please tell us if we missed a String. There is also a sample translation added that just adds RS_ to every string. Users are able to change the language in the options.
I will also try and create a German translation as soon as this is merged into master. intnick will probably make a Serbian translation.


ui/js/rewards.js
```diff
@@ -7,13 +7,13 @@ import {
...
 function rewardMessage(type, amount) {
   return {
-    new_developer: `You earned ${amount} for registering as a new developer.`,
-    new_user: `You earned ${amount} LBC new user reward.`,
-    confirm_email: `You earned ${amount} LBC for verifying your email address.`,
-    new_channel: `You earned ${amount} LBC for creating a publisher identity.`,
-    first_stream: `You earned ${amount} LBC for streaming your first video.`,
-    many_downloads: `You earned ${amount} LBC for downloading some of the things.`,
-    first_publish: `You earned ${amount} LBC for making your first publication.`,
+    new_developer: __(`You earned %s for registering as a new developer.`, amount),
```
kauffj: Tick marks are only used around interpolated strings. If interpolation is being dropped, please switch to double quotes.

ui/js/selectors/app.js
```diff
@@ -39,34 +39,37 @@ export const selectPageTitle = createSelector(
...
-        return 'Report'
+        return __('Report')
       case 'wallet':
+        return __("Wallet")
       case 'send':
+        return __("Send")
       case 'receive':
+        return __("Receive")
       case 'rewards':
-        return page.charAt(0).toUpperCase() + page.slice(1)
```
kauffj: This code was correct. Yes, it means it cannot easily be translated. For tricky ones like this, just make a note and we'll figure out how to resolve them.

ui/watch.bat
```diff
@@ -0,0 +1,33 @@
+rmdir %~dp0node_modules /s /q
```
kauffj: Please update the README and add a heading titled "Internationalization" that explains a bit about this, and anything else that would be relevant to set up or programmers new to the repo. Should be a couple paragraphs tops, and possibly just a sentence or two.

ui/js/page/help/view.jsx
```diff
@@ -65,70 +65,72 @@ class HelpPage extends React.Component {
...
+         <section className="card">
+           <div className="card__title-primary"><h3>{__("About")}</h3></div>
+           <div className="card__content">
            {this.state.appVersionInfo ?
-            (ver.lbrynet_update_available || ver.lbryum_update_available ?
-               <p>A newer version of LBRY is available. <Link href={newVerLink} label={`Download LBRY ${ver.remote_lbrynet} now!`} /></p>
-               : <p>Your copy of LBRY is up to date.</p>) : null}
-           { ver ?
+             (ver.lbrynet_update_available || ver.lbryum_update_available ?
+              <p>{__("A newer version of LBRY is available.")} <Link href={newVerLink} label={__("Download LBRY %s now!"), ver.remote_lbrynet} /></p>
```
kauffj: Misstructured call?

ui/js/main.js
```diff
@@ -2,13 +2,19 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import lbry from './lbry.js';
 import lbryio from './lbryio.js';
+
+const i18n = require('y18n')({directory: 'app/locales'/*, locale: lbry.getClientSetting('language')*/});
+window.__ = i18n.__;
+window.__n = i18n.__n;
```
kauffj: Is this used? If not, no need to global.

ui/js/main.js
```diff
@@ -2,13 +2,19 @@ import React from 'react';
...
+window.i18n = i18n;
```
kauffj: This can probably be imported when needed, rather than globalled.

kauffj:
@lyoshenka might want to take a look at this one too

ui/js/selectors/app.js
MaxiBoether: @intnick I think you translated this, could you please take a look?

ui/watch.bat
MaxiBoether: Ohhh, this were my Windows compilation files. They were not supposed to be made public because most ppl are probably working with unix based systems. I am gonna remove them

ui/js/main.js
MaxiBoether: @intnick can you please have a look at those?

ui/js/rewards.js
MaxiBoether: should be fixed

ui/js/page/help/view.jsx
MaxiBoether: should be fixed

ui/js/selectors/app.js
intnick: OK, do you want to return to the initial solution ? We could possibly translate the return of the function, I just found it not to be an elegant or easily readable solution so I separated them

ui/js/main.js
intnick: Yes, the translate functions are used on every page. __n is used rarely and we could use it as i18.__n() instead, though to make it equivalent and available.
I tried it first with i18n component in local scope but problems start with language switching. It wouldn't change the language strings if on the settings page we load the i18n again and do a setLanguage(language) method call. We could tell the user to restart the app, which would have been much easier for me, but I didn't like the idea of app restart if you changed the language. More than open for other suggestions ?

ui/js/selectors/app.js
kauffj: Yes, just leave it the way that it was. I'm fine if a few problem strings aren't translated to start (though it'd be good to build a list of these).

ui/js/main.js
kauffj: I think it'd make sense for i18n to be set in `ui/js/app.js` then?

kauffj:
Please make sure to ping me once the few remaining minor changes are made -- I'll put this in master than so we stop generating conflicts with ongoing work.

ui/js/selectors/app.js
intnick: On it. I think a few more have the same case and name so I'll aggregate those as well.

ui/js/main.js
intnick: Trying it out

ui/js/selectors/app.js
MaxiBoether: I don't know if it helps, but wrapping __() around the output, even if it's not a build-in String (e.g. rewards which are loaded from the server) also work with y18n, you just need to run the app once and then it's added to the language file automatically

kauffj:
Please make the changes discussed in Slack.

MaxiBoether:
Changes should be made, but please review carefully

PR #189, opened 2017-06-02
Render all file types that render-media can handle

6ea86b96:
I've added all the other types that `render-media` can play @kauffj.

It renders documents into an iframe. This could be a security vulnerability. I tried adding some HTML with JS and couldn't get anything to run, but looking at the `render-media` source, it seems that it does allow scripts somehow https://github.com/feross/render-media/blob/master/index.js#L282.

filetypes

### DIFF
--- a/ui/js/component/video/view.jsx
+++ b/ui/js/component/video/view.jsx
@@ -49,7 +49,7 @@ class VideoPlayButton extends React.Component {
      */
 
     const disabled = isLoading || fileInfo === undefined || (fileInfo === null && (!costInfo || costInfo.cost === undefined))
-    const icon = mediaType == "image" ? "icon-folder-o" : "icon-play"
+    const icon = ["audio", "video"].indexOf(mediaType) !== -1 ? "icon-play" : "icon-folder-o"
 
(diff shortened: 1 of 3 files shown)


kauffj:
Would the scripts would still be sandboxed from accessing the daemon because they're in an iframe?

6ea86b96:
I think so but not 100% certain. We would need to try some malicious files.

6ea86b96:
Seems to be ok @kauffj. It's not possible to access the parent in any way.

2017-06-03 at 19 28

ui/js/component/video/view.jsx
```diff
@@ -119,6 +119,8 @@ class Video extends React.Component {
     if (mediaType === "video") {
       klassName += "video-embedded video"
       klassName += isPlaying ? " video--active" : " video--hidden"
+    } else if (mediaType === "application") {
+      klassName += "video-embedded"
```
kauffj: Is all of the class checking correct here? I may not understand the conditions properly, but seems like you could end up with repeated classes and/or classes stacked without a space.

PR #191, opened 2017-06-03
Add timeout for jayson client requests

### LINKED ISSUE #19: LBRY daemon does not always launch when the app runs
## The Issue
I've heard of this issue on Reddit and from several people on Slack, but I have not experienced it myself.

Several people report that the daemon does not launch when the app is run and that they have to launch it manually to get it to work.

### Steps to reproduce
I can't :(

### Expected behaviour
App launches daemon.

### Actual behaviour
Sometimes it doesn't, apparently.

### LINKED ISSUE #188: lbry-app hangs indefinitely when daemon hangs
## The Issue
If `lbrynet-daemon` hangs on the initial `status` call made by `app/main.js`, then the app will hang indefinitely.

### Steps to reproduce
1. Launch a copy of `lbrynet-daemon` that hangs (not sure how to do this reliably)
2. Launch the app

### Expected behaviour
App at least errors cleanly, if not killing and restarting daemon.

### Actual behaviour
App sits at 'starting daemon' forever...

### Anything Else
`launchDaemonIfNotRunning` needs to, at a minimum, detect a hanging `status` call.

krrishnarraj:
request() didn't hit the callback sometimes
fixes #19
fixes #188

Timeout is currently set to 1000 ms, please choose an optimal number

### DIFF
--- a/app/main.js
+++ b/app/main.js
@@ -21,7 +21,13 @@ const VERSION_CHECK_INTERVAL = 30 * 60 * 1000;
 const LATEST_RELEASE_API_URL = 'https://api.github.com/repos/lbryio/lbry-app/releases/latest';
 
 
-let client = jayson.client.http('http://localhost:5279/lbryapi');
+let client = jayson.client.http({
+  host: 'localhost',
+  port: 5279,
+  path: '/lbryapi',
+  timeout: 1000
+});
+
 // Keep a global reference of the window object, if you don't, the window will
 // be closed automatically when the JavaScript object is garbage collected.
 let win;


krrishnarraj:
I still see one more problem.
The daemon when launched by app, takes around 10-12 secs to initialize completely. By then app timeouts and reports 'Connection Failure'. Can this time be increased or retried?

kauffj:
@krrishnarraj Thanks for fixing this. I think there is still a problem with detecting and handling unresponsive daemons, but this should address the tickets you mention.

The app will later poll that endpoint for up to about a minute, so 1 second as a timeout works fine for this even with a 10-12 second initialize time.

Please DM me as kauffj in our Slack if you'd like a tip for this PR.

PR #192, opened 2017-06-04
Store claims by claim_id and fileinfo by sd_hash

6ea86b96:


kauffj:
@6ea86b96 unfortunately `sd_hash` is not guaranteed to be unique per `fileInfo`, while `outpoint` is. Technically, any number of claim entries could resolve to the same `sd_hash`.

Storing claims by id is definitely a smart change, please reduce this PR to just that and re-assign to me.

6ea86b96:
Hmm, that's a bummer with the `sd_hash`. I did that to be able to store pending uploads in the state. Will think about a different way to tackle pending uploads then.

6ea86b96:
Ok, rebased out the fileinfo stuff.

6ea86b96:
I also have a question about the way claims/certificates work. I just commented this out for now as I didn't really understand what it's doing. https://github.com/lbryio/lbry-app/pull/192/files#diff-152922ea714e848ca916a0fde611d75eR25

6ea86b96:
Also, it seems that this PR introduces a bug for uris that have no claim. Happens when searching. We end up in an infinite resolve loop.

6ea86b96:
The loop is caused by the `setTimeout` that keeps trying to resolve the claim in `doFetchCostInfoForUri`

6ea86b96:
nvm, fixed

6ea86b96:
hmm, ok, so I see that the certificates stuff I commented out has broken the channel pages.

6ea86b96:
ok, nvm, I fixed those. I think this is good for review now.

kauffj:
Would you mind cleaning up these merge conflicts? Sorry to have created extra work for you. I was working backwards off of my email and thought this wasn't ready, or I would have merged it first.

6ea86b96:
rebased

PR #194, opened 2017-06-04
Cache state in IndexedDB

6ea86b96:
We can cache parts of the state in local storage to achieve almost instantaneous load times.

fast-load

We can also compress that data nice and small:

2017-06-05 at 00 35

It looks a bit like this:

2017-06-05 at 00 37


kauffj:
This is exciting! One thing to consider is what this behavior should be like as the landing page gets more dynamic.

6ea86b96:
Yes, I was trying to cache the minimum necessary data to speed everything up. This branch was forked from the claim ids branch. I'm actually only caching a very small subset of the state. Just the `byId`, `claimsByUri` and `myClaims` properties of the `claims` store property. This way we still lookup featured content, uploaded/downloaded files, etc, and just load the claim from cache. Hopefully we can also minimize situations where we change the structure of the data and it's different in the cache like this too.

Remaining to do here is investigate cache expiry. I'm looking at adding at time based expiration to individual claims. I'm not too sure what happens if the cache size hits the localStorage 10MB limit. I couldn't find much information on how `redux-persist` handles that case. Even compressed, the cache does grow fairly rapidly. I was able to grow it to 230KB after 30 minutes of searching.

We will also need to refresh claim data while navigating to ensure updated claims are synced to the store/cache.

kauffj:
A few comments:
1) Rather than worrying about date based time outs, we could just always refresh when serving from a cached state, possibly with some minor indicator that potentially stale data is being rendered in the interim.
2) We can invalidate cached data via the build process as well. For example, it could be directly tied to `lbryschema` or `lbry` version numbers.
3) The slowness of `file_list` and `claim_list_mine` and other API calls is a relatively high priority for improvements to the daemon. This doesn't really affect anything here, just letting you know.

6ea86b96:
I think the problem is more about users using the app for long enough that localstorage limits are reached, which seems possible within around 1 full day of heavy usage. Of course this also makes date based timeouts not particularly useful. Seems like a FIFO strategy would be great here, but not particularly easy to implement. 100% agree we should refresh when serving from cache. Unless I'm missing something I don't think invalidating from builds is going to help that much in this case, at least once releases become a bit less frequent. I'm kind of hoping `redux-persist` just handles storage full out of the box. Going to look into that today.

Those daemon improvements will be great. `claim_list_mine` results are actually already in the cache and refreshed as you navigate. `file_list` did seem like a good caching target too. Not sure how much optimization can be done when you need to read from the file system? I didn't add those to the cache yet though, just to keep it simple.

kauffj:
Most slow API commands are driven by remote calls to validate signatures. File list is slow not from disk reads but because it is re-validating the claim info every time it is fetched.

Yes, this is dumb.

6ea86b96:
Ok @kauffj, so there are 2 separate issues here.

1) localStorage filling up. I've fixed this by switching to `localForage` which uses `indexedDB` under the covers. Electron has much larger size limits for this and when it does hit them `indexedDB` handles eviction itself on a LRU basis.

2) Making sure that the claims/prices/fileInfo/etc are updated. This is a general issue now that everything is in redux and we need to figure out a nice way to sync data. I think we tackle that separately.

6ea86b96:
Just rebased on master and squashed into one commit.

6ea86b96:
Realized that this would cause some problems on the discover page. Cached claim data wouldn't get updated so people wouldn't see updated community content.

I've added some code to force claims to be refreshed on the discover page and on the show page when it's mounted.  Search already seems to resolve claim data so I just made that a little more efficient with batch actions.

@kauffj

kauffj:
I'd like to wait to merge into master until after we release 0.12.

6ea86b96:
Closing this in favour of https://github.com/lbryio/lbry-app/pull/225

PR #204, opened 2017-06-06
Some package tweaks for npm 5x

6ea86b96:
Just upgrade to NPM 5 and it now generates these `package-lock` files. I think that's backwards compatible. Would be good if someone with a lower NPM version could double check. If not I'll try downgrading my local version in 8-10 hours.


kauffj:
Just trying to rack up the line changes I see.

Works for me on NPM 3.5, so merged.

PR #206, opened 2017-06-07
Update video player loading screen

seanyesmunt:
### Summary

The goal of this is to add a slightly better-looking video player loading screen and set a folder tree structure for internal components. In this PR I have created a folder called `internal` inside of the `video` component folder. Currently, the components are separated by name into folders, but they only have a single `.jsx` file for the main component, as well as any other internal components needed to build the main component.

This leads to huge React files with many components which can make it hard to follow the component tree. I think we should set a style now and try to adhere to it as components grow.

The structure added in this PR is:
```
component/
  ...
  video/
      internal/
          some-internal-comp.jsx // some component that is imported by view.jsx
      view.jsx // the base component
      index.js // connects base component to store

  ...
```

### How it looks

loading

### DIFF
--- a/ui/js/component/video/internal/loading-screen.jsx
+++ b/ui/js/component/video/internal/loading-screen.jsx
@@ -1,11 +1,11 @@
 import React from "react";
 
-const LoadingScreen = ({ statusMessage }) =>
+const LoadingScreen = ({ status }) =>
   <div className="video--loading--screen">
     <div className="loading--spinner" />
 
     <div className="video--loading--status">
-      {statusMessage}
+      {status}
     </div>
   </div>;
 
--- a/ui/js/component/video/internal/loading-screen.jsx
+++ b/ui/js/component/video/internal/loading-screen.jsx
@@ -1,0 +1,12 @@
+import React from "react";
+
+const LoadingScreen = ({ statusMessage }) =>
+  <div className="video--loading--screen">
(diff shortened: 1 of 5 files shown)


6ea86b96:
:+1: from me. I'd love us to break the components up more like this.

ui/js/component/video/internal/loading-screen.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+const LoadingScreen = ({ status }) =>
+  <div className="video--loading--screen">
```
kauffj: We're using [BEM](http://getbem.com/) as our CSS naming pattern. If our top level block here is `video`, then these class names should be something like:

video__loading-screen
video__loading-screen-content
video__loading-status

`--` should only be used as a modifier, e.g. `video__loading-status--error`.

ui/scss/component/_video.scss
```diff
@@ -28,6 +27,69 @@ video {
...
     /*background: none;*/
   }
+
+  .video--loading--screen {
+    height: 100%;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    .loading--screen--content {
```
kauffj: Since BEM generates unique class names, we don't typically nest CSS styles, particularly classes.

PR #218, opened 2017-06-09
Fix display of info about user's own claims on Publish page (WIP)

alexliebowitz:


kauffj:
This is better than not fetching at all, so I merged it.

PR #229, opened 2017-06-12
Cleaning up publishing (WIP)

6ea86b96:
Some publishing cleanup. Fixes looking up claims but needs work adding new channels and getting publishing into redux.

publishing-lookup

- [x] Refactor publish page
- [x] Fix markdown support
- [x] Fix prefilling claim data
- [x] Switch to my claims for publishes page
- [x] Abandoned files show up in publishes until the component is reloaded
- [x] Show page is blank for unconfirmed files
- [x] Move pending files to pending claims
- [x] Remove pending claim when it is accepted, or the confirmed one is fetched.
- [x] Remove old failed pending claims automatically
- [x] Fix name for pending claims on publishes page


6ea86b96:
This is getting a bit better now. Just having one problem where if I publish to `lbry://@mychannel/testing` for example, then also the file at `lbry://testing` appears in my uploaded files list until I clear the cache. Haven't figured out how it's getting in there in the first place yet.

6ea86b96:
ok @kauffj, this is finally working now but needs some cleanup. I'll get that done tomorrow or Tuesday morning. There is one bug which will be fixed by https://github.com/lbryio/lbry-app/issues/315.

Currently the fake pending publish channel/name are correct, and the confirmed file channel/name are also correct, but the files returned by `file_list` which belong to an unconfirmed change have no channel id. So there's a period of time where the file appears wrong.

6ea86b96:
Still having problems with this. Found several bugs while testing. Switching to using claim_list_mine for published files will fix. Made some good progress with that and hoping to get it wrapped up by tomorrow or the day after. :sweat_smile:

6ea86b96:
Ok, this is actually working now. However, I don't really like it :/

I put the pending publishes code into the cached redux data at the beginning but it doesn't feel right if we're going to end up dropping `redux-persist` altogether soon. Also, I switched to using `claim_list_mine` for published files but am still using the `FileList` component at the moment.

I kept the code for removing caching from `lbry.js` in it's own commit. Going to create a second PR tomorrow with that rebased out, and make the app reload the claims when publish succeeds/fails.

6ea86b96:
Superseded by https://github.com/lbryio/lbry-app/pull/346

PR #232, opened 2017-06-13
Regenerate lock files

6ea86b96:
Beginning to wonder how these lock files actually help. However, I can't run the app without them right now.


kauffj:
If this is a perpetual issue, we should probably look at a commit hook.

PR #242, opened 2017-06-15
Support markdown makeup in claim description

longle255:
Allow user to put rich content in claim's description by using Markdown language.

Without markdown support:
Without markdown support

With markdown
With markdown support


kauffj:
@longle255 I think adding markdown support is a great idea! However, I think if we're going to do it we should flesh it out a bit more:

- Switch the publish description editor to something like https://simplemde.com/
- Either disallow h1, h2 (and maybe h3?) or style them to be smaller
- Handle case of invalid or poorly formed markdown when rendering (someone can theoretically put *anything* in the description metadata field - will `ReactMarkdown` handle all of this?)

Feel free to discuss with me on Slack as well, this is definitely a valuable addition!

longle255:
sure, that makes sense. I was going to update the description editor too. I'll see if I can address all suggestions

longle255:
added:

- editor for claim's description in publish page
- limit markdown type: remove heading, html block

Markdown editor in publish page

lyoshenka:
Another consideration: this opens up the possibility of someone embedding javascript in the description, which is a serious security hole. Please make sure to handle this. We probably don't want CSS either.

longle255:
yes that was considered too, all HTML block which could possibly include JS and CSS will not be rendered

longle255:
For example if I put a markdown like this
```
Normal markdown
*italic*
**bold**
<b> This should not be bold </b> <script> alert("security break");</script>
another text
```
those HTML tag will be escape as screenshot

escaped html

longle255:
@kauffj ready for review 😁

ui/js/component/common.js
```diff
@@ -42,6 +42,59 @@ export class TruncatedText extends React.PureComponent {
...
+export class TruncatedMarkdown extends React.PureComponent {
+  static propTypes = {
+    lines: React.PropTypes.number,
+  };
+
+  static defaultProps = {
+    lines: null,
+  };
+
+  truncateMarkdown(text) {
```
kauffj: This entire function terrifies me. I'm guessing the proper way to do this is to fully parse the markdown, then return the plain text.

ui/js/component/fileCard/view.jsx
```diff
@@ -96,7 +101,7 @@ class FileCard extends React.PureComponent {
                 style={{ backgroundImage: "url('" + metadata.thumbnail + "')" }}
               />}
             <div className="card__content card__subtext card__subtext--two-lines">
-              <TruncatedText lines={2}>{description}</TruncatedText>
+              <TruncatedMarkdown lines={3}>{description}</TruncatedMarkdown>
```
kauffj: Why did lines switch?

ui/js/component/form.js
```diff
@@ -106,26 +113,45 @@ export class FormField extends React.PureComponent {
...
+          }
+          name={this.props.name}
+          options={{
+            hideIcons: ["heading"],
+          }}
+          {...otherProps}
+        >
+          {this.props.children}
+        </SimpleMDE>
+      : <this._element // can't be applied to SimpleMDE idk why
```
kauffj: I'm decently confident this is possible, maybe with `this._element = <SimpleMDE>` or `this._element = SimpleMDE`? Worth looking into. If it's not, this can still be cleaned up by compacting the shared properties to one declaration and passing them in with an extract operator.

ui/scss/component/_form-field.scss
```diff
@@ -163,4 +163,8 @@ input[type="text"].input-copyable {
 }
 .form-field__helper {
   color: $color-help;
+}
+
+.CodeMirror-scroll {
+  height: auto !important;
```
kauffj: Is this `!important` unavoidable?

kauffj:
There's also some discussion among LBRY devs that this will require an additional metadata field.

kauffj:
This is merged into `publishing` branch, which will eventually be either 0.12.1 or 0.13. Either way, this is merged and considered complete, but is not going into `master`.

PR #248, opened 2017-06-16
Add a way for users to clear their own cache

6ea86b96:
In anticipation of caching problems. Not sure if there's a nicer way to reload. We need to do a full page load though, otherwise the store will just get synced back into the cache again and nothing will have changed.


kauffj:
Perfectly fine for a beta and good to have.

PR #273, opened 2017-06-22
Fix refreshing

6ea86b96:
Stayed up until 3am to watch the live feed so figured I should use the time to fix something too :)

I used hash based routing as it seems easier than trying to intercept refreshes, reload the app and then redirect back to the previous path. Might be stupid, I'm tired.


6ea86b96:
Not sure why this build failed...

PR #281, opened 2017-06-23
UI concepts #1. 23rd June, 2017

akinwale:
- UI font changed to Lato.
- Implemented auto thumbnails for cards without thumbnails. A substring of the title (5 characters or the length of the title, whichever is less) will be displayed over a random background colour from a set of predefined colours.
- Moved thumbnails to the top of the card.
- Moved the title, publisher and price to the bottom of the card.
- Moved the description to a semi-transparent overlay over the thumbnail, which will be displayed on card mouse over
- Reduced card width.

### DIFF
--- a/ui/dist/index.html
+++ b/ui/dist/index.html
@@ -5,7 +5,7 @@
         <title>LBRY</title>
 
         <link href='https://fonts.googleapis.com/css?family=Raleway:600,300' rel='stylesheet' type='text/css'>
-        <link href="https://fonts.googleapis.com/css?family=Lato:400,400i" rel="stylesheet">
+        <link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,400italic,600italic,600' rel='stylesheet' type='text/css'>
         <link href="./css/all.css" rel="stylesheet" type="text/css" media="screen,print" />
         <link rel="stylesheet" href="https://cdn.plyr.io/2.0.12/plyr.css">
         <link rel="icon" type="image/png" href="./img/fav/favicon-32x32.png" sizes="32x32">
--- a/ui/dist/index.html
+++ b/ui/dist/index.html
@@ -5,7 +5,7 @@
         <title>LBRY</title>
 
         <link href='https://fonts.googleapis.com/css?family=Raleway:600,300' rel='stylesheet' type='text/css'>
-        <link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,400italic,600italic,600' rel='stylesheet' type='text/css'>
+        <link href="https://fonts.googleapis.com/css?family=Lato:400,400i" rel="stylesheet">
         <link href="./css/all.css" rel="stylesheet" type="text/css" media="screen,print" />
         <link rel="stylesheet" href="https://cdn.plyr.io/2.0.12/plyr.css">
         <link rel="icon" type="image/png" href="./img/fav/favicon-32x32.png" sizes="32x32">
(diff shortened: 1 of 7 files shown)


kauffj:
Thanks for this @akinwale! Will have some feedback on this Monday.

ui/dist/index.html
```diff
@@ -5,7 +5,7 @@
         <title>LBRY</title>
 
         <link href='https://fonts.googleapis.com/css?family=Raleway:600,300' rel='stylesheet' type='text/css'>
-        <link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:400,400italic,600italic,600' rel='stylesheet' type='text/css'>
+        <link href="https://fonts.googleapis.com/css?family=Lato:400,400i" rel="stylesheet">
```
kauffj: I'm not prepared to change the font as part of these design changes.

ui/js/component/fileCard/view.jsx
```diff
@@ -8,6 +8,23 @@ import UriIndicator from "component/uriIndicator";
 
 class FileCard extends React.PureComponent {
   componentWillMount() {
+    const autoThumbClasses = [
```
kauffj: This behavior should be standardized and used by at least FileTile (and FileShow? not sure about this). Probably involves making `card__media` into a component.

ui/scss/component/_card.scss
```diff
@@ -63,7 +65,19 @@ $padding-card-horizontal: $spacing-vertical * 2/3;
   white-space: pre-wrap;
 }
 .card__subtext--two-lines {
-  height: $font-size * 0.9 * $font-line-height * 2;
+  height: ($font-size * 0.9 * $font-line-height + 6) * 2;
```
kauffj: "+6" makes this magic CSS, if this 6 references something else, it should be a variable, if it was determined by trial and error, it's a hack

ui/scss/component/_card.scss
```diff
@@ -31,6 +32,7 @@ $padding-card-horizontal: $spacing-vertical * 2/3;
   padding: 0 $padding-card-horizontal;
   margin-top: $spacing-vertical * 1/3;
   margin-bottom: $spacing-vertical * 1/3;
+  font-size: 80%
```
kauffj: I could possibly favor a reduction in card size, but I think as-is this font up being too small. Currently, the app is sort of inconsistently following material design guidelines (but trending towards consistency!). You can find these at [material.io](https://material.io).

akinwale:
@kauffj Pushed a new commit. Here's a summary of the changes.
- Added a CardMedia component with title and thumbnail properties. Can be included in any component where a title and a thumbnail can be specified. Currently used in the FileCard and FileTile components.
- Changed the font back to Source Sans Pro.
- Replaced the percentage font sizing with em font sizing and increased the minimum font size. The font sizes may still require some tweaking based on the average length of title strings that need to be displayed on the cards.
- Removed magic CSS.

kauffj:
I took another look at this. I like the `CardMedia` change a lot and the code looks shippable.

I like the idea of reducing the card with, but am not sold on the current hover behavior. I'd like to look at more options here, Including looking at a different hover style that introduces more additional info on hover.

Could you split the CardMedia part out into it's own PR? And then either open the remainder in another PR, or discuss with me on Slack and we can come up with some other options.

akinwale:
@kauffj I've created a new pull request for the CardMedia implementation with auto thumbnails at https://github.com/lbryio/lbry-app/pull/349. I suppose you could close this pull request and then I'll discuss with you on Slack regarding options for hover display, and the card size before I create another one.

PR #282, opened 2017-06-24
Add a badge to the dock for downloaded files and show progress bar for download progress

6ea86b96:


6ea86b96:
Also added desktop notifications when a download completes

kauffj:
@6ea86b96 the fact that this is merged is the true comment, but:

- The notifications were not clickable for me. This may be an Ubuntu thing.
- I am hesitant about the way the badge functions were added to `utils.js`. Perhaps this code should go in an actions file?

6ea86b96:
@kauffj I was also not really sure about the util functions. Seemed a bit yagni. I will move to actions. The notifications are clickable on mac, but I didn't test on linux or windows

PR #287, opened 2017-06-25
Discover Page - Replace horizontal scrollbar with arrows (WIP)

seanyesmunt:
https://github.com/lbryio/lbry-app/issues/250

### DIFF
--- a/ui/js/component/discover/featuredCategory.jsx
+++ b/ui/js/component/discover/featuredCategory.jsx
@@ -2,33 +2,75 @@ import React from "react";
 import FileCard from "component/fileCard";
 import ToolTip from "component/tooltip.js";
 
-const FeaturedCategory = props => {
-  const { category, names } = props;
+/*
+Only show scroll arrow if cursor is inside of row
 
-  return (
-    <div className="card-row card-row--small">
-      <h3 className="card-row__header">
-        {category}
-        {category &&
-          category.match(/^community/i) &&
-          <ToolTip
-            label={__("What's this?")}
-            body={__(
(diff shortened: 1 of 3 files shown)


kauffj:
@seanyesmunt since this only contains a refactor so far, I'm inclined to close. Would love to see what you come up with here though! 

And would also be open to non-master commit rights if it makes workflow easier.

seanyesmunt:
@kauffj I just got back from a trip, so haven't had much time to work on it so far, but I plan to work on this today/following days. That refactor commit was just an initial commit to get the PR made. I would appreciate if you could reopen the PR (if possible). If not I can just open a new PR.

seanyesmunt:
@kauffj What do you think about using an open source carousel component for this? https://github.com/akiran/react-slick is one I've found that looks especially good (a ton of options).

kauffj:
@seanyesmunt I'm very much in favor of looking at existing solutions, but I'm conservative about adding additional code and dependencies.

My preferred approach here would be to look at the various styles, pick one, and then look at how much of the code we actually need to implement just that specific solution.

Sometimes it ends up that you only need a small minority of the JS/CSS used by the entire component, and you're better off bringing just this specific part in (it also means it will match your naming conventions, styles, etc.).

kauffj:
OTOH if using an off the shelf carousel delivers something more usable than the current version, I'd probably just take it :)

seanyesmunt:
@kauffj sorry to close after I asked you to reopen. I don't really have the time to finish this right now, at least not something I would be happy with. I will continue to look at open issues, and if this is still an outstanding feature request when I am able to continue working I would love to take it back up.

PR #289, opened 2017-06-26
"Fixed NSFW content showing on filePage"

### LINKED ISSUE #286: NSFW content isn't obscured on show page
## The Issue

NSFW content isn't obscured on show page

### Steps to reproduce
1. Go to lbry://jacki with NSFW content turned off

### Expected behaviour
The content should be obscured

### Actual behaviour
The NSFW thumbnail image is visible

hackrush01:
NSFW content is obscured on show page and gives an overlay to inform
the user. Fixes #286

Added one additional format(epub) in lbry.js.

### DIFF
--- a/ui/js/component/fileCard/view.jsx
+++ b/ui/js/component/fileCard/view.jsx
@@ -97,22 +97,22 @@ class FileCard extends React.PureComponent {
               <TruncatedText lines={2}>{description}</TruncatedText>
             </div>
           </Link>
-          {obscureNsfw &&
-            this.state.hovered &&
-            <div className="card-overlay">
-              <p>
-                {__(
-                  "This content is Not Safe For Work. To view adult content, please change your"
-                )}
-                {" "}
-                <Link
-                  className="button-text"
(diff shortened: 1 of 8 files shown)


kauffj:
This PR doesn't require changes, since it's all minor cleanup I can do as I merge, but here's feedback:

- Since this code runs in Electron, which is essentially Chrome, it's the only browser that needs to be supported.
- SCSS variables that are not intended or unlikely to be shared across files should be in the file itself, not global.scss.
- Even small stuff like epub should be a separate PR.
- Indentation should be handled by `Prettier`, is this running in your local environment?

hackrush01:
Ok, thanks for the feedback, I will keep this in mind from next time onwards.

6ea86b96:
@hackrush01 `prettier` should be getting run when you `git commit` (https://github.com/lbryio/lbry-app/blob/master/ui/package.json#L8). If that's not working then you can still run manually from `ui` directory with `yarn run prettier`

kauffj:
Merged in master by hand so I could squash some additional fixes.

PR #301, opened 2017-06-28
Move client settings into redux

6ea86b96:
Toggling NSFW changes immediately now instead of after restart.


kauffj:
Nice work, thanks!

PR #305, opened 2017-06-29
Cancel resolving URis when unmounting discover page

6ea86b96:
Play button is now appearing with no delay again @kauffj 
cancel-resolve


kauffj:
See, cancelling was useful :)

PR #306, opened 2017-06-29
Add option to abandon claim when deleting published file

6ea86b96:
Just a quick idea. I got sick of constantly having to abandon my claims on the cli while testing publishing refactoring.

2017-06-29 at 15 42


kauffj:
Great stuff, this was on the punch list to add abandoning claims. I'd be happy to merge this as-is, though I also think it'd be good to separate this from deletion.

Or maybe this action isn't deletion at all? Maybe inside of the publish area the action should be called unpublish, with the main prompt being to abandon the claim. Deleting could be a sub-option inside of that.

Either way, I like it, but may try to tweak the UX a bit before merging. If you want to try it yourself, I won't be coming back for 3-4+ hours.

6ea86b96:
Maybe we should add this in the published files section as well as where it is in this PR? If someone deletes a file right now then it's not going to show up in published area anymore so there will be no way to abandon the claim. At least I think that's the case.

kauffj:
Part of what I'm saying is that "delete" is a poor option for content you've published, because it's unclear what it does. Deleting the file does nothing if it doesn't remove the claim and the content is rehosted, but it feels like the content is gone.

Instead, delete probably shouldn't even be an option for published content. It should be labeled "Unpublish", with the primary action being to revoke the claim, and deleting the file a secondary additional option.

kauffj:
I'd like to do more with this, but it's an improvement as-is for sure, so I merged it. 

I did switch all modal components to start with `Modal` (e.g. `ModalUpgrade` rather than vice versa) and split this modal out of `FileActions`.

I'd also like to start having each modal represented by a constant, I started this in `constants/modal_types`, but let me know if you think another approach is better.

kauffj:
I hate magic strings!

Issue #311, opened 2017-06-29
Space bar and video click should both play and pause video 


tzarebczan:
## The Issue
The space bar on my keyboard should pause & play the video. This is a universal with good media players, and is frustrating when it doesn't work.
### Steps to reproduce
1. Go to video, click spacebar
2. Nothing happens

### Expected behaviour
Video should pause/play

### Actual behaviour
Nothing

## System Configuration

 

- LBRY Daemon version:
- LBRY App version: 0.12.2RC9
- LBRY Installation ID:
- Operating system: Windows

PR #317, opened 2017-07-01
Issue #311 playback input improvements

akinwale:
- Pressing the space bar can initiate video playback
- Pressing the space bar can pause and resume video playback
- Clicking on the video area (above the video controls) after initial playback can pause and resume video playback

### DIFF
--- a/ui/js/component/video/internal/play-button.jsx
+++ b/ui/js/component/video/internal/play-button.jsx
@@ -4,10 +4,21 @@ import Link from "component/link";
 import Modal from "component/modal";
 
 class VideoPlayButton extends React.PureComponent {
+  componentDidMount() {
+    document.addEventListener("keydown", this.onKeyDown.bind(this));
+  }
+
   onPurchaseConfirmed() {
     this.props.closeModal();
     this.props.startPlaying();
     this.props.loadVideo(this.props.uri);
+  }
+
+  onKeyDown(event) {
+    if (event.keyCode === 32) {
+      event.preventDefault();
+      this.onWatchClick();
+    }
   }
 
   onWatchClick() {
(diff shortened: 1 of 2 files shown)


ui/js/component/video/internal/play-button.jsx
```diff
@@ -4,12 +4,23 @@ import Link from "component/link";
 import Modal from "component/modal";
 
 class VideoPlayButton extends React.PureComponent {
+  componentDidMount() {
+    document.addEventListener("keydown", this.onKeyDown.bind(this));
```
kauffj: this needs to be undone in unmount

ui/js/component/video/internal/play-button.jsx
```diff
@@ -4,12 +4,23 @@ import Link from "component/link";
...
+  }
+
   onPurchaseConfirmed() {
     this.props.closeModal();
     this.props.startPlaying();
     this.props.loadVideo(this.props.uri);
   }
 
+  onKeyDown(event) {
+    if (event.keyCode === 32) {
```
kauffj: Can we use a code or constant here? Possible https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code?

ui/js/component/video/internal/player.jsx
```diff
@@ -29,10 +29,11 @@ class VideoPlayer extends React.PureComponent {
     player.append(
       this.file(),
       container,
-      { autoplay: false, controls: true },
+      { autoplay: false, controls: true, enableKeyboard: true },
       renderMediaCallback.bind(this)
     );
 
+    document.addEventListener("keydown", this.togglePlay.bind(this));
```
kauffj: This needs to be unbound when unmounted.

akinwale:
@kauffj Pushed a new commit with the requested changes.

kauffj:
@akinwale changes look good, thanks! One issue I noticed while testing is that the "keydown" bind hijacks the ability to type space into the omnibar.

Either the event needs to be bound to an element deeper in the DOM, or we need to check if the keypress is happening inside of a form element and allow it to proceed rather than treating is a possible play press.

akinwale:
@kauffj Added a form control check in the new commit.

Issue #310, opened 2017-06-29
Volume control resets when you going from one video to another


tzarebczan:
## The Issue
When watching a video and adjusting the volume, the setting is lost when going to the next video. 
### Steps to reproduce
1. Watch Video 1
2. Adjust volume
3. Go to video 2, volume is back to 100%

### Expected behaviour
Volume should persist

### Actual behaviour
Volume resets to full

## System Configuration

 

- LBRY Daemon version:
- LBRY App version: 0.12.2RC9
- LBRY Installation ID:
- Operating system:

kauffj:
https://github.com/lbryio/lbry-app/commit/444f2308772f49b5209b21caa6a393e19c5c4958

PR #318, opened 2017-07-01
Issue #310 video volume reset fix

akinwale:
- Media player volume is persisted to local storage when the volume is changed by the user.
- The volume from local storage will be set as the media player start volume whenever the user tries to watch a new video. It defaults to the max volume if an invalid value was retrieved from local storage.

### DIFF
--- a/ui/js/component/video/internal/player.jsx
+++ b/ui/js/component/video/internal/player.jsx
@@ -42,7 +42,16 @@ class VideoPlayer extends React.PureComponent {
           once: true,
         }
       );
+      mediaElement.addEventListener("volumechange", () => {
+        localStorage.setItem("prefs_volume", mediaElement.volume);
+      });
+      mediaElement.volume = this.getPreferredVolume();
     }
+  }
+
+  getPreferredVolume() {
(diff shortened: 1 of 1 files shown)


6ea86b96:
Nice work @akinwale!

I think it would be cool to get this into the redux store. Something like this:

- on volume change dispatch a `VOLUME_CHANGED` event which updates the volume.
- a selector like `selectVolume` which is passed into the video component

then either
1) we store the volume in IndexedDB (https://github.com/lbryio/lbry-app/blob/master/ui/js/store.js#L89).
or
2) the `defaultState` of the `appReducer` is `{ volume: localStorage.getItem('volume') || 1 }` (and we make the `VOLUME_CHANGED` action also set localStorage.

What do you think @kauffj?

kauffj:
I agree with @6ea86b96 that this is not the React/Redux way to do this and would prefer a solution along the lines he outlined (though I prefer persisting the state rather than using localStorage).

That said, it's basically impossible for this code to add side-effects, and it seemed silly to delay this feature on the grounds that it should be done in 40 lines of code instead of 4, so I merged it.

I also switched it to just use the session rather than persist.

1dea560

Issue #308, opened 2017-06-29
Top menu not suppressed in full screen video mode


tzarebczan:
## The Issue
When playing a video in the app full screen, the Edit / Help / Developer menu bar is stiill shown. 
### Steps to reproduce
1. Go to claim with video (lbry://@IGNonLBRY/linked-ep5)
2. Click Play, go full screen
3. Menu bar visible

### Expected behaviour
Video should be in full screen

### Actual behaviour
Menu bar shown

## System Configuration

 

- LBRY Daemon version:
- LBRY App version: 0.12.2.rc9
![toolbar](https://user-images.githubusercontent.com/8120721/27690334-13838ff8-5caf-11e7-95f7-cebb97b3a2c7.png)

- LBRY Installation ID:
- Operating system: Windows 10

kauffj:
https://github.com/lbryio/lbry-app/commit/d8e867aa5ae8e9872b7943a0fd8593f9fc68d826

PR #319, opened 2017-07-02
Issue #308 hide menu in fullscreen mode on Windows

akinwale:
This hides the menu when a video enters full screen mode on Windows. I added the win32 platform check in the `webkitfullscreenchange` event handler since I was unable to reproduce the issue on Linux (Ubuntu 16.04), so I assumed it's probably Windows-specific.


kauffj:
Good stuff! Merging.

Issue #253, opened 2017-06-16
Seeking in a MP3 doesn't work until the file plays all the way through


benhohner:
## The Issue
Seeking in a MP3 (possibly other audio files too) doesn't work until the file plays all the way through

### Steps to reproduce
1. Download lbry://curvedspace-scatter
2. Play file
3. Try to seek, seeking fails
4. Finish listening to file
5. Seeking now works.

### Expected behaviour
Users should be able to seek as soon as they've downloaded the whole file (if not as the file is downloaded, but I know we have some issues with that ATM)

### Actual behaviour
User isn't able to seek until after the file plays all the way to the end.

## System Configuration
daemon (lbrynet)	0.11.0
wallet (lbryum)	2.7.25
interface	0.12.0
Platform	Windows (Windows-10-10.0.15063)
Installation ID	2BX4AhDAHSJoPeyAjAU24B5H6qgRPUgWf92uX3iwLMvxPHPZSq9Vw4FmyXgysLG2hx
- Operating system: Win 10 x64

6ea86b96:
@benhohner this is because in some files the metadata is at the end and so unavailable when streaming.

benhohner:
@6ea86b96 I'm aware there's that issue when streaming, but this happens after the entire file has already been downloaded in a previous browsing session, and I'm rewatching it.

6ea86b96:
@kauffj this one appears to be fixed.

PR #320, opened 2017-07-02
Issue #253 mp3 audio seek fix

akinwale:
Implemented a workaround to load the entire media file using xhr if an audio media element is detected after the metadata is loaded.

I published a test MP3 file to lbry://beyed.


6ea86b96:
What is that actual cause of this issue? It's only a subset of audio files that are not seekable afaik. Seems a bit unfortunate if we have to download the entire file for all audio files to fix the bad ones.

akinwale:
According to the issue at https://github.com/lbryio/lbry-app/issues/253, it appears that the player cannot seek through the file until it has finished playing at least once, which indicates that the length is not determined when the media file is first loaded. I can try to limit this to just files with the have the content type, "audio/mpeg3" or "audio/mpeg". Although it's not perfect, it limits the full load to just MP3 audio files.

kauffj:
I'd like confirmation that the [render-media](https://github.com/feross/render-media) plugin has this issue for all MP3s before merging this. @akinwale could you try confirming this issue exists with the plugin directly?

akinwale:
@kauffj From further testing, I tried adding the audio element to the media container directly using `downloadPath` for the `src` attribute, and the audio was seekable, so it looks like the problem could be with the render-media plugin, perhaps with the way the file stream is being created. Considering this is the case, I think we can handle the `audio/mpeg` content type differently by appending the HTML5 audio element directly, while every other type content type will be handled by render-media. What do you recommend?

akinwale:
@6ea86b96 I tried to use this but it doesn't work. The seekable time range end returns a value (which is less than the audio duration), but I am still unable to seek when I try to drag the control.

kauffj:
@akinwale Bypassing render-media and using the audio element directly sounds like a reasonable solution to me.

akinwale:
@kauffj Pushed a new commit.
* Added the `contentType` property to the `VideoPlayer` component.
* Added a `renderAudio` method which is used instead of render-media if the `contentType` is `audio/mpeg3` or `audio/mpeg`.
* Merged with master to resolve conflicts.

Issue #255, opened 2017-06-18
Purchase confirmation box - display issue


tzarebczan:
## The Issue
The message box that pops up after clicking download/play on paid content cuts off text if the URI is too long. Also, the message prompt is different when clicking the thumbnail vs the download button - clicking the thumbnail shows the title, not URI.  If Download is fixed to show Title, still need to ensure a longer title doesn't cause cut off issue. 

### Steps to reproduce
1.  Go to lbry://toughentheroots#9923383d3ded58658cf7bf510c23c5cf05db84dc
2.  Click download, msg box is cut off (see screenshot). 
3. Click cancel, then click thumbnail to stream, message box shows title as opposed to URI when clicking download

### Expected behaviour
Be consistent between both popups, resize box to text

### Actual behaviour
pops up have different text, pop up cuts off

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

## Screenshots

![msg box](https://user-images.githubusercontent.com/8120721/27257226-2cd68490-539c-11e7-96e8-5a18f7a8f5ad.PNG)
![msg box 2](https://user-images.githubusercontent.com/8120721/27257227-2cd771b6-539c-11e7-8303-ad4428c22bb6.PNG)

tzarebczan:
@akinwale was this fixed? needs a PR?

tzarebczan:
fixed by https://github.com/lbryio/lbry-app/pull/399

PR #321, opened 2017-07-02
Issue #255 purchase dialog confirmation content fix

akinwale:
- Added the `word-break: break-word` CSS style to the modal dialog to force break long overflowing text or URLs.
- Added a property called `mediaTitle` to the `FileActions` component so that the purchase confirmation dialog can display the title in cases where the `fileInfo` property or `fileInfo.metadata` may not be set.

### DIFF
--- a/ui/js/component/fileActions/view.jsx
+++ b/ui/js/component/fileActions/view.jsx
@@ -50,6 +50,7 @@ class FileActions extends React.PureComponent {
 
   render() {
     const {
+      mediaTitle,
       fileInfo,
       isAvailable,
       platform,
@@ -72,7 +73,7 @@ class FileActions extends React.PureComponent {
         ? __("Open in Finder")
         : __("Open in Folder"),
       showMenu = fileInfo && Object.keys(fileInfo).length > 0,
-      title = metadata ? metadata.title : uri;
+      title = mediaTitle ? mediaTitle : metadata ? metadata.title : uri;
 
     let content;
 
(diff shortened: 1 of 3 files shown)


kauffj:
What's the root cause of this bug? Why doesn't metadata.title have a value?

akinwale:
The `FileActions` component expects a `fileInfo` property which should contain the metadata. However, in the `FilePage` component at lbry-app/ui/js/page/filePage/view.jsx#118, `fileInfo` is not specified, hence no metadata to obtain the title from.

kauffj:
FileActions takes a uri, which is passed in both places. It looks to me that the issue is that it is trying to read metadata from file info, which will only exist if `get` has been called, instead of getting the metadata from the claim info.

kauffj:
Also, `break-word` is not a valid CSS property. https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

kauffj:
I'm going to close this. I'm fairly confident the correct fix is not to pass the title in, but to refactor the component logic so that it works properly. If you want to re-submit this, you can with valid CSS to fix the wrapping issue, or a refactor to properly get the title in.

akinwale:
@kauffj `word-break: break-word` is Webkit specific which gracefully handles long overflowing text and doesn't split regular words in the middle like `break-all` does. You can refer to https://miketaylr.com/posts/2014/01/word-break-break-word.html for more info on this.

PR #325, opened 2017-07-02
Added Language Downloading and Settings

filipnyquist:
Added basic downloading of languages from our language server and a dynamic settings panel which allows the user to change the language between the downloaded ones.

With the help of the awesome (which wrote parts of the original code, woop woop), @MaxiBoether !


ui/js/langs.js
```diff
@@ -0,0 +1,196 @@
+/*
```
kauffj: We're working to reduce top-level files, though there's not an obvious place for this. I guess `js/util` for now?

ui/js/main.js
```diff
@@ -130,4 +133,40 @@ var init = function() {
   }
 };
 
+const download = (url, dest, lang, cb) => {
```
kauffj: I'd like this to follow more of the standard React/Redux model that this rest of the app is using. You can model the existing settings logic (`page/settings/view.jsx`, `page/settings/index.js`, `selectors/settings.js`, `reducers/settings.js`, `selectors/settings.js`). If this isn't making sense, chat with me.

ui/js/page/settings/view.jsx
```diff
@@ -85,6 +85,12 @@ class SettingsPage extends React.PureComponent {
     this.props.setClientSetting("showNsfw", event.target.checked);
   }
 
+  onLanguageChange(language) {
+    lbry.setClientSetting('language', language);
+    window.app.i18n.setLocale(language);
+    this.setState({language: language})
```
kauffj: After the refactor, language would be a property, not a part of the state. And the language setting would happen inside of an action. In this model, components never change the state of the app directly, they only dispatch actions.

ui/js/langs.js
filipnyquist: I was thinking of adding it there, but placed it in its own, thanks for great idea, i´ll implent that later today

ui/js/main.js
filipnyquist: Sounds like a good plan, noticed and working on refactoring :+1:

filipnyquist:
Working on making it react/redux.

kauffj:
Closing in favor of #455.

Issue #312, opened 2017-06-29
Retain page scroll position on backwards navigation


tzarebczan:
## The Issue
The home page loses its place when you go back to it. If the home page had hundreds of options on it, it would be really annoying to have to scroll all the way down to wherever you were last browsing before you clicked on something.

### Steps to reproduce
1. Scroll down on home page, enter a video/media
2. Click back
3. Now you at back at the top of the home page

### Expected behaviour
Home page should return to where you were before going into media

### Actual behaviour
Home page goes back to top

## System Configuration

 

- LBRY Daemon version:
- LBRY App version: 0.12.2RC9
- LBRY Installation ID:
- Operating system:

akinwale:
Should only the vertical scroll position be maintained? Or should the horizontal scroll position for each individual category be maintained as well?

kauffj:
Vertical only

kauffj:
There is a PR for this.

PR #331, opened 2017-07-06
Issue #312 home page scroll position on back navigation

akinwale:
This was a bit difficult to figure out, since `componentDidMount` is meant to execute after `render` (based on what I have read about the React component life cycle), but for some reason, calling `window.scrollTo` directly in this method did not work. I also tried in the `componentDidUpdate` method. I ended up using a `setTimeout` to call the `scrollTo` method for the page to actually scroll to the previous position.


kauffj:
@akinwale Is there any way to use Chrome's native scroll position retention? I suspect not, since we're controlling navigation, but perhaps it's possible.

Additionally, I'm considering whether or not we want this on all pages. I think most users would expect "back" to take them to the position they were previously in (I understand the ticket didn't stipulate this.)

Finally, I'd like to avoid using the session for this in favor of the redux store.

kauffj:
@akinwale want to come back to this one? I thought about merging as-is, but at a minimum it needs to only happen on "back" events, not all navigations to Discover.

akinwale:
@kauffj I haven't yet figured out how to detect navigation back events in React, but I've made changes to make use of the Redux store. I'll push a new commit after I'm done.

kauffj:
@akinwale look at `doHistoryBack` in `ui/actions/app.js`

akinwale:
@kauffj Pushed a new commit.

ui/js/actions/app.js
```diff
@@ -64,6 +64,17 @@ export function doHistoryBack() {
     if (!history.state) return;
 
     history.back();
+    dispatch({
+      type: types.HISTORY_BACK,
```
kauffj: Is this supposed to be `HISTORY_BACK_COMPLETED`? If not, I don't think `doHistoryBackCompleted` ever happens. If so, I don't think we're guaranteed the render happens while `navigatingBack` is true.

The other relevant code to know here (if you haven't discovered it already) is the popstate event listener in `ui/js/main.js`. This is what history.back() is triggering.

ui/js/page/discover/view.jsx
```diff
@@ -37,10 +38,31 @@ const FeaturedCategory = props => {
...
+    if (this.props.isNavigatingBack) {
+      const scrollY = parseInt(lbry.getClientSetting("prefs_scrolly"));
+      if (!isNaN(scrollY)) {
+        const restoreScrollPosition = () => {
+          window.scrollTo(0, scrollY);
+        };
+        setTimeout(restoreScrollPosition, 100);
+      }
+
+      this.props.finishedNavigatingBack();
```
akinwale: @kauffj  This delegates to `doHistoryBackCompleted()`. I called it at this point because the `popstate` event was not being fired after clicking the back button. Also, `HISTORY_BACK` occurs before `CHANGE_PATH`, and this happens before `componentDidMount`. I needed a way to reliably detect that the user navigated using the back button and being able to check the flag when the component mounts was how I could do it (although I figure there might be a better way but I'm not seeing it at this point).

6ea86b96:
I think this can be simplified. We should be able to monitor the scroll position globally from the app component right? All back events are handled by `ui/js/main.js`. In there we can just pass a `scroll` variable to the `doChangePath` action to tell it whether to scroll to a previous position or not.

6ea86b96:
In fact actually we should just be able to store the scroll position in the history state and scroll ourselves, on all pages.

filipnyquist:
@6ea86b96 seems like a easier and better way to do it, any downsides?

6ea86b96:
@filipnyquist @kauffj I don't see any immediate downsides. I've implemented it here https://github.com/lbryio/lbry-app/pull/365

Now scrollY is retained over the full history.

Issue #330, opened 2017-07-03
Require minimum daemon version in app


lyoshenka:
Story: As a user, I want to know if the running daemon will not work with the app, so that I don't get weird bugs.

kauffj:
https://github.com/lbryio/lbry-app/commit/444f2308772f49b5209b21caa6a393e19c5c4958

tzarebczan:
Also related to: https://github.com/lbryio/lbry.io/pull/183

kauffj:
Fixed in https://github.com/lbryio/lbry-app/commit/38fd451e72c60b0373a8e337bb1de1da37f7942a

PR #335, opened 2017-07-07
Warn user if daemon version mismatches daemon bundled with app (WIP)

alexliebowitz:
Addresses issue #330. To work with dev builds, it needs this tweak to lbrynet: lbryio/lbry#763


build/DAEMON_VER
```diff
@@ -0,0 +1 @@
+0.13.1
```
kauffj: This is the type of progressive, positive refactoring that is great to see.

build/build.sh
```diff
@@ -79,11 +79,23 @@ if $OSX; then
...
 unzip "$BUILD_DIR/daemon.zip" -d "$ROOT/app/dist/"
 rm "$BUILD_DIR/daemon.zip"
 
+
+###################
+#   Config file   #
+###################
+
+cat << EOF > "$ROOT/app/config.json"
+{
```
kauffj: Is there a way to do this without adding another config file unique to tracking the daemon version? In this design, both config.json and DAEMON_VER exist to serve this simple purpose, when ideally there would be zero files unique to this purpose and it would be part of... something else (package.json?)

ui/js/app.js
```diff
@@ -2,7 +2,10 @@ import store from "store.js";
 import lbry from "./lbry.js";
 
 const env = ENV;
-const config = require(`./config/${env}`);
+const config = {
+  ...require(`./config/${env}`),
+  appConfig: require("../../app/config.json"),
```
kauffj: This seems over-engineered, isn't app/config.json a new file?

ui/js/component/incompatibleDaemonModal/index.jsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
```
kauffj: Please rename these to start with `modal`.

ui/js/component/app/view.jsx
```diff
@@ -32,6 +35,7 @@ class App extends React.PureComponent {
         <div id="main-content">
           <Router />
         </div>
+        {modal == "incompatibleDaemon" && <IncompatibleDaemonModal />}
```
kauffj: Please add to `config/modal_types`

ui/js/component/incompatibleDaemonModal/view.jsx
```diff
@@ -0,0 +1,26 @@
...
+        isOpen={true}
+        contentLabel={__("Incompatible daemon running")}
+        type="confirm"
+        confirmButtonLabel={__("Exit")}
+        abortButtonLabel={__("Continue")}
+        onConfirmed={quit}
+        onAborted={skipWrongDaemonNotice}
+      >
+        {__(
+          "An incompatible version of the LBRY daemon is running. If you continue to receive this message after starting LBRY again, try restarting your computer."
```
kauffj: Let's make this a link to a KB page, so we can manage the process and provide more information:

An incompatible version of the LBRY daemon is running. Please see [this article](https://lbry.io/faq/incompatible-daemon on how to fix this problem.

@tzarebczan please create a KB article at `lbry.io/faq/incompatible-daemon` on what users should do when their daemon is out of date (I think it's basically kill all processes and/or restart and if that doesn't fix re-install.)

app/main.js
```diff
@@ -27,7 +27,9 @@ const {version: localVersion} = require(app.getAppPath() + '/package.json');
 
 const VERSION_CHECK_INTERVAL = 30 * 60 * 1000;
 const LATEST_RELEASE_API_URL = 'https://api.github.com/repos/lbryio/lbry-app/releases/latest';
+const DAEMON_PATH = process.env.LBRY_DAEMON || path.join(__dirname, 'dist', 'lbrynet-daemon');
 
+global.daemonPath = DAEMON_PATH; // For easy access from render process
```
alexliebowitz: I'm going to scratch all the daemon-related code completely. I just didn't want to clean it all out until I was sure we're nixing the original approach of calling the daemon.

build/build.sh
alexliebowitz: The reason I left it as separate files is that they're conceptually different things. `config.json` is the configuration file for the *app*, which is *generated* by the build process. Some parts of `config.json` might even be dynamically generated eventually.

There would definitely be a case for making a `build_config.json` that in turn builds this file. We could also put all of the other build constants in there, like the URL format.

npm does let you put configuration in package.json. Some Node projects use that, some use a `config.json` file. `config.json` seemed more simpler and explicit, especially because we may end up needing to dynamically generate the file at build time.

On the other hand, I just looked into it some more and it turns out NPM lets you programmatically modify the configuration in package.json via `npm config`. It would be really nice if we could do that in the build script. I'll look into it.

ui/js/app.js
alexliebowitz: So there are 2 things about this that could make it feel over-engineered:

1. `config.json` is a separate file, and it's at the app level, not the UI level
2. We're importing it into the Redux config object as a sub-object

#1 is because conceptually these are settings that apply to the whole app, even though currently we only have a single one that's used in the UI. Even with this one setting, later on there could easily be app level stuff that wants to know the UI version.

#2 is just to make it easy to access app-level settings in the UI without having to require `config.json` every time. The Redux config object seemed like a logical enough place.

Hopefully if we switch to using `package.json`, that will make it easy to access config settings within the render process, and we won't need this. Not sure yet.

build/build.sh
kauffj: I think `npm` is dying in favor of `yarn`? Unfortunately, I do not understand the current packaging and config process very well.

alexliebowitz: I only half understand it myself. But it looks like yarn is supposed to be backwards compatible with npm and there is a `yarn config` command, so everything I just said goes for yarn too.

alexliebowitz:
Converted this to store in `package.json`, eliminating the two new files. I decided that `yarn config` isn't really right for this after all; the behavior is a little weird and it stores the settings in some hidden file. It's really more for package manager level settings.

I used a top-level key in `package.json`  instead of storing it in the `config` key. I've found examples of both approaches, but the top-level key approach is more popular from what I can tell.

ui/js/actions/app.js
```diff
@@ -202,6 +201,27 @@ export function doCheckUpgradeAvailable() {
...
 }
 
+export function doCheckDaemonVersion() {
+  return function(dispatch, getState) {
+    lbry.version().then(({ lbrynet_version }) => {
+      if (config.lbrynetDaemonVersion != lbrynet_version) {
+        dispatch({
+          type: types.OPEN_MODAL,
+          data: {
+            modal: "incompatibleDaemon",
```
kauffj: add to `constants/modal_types.js`

ui/js/actions/app.js
```diff
@@ -202,6 +201,27 @@ export function doCheckUpgradeAvailable() {
...
+          },
+        });
+      }
+    });
+  };
+}
+
+export function doSkipWrongDaemonNotice() {
```
kauffj: I'm inclined to just not allow this to be skipped. If the wrong daemon version is running, something went wrong with your install.

ui/js/component/modalIncompatibleDaemon/view.jsx
```diff
@@ -0,0 +1,26 @@
...
+        isOpen={true}
+        contentLabel={__("Incompatible daemon running")}
+        type="confirm"
+        confirmButtonLabel={__("Exit")}
+        abortButtonLabel={__("Continue")}
+        onConfirmed={quit}
+        onAborted={skipWrongDaemonNotice}
+      >
+        {__(
+          "An incompatible version of the LBRY daemon is running. If you continue to receive this message after starting LBRY again, try restarting your computer."
```
kauffj: "This browser is running with an incompatible version of the LBRY protocol and your install must be repaired."

[Quit and Learn More]

.
.
.
.

(or if that's bad UX, maybe:)

[Continue]

&lt;meta&gt;Clicking continue will close the app and provide further instructions.&lt;/meta&gt;

kauffj: Either way, this should be a link to lbry.io/faq/incompatible-protocol-version

kauffj:
Merged https://github.com/lbryio/lbry-app/commit/38fd451e72c60b0373a8e337bb1de1da37f7942a

PR #341, opened 2017-07-09
Update _card.scss

98farhan94:
Added transition to cards for smoother experience.

### DIFF
--- a/ui/scss/component/_card.scss
+++ b/ui/scss/component/_card.scss
@@ -126,6 +126,7 @@ $height-card-small: $spacing-vertical * 15;
     vertical-align: top;
     display: inline-block;
     margin-right: $spacing-vertical / 3;
+    transition: all .1s ease-in-out;
   }
   + .card-row {
     margin-top: $spacing-vertical * 1/3;


kauffj:
Good idea to add more transitions! However, I would like to avoid transition all. What effects actually need transitions? 

Additionally, there is a `$transition-standard` defined in `_global.scss`, I'd like to continue using this. I'd be fine with changing it from `0.225s ease`, as I don't think that was an especially considered choice.

Smurphonomics:
This works well

`section.card.card--small.card--link:hover {
    z-index: 2;
    -webkit-transition: all 120ms ease-in;
    -webkit-transform: scale(1.1);
    -ms-transition: all 120ms ease-in;
    -ms-transform: scale(1.1);
    -moz-transition: all 120ms ease-in;
    -moz-transform: scale(1.1);
    transition: all 120ms ease-in;
    transform: scale(1.1);
}`

kauffj:
@Smurphonomics we are always in a current version of chrome in this app, so no need for additional browser rules. And I would like to avoid `transition: all`.

98farhan94:
I think we can just do:

` transition: transform .1s ease-in-out;`c

The scale transform feels very abrupt right now, that's all I wanted to make smoother.

kauffj:
https://github.com/lbryio/lbry-app/commit/7c380ce01b7e9d45399f993a4ade39f9a4adce9d

Smurphonomics:
I saw cross browser SCSS in the app code, so I assumed it was being coded for a web based version also?

A good idea to remove "ALL", as it can have performance issues with it containing hidden code you don't need, catering for ALL properties.

@98farhan94 transition: transform .1s ease-in-out; is much cleaner :)

Smurphonomics:
@98farhan94 For even cleaner and more optimised code, we can omit the "transform" part as everything has already been declared. 

`` 
.card-row > .card {transition: .1s ease-in-out;}

98farhan94:
@Smurphonomics Okay, I'll append it in a new merge.

Smurphonomics:
Nope. Forget it. I thought omitting ALL would mean it just used the animation on the scale hover event. But it looks like omitting ALL just sets it to the initial state, which is ALL :(

So keep it to `transition: transform .1s ease-in-out;` after all :)

kauffj:
Yep, that's basically what we added @Smurphonomics.

Re: cross-browser, it's from the pre-Electron days and should probably be removed.

PR #346, opened 2017-07-10
Publishing Take 2

6ea86b96:


ui/js/component/common.js
```diff
@@ -42,6 +44,40 @@ export class TruncatedText extends React.PureComponent {
   }
 }
 
+export class TruncatedMarkdown extends React.PureComponent {
```
kauffj: Probably merits its own component?

ui/js/component/form.js
```diff
@@ -38,6 +40,12 @@ export class FormField extends React.PureComponent {
     } else if (this.props.type == "text-number") {
       this._element = "input";
       this._type = "text";
+    } else if (this.props.type == "SimpleMDE") {
+      this._element = SimpleMDE;
+      this._type = "textarea";
+      this._extraElementProps.options = {
+        hideIcons: ["guide", "heading", "image", "fullscreen"],
```
kauffj: I don't think this needs to change now, and I know this wasn't your work, but this is a pattern worth reconsidering before continuing it further.

ui/js/component/publishForm/internal/channelSection.jsx
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  handleNewChannelBidChange(event) {
+    this.setState({
+      newChannelBid: event.target.value,
+    });
+  }
+
+  handleCreateChannelClick(event) {
+    if (this.state.newChannelName.length < 5) {
```
kauffj: I'm inclined to drop this restriction (https://github.com/lbryio/lbry/issues/769). But I suppose we ought to keep it for now, especially since the error the daemon gives is unclear.

ui/js/component/publishForm/internal/channelSection.jsx
```diff
@@ -0,0 +1,174 @@
...
+      creatingChannel: true,
+    });
+    const success = (() => {
+      this.setState({
+        creatingChannel: false,
+        addingChannel: false,
+        channel: newChannelName,
+      });
+      this.props.handleChannelChange(newChannelName);
+    }).bind(this);
```
kauffj: Are these binds if necessary? The `() => { }` syntax automatically keeps the `this` context.

ui/js/component/publishForm/view.jsx
```diff
@@ -0,0 +1,924 @@
...
+      this.claim() === undefined
+    ) {
+      return __("Checking...");
+    } else if (!this.state.name) {
+      return __("Select a URL for this publish.");
+    } else if (!this.claim()) {
+      return __("This URL is unused.");
+    } else if (this.myClaimExists() && !this.state.prefillDone) {
+      return (
+        <Notice>
```
kauffj: The `Notice` component was only used here and the style does not match the application. I dropped it.

ui/js/page/filePage/view.jsx
```diff
@@ -119,7 +120,11 @@ class FilePage extends React.PureComponent {
               </div>
             </div>
             <div className="card__content card__subtext card__subtext card__subtext--allow-newlines">
-              {metadata && metadata.description}
+              <ReactMarkdown
+                source={(metadata && metadata.description) || ""}
+                escapeHtml={true}
+                disallowedTypes={["Heading", "HtmlInline", "HtmlBlock"]}
```
kauffj: The dual view button needs to be added to this list.

ui/js/component/publishForm/internal/channelSection.jsx
```diff
@@ -0,0 +1,174 @@
...
+    this.state = {
+      newChannelName: "@",
+      newChannelBid: 10,
+      addingChannel: false,
+    };
+  }
+
+  handleChannelChange(event) {
+    const channel = event.target.value;
+    if (channel === "new") this.setState({ addingChannel: true });
```
6ea86b96: Braces aren't required if there's only one statement. So

```
if (something) ...
else if (somethingElse) ...
else ...
```

is valid

ui/js/component/publishForm/internal/channelSection.jsx
6ea86b96: They are not necessary :)

ui/js/constants/action_types.js
```diff
@@ -64,6 +64,15 @@ export const FETCH_AVAILABILITY_COMPLETED = "FETCH_AVAILABILITY_COMPLETED";
 export const FILE_DELETE = "FILE_DELETE";
 export const ABANDON_CLAIM_STARTED = "ABANDON_CLAIM_STARTED";
 export const ABANDON_CLAIM_COMPLETED = "ABANDON_CLAIM_COMPLETED";
+export const FETCH_CHANNEL_LIST_MINE_STARTED =
+  "FETCH_CHANNEL_LIST_MINE_STARTED";
+export const FETCH_CHANNEL_LIST_MINE_COMPLETED =
```
6ea86b96: Good call, I like it.

ui/js/actions/content.js
```diff
@@ -339,3 +340,68 @@ export function doFetchClaimListMine() {
...
+          channelClaim => {
+            channelClaim.name = name;
+            dispatch({
+              type: types.CREATE_CHANNEL_COMPLETED,
+              data: { channelClaim },
+            });
+            resolve(channelClaim);
+          },
+          err => {
+            resolve(err);
```
6ea86b96: oops, yes it should.

ui/js/constants/action_types.js
6ea86b96: maybe `_SUCCEEDED`

ui/js/page/filePage/view.jsx
6ea86b96: dual view button?

6ea86b96:
Sorry @kauffj, I accidentally force pushed over a couple of changes you made :(

ui/js/page/filePage/view.jsx
kauffj: The right most button that shows both editor and preview.

kauffj:
@6ea86b96 np re: force push

ui/js/component/publishForm/view.jsx
6ea86b96: ok, dropped.

ui/js/page/filePage/view.jsx
6ea86b96: I don't see that in the list of possible types to disallow.

2017-07-12 at 14 14

6ea86b96:
I've added some extra code to disable the submit button if the bid is not higher than the current claim @kauffj

ui/js/page/filePage/view.jsx
6ea86b96: However, I have now noticed that the dual view mode in the publish form is horribly broken.

6ea86b96: For now I've just disabled `side-by-side` mode.

ui/js/component/form.js
6ea86b96: Definitely need to think about a better solution.

6ea86b96:
Found and fixed one other bug where if you overwrite an existing claim the old fileinfo shows up in downloaded files.

6ea86b96:
One more bug I found but have not fixed though.

- publish a claim
- update that claim
- delete the claim
- now the fileinfo from the first publish is displaying in the downloaded files

This is because even though the claim was abandoned the daemon is still returning the first file from the `file_list` command (but not returning the claim anymore in `claim_list_mine`).

Issue #333, opened 2017-07-06
Switching from one media type to another using the omnibar does not work properly


akinwale:
## The Issue
Switching directly from a video to audio, or vice versa after the media has started playing results in the previous media still playing, while the new media does not load.

### Steps to reproduce
1. Open a video that has been downloaded and then play the video.
2. Type the URL for an audio file that has been downloaded and press Enter. You have to input all of it at once before the search-as-you-type timeout kicks in.
3. These steps can also be reversed by playing an audio file first, then opening a video.

### Expected behaviour
The previously loaded media should stop playing and the play button for the new media should be displayed.

### Actual behaviour
The previously loaded media keeps playing while the new media does not load.

## System Configuration
- LBRY Daemon version: 0.13.1
- LBRY App version: 0.13.0
- LBRY Installation ID: 82AH5JjzwZRxw6KreYehogwMipUDSyVsiMXXefnRKFmnKjQJyhN2tF16KmwwJTuyGm
- Operating system: Windows (Windows-10-10.0.15063)

## Screenshots
- Switching from video to audio: https://www.dropbox.com/s/f2jfbsf1xlgfkg1/lbry-video-to-audio.png?dl=0
- Switching from audio to video: https://www.dropbox.com/s/0hgwak7j992cifm/lbry-audio-to-video.png?dl=0

tzarebczan:
I'm pretty sure I  had this happen to me. Wrote down the steps but didn't get a chance to reproduce and document, nice job!

tzarebczan:
@akinwale this was fixed right?

akinwale:
Yes, it was.

PR #348, opened 2017-07-12
Issue #333 media switch fix

akinwale:
Fix for switching between audio and video media types after playback has started.

I had to set the state's `isPlaying` property directly in the `componentWillReceiveProps` method, since I tried to make use of `setState` which resulted in a no-op error: `Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op.`

### DIFF
--- a/ui/js/component/video/view.jsx
+++ b/ui/js/component/video/view.jsx
@@ -14,6 +14,13 @@ class Video extends React.PureComponent {
     };
   }
 
+  componentWillReceiveProps(nextProps) {
+    // reset playing state upon change path action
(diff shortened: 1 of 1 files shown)


kauffj:
@akinwale I believe this should check if the deterministic property (fileInfo.download_path? fileInfo.outpoint?) has changed before setting the state. Theoretically this component could re-render without the file changing.

akinwale:
@kauffj Pushed a new commit to address this. Made use of outpoint since this is guaranteed to not change.

Issue #351, opened 2017-07-16
No way to view more than 10 claims of a channel.


hackrush01:
## The Issue
There is no next button when when resolving a channel in the app.

### Steps to reproduce
1. Open app
2. Search for lbry://@UCBerkeley and then press enter
3. Scroll to the bottom

### Expected behaviour
A Next/Previous button for navigating between the pages to view all of content.

### Actual behaviour
Only 10 claims can be seen(whereas there should be around 2815)

## System Configuration

- LBRY Daemon version: 0.14.1
- LBRY App version: latest clone from github

## Screenshots
http://imgur.com/a/mmdNW

PR #354, opened 2017-07-17
Add pagination to channel pages

6ea86b96:
pagination

- [x] Switched channel claims to lookup from `claims.byId`
- [x] Added pagination
- [x] Added ability to use pagination params in the URL

For some reason the container doesn't auto resize though and I haven't been able to figure out why yet.


hackrush01:
@6ea86b96 @kauffj @lyoshenka 

I have some suggestions regarding this.
1. Instead of auto-scrolling it should be paginated, as infinite scrolling is a bit tedious if you have to find some thing for the n<sup>th</sup> time. So there should be page numbers at the bottom for navigation.
2. It should append some prompt on the address bar such as `lbry://url&page=3` or `lbry://url:3` so that user can type in that and directly reach to that page, or share the page with people.
3. (Super Optional) Maybe the user could set the no. of results to be displayed on one page?

P.S. #351 is dependent on this.

6ea86b96:
Good point. Pagination would be a lot more usable.

kauffj:
@6ea86b96 did you want me to review/merge this as-is? I got the impression you were planning to come back to it.

6ea86b96:
@kauffj Removed you as a reviewer for now. I like the pagination suggestion. I'm going to come back to this after the channel surfing rewards fix.

6ea86b96:
Added pagination, just need to get the css working and it's good to go.

6ea86b96:
@kauffj pagination is working now. Doesn't look that amazing. CSS is not my strong point :)

hackrush01:
@6ea86b96 I was just trying it out, but I found one issue, it seems to work with the provided version of the lbrynet-daemon but not with the latest version(with the latest lbryum and schema). One possible reason I've singled out is that it may be happening due to [this PR](https://github.com/lbryio/lbryum/pull/138). Can you please check it out once?

kauffj:
@hackrush01 it worked fine for me after I rebased onto master. 

@6ea86b96 merged. Added my own CSS, which I also don't like. I did clean up the class names, though.

PR #357, opened 2017-07-17
[WIP] Discover page UI tweaks

akinwale:
- Reduced the card size
- Reduced the font size of the card titles
- Reduced the font size of the card description
- Moved the title, publisher and price below the card thumbnail.


kauffj:
@akinwale maybe it's just that these changes grew on me, but I decided to merge. See https://github.com/lbryio/lbry-app/commit/53168074e09b855c515ef09ff72a6154ace56555 for some additional changes I made. I'd still like to be more careful about haphazardly introducing additional colors and font-sizes. I also try to avoid nesting CSS rules whenever possible, though this version still does that.

PR #362, opened 2017-07-19
further language improvements

reillysmith:
extended tool tip with examples for content URL


kauffj:
Fixed in https://github.com/lbryio/lbry-app/commit/4632c4f919094045d83f5a1e320d3db397bf8faa

@reillysmith note that there are no spaces between blocks of text, so if you add two lines like:

{ __("blah blah blah") } 
{ __("even more blah") }

There will be no space between the two blocks unless you add `{ " " }`.

PR #363, opened 2017-07-19
move channel to top

reillysmith:
added suggestion: if last @channelname is true, then autofill, else fill with "anonymous"


kauffj:
I tried this, and I'm not convinced this will improve the UX. If we want to put channel at the top, we need to change the copy too. Will take another look at this later.

kauffj:
Closing this. Requires more nuanced changes. @reillysmith please keep advocating for this.

PR #366, opened 2017-07-20
Add identity verification to app

kauffj:


kauffj:
I'm planning to review this more myself, and just skimming the PR there are a few things I ought to change, but this is 95% ready.

PR #369, opened 2017-07-21
Add access token to help page

6ea86b96:
2017-07-21 at 13 23

Not sure if this is the best place to add it?


kauffj:
Nice touch adding the show toggle, btw @6ea86b96

PR #370, opened 2017-07-21
Track downloading files by outpoint

6ea86b96:
Now we have one unique place to track downloading files, instead of by URI. This fixes files in the downloaded file list not showing download progress along with some other downloading weirdness.

### DIFF
--- a/ui/js/actions/file_info.js
+++ b/ui/js/actions/file_info.js
@@ -10,8 +10,11 @@ import {
   selectIsFetchingFileList,
   selectFileInfosByOutpoint,
   selectUrisLoading,
+  selectTotalDownloadProgress,
 } from "selectors/file_info";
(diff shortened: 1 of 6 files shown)


ui/js/actions/file_info.js
```diff
@@ -112,12 +114,16 @@ export function doDeleteFile(outpoint, deleteFromComputer, abandonClaim) {
       }
     }
 
-    dispatch({
-      type: types.FILE_DELETE,
-      data: {
-        outpoint,
-      },
-    });
+    setTimeout(() => {
```
kauffj: This seems like a hack. Can the update to the progress instead be based on an event?

6ea86b96: Figured out a more sane workaround.

PR #375, opened 2017-07-22
The URL name is automatically suggested.

### LINKED ISSUE #347: Auto-suggest URL from selected file on Publish
The name field on the Publish page can be auto-populated by the file name sans extension.

hackrush01:
Fixes #347


hackrush01:
I found use of `var` in L106, L107 and L206 also, should that be changed to `let/const` also?

kauffj:
Merged. Yes, @hackrush01, you can replace `var` with `const` or `let` as appropriate. The app was started prior to using ES2016, so I'm sure there are plenty of `var` remaining. However, it's not safe to just do a find and replace, since I believe `let`/`const` use block level scoping, while var is function level. I doubt we're relying on function level variable scoping, but I still wouldn't global find/replace without inspection, just in case.

PR #378, opened 2017-07-22
Check if daemon version is greater than or equal to recommended, and modal edit.

hackrush01:
Some quickfixes which enhance the user/developer experience.

### DIFF
--- a/ui/js/actions/app.js
+++ b/ui/js/actions/app.js
@@ -224,7 +224,7 @@ export function doCheckDaemonVersion() {
   return function(dispatch, getState) {
     lbry.version().then(({ lbrynet_version }) => {
       dispatch({
-        type: config.lbrynetDaemonVersion == lbrynet_version
+        type: lbrynet_version.localeCompare(config.lbrynetDaemonVersion) >= 0
           ? types.DAEMON_VERSION_MATCH
           : types.DAEMON_VERSION_MISMATCH,
       });
--- a/ui/js/actions/app.js
+++ b/ui/js/actions/app.js
@@ -280,7 +280,13 @@ export function doClearCache() {
   };
 }
 
-export function doQuitAndLaunchDaemonHelp() {
+export function doQuit() {
+  return function(dispatch, getState) {
+    remote.app.quit();
+  };
+}
+
+export function doLaunchDaemonHelp() {
   return function(dispatch, getState) {
     shell.openExternal("https://lbry.io/faq/incompatible-protocol-version");
     remote.app.quit();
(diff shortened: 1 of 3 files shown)


kauffj:
This is a good change generally, but I'm not convinced we want to allow greater version numbers on the daemon. If the app is shipped linked to a specific daemon version, then higher daemon versions are not necessarily backwards compatible, and could have unforeseen consequences.

hackrush01:
I did consider that scenario, but IMO that is going to happen rarely, so for only those releases we would make it an equality check(one line modification),  like the way it is done in other softwares.
Does this sound good?

tzarebczan:
Or have it be a setting, default value to equals. Maybe "developer mode"?

hackrush01:
Yes, I always thought a developer mode was long due(that could be done using a `--dev` flag), but I am not sure how many changes would a step like this require. But surely that would streamline the process for devs as well as users. But in the interim some sort of workaround could be implemented.

kauffj:
Can you help me see the use case for a developer testing a higher version number of the daemon, but not wanting to update the packaging settings? Changing the daemon version is easy and being strict ensures that the developer is always coding against the daemon they intend to be.

tzarebczan:
Recently with the various daemon big fixes, I've been using updated daemons in my Windows app installs in order to test and verify. As a windows user, it's nice to be able to drag and drop the daemon files for quick testing. Otherwise I would need to build from source or have the devs issue an app update along with daemon. 

Also, from a support perspective, it would be nice to have a quick solution for users to test that a certain bugfix is working in a later daemon. Although this shouldn't happen frequently, it's convenient. 

All of this assumes that the higher level daemon does not break anything app related.

kauffj:
@tzarebczan part of my hesitancy here is that a higher daemon version is basically _promising_, or at least strongly hinting, that it will break something app related.

However, you raised some good points, and I could see how testing a higher daemon via the app could be a quick way to assess potential issues.

Will make a call here soon. 

(Also, I did notice on second glance that the version check is a straight string check, this also probably needs to change.)

hackrush01:
I think this can be closed as allowing for a new daemon version could create problems even for devs, so a manual allowance should only be done (if required), as pointed out by @kauffj. The second commit was broken into a new PR [here](https://github.com/lbryio/lbry-app/pull/411).

kauffj:
I'm closing this since the modal changes got merged in the other PR.

@tzarebczan raised a good point, so I'm fine with adding this. But the following would need to change:

1) It has to be a proper version check. String checking on version numbers can fail ("0.10" < "0.9").
2) It has to give some kind of indicator or acknowledgement that you are running a future daemon. It shouldn't be possible to do silently.

PR #381, opened 2017-07-24
Display the download button for files that are not downloading and have 0 progress

6ea86b96:
Seems to be a problem where some files start to download but never receive any data. Then the "open" link appears when viewing them after the app has been restarted. This switches it to display the download link still.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
   * Fixed files on downloaded tab not showing download progress
   * Fixed downloading files that are deleted not being removed from the downloading list
   * Fixed download progress bar not being cleared when a downloading file is deleted
+  * Fixed app thinking downloads with 0 progress were downloaded after restart
 
 ### Deprecated
   *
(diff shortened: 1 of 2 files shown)


kauffj:
I'm merging this because it seems better than the alternative, but both still seem wrong. If the stream descriptor blob has been downloaded, then the daemon will continue to try to download the file without any additional actions by the user. This should really be sitting at 0%, I think.

6ea86b96:
@kauffj agree, I didn't write a note as it was like 1:30am when I pushed this. It seemed like the simplest possible quick fix. Perhaps we should check if `fileInfo.written_bytes < fileInfo.total_bytes && !downloading` when this component is mounted/receives props. If the above is true then add it to the downloading files and restart the update download progress timeout code.

kauffj:
@6ea86b96 it seems like we could possibly have the `FILE_LIST_SUCCEEDED` reducer update the downloading status (or directly read the downloading status directly from all `FileInfo`s)?

Either way, my inclination would be to see if we could get to the proper state from the initial `file_list` call, rather than re-checking on receive props.

PR #387, opened 2017-07-26
Settings fix

6ea86b96:
Currently reloading the settings page blows up the app trying to access `daemonSettings.max_key_fee.amount` before the settings have been loaded.


kauffj:
Merged. Though I do support defaulting fetchable objects to undefined rather than empty objects, I think it makes the state clearer.

Issue #355, opened 2017-07-17
Tipping button in pages.


hackrush01:
Story: As a publisher I would like if the content page has a tipping button, so that people could send me some LBCs without first resolving my content uri, then searching for my address, and then tipping me either via cli or UI.

kauffj:
The tip can be sent to the same address of the payment, or the address of the claim if no payment details are specified.

tzarebczan:
Possibly related to https://github.com/lbryio/lbry-app/issues/247 based on team discussions.

tzarebczan:
I forget where the conversation took place, but we mentioned using claim supports for tipping/karma that then could be redeemable (maybe a timelock feature?). 

Just wanted to mention it here so the idea doesn't get lost. The duplicate claim_id support issue would need to be fixed first (https://github.com/lbryio/lbry/issues/795)

kauffj:
I'm guessing the first iteration of tips will just be sends to the address since I don't think issuing tips as supports is possible via the existing LBRY API. (Please correct this @jackrobison)

tzarebczan:
When we tested tipping, the tipped amount shows up in wallet but gets inserted as a support, which is not spendable.  This issue may need to be expanded so that it's clearer to the users what is happening (WHERE IS MY LBC??!?, WHY IS MY BALANCE INCORREC?T?!?): https://github.com/lbryio/lbry-app/issues/379

Also, claim_abandon support: https://github.com/lbryio/lbry/issues/795 will be needed to fully take advantage of the tips.

hackrush01:
After having a discussion on hangouts regarding the working of the `send_claim_tip` command, there were several other issues that prevented this from working properly. One of which is pointed in the last comment. As for the solution part, it would need a new kind of wallet and better claim management system to get it working properly. Right now the feature is not working in the way it was intended in the daemon itself. So this is more of the daemon issue but that will be blocking he PR here.

If we want this feature to be out there now, then it has to be done in the more straight-forward way, like a direct wallet transaction.

> tl;dr `send_claim_tip` requires deeper daemon changes to work properly. A basic tip/support implementation is possible.

tzarebczan:
@jackrobison may want to chime in here, but this was what we discussed on the call this morning.  If we want tipping ASAP, it just needs to be a simple LBC send tx as opposed to the new supports mechanism.

kauffj:
I caught up on this and discussed with Jack. I think we can still move forward with this. @hackrush01 let's sync on Slack.

PR #391, opened 2017-07-26
Add tipping feature to app.

hackrush01:
- [x] Button add to `FileActions` component, from a new component file.
- [x] Use icon `gift`
- [x] Once clicked, the user is displayed a simple form with a single input for LBC, and "Confirm" / Cancel buttons (confirm should affirm the action, e.g. "Send Tip")
- [x] The forms should include some help copy, possibly with a link to a FAQ.
- [x] The initiated action should be shown inline
- [x] Add snack bar for successful tip.
- [x] Confirm that sending transaction is not causing weird behaviour(checked as no problem was caused during preliminary testing).
- [x] Change modals to use constants.
- [x] Hide Nearby buttons when tip form is shown.
- [ ] Treat myself at KFC once merged.

For issue #355 

New list copied from below to track
- [x] Rework tipping to use the `wallet_send` API command
- [x] Render tipping UI in `FilePage` instead of `FileActions`
- [x] ~Extract `Wallet_Send` form and use that for tipping instead of the `FormFieldPrice` as one can tip in LBC only~
- [x] Something with controlled and uncontrolled components
- [x] ~Borrow things from `tip_and_supports` branch?????? (Not at sure what to do with that)~
- [x] Display a better `Transaction List` with columns for types of transaction(Support, claim, update, unbound), a link to claim if applicable(this might need some changes depending on how lbryio/lbryum#148 goes)
- [x] Use a better name instead of "Unbound" transaction. **Using "Unfiltered"**
- [x] Do a whole lot of conflict fixing in the end.


ui/js/component/publishForm/view.jsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import lbry from "lbry";
 import lbryuri from "lbryuri";
-import { FormField, FormRow } from "component/form.js";
+import { FormField, FormRow } from "component/form";
```
hackrush01: Changed for consistency

ui/js/component/fileActions/view.jsx
```diff
@@ -206,6 +277,36 @@ class FileActions extends React.PureComponent {
             outpoint={fileInfo.outpoint}
             title={title}
           />}
+        {modal == "insufficientBalance" &&
```
kauffj: We've also changed the way modals work to use constants. I also think this code was copy and pasted, which is a sign it ought to be refactored. These should probably be moved into `app/view.jsx`.

You can see the FIRST_REWARD modal here for reference https://github.com/lbryio/lbry-app/commit/29553bc391e594c0d918179345cf1e33a38c9c40

ui/js/page/filePage/view.jsx
```diff
@@ -116,7 +116,7 @@ class FilePage extends React.PureComponent {
                   : uriIndicator}
               </div>
               <div className="card__actions">
-                <FileActions uri={uri} />
+                <FileActions uri={uri} claim={claim} />
```
kauffj: Shouldn't this should be obtainable from the URI without being passed in?

ui/js/component/fileActions/view.jsx
```diff
@@ -83,6 +119,40 @@ class FileActions extends React.PureComponent {
...
+    let tipLink = (
+      <Link
+        label={__("Tip Publisher")}
+        button="text"
+        icon="icon-gift"
+        onClick={this.handleTipPublisherButtonClicked.bind(this)}
+      />
+    );
+
+    let tipBox = (
```
kauffj: All of the tip box logic should be it's own component.

ui/js/component/fileActions/index.js
```diff
@@ -48,6 +53,9 @@ const perform = dispatch => ({
   startDownload: uri => dispatch(doPurchaseUri(uri, "affirmPurchase")),
   loadVideo: uri => dispatch(doLoadVideo(uri)),
   restartDownload: (uri, outpoint) => dispatch(doStartDownload(uri, outpoint)),
+  sendToAddress: () => dispatch(doSendDraftTransaction()),
+  setAmount: amount => dispatch(doSetDraftTransactionAmount(amount)),
```
kauffj: If these source values are shared across forms, could this cause any weird behavior?

hackrush01: I'm unsure. This might need some testing/confirmation from your side. Although to make an educated guess, we _are_ dispatching states in `doSendDraftTransactions`, so it must be to avoid these weird behaviour. But I am not sure if that's how it works.

ui/js/page/filePage/view.jsx
hackrush01: It is(by resolving it). But I think re-resolving the same URL, would be a useless overhead, when we could simply use an existing data-set. This also helps the app to remain fast and fluid in its operation. I suggest keeping this as is. What's your stance on this?

ui/js/component/fileActions/view.jsx
hackrush01: I will look into it.

ui/js/component/fileActions/view.jsx
hackrush01: Ok. I'll do it.
That's how it was originally done(the modal one), but I got confused with "add button to `fileActions`" part. A slight misinterpretation on my part.

hackrush01:
Updated task list to reflect the review.

ui/js/actions/wallet.js
```diff
@@ -85,13 +86,13 @@ export function doSendDraftTransaction() {
         dispatch({
           type: types.SEND_TRANSACTION_COMPLETED,
         });
-        dispatch(doOpenModal("transactionSuccessful"));
+        dispatch(doOpenModal(modals.TRANSACTION_SUCCESSFUL));
```
kauffj: Let's drop this modal and replace with a snackbar showing:

You sent ${amount} LBC. [History]

History is a link to the Transaction History page.

ui/js/component/tipLink/view.jsx
```diff
@@ -0,0 +1,124 @@
...
+    });
+  }
+
+  render() {
+    const { modal, closeModal } = this.props;
+    const { showTipBox } = this.state;
+
+    let tipLink = (
+      <Link
+        label={__("Tip Publisher")}
```
kauffj: Just "Tip"

kauffj: And move inside of the "..." menu button

ui/js/component/tipLink/view.jsx
```diff
@@ -0,0 +1,124 @@
...
+          button="alt"
+          onClick={this.handleTipCancelButtonClicked.bind(this)}
+        />
+      </span>
+    );
+
+    return (
+      <div className="menu-container">
+        {showTipBox ? tipBox : tipLink}
+        {modal == "insufficientBalance" &&
```
kauffj: All of this modal code can be dropped from here now.

ui/js/component/tipLink/view.jsx
```diff
@@ -0,0 +1,124 @@
...
+          <Modal
+            isOpen={true}
+            contentLabel={__("Insufficient balance")}
+            onConfirmed={closeModal}
+          >
+            {__(
+              "Insufficient balance: after this transaction you would have less than 1 LBC in your wallet."
+            )}
+          </Modal>}
+        {modal == "transactionSuccessful" &&
```
kauffj: Turn this into a snackbar (mentioned elsewhere).

ui/js/page/filePage/view.jsx
kauffj: Are you sure an actual second resolve is triggered? I.e. is an API call actually made or are you just seeing the action dispatch?

Either way, the claim shouldn't be passed. If we are doing an unnecessary extra resolve, that should be fixed elsewhere.

ui/js/component/tipLink/view.jsx
```diff
@@ -0,0 +1,124 @@
...
+        label={__("Tip Publisher")}
+        button="text"
+        icon="icon-gift"
+        onClick={this.handleTipPublisherButtonClicked.bind(this)}
+      />
+    );
+
+    let tipBox = (
+      <span>
+        <FormField
```
kauffj: An LBC/USD input now appears:
- Here
- On Publish
- On Settings

It should probably just be turned into it's own component.

ui/js/component/fileActions/view.jsx
```diff
@@ -180,6 +181,7 @@ class FileActions extends React.PureComponent {
               />
             </DropDownMenu>
           : ""}
+        <TipLink claim={this.props.claim} />
```
kauffj: This should hide everything else in this component when activated.

ui/js/actions/wallet.js
hackrush01: Ok

ui/js/component/tipLink/view.jsx
hackrush01: You mean inside the drop-down menu in open?

ui/js/component/tipLink/view.jsx
hackrush01: Not quite sure what you mean by this. Please elaborate.

kauffj: A text input coupled with an LBC/USD select box appears in a few places. This adds another place it should appear (even though you did it without the currency select).

I was proposing that the text input and lbc/usd select could become it's own component, then that component could be used in all 3 places.

ui/scss/component/_form-field.scss
```diff
@@ -165,4 +165,8 @@ input[type="text"].input-copyable {
 
 .form-field__input.form-field__input-SimpleMDE .CodeMirror-scroll {
   height: auto;
+}
+
+.form-field--tip {
```
hackrush01: I've added this css, is this okay?

ui/js/page/filePage/view.jsx
hackrush01: Done, now it is being done using selectors in the FileActions

ui/js/component/fileActions/view.jsx
hackrush01: Done

ui/js/component/tipLink/view.jsx
hackrush01: Done

ui/js/component/tipLink/view.jsx
hackrush01: Done

ui/js/component/tipLink/view.jsx
hackrush01: Done, please check slack

ui/js/component/tipLink/view.jsx
hackrush01: done

ui/js/actions/wallet.js
hackrush01: Done

ui/js/component/fileActions/view.jsx
hackrush01: Done

ui/js/component/fileActions/view.jsx
hackrush01: Done

ui/js/component/tipLink/view.jsx
```diff
@@ -0,0 +1,95 @@
...
+    this.props.onTipShow();
+  }
+
+  handleSendButtonClicked() {
+    let claim_id = this.props.claim_id;
+    let amount = this.state.feeAmount;
+
+    this.props.setClaimID(claim_id);
+    this.props.setAmount(amount);
+    this.props.sendSupport();
```
kauffj: It's fine for this to be `this.props.sendSupport(claim_id, amount)`.

ui/js/component/fileActions/view.jsx
```diff
@@ -165,8 +182,14 @@ class FileActions extends React.PureComponent {
 
     return (
       <section className="file-actions">
-        {content}
-        {showMenu
+        {showTipBox ? "" : content}
+        <TipLink
```
kauffj: This shouldn't get rendered here.

kauffj: I think it probably needs to go into `FilePage`.

ui/js/component/tipLink/view.jsx
```diff
@@ -0,0 +1,95 @@
...
+    );
+
+    let tipBox = (
+      <span>
+        <FormFieldPrice
+          min="0"
+          placeholder="1.00"
+          step="0.1"
+          onChange={value => this.handleSupportPriceChange(value)}
+          defaultValue={{ amount: feeAmount, currency: currency }}
```
kauffj: Please review React docs on controlled and uncontrolled components. I suspect this may not reset to defaults the way you expect.

`defaultValue` can only be used when rendering the element for the first time.

ui/js/reducers/claims.js
```diff
@@ -190,6 +197,55 @@ reducers[types.CREATE_CHANNEL_COMPLETED] = function(state, action) {
   });
 };
 
+reducers[types.SET_SUPPORT_AMOUNT] = function(state, action) {
```
kauffj: All of this can be simplified by having the component send these values directly to the action, unless I'm missing the need for these to be in Redux state.

ui/js/component/fileActions/view.jsx
hackrush01: Yes, UI changes will be made in subsequent commits. I have a list of them.

ui/js/component/tipLink/view.jsx
hackrush01: Amount's value in store is needed for [this](https://github.com/lbryio/lbry-app/pull/391/files#diff-10945e96d61f083b2a59634a088e8c7aR16)

As for `claim_id`; currently it being passed as a prop [here](https://github.com/lbryio/lbry-app/pull/391/files#diff-1fa5cb7e26763162ff46465cbb589c6eR190), which was possible due to having the `claimInfo` in FileActions, but since this component will be moved to `FilePage`, `claim_id`(for support) will be set during the loading of the page by directly fetching the values from the store(`FETCH_FILE_INFO_COMPLETED` action to be specific). Seems to me a neater and React way of doing this. Also it will ensure no redundant data is being passed between the components.

ui/js/component/tipLink/view.jsx
hackrush01: As far as I understood by reading the [docs](https://facebook.github.io/react/docs/forms.html#controlled-components), controlled components are not a problem, so long all of them have associated `setState` handlers, as the component will re-render. So I'm inclined to keep it here.
But If you have some particular scenario in mind, I'll be happy to test it out though.

ui/js/reducers/claims.js
hackrush01: Clarified in first comment, above.

hackrush01:
- [x] Rework tipping to use the `wallet_send` API command
- [ ] Render tipping UI in `FilePage` instead of `FileActions`
- [ ] Extract `Wallet_Send` form and use that for tipping instead of the `FormFieldPrice` as one can tip in LBC only
- [ ] Something with controlled and uncontrolled components
- [ ] Borrow things from `tip_and_supports` branch?????? (Not at sure what to do with that)
- [x] Display a better `Transaction List` with columns for types of transaction(Support, claim, update, unbound), a link to claim if applicable(this might need some changes depending on how lbryio/lbryum#148 goes)
- [ ] Use a better name instead of "Unbound" transaction.
- [ ] Do a whole lot of conflict fixing in the end.

hackrush01:
Made the merge to master, definitely must have broken something.

kauffj:
@hackrush01 there are changes in master that affect the `TransactionList` component. I'd advise rebasing this onto master before continuing.

hackrush01:
@kauffj Ok, I'll look into it.

ui/js/component/transactionList/view.jsx
```diff
@@ -1,57 +1,57 @@
...
+        </div>
+      );
+    }
 
-  if (!transactions || !transactions.length) {
     return (
-      <div className="empty">
-        {emptyMessage || __("No transactions to list.")}
+      <div>
+        <span className="sort-section">
```
hackrush01: I was not sure about the css part so I left it as is.

kauffj:
Ran the latest today. Here's what ought to change. It might seem long but it's all very minor -- very close to shippable.

### Tipping
- Clicking "Support" ought to hide _everything_ below `FileActions` (description and report I noticed stay)
- The title of the form should not use the word claim. Maybe just "Support"?
- The form should have a one sentence description (probably just a `<p>`) with a link to lbry.io/faq/tipping (to be created). First draft: "Support the creator and the success of their content by sending a tip. <a>Learn more.</a>"
- Submitting `-1` errors uncleanly (possibly out of scope, don't write specific code for this in the form).
- _Ideally_ the support form would have an additional select with two options: Permanent and Temporary. Permanent issues a tip, temporary issues a support. But this also isn't a hard requirement.

### Transaction List
- `TransactionTableXXX` probably ought to be moved into an `internal` folder inside of `TransactionList`, since they are strongly tied to this component rather than likely to be used individually. See the `Video` component for an example of this.
- The empty message needs to be displayed when no items match the transaction filter.
- The "Claim Name" header shouldn't toggle visibility. This should always be displayed, and should be filled even on the default view when relevant. 
- Column widths shouldn't jump on filter. We probably should set some widths specifically for a transaction table.
- Change some labels and orders on filtering:
  - Unfiltered -> All
  - Claims -> Publishes
  - Drop updates if possible. Updates should be in-line with others of the type (presumably publishes / supports). 
- Let's try displaying the fee below the amount. Display it in class "meta" and without a label for the unit.
- Let's try dropping all labels on `CreditAmount` components in the table (i.e. `label=false`).

PR #394, opened 2017-07-27
[WIP] Discover page horizontal scroll implementation

akinwale:
* Removed horizontal scroll bars
* Added scroll left and right buttons

Ideally, the card scaling on the horizontal axis should be able to push the siblings as required (quite similar to how the Netflix carousel works), but I haven't been able to figure out how to do this yet with the CSS transform, so you have the elements being clipped at the beginning or the end of each category.

### DIFF
--- a/ui/js/page/discover/view.jsx
+++ b/ui/js/page/discover/view.jsx
@@ -15,22 +15,13 @@ class FeaturedCategory extends React.PureComponent {
     });
   }
 
-  componentDidMount() {
-    const prevIcon = ReactDOM.findDOMNode(this.refs.scrollPrevious);
-    const nextIcon = ReactDOM.findDOMNode(this.refs.scrollNext);
-    this.scrollPreviousHandler = this.handleScrollPrevious.bind(this);
-    this.scrollNextHandler = this.handleScrollNext.bind(this);
-    prevIcon.addEventListener("click", this.scrollPreviousHandler);
-    nextIcon.addEventListener("click", this.scrollNextHandler);
-  }
-
   handleScrollPrevious() {
     const cardRow = ReactDOM.findDOMNode(this.refs.rowitems);
     if (cardRow.scrollLeft > 0) {
       // check the visible cards
       const cards = cardRow.getElementsByTagName("section");
-      var firstVisibleCard = null;
-      var firstVisibleIdx = -1;
+      let firstVisibleCard = null;
+      let firstVisibleIdx = -1;
       for (var i = 0; i < cards.length; i++) {
         if (this.isCardVisible(cards[i], cardRow, false)) {
           firstVisibleCard = cards[i];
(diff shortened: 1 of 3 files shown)


akinwale:
Sorted out (and perhaps improved) the hover display using the translate3d transform. I've also implemented simple animation for the scroll effect. I suppose it could be improved as I'm not sure multiple recursive calls to setTimeout (10 in my current implementation) is very efficient.

ui/js/page/discover/view.jsx
```diff
@@ -1,38 +1,212 @@
...
+  componentWillMount() {
+    this.setState({
+      numItems: this.props.names.length,
+      canScrollPrevious: false,
+      canScrollNext: true,
+    });
+  }
+
+  componentDidMount() {
+    const prevIcon = ReactDOM.findDOMNode(this.refs.scrollPrevious);
```
kauffj: Would it be cleaner to handle this just by wrapping the `<Icon>` in something like `<a onClick={this.handleScrollNext.bind(this)} />`?

ui/js/page/discover/view.jsx
```diff
@@ -1,38 +1,212 @@
...
+      );
+    }
+  }
+
+  handleScrollNext() {
+    const cardRow = ReactDOM.findDOMNode(this.refs.rowitems);
+
+    // check the visible cards
+    const cards = cardRow.getElementsByTagName("section");
+    var lastVisibleCard = null;
```
kauffj: No `var` :P

ui/scss/component/_card.scss
```diff
@@ -183,26 +187,65 @@ $height-card-small: $spacing-vertical * 15;
...
+}
+
+.card-row__nav {
+  display: none;
+  position: absolute;
+  padding: 0 $spacing-vertical / 1.5;
+  top: ($font-size * 1.4 * $font-line-height) + ($spacing-vertical / 3);
+  height: ($width-card-small * 9 / 16) + ($font-size * 0.82 * $font-line-height * 4) + ($spacing-vertical * 4/3)
+}
+.card-row__nav .icon {
```
kauffj: Adding a wrapper element to `<Icon>` could also eliminate some of this potentially undesirable styling.

(Overriding styles introduces the ability for future changes to have difficult to see side-effects.)

kauffj: A hover effect on this would also be nice. Maybe slightly larger size and icon size, as well as changing the icon color?

kauffj:
Really like the behavior on the whole @akinwale!

akinwale:
> Would it be cleaner to handle this just by wrapping the `<Icon>` in something like `<a onClick={this.handleScrollNext.bind(this)} />`?
Well, this is why sleep is important. I was up late working and completely forgot that this was possible.

Pushed a new commit which addresses this and the other issues. Also made the arrows 50% transparent, and fully opaque with a transition effect on hover.

kauffj:
@akinwale Merged. See my additional changes here: https://github.com/lbryio/lbry-app/commit/00f2413597dc7ba713f17d2b381588fe4b943cee

PR #406, opened 2017-07-30
[WIP] Support claim UI implementation

akinwale:
Added the Support Claim link and inline form to the view media page.

I'm still not exactly sure where total claim support should be displayed. Should it be enclosed within brackets suffixed to the button label or in the inline form?


6ea86b96:
I think it might be good to create a `actions/claims.js`. There should be one already, this is my fault. There's also a bunch of code in `actions/content.js` which really should not be there too.

akinwale:
@6ea86b96 I guess this means I should also move the corresponding selectors and reducers from `selectors/file_actions.js` and `reducers/file_actions.js` to `selectors/claims.js` and `reducers/claims.js` respectively?

6ea86b96:
IMO yes, the supports stuff fits into claims better than file actions.

kauffj:
This is in `tip_and_supports` branch and will come in via #391.

PR #408, opened 2017-07-30
Set window title to filePage metadata title

akinwale:
Currently, the window title is set to the item's lbry URL after opening a media item. This updates the title in the `componentDidMount` method if metadata is present.


kauffj:
I don't like this change because it introduces two different ways titles get set. And I don't think document.title directly inside of components is a "React" way of doing things (for one, it relies on lifecycle components, when the trend is towards stateless).

Maybe a pattern like this if you wanted to refactor? https://stackoverflow.com/a/26266720/82124

But I haven't thought about this or researched this for very long. 

Either way, I'm going to reject this as-is.

PR #413, opened 2017-07-31
Fix for blurry text and image on hover

akinwale:
Replaced translate3d with translateX in CSS since the cards only need to shift on the x-axis (horizontally). More details at https://stackoverflow.com/questions/8024061/webkit-blurry-text-with-css-scale-translate3d.

### DIFF
--- a/ui/scss/component/_card.scss
+++ b/ui/scss/component/_card.scss
@@ -92,7 +92,6 @@ $card-link-scaling: 1.1;
 }
 .card--link {
   transition: transform 120ms ease-in-out;
-  backface-visibility: visible
 }
 .card--link:hover {
   position: relative;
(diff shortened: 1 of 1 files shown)


akinwale:
Please hold off on this. Still seeing some issues on Windows.

kauffj:
If it's just the `translate` call causing the issue, could it be fixed by applying just the `scale` call and then a `left` or `margin-left` to adjust it's x positioning?

kauffj:
Also, please include a comment with a link to the Stack Overflow post, so a future dev (or future @akinwale ! :) ) can change this when the root problem is addressed.

kauffj:
Any word on this @akinwale?

akinwale:
@kauffj I apologise for the lack of updates. I was a little under the weather up till Wednesday, and I haven't had a chance to look at it yet. I've seen some suggestions to use zoom instead of scale, so I'll try this out to see if it works.

MSFTserver:
you trying using zoom instead?

akinwale:
Finally figured this out. I had to change the other `translate3d` call (in the class `.card--link:hover ~ .card--link`) to `translateX` since the cards only need to move on the x-axis. This apparently affected the rendering of the currently hovered card which resulted in the text still being blurry.

PR #431, opened 2017-08-03
Created a more generic PriceForm component

### LINKED ISSUE #426: Placeholder fee is confusing in Publish
## The Issue
When publishing content for a Fee, you get a placeholder of 1.00USD or 1.00 LBRY but it isn't an actual value that's taken into account during the publish. Need to make this less confusing. The same is happening for bid amounts, but that is discussed here: https://github.com/lbryio/lbry-app/issues/373

Log shows: 
`2017-08-03 08:08:39,329 WARNING  lbrynet.daemon.Daemon:1812: Stripping empty fee from published metadata`

### Steps to reproduce
1. Go to Publish
2. Choose fee, leave as is, Publish
3. Content shown as free

### Expected behaviour
Use placeholder, show place holder differently?

### Actual behaviour
placeholder not take into account

## System Configuration

 

- LBRY Daemon version: 0.14.2
- LBRY App version: 0.14.0

- LBRY Installation ID:
- Operating system:

## Screenshots

feee

hackrush01:
- [ ] Fixes #426

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
 
 ## [Unreleased]
 ### Added
-  *
+  * Added a new component, "PriceForm" which is now used in Publish and Settings
   *
 
 ### Changed
@@ -16,7 +16,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
   *
 
 ### Fixed
-  *
+  * Erraneous handling of placeholder value in Publish Form
   *
 
 ### Deprecated
@@ -24,7 +24,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
   *
 
 ### Removed
-  *
+  * Removed one instance of string "Max Purchase Price" from settings page, it's redudant.
   *
 
 ## [0.14.3] - 2017-08-03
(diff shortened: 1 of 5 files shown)


kauffj:
Good idea to separate this out. Can you make the following changes?

- Rebase to master
- Rename to `FormFieldPrice` (we'll establish convention that form widgets are named `FormFieldXXX`, since they will likely become extensions of `FormField`)
- Reduce component from taking two callbacks to a single `onChange` callback, that provides both the currency and the value.

(It is likely that this will be refactored further in the future as we establish a proper form widget pattern, but this is good progress!)

hackrush01:
I've done all the changes you've asked for, but I'm not in love with the last idea, it just reduces function calls, but rest of the code remains same-ish(might be even less pretty to read).

hackrush01:
There's some problem with rebasing.

kauffj:
@hackrush01 misunderstanding here. I was suggesting that the single event always include _both_ values. I.e. rather than `onChange` being passed to the two independent fields inside of it, `FormFieldPrice` handles them both independently and locally, and then calls `onChange` with an object containing both values when _either_ changes.

If unclear, DM me.

hackrush01:
ONLY MEANT FOR TESTING, NOT MERGING.

Some references taken from here.
https://stackoverflow.com/questions/33867474/pass-form-data-to-parent-class-in-react

kauffj:
See https://github.com/lbryio/lbry-app/commit/a299ca2dddffe73bab11f3136d3cc8ee50f427fd

Focus on just the changes I made to the component. You'll see what I meant now.

PR #434, opened 2017-08-04
Fixes alignment problems in some videos

### LINKED ISSUE #419: Videos inappropriately aligned if 4:3 or vertical
## The Issue
Take this video as an example: lbry://nightofthelivingdead#e59cb8ee903d9c38c2563e2200026aec1be638fd 

The streaming screen could look a bit better in terms of aesthetics for videos with odd resolutions. It almost looks like it's cutting off. Maybe center or size the player accordingly? See screenshot below

### Steps to reproduce
1. Stream lbry://nightofthelivingdead#e59cb8ee903d9c38c2563e2200026aec1be638fd
2.
3.

### Expected behaviour
Can be displayed better

### Actual behaviour
looks broken

## System Configuration

 

- LBRY Daemon version: 0.14.2
- LBRY App version: 0.14.2
- LBRY Installation ID:
- Operating system: Windows 10

## Screenshots

video cutting off

hackrush01:
Couldn't break anything else after this fix, and the broken videos
_were_ fixed. So I guess this is a proper fix, but not sure(if it was
this simple or not).

Fixes #419

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
   *
 
 ### Fixed
-  *
+  * Fixed the alignment of video, if it is 4:3 or vertical
   *
 
 ### Deprecated
--- a/ui/scss/component/_video.scss
+++ b/ui/scss/component/_video.scss
@@ -20,7 +20,6 @@ video {
   position: relative;
   video {
     height: 100%;
-    position: absolute;
     left: 0;
     top: 0;
   }


kauffj:
@hackrush01 if a line of code seems to make no sense, it's a good idea to check `git blame` to see if the PR that added it provides insight.

In this case, we see that it is this one: https://github.com/lbryio/lbry-app/commit/99cb1a0a6502a2157fcd46b0c61017f75688a8b1

Now, despite writing that code just a month ago, I actually don't remember exactly what sizing issue it fixed. Perhaps videos less than the minimum height?

Regardless, it's clear that that code did _something_ and we can't remove it without understanding what, or at least thoroughly testing different cases to ensure everything still works.

Welcome to the real-world problems of software engineering that aren't taught in school.

(Additionally, if we were remove absolute, we'd want to remove the other 2 lines.)

btzr-io:
Removing absolute works because `video` gets wrapped inside two `divs` and not directly in `.video-embedded`

btzr-io:
Also no idea why this was closed, it was going in the right direction :stuck_out_tongue:

btzr-io:
Alternative don't wrap `video` inside those two divs ^^ 
and the current `style` should work :upside_down_face:

btzr-io:
@hackrush01 ^^

kauffj:
@btzr-io I'm an aggressive closer. If I close a PR, I'm typically not rejecting it permanently. I just don't like lots of open/sitting PRs.

PR #454, opened 2017-08-08
Forward history

btzr-io:
> Add forward history navigation
### Fixes
Fix for https://github.com/lbryio/lbry-app/issues/443
### Todo
- [x] Disable / style  history buttons when aren't needed. https://github.com/lbryio/lbry-app/pull/481
- [x] Update `CHANGELOG.md`


ui/js/actions/app.js
```diff
@@ -84,6 +84,15 @@ export function doHistoryBack() {
   };
 }
 
+export function doHistoryForward() {
+  return function(dispatch, getState) {
+    //if (!history.state) return;
```
btzr-io: Not sure if this is need it :confused:

ui/js/actions/app.js
```diff
@@ -84,6 +84,15 @@ export function doHistoryBack() {
...
+    //if (history.state.index === 0) return;
+
```
btzr-io: not sure if needs this ? :confused:

btzr-io:
any feedback  / review would be helpful... ^^

tzarebczan:
@lyoshenka Is this something you can take a look at, or wait for @kauffj 's review?

@btzr-io - Jeremy will not be around for a day or two, I know you have a couple things that need to be looked at.

ui/js/actions/app.js
btzr-io: I'm really confused about this ^^

btzr-io: ...someone care to explain me? :upside_down_face:

hackrush01: Well you could comment out the same statements in the "back" function and check what breaks(most probably clicking back when the history is empty), and then try to replicate the similar problem with the "forward" button. Although this statement `if (!history.state) return;` seems kind of necessary, the next one, not so much...

kauffj: As hackrush says, this is to prevent a back button causing a blank screen when there is no history.

kauffj:
I don't want to add this without disabling the button when it can't be used.

I thought this wouldn't be that bad to do, but I already spent more time on this than I can relative to other improvements.

Can you take a look at this branch and see if there's a way to get the back and forward buttons disabled properly? 

https://github.com/lbryio/lbry-app/tree/history

If we can get `HISTORY_NAVIGATE` reducer in `reducers/app.js` working properly, we're good to go.

btzr-io:
Ok, I tried the easy way,  but never figured out :disappointed: ^^

kauffj:
@6ea86b96 any idea on how to get these properties set correctly?

https://github.com/lbryio/lbry-app/blob/history/ui/js/reducers/app.js#L168

jsigwart:
I didn't think a forward button would be a big deal and I'm sorry it's taken this much time. Reading this thread, I'm wondering if the browser even needs history navigation? Ideally users will only navigate forward in the app, going between search/browse, file library, and account information.

6ea86b96:
Looking now @kauffj

6ea86b96:
@kauffj I have something basically working, but got an off-by-one error on disabling buttons. Also need to figure out how  to keep the current page index between refreshes. I didn't have long to look but should be able to find some time tomorrow morning.

btzr-io:
fix conflicts 748bdd5

btzr-io:
> The user agent might disallow calls to these methods that are invoked on a timer, or from event listeners that are not triggered in response to a clear user action, or that are invoked in rapid succession

btzr-io:
Not sure if this help but:
https://github.com/ReactTraining/history

btzr-io:
@kauffj ^^

kauffj:
@btzr-io somewhat confused by the PR you linked. I'd prefer to avoid a library unless we really need it -- it seems like if `history.state` isn't serving our purposes that it'd be pretty easy to replace with a simple array.

It sounds like @6ea86b96 is pretty close with this commit anyway: https://github.com/lbryio/lbry-app/commit/99999a93e3b37d6160aeaf695471ab87b070e20a

btzr-io:
Closed by: https://github.com/lbryio/lbry-app/pull/481

hackrush01:
I think the work regarding this is continued in #481 so I am closing this. Re-open if that's not the case.

btzr-io:
@hackrush01 Oh, yes I forgot to close this, thanks :+1:

PR #455, opened 2017-08-08
Reduxify pull request for Language Downloading and Settings

akinwale:
I encountered an issue which I haven't been able to figure out yet. The individual language downloads work after a fresh launch of the app. However, if I use Reload from the Developer menu, the download requests never complete (even after adding the error handlers I can think of), and the created language files in the `app/locales` folder end up being zero bytes. I am not exactly sure if this is a result of how the batch actions are being handled, but I am still looking into it.


ui/js/actions/app.js
```diff
@@ -297,6 +297,14 @@ export function doClearCache() {
   };
 }
 
+export function doChangeLanguage(newLanguage) {
+  app.i18n.setLocale(newLanguage);
+  return {
```
kauffj: If there is no reducer for this, do we need it?

ui/js/actions/settings.js
```diff
@@ -41,3 +44,296 @@ export function doSetClientSetting(key, value) {
...
+
+          response.on("data", chunk => {
+            str += chunk;
+          });
+
+          response.on("end", () => {
+            const files = JSON.parse(str);
+            const actions = [];
+            files.forEach(file => {
+              actions.push(doDownloadLanguage(file, `app/locales/${file}`));
```
kauffj: Do we actually want to download all language files at first-run? Maybe we should instead just download the current language, and download new languages when the language is switched?

ui/js/reducers/settings.js
```diff
@@ -5,6 +5,7 @@ const reducers = {};
 const defaultState = {
   clientSettings: {
     showNsfw: lbry.getClientSetting("showNsfw"),
+    language: lbry.getClientSetting("language"),
```
kauffj: Out of scope, but we should really move settings to a constants file.

ui/js/actions/app.js
kauffj: (I meant the event return)

kauffj:
Additionally, could you eliminate the hardcoded-references to app/locales? It looks like this can be read directly from the `i18n` object. If not, it should be defined as a constant... somewhere (I'm so helpful).

ui/js/actions/app.js
akinwale: I guess we could leave it there in case there are some actions that need to be performed in the future upon language change. @filipnyquist Any thoughts on this?

ui/js/actions/settings.js
akinwale: Based on the original logic, it looks like the downloaded language files are used to determine the languages that are displayed on the Settings page (except English which is the default). I feel changing the display logic here to make use the language index downloaded from http://i18n.lbry.io/ could be done in another pull request.

Also, if we have to download the language on demand, I suppose we have to have some sort of wait dialog (while the file downloads) on the Settings page before the language change is committed.

ui/js/actions/app.js
kauffj: Just remove this. We can add it when we need it.

akinwale:
@kauffj Pushed a new commit.

kauffj:
@akinwale just checked this out again and I'm not seeing any language selections other than English.

Additionally, whatever `http` is doing under the hood, it somehow makes requests without them showing up on the Chrome debug toolbar?! This makes no sense to me, but I don't see any requests at all, despite data being returned. Any idea what is going on there?

akinwale:
@kauffj Does this happen if you launch the app or if you reload using the Reload option from the Developer Menu? There seems to be an issue where the latter doesn't work properly for some reason. Launching the app seems to be fine, however.

Regarding the `http` requests, this may be related: https://stackoverflow.com/a/44890425/61385

kauffj:
1. Please switch to the i18n branch. All code, including your latest code, and some refactoring by me, is in that branch.
1. This code is extremely difficult to test and basically unshippable until the reloading issue is worked out. I spent a significant amount of time looking into it without resolution. What I learned:

- The requests will eventually timeout.
- The requests do actually get made and sent to the correct URL

akinwale:
@kauffj I was also stuck on the same reloading issue. I can try to move the language load logic to the settings page component `componentDidMount` method. I'll probably also have to add some sort of loading indicator to the settings page. What do you think?

kauffj:
@akinwale I think the load issue is deeper than that. The requests are getting made properly, they are just timing out. I suspect it is somehow related to the previous request not getting properly closed or cleaned up.

You can observe the requests happen with something like:
`sudo tcpflow -p -c -i wlp4s0 port 80 | grep -oE '(GET|POST|HEAD) .* HTTP/1.[01]|Host: .*'`

(Use `ifconfig` to determine the `-i` argument)

One fix that would definitely fix it is to replace the use of Node's `http.request` with browser-native xhr requests. But `http.request` shouldn't be behaving the way it is and I hate giving up without understanding :P

akinwale:
Pushed commit https://github.com/lbryio/lbry-app/commit/efd2b99cda762496dc718fca0132643f8e28125f with `doDownloadLanguages` modified to use `XMLHttpRequest` instead of `http.request` which seems to have fixed reload issues.

kauffj:
See comment on https://github.com/lbryio/lbry-app/commit/efd2b99cda762496dc718fca0132643f8e28125f

akinwale:
Follow up comment at https://github.com/lbryio/lbry-app/commit/efd2b99cda762496dc718fca0132643f8e28125f#commitcomment-23806852

Issue #295, opened 2017-06-27
Video playback controls displayed on video description page


trickyhunter2017:
## The Issue
Video playback controls flash briefly at the bottom of the screen when you try to play an already downloaded video.  If the app can't retrieve the metadata, the video fails to play and the playback controls are displayed permanently at the bottom of the page.  See screenshot below.  

### Steps to reproduce
1.  Install the latest 0.12.2rc7 app, Windows client on Windows 10
2.  Play any already downloaded video
3.  Issue visible on top left of hompage, first vdieo - Cassually Explained: How to make a successful youtube channel.  On other videos it flashes briefly at the bottom of the page.

### Expected behaviour
Video playback controls should not be visible on the video description page.

### Actual behaviour
You'll see the controls either flash briefly or be displayed permanently at the bottom of the page.
(See screenshot below.)

## System Configuration

 

- LBRY Daemon version:  0.13.1
- LBRY App version: 0.12.2rc7
- LBRY Installation ID: 6hR14gSUQpdL8aVk5s9eM4hiHjHzMRKT8PFYd2wGJ3kSrGHaJ2SymZ1oZDPvEQaGzm
- Operating system: Windows (Windows-10-10.0.14393)

## Anything Else

Clearing the application cache didn't help.
Closing and restarting the app didn't help.
I ensured no LBRY processes were running in memory between closing and restarting the app.

## Screenshots

![lbry___ casuallyexplained_casually-successful 2017-06-27 16 59 48](https://user-images.githubusercontent.com/4589923/27599297-3775e2b2-5b60-11e7-8c26-c80ded3f4b43.png)

tzarebczan:
I'm able to confirm this happens - very brief but does occur.

akinwale:
I was able to reproduce this on Ubuntu 16.04. It is very brief. Working on a fix.

kauffj:
I believe this was fixed.

PR #457, opened 2017-08-09
Fix video width

btzr-io:
how_to_fix_video_aligment

The real problem is not the alignment but the the width of the `video` element. 
~Since it's using a container there is no need for `position: absolute`,~
setting the width to `100% `  should do the trick :smile: 

### Fixes
Fix: https://github.com/lbryio/lbry-app/issues/419
Prevent regression:  #295

### DIFF
--- a/ui/scss/component/_video.scss
+++ b/ui/scss/component/_video.scss
@@ -21,6 +21,9 @@ video {
   video {
     height: 100%;
     width: 100%;
+    position: absolute;
+    top: 0;
+    left: 0;
   }
   &.video--hidden {
     height: $height-video-embedded;
--- a/ui/scss/component/_video.scss
+++ b/ui/scss/component/_video.scss
@@ -20,9 +20,7 @@ video {
   position: relative;
   video {
     height: 100%;
-    position: absolute;
-    left: 0;
-    top: 0;
+    width: 100%;
   }
   &.video--hidden {
     height: $height-video-embedded;


btzr-io:
I can't request reviews :disappointed:  ^^

jsigwart:
Wouldn't this cause horizontal stretching that would look odd on narrow pictures?

btzr-io:
@jsigwart  Nope, because it's using `object-fit: contain`
> contain: increases or decreases the size of the image to fill the box whilst preserving its aspect-ratio.

jsigwart:
I think what you want to do is this: 
```
video {		   
   height: 100%;
   margin: 0 auto;
}
```

hackrush01:
@btzr-io Could you give a look [here](https://github.com/lbryio/lbry-app/pull/434#issuecomment-320514720).

btzr-io:
@jsigwart you don't need `margin: 0 auto`

jsigwart:
Yup, you're exactly right. And that should fix the case of a video being too short, too.

hackrush01:
This is further along and with a better explanation, so I'm inclined to keep this. Also I'm linking some important comments from that PR to here so the discussion can continue here.
[Comment#1](https://github.com/lbryio/lbry-app/pull/434#issuecomment-321133778)

As for it being rendered inside multiple `<div>` tags, it could be because of the way react handles components(everything can be returned only in a single wrapper tag).

btzr-io:
Well this is the most easy fix,not sure if someone wants to go deep inside that react component

btzr-io:
I did a quick look, and didn't fine nothing wrong,so that means

PR #463, opened 2017-08-09
Added wallet backup guide reference.

hackrush01:
It also displays the current wallet path so user doesn't have to "go places"

Could use a different/better message though, this was more of a place-holder.

For reference check [this](https://github.com/lbryio/lbry/issues/773#issuecomment-316427167).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
 ## [Unreleased]
 ### Added
   * Added a new component, `FormFieldPrice` which is now used in Publish and Settings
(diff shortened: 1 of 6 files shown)


kauffj:
Merged in https://github.com/lbryio/lbry-app/commit/a21eaf0f5772c2aadc818dedafe122df41cfef87

See https://github.com/lbryio/lbry-app/commit/622f1aaa37b7eddc8380d830f3eb8f8b03883ae8 for my changes.

I dropped it as a navigation item and made it text under the balance.

@hackrush01 my only PR feedback is use `code` tag for something like a folder path and avoid adding inline CSS.

Issue #462, opened 2017-08-09
Can Publish without accepting TOS


tzarebczan:
## The Issue
Able to publish something without clicking the TOS checkbox at the bottom. 

### Steps to reproduce
1. Go to Publish, fill out info, don't check TOS
2. Click Publish 
3. Success!

### Expected behaviour
Prompt to check the TOS box

### Actual behaviour
Does not validate that TOS is not checked off

## System Configuration

 

- LBRY Daemon version: 0.14.2
- LBRY App version: 0.14.3
- LBRY Installation ID:
- Operating system: Windows 10

PR #464, opened 2017-08-09
(quickfix) Check for required fields during publishing

hackrush01:
I'm not sure if this is the best way to fix this but it works without breaking anything(based on some testing).

A quickfix for issue #462


tzarebczan:
I just opened this which may be related: https://github.com/lbryio/lbry-app/issues/467

PR #470, opened 2017-08-11
Configure bumpversion to avoid modifying daemon version in package.json

alexliebowitz:
Before, it was matching *any* instance of the current app version string, so if the daemon version configured in app/package.json happened to match the app version, it would get bumped.

This updates the bumpversion config to only update the line containing the package version itself.


kauffj:
I figured it was this. Fun corner case :)

PR #481, opened 2017-08-17
Fix history

btzr-io:
### Fixes
Fix for history forward / backward buttons: `action / reducer / component`
See -> https://github.com/lbryio/lbry-app/pull/454

### Todo
- [x] Tidy up.
- [x] Simplify.
- [x]   Create `historyStack` and `historyStackIndex` state.
- [x] Create `getForward` and `getBackward`  selectors.
- [x] Stop using utils and follow `React` / `Redux` style. :stuck_out_tongue:


kauffj:
@btzr-io you are making great progress with this and on the right track. This is the right fundamental approach.

However, it needs to be done in the React/Redux style used by the rest of the app. Rather than a utility file and class, the historyStack and historyStackIndex can be direct state properties stored by a reducer. Similarly, the `getForward` and `getBackward` functions can be selectors.

btzr-io:
Ok, I'll update the todo-list  ^^

btzr-io:
@kauffj Well now it's more complex but here is it:  01fbe81b5b49f581aac75665a38f0e2bdf893da9^^

btzr-io:
this fix -> https://github.com/lbryio/lbry-app/issues/443

PR #483, opened 2017-08-17
update setup

tzarebczan:
-add node / nvm link
-add keytar/libsecret step
-add yarn step

Feel free to adjust directions/wording


kauffj:
This one is for @lyoshenka.

We really ought to get the missing packages for this in the install/build process.

PR #489, opened 2017-08-22
Check for valid address when sending credits on Wallet page

### LINKED ISSUE #445: Check Address is valid before sending
## The Issue
issue comes about when trying to send funds to a non legit account like "aaaaaaaa"
### Steps to reproduce
1.enter address aaaaaaaaaa
2.enter amount
3.send

### Expected behaviour
i would expect this to do an api call to see if this address is existence and has funds linked to it or if it does not have a public or private key don't send the funds, and have error pop up if it doesnt work 
### Actual behaviour
it claims it sent the funds to the non existence account

## System Configuration

 

- LBRY Daemon version: 0.14.2
- LBRY App version: 0.14.3
- LBRY Installation ID: 9JKT6nrdGYoFKshVLmjnbfUdqNvHRkzWmsLXk5xNdbXzGEwpraZ6MVg1KoD5jrBrf2
- Operating system: Windows 10

alexliebowitz:
Fixes #445. Ready to go, but https://github.com/lbryio/lbry/pull/867 needs to be merged first

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
   * Fixed pagination styling for pages > 5 (#416)
   * Fixed sizing on squat videos (#419)
   * Support claims no longer show up on Published page (#384)
+  * Check for valid address when sending credits (#445)
 
 ### Deprecated
   *
--- a/ui/js/actions/wallet.js
+++ b/ui/js/actions/wallet.js
@@ -2,6 +2,7 @@ import * as types from "constants/action_types";
 import lbry from "lbry";
 import {
   selectDraftTransaction,
+  selectDraftTransactionAddress,
   selectDraftTransactionAmount,
   selectBalance,
 } from "selectors/wallet";
@@ -69,6 +70,7 @@ export function doSendDraftTransaction() {
   return function(dispatch, getState) {
     const state = getState();
     const draftTx = selectDraftTransaction(state);
+    const address = selectDraftTransactionAddress(state);
     const balance = selectBalance(state);
     const amount = selectDraftTransactionAmount(state);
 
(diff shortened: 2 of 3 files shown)


kauffj:
`is_address` will not be an API call. See discussion elsewhere.

More importantly, this is not solved as suggested by #445. See my comment [here](https://github.com/lbryio/lbry-app/issues/445#issuecomment-320525524).

Please solve this by simply adding a regexp parameter to `FormField` or a descendant of `FormField`. Then an invalid address can be detected and shown to the user entirely in widget-land -- no need for a modal.

I'm closing this PR as I think very little of it will apply to the appropriate solution.

(Btw, if you think the proposed approach is incorrect, I am, as always, open to discussion. I suspect in this base it simply wasn't seen.)

alexliebowitz:
Ugh, yes, you're right, I missed it. I skimmed the whole thread but somehow took away that your idea was about implementation details (not UX). Should have gone over it again before I started work.

Is it really just a regex check to confirm a valid address, though? The JS libraries seem to do some base58 voodoo to check: https://github.com/cryptocoinjs/coinstring/blob/master/lib/coinstring.js#L53-L61 -> https://github.com/cryptocoinjs/coinstring/blob/master/lib/coinstring.js#L24-L39

In which case we could use the validator function from one of libraries (or write our own small validator function) and add a validator function param to form fields instead of a regex param.

kauffj:
@alexliebowitz I think it is just `/^b[HJK][1-9A-Za-z][^OIl]{33}/` or something close to that...

alexliebowitz:
@kauffj: @kaykurokawa says "regex can tell you if its base 58 encoded, but it won't validate it because there 's a checksum and a header byte". Checksum seems important, since it will notice typos where 1 character is off but it's still base58.

PR #494, opened 2017-08-25
Indicate downloaded/published files in file tiles

alexliebowitz:
Adds the same folder icon that's there in file cards.

Also includes some minor CSS improvements and JS refactoring.

### DIFF
--- a/ui/js/component/iconDownloaded/view.jsx
+++ b/ui/js/component/iconDownloaded/view.jsx
@@ -1,0 +1,15 @@
+import React from "react";
+import { Icon } from "component/common.js";
+
+const IconDownloaded = props => {
+  return (
+    <span
+      className="icon-downloaded"
+      title={__("You've already downloaded this content.")}
+    >
+      <Icon icon="icon-folder" fixed className="card__icon-downloaded" />
+    </span>
+  );
+};
+
+export default IconDownloaded;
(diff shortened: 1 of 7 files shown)


alexliebowitz:
The alignment between the icons looks weird:

screen shot 2017-08-25 at 12 27 55 am

There's already the same problem for FileCard, but it's not as noticeable at the smaller font size:
screen shot 2017-08-25 at 12 29 19 am

I'm thinking we should add some CSS to either `<IconFeatured />` or `<IconDownloaded />` to make the alignment match? Hopefully using a percentage or ems to make it work at any size?

ui/js/component/iconDownloaded/view.jsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { Icon } from "component/common.js";
+
+const IconDownloaded = props => {
```
alexliebowitz: Technically currently applies to both published stuff and downloaded stuff. If that's what we want, we could rename this to `<IconMyFile />` or `<IconMine />` or something like that. But I'm wondering if we should have a separate icon for published files anyway?

kauffj:
1. If you want to fix the alignment issue and it's not a super clean/true fix, please apply it is a fix scoped within `file-tile__icons`.
1. It's worth searching "font awesome" and "vertical alignment" -- others have definitely experienced this.
1. Maybe the proper DRY fix is a shared `FileIcons` component?

kauffj:
I'm closing this. 

I also think creating a different class for every Icon is a bad idea, though I understand `IconFeatured` started this pattern.

Instead it is probably better to properly define icon types as global constants, and then use these to standardize titles.

Issue #460, opened 2017-08-09
Published URIs with Channel getting incorrect claim_id


tzarebczan:
## The Issue
This issue originally stemmed from here: https://github.com/lbryio/lbryum/issues/143
After Publishing a file with an identity, the user is presented a link in their Publishes area.   This link has the incorrect channel id in this format:
lbry://@jiggytom#6e477eded9468613c9c5b45eb58662ea302bccea/index

The channel is actually lbry://@jiggytom#e48d2b50501159034f68d53321f67b8aa5b1d771 and the index claim is lbry://index#6e477eded9468613c9c5b45eb58662ea302bccea

There is a bug in lbryum that allows it to resolve correctly, but a get fails.  Wasn't sure what the root cause is, so opening an app issue on top of the lbryum one.  The get returns a failure to resolve error. 

### Steps to reproduce
1. Publish content under an identity, let it confirm
2. Observe URI on the Publishes screen 
3. Notice that it's showing the claim id and not the channel id in the URI

### Expected behaviour
Should show the channel id in the URI

### Actual behaviour
Shows claim id in the URI

## System Configuration

 

- LBRY Daemon version: 0.14.2
- LBRY App version: 0.14.3 
- LBRY Installation ID:
- Operating system: Win 10

PR #501, opened 2017-08-26
Return proper URI when channel name is present.

hackrush01:
FIx for #460. Now the channel's claim_id is appended in after the channel_name in the URI.


kauffj:
@hackrush01 FYI you should be able to branch off the main `lbryio` repo rather than fork.

kauffj:
@hackrush01 this fix had a bug, see 90c83c6

kauffj:
And my fix had a bug, proper fix is 8fd9d64

PR #502, opened 2017-08-26
Invites and rewards (WIP)

kauffj:
- Adds invites page
- Adds reward list component
- Significant refactor of wallet area, including new summary components
- Significant refactor of first run and credit introduction (not finished)


ui/js/component/publishForm/view.jsx
```diff
@@ -318,7 +318,7 @@ class PublishForm extends React.PureComponent {
   handleFeePrefChange(feeEnabled) {
     this.setState({
       isFee: feeEnabled,
-      feeAmount: this.state.feeAmount == "" ? "5.00" : this.state.feeAmount,
+      feeAmount: this.state.feeAmount == "" ? "0.01" : this.state.feeAmount,
```
akinwale: It makes sense to define `0.01` as a global constant, assuming it would also be the placeholder / step value for the tipping field.

ui/js/component/userVerify/view.jsx
```diff
@@ -26,37 +26,125 @@ class UserVerify extends React.PureComponent {
...
+
+          </div>
+          <div className="card__actions">
+            <Link
+              onClick={() => navigate("/discover")}
+              button="alt"
+              label={__("Skip Rewards")}
+            />
+          </div>
+        </section>
```
akinwale: More of a personal opinion, but I feel verification by chat should be the first option, although I assume this would increase the workload for community moderators. I think requesting for credit / debit card info first will put some people off. Perhaps, credit card or debit card verification should even be the last option.

ui/js/constants/modal_types.js
```diff
@@ -7,4 +7,5 @@ export const UPGRADE = "upgrade";
 export const WELCOME = "welcome";
 export const FIRST_REWARD = "first_reward";
 export const AUTHENTICATION_FAILURE = "auth_failure";
+export const REWARD_APPROVAL_REQUIRED = "REWARD_APPROVAL_REQUIRED";
```
akinwale: Value should be lower case to maintain consistency with the other constants.

ui/js/constants/settings.js
```diff
@@ -1,5 +1,6 @@
+/*hardcoded names still exist for these in reducers/settings.js - only discovered when debugging*/
 export const CREDIT_INTRO_ACKNOWLEDGED = "credit_intro_acknowledged";
-export const FIRST_RUN_ACKNOWLEDGED = "welcome_acknowledged";
+export const NEW_USER_ACKNOWLEDGED = "welcome_acknowledged";
```
akinwale: Shouldn't the value for this be changed to `new_user_acknowledged` or is there a reason for sticking with `welcome_acknowledged`?

ui/js/modal/modalCreditIntro/view.jsx
```diff
@@ -1,40 +1,64 @@
 import React from "react";
 import { Modal } from "modal/modal";
-import { CreditAmount } from "component/common";
+import { CreditAmount, CurrencySymbol } from "component/common";
```
akinwale: Although not too significant, defining `LBC` as a global constant instead of a component would remove the overhead of instantiating the component every time it's used and then calling the `render()` method.

ui/js/modal/modalRewardApprovalRequired/view.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+    return (
+      <Modal
+        isOpen={true}
+        contentLabel={__("Human Verification Required")}
+        onConfirmed={doAuth}
+        onAborted={closeModal}
+        type="confirm"
+        confirmButtonLabel={__("I'm Totally Real")}
+        abortButtonLabel={__("Nevermind")}
```
akinwale: "Never mind" are two words.

ui/js/selectors/wallet.js
```diff
@@ -5,7 +5,7 @@ export const _selectState = state => state.wallet || {};
 
 export const selectBalance = createSelector(
   _selectState,
-  state => state.balance || 0
+  state => 0 //state.balance || 0
```
akinwale: I'm guessing this was supposed to be a temporary comment? The displayed wallet balance is always 0.

akinwale:
@kauffj Also, I like the new layout for the Wallet overview and the rewards pages. Looks cleaner. However, on the Rewards tab, the Referral reward card width is less than the combined width and padding of the 2-column cards.

ui/js/actions/rewards.js
```diff
@@ -72,7 +74,7 @@ export function doClaimReward(reward, saveError = false) {
         type: types.CLAIM_REWARD_FAILURE,
         data: {
           reward,
-          error: saveError ? error : null,
+          error: error ? error : null,
```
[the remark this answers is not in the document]
kauffj: :+1:

ui/js/component/publishForm/view.jsx
kauffj: I don't think these are the same value. I would agree to constant this if it was being repeated (I love constants) but right now this is the only place. I did make it a variable to make it clearer.

ui/js/component/userVerify/view.jsx
kauffj: Obviously the real solution is to randomize them and see if it makes a difference :)

I'm going to leave this as-is since it's a smaller change, but I don't feel strongly about the order.

ui/js/constants/settings.js
kauffj: Yes, the value locally stored in the app by the constant value. Thus, changing the value would invalidate existing settings. I added a comment to settings.js about this, since it applies to many settings.

ui/js/modal/modalCreditIntro/view.jsx
kauffj: used a component since I thought it likely it will be a richer element (e.g. title text at least?). This was perhaps premature and it should have been a constant, but I'm going to leave this as-is.

ui/js/modal/modalRewardApprovalRequired/view.jsx
kauffj: It's two in this case, good catch. It's confusing for everyone :)

ui/js/selectors/wallet.js
kauffj: Yes and fixed.

kauffj:
@akinwale thanks for the detailed look. Addressed all but the width issue since I'm having trouble getting the app to open atm.

PR #504, opened 2017-08-27
Fix history logic

btzr-io:
### Todo

~Fix / debug https://github.com/lbryio/lbry-app/issues/499~
- [ ] Test this.
- [ ] Try the hard way:  https://github.com/lbryio/lbry-app/issues/499#issuecomment-325423807
- [x] Fix / debug https://gihub.com/lbryio/lbry-app/issues/499#issuecomment-325147682 
- [x] Fix duplicated history bug.
- [x] Update changelog.md.


btzr-io:
please test this  :upside_down_face:  ^^

btzr-io:
Fix conflicts `#49ab36d`

kauffj:
As I suspected, this is possible to do with just one stack. I found some other simplifications as well. You can see the full changeset here:

https://github.com/lbryio/lbry-app/commit/f6976c3feb5a049e5071d027f520a11dbba3b4f0

The total line changes make it worse than it was, as I decided we ought to split navigation into it's own set of actions and reducers. If you look at just the reducer for `HISTORY_NAVIGATE`, you'll see it's not that bad.

Most importantly, I noticed a reducer that was directly modifying the state of an object. In React/Redux reducers, object and array state ought not to be directly mutated, but instead the objects should be copied and updated. See: https://github.com/reactjs/redux/blob/master/docs/recipes/reducers/ImmutableUpdatePatterns.md

btzr-io:
@kauffj  Ok, sorry I didn't have time to finish this, thanks for fixing my mess :upside_down_face:  ^^
btw your implementation looks awesome :+1:

PR #506, opened 2017-08-27
Show most recent claims first on channel pages

### LINKED ISSUE #479: channel content list is backward
## The Issue
most recent claims show on last page, not front page
### Steps to reproduce
1. publish 20 pieces of content to a channel
2. observe pagination order
3.

### Expected behaviour
Newest videos should be on top of the channel

### Actual behaviour
Newest videos (in tx order) show up from end of channel pages

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

alexliebowitz:
@kauffj: Seems like this is all it will take to fix #479, unless we want lbrynet to return the content newest first?

### DIFF
--- a/ui/js/page/channel/view.jsx
+++ b/ui/js/page/channel/view.jsx
@@ -48,7 +48,7 @@ class ChannelPage extends React.PureComponent {
       contentList = <BusyMessage message={__("Fetching content")} />;
     } else if (claimsInChannel) {
       contentList = claimsInChannel.length
-        ? claimsInChannel.map(claim =>
+        ? claimsInChannel.reverse().map(claim =>
             <FileTile
               key={claim.claim_id}
               uri={lbryuri.build({
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
   * Updated several packages and fixed warnings in build process (all but the [fsevents warning](https://github.com/yarnpkg/yarn/issues/3738), which is a rather dramatic debate)
 
 ### Fixed
+  * Show most recent content first on channel pages (#479)
   * Tiles will no longer be blurry on hover (Windows only bug)
   * Removed placeholder values from price selection form fields, which was causing confusion that these were real values (#426)
   * Fixed showing "other currency" help tip in publish form, which was caused due to not "setting" state for price


ui/js/page/channel/view.jsx
```diff
@@ -48,7 +48,7 @@ class ChannelPage extends React.PureComponent {
       contentList = <BusyMessage message={__("Fetching content")} />;
     } else if (claimsInChannel) {
       contentList = claimsInChannel.length
-        ? claimsInChannel.map(claim =>
+        ? claimsInChannel.reverse().map(claim =>
```
hackrush01: I think this only reverses the claims in the current page(as `claimsInChannel` only gets the results, see [here](https://github.com/lbryio/lbry-app/blob/master/ui/js/page/channel/index.js#L19)) I think it should reverse all the claims. Might require a daemon change IMO. @kauffj??

alexliebowitz: Good catch, thanks. I feel like that var should be called claimsInPage or something.

I don't _think_ we have to modify the API to newest first, unless we think it would be a better API anyway. We could simply make a one line wrapper function like `lbry.getClaimsByChannelNewestFirst()` and call that instead of calling the API directly.

If we don't want to add any more junk to the `lbry` module, we could also just modify the spot where we're processing it in the reducer: https://github.com/lbryio/lbry-app/blob/master/ui/js/reducers/claims.js#L116
and below it add:
```
claims = claims.reverse();
```
(or use another var, you get the idea)

There's some hairy state management going on, but it all flows from that one call to `get_claims_by_channel`, so as long as we reverse the results before we do any processing we should be fine.

@kauffj

hackrush01: @alexliebowitz 

The var name should indeed better reflect what is it doing. `claimsInPage` has a nice ring to it.

The `claim_list_by_channel` channel commands takes in `uri` and `page` as parameters. So the API itself returns paginated results(in this case it only returns 10 results per API call). The app is not doing any preprocessing here I guess. It just parses the result of the API call: [link](https://github.com/lbryio/lbry/blob/master/lbrynet/daemon/Daemon.py#L2013)

kauffj: Yes, the API is poorly named, but it's what got shipped.

The fix here without changing the API is simply to paginate backwards.

hackrush01: @kauffj If a channel has say 15 claims, then reverse paginating would show 5 claims in first page and 10 in the next.

kauffj: Bah, you're right. Then it should probably be fixed lower down, and if we're doing that, then we can rename the count field and fix the return signature more generally (no need for the return to include an extra nested element).

alexliebowitz: This is a design discussion that's probably already happened, but might it be possible/better to have the API call take a start and end offset instead of a page number? That would allow apps to do forward _or_ backward paging, load arbitrary numbers of claims at once for caching, etc.

I know it would require some rejiggering on the front end, but probably not _too_ bad, as could can keep managing pages in state the same way, but have it do the math on the client side to figure out what range to request for each page.

kauffj: You're right, but I wouldn't also make it a priority. Page, page size, and sort order would also be equivalent, FWIW.

alexliebowitz: Good point. In that case we definitely ought to look at adding a sort order param to the API instead of having it always return reverse—that way apps could do either type of paging, _and_ would be fairly easy to request arbitrary ranges from either side.

kauffj:
For now, this will be fixed deeper in `lbryum` (PR is already open), so I'm going to close this.

PR #507, opened 2017-08-27
Fixed regression in #501

hackrush01:
- [x] Beat @wallermadev to the PR
- [x] Fix the regression

This is a temporary fix, but I think the daemon should return consistent result whether it be `file_list` or `claim_list_mine`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
 
 ### Fixed
  * Corrected improper pluralization on loading screen
+ * Regression in #501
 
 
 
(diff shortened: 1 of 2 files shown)


wallermadev:
Haha -- Cheers mate, boxing go
It in the way will do build on windows/Linux and test now.

hackrush01:
@wallermadev How do you build it for windows?

wallermadev:
I would be more than happy to add a wiki page to the lbry-app repository as this will probably be useful.

I have tested your fix both with and without items published on the following operating systems;

- Windows 10 (Working)
- Ubuntu Budgie (Working)

@kauffj Seems to all work fine.

W

kauffj:
@wallermadev adding Windows build instructions to the README would be much appreciated

wallermadev:
@kauffj [566](https://github.com/lbryio/lbry-app/pull/566) Added.

PR #513, opened 2017-08-29
Fix Grammar

MSFTserver:


MSFTserver:
not sure why it has so many conflicts i didnt do anything it was all jeremy i swear

kauffj:
I had already fixed this in parallel.

kauffj:
@MSFTserver it's because I rebased onto master and force pushed before opening the PR.

PR #515, opened 2017-08-29
Stripe $1 Charge Notice

MSFTserver:


kauffj:
Addressed in 6be141b. Thanks for the prompt.

PR #520, opened 2017-08-31
Card row autoadjust on window resize and scroll arrow clipping css fix

akinwale:


kauffj:
I didn't run this :pray:

PR #522, opened 2017-09-01
Validate address on Send page

alexliebowitz:
Another crack at fixing #445.

### DIFF
--- a/ui/js/component/formField/view.jsx
+++ b/ui/js/component/formField/view.jsx
@@ -12,6 +12,15 @@ class FormField extends React.PureComponent {
     prefix: React.PropTypes.string,
     postfix: React.PropTypes.string,
     hasError: React.PropTypes.bool,
+    trim: React.PropTypes.bool,
+    regexp: React.PropTypes.oneOfType([
+      React.PropTypes.instanceOf(RegExp),
+      React.PropTypes.string,
+    ]),
+  };
+
+  static defaultProps = {
+    trim: false,
   };
 
   constructor(props) {
@@ -77,6 +86,13 @@ class FormField extends React.PureComponent {
     });
   }
 
+  clearError() {
+    this.setState({
+      isError: false,
+      errorMessage: "",
+    });
+  }
+
   focus() {
     this.refs.field.focus();
   }
@@ -87,7 +103,9 @@ class FormField extends React.PureComponent {
     } else if (this.props.type == "SimpleMDE") {
       return this.refs.field.simplemde.value();
     } else {
-      return this.refs.field.value;
+      return this.props.trim
+        ? this.refs.field.value.trim()
+        : this.refs.field.value;
     }
   }
 
@@ -97,6 +115,16 @@ class FormField extends React.PureComponent {
 
   getOptions() {
     return this.refs.field.options;
+  }
+
+  validate() {
+    if ("regexp" in this.props) {
+      if (!this.getValue().match(this.props.regexp)) {
+        this.showError(__("Invalid format."));
+      } else {
+        this.clearError();
+      }
+    }
   }
 
   render() {
@@ -116,6 +144,8 @@ class FormField extends React.PureComponent {
     delete otherProps.postfix;
     delete otherProps.prefix;
     delete otherProps.dispatch;
+    delete otherProps.regexp;
+    delete otherProps.trim;
 
     const element = (
       <this._element
@@ -124,6 +154,7 @@ class FormField extends React.PureComponent {
         name={this.props.name}
         ref="field"
         placeholder={this.props.placeholder}
+        onBlur={() => this.validate()}
         className={
           "form-field__input form-field__input-" +
           this.props.type +
--- a/ui/js/component/walletSend/view.jsx
+++ b/ui/js/component/walletSend/view.jsx
@@ -2,6 +2,7 @@ import React from "react";
 import Link from "component/link";
 import Modal from "modal/modal";
 import { FormRow } from "component/form";
+import lbryuri from "lbryuri";
 
 const WalletSend = props => {
   const {
@@ -12,6 +13,7 @@ const WalletSend = props => {
     setAddress,
     amount,
     address,
+    error,
   } = props;
 
   return (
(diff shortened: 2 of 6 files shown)


ui/js/component/formField/view.jsx
```diff
@@ -99,6 +121,16 @@ class FormField extends React.PureComponent {
     return this.refs.field.options;
   }
 
+  validate() {
+    if ("regexp" in this.props) {
+      if (!this.getValue().match(this.props.regexp)) {
+        // Handle invalid input here
```
alexliebowitz: Do we want to just display an error and cancel the form submit ourselves? Feels out of sync with how we've been handling form validation up to now (where the form itself does the check and tells the input to show an error). Two different ways of managing error state in the same component.

ui/js/lbryuri.js
```diff
@@ -5,6 +5,8 @@ const lbryuri = {};
 
 lbryuri.REGEXP_INVALID_URI = /[^A-Za-z0-9-]/g;
 
+lbryuri.REGEXP_ADDRESS = /^b[HJK][1-9A-HJ-Za-km-z]{32}/;
```
alexliebowitz: Kinda ugly, but you can't mix negative and positive character ranges. I tried to do it with lookaheads, since you can apply two of those to the same character, but ended up with a 40+ char monstrosity.

MSFTserver: i think it would need more then reg expression i could enter 32 a's in this case right?

alexliebowitz: We decided to keep it simple and do the client-side address validation with a simple regex, and have the daemon return an error if the address isn't actually valid. So we're going to handle that case, but it's going to be a separate chunk of logic (probably in this same PR).

alexliebowitz: Found a way to do it with lookahead that I guess is better? `/^b[HJK](?=[^0Il]{32})[0-9A-Za-z]{32}$/`

ui/js/component/formField/view.jsx
```diff
@@ -51,6 +56,13 @@ class FormField extends React.PureComponent {
   }
 
   componentDidMount() {
+    if (this.props.regexp) {
+      this._validateCallback = () => this.validate();
+    }
+
+    this.refs.field.addEventListener("blur", this._validateCallback);
+    this.refs.field.form.addEventListener("submit", this._validateCallback);
```
kauffj: How does `this.refs.field.form` always exist?

ui/js/component/formField/view.jsx
kauffj: It should use the existing way of rendering the form error here.

I agree that it should not abort a form submit. Really, this should not know about the form at all.

Likely, the right pattern is:

- Forms don't submit when they have a child component with an error.
- Forms can trigger child form fields to validate themselves one last time on submit

I think the validation in the second case should only happen if the value has changed since last validation.

ui/js/lbryuri.js
kauffj: I'm open to `trim` as a high-level prop on inputs that always trims spaces on the value.

ui/js/component/formField/view.jsx
alexliebowitz: Don't all form inputs have a `.form` property, right? Even `<select>` and such? So as long as `this.refs.field` is some kind of form input, it should work. And because it's in `componentDidMount`, we don't have to worry about the ref not existing.

ui/js/lbryuri.js
alexliebowitz: That makes sense. Although just to be clear, the lookahead regex is just meant to be more readable; it doesn't address the whitespace issue.

ui/js/component/formField/view.jsx
kauffj: 1) I think that is not true. Fields can be rendered outside of forms.
2) Why would being in `componentDidMount` avoid an exception from calling a method on undefined?

alexliebowitz: 1. True. I had the sense that having inputs outside of forms was supposed to be a bit of an anti-pattern, since wrapping everything in a `<form>` is more semantic even when you're not submitting to the server. But it looks like there are uses that are considered valid. It sounds like we're going to nix this altogether, but if we somehow end up keeping it, I'll make sure that it doesn't try to interact with the form if the form doesn't exist.
2. It doesn't. I was assuming there'd be a form, being in `componentDidMount` just assures that the ref will already exist.

ui/js/component/formField/view.jsx
```diff
@@ -99,6 +117,16 @@ class FormField extends React.PureComponent {
     return this.refs.field.options;
   }
 
+  validate() {
+    if ("regexp" in this.props) {
+      if (!this.getValue().match(this.props.regexp)) {
+        this.showError("Invalid address.");
```
alexliebowitz: This will not be detected on submit, only blur. The Right Way to do this is probably to have the form query each of its inputs about whether there's an error, but that's tricky to do with React/Redux for now. In this case the user will still find out because the form reports errors from the daemon.

ui/js/component/formField/view.jsx
```diff
@@ -77,6 +86,13 @@ class FormField extends React.PureComponent {
     });
   }
 
+  clearError() {
```
alexliebowitz: This already exists in `<FormRow />`, and we probably shouldn't have it in 2 places, but I figure it's a 4-line function and we can worry about that later. (`showError` is already there in both components.)

ui/js/component/walletSend/view.jsx
```diff
@@ -77,7 +81,7 @@ const WalletSend = props => {
           contentLabel={__("Transaction failed")}
           onConfirmed={closeModal}
         >
-          {__("Something went wrong")}:
+          {__(error)}
```
alexliebowitz: Two questions here:
  1) Are we OK with displaying the error message from the daemon directly? Without an error code, my only other option would be to check the text string of the error message and display our own message based on that. Right now this is the message it's passing through from the daemon:
screen shot 2017-09-04 at 6 44 16 pm

  2) In the meanwhile, should this be wrapped in an `__` function or not? Feels brittle since the message on the daemon side might change, OTOH we don't want to just always have it be in English.

kauffj: 1) Yes, it is fine to display daemon error messages.
2) Do not i18n daemon errors. This is an unsolved problem.
3) This doesn't need to block PR (it already worked this way), but form errors should be inline when possible.

Issue #523, opened 2017-09-01
Add a message about token security to auth_token display


lyoshenka:
As a user, I don't want to ever share my token with anyone, or else they can impersonate me.

We need a message next to the auth token field that says something like "Auth token is for API use only. Never share it with anyone. Keep it secret - treat it like your password."

MSFTserver:
the access token im assuming?

MSFTserver:
something like this?
![image](https://user-images.githubusercontent.com/18357967/29991310-9353f10c-8f39-11e7-8f29-177d8a267723.png)

MSFTserver:
Pull-Request https://github.com/lbryio/lbry-app/pull/524

kauffj:
https://github.com/lbryio/lbry-app/commit/19165895533dc3c6b5a092b2c22d14730d11bb14

PR #524, opened 2017-09-02
Add a message about token security [issue #523]

MSFTserver:
image

### DIFF
--- a/ui/js/page/help/view.jsx
+++ b/ui/js/page/help/view.jsx
@@ -199,6 +199,10 @@ class HelpPage extends React.PureComponent {
                           this.props.accessToken}
                       </td>
                     </tr>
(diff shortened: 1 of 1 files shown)


MSFTserver:
image

MSFTserver:
lol i still cant figure out how to request reviews @btzr-io

MSFTserver:
could even give it its own class and add red highlight or something to make it stand out more

kauffj:
Fixed in https://github.com/lbryio/lbry-app/commit/19165895533dc3c6b5a092b2c22d14730d11bb14

kauffj:
Also, FYI, error messages and styles should never be used stylistically unless an error actually occurred. A notice or warning style would instead be a candidate here.

PR #525, opened 2017-09-02
Display published date

btzr-io:
issue: https://github.com/lbryio/lbry-app/issues/59

### Todo
- [ ] Test ( please test this ) :upside_down_face: 
- [x] Better formating for date.

### Image
published-date


ui/js/constants/action_types.js
```diff
@@ -83,6 +83,7 @@ export const CREATE_CHANNEL_COMPLETED = "CREATE_CHANNEL_COMPLETED";
 export const PUBLISH_STARTED = "PUBLISH_STARTED";
 export const PUBLISH_COMPLETED = "PUBLISH_COMPLETED";
 export const PUBLISH_FAILED = "PUBLISH_FAILED";
+export const FETCH_DATE = "FETCH_DATE";
```
btzr-io: A better name for this ?

kauffj:
Merged, but changed rather substantially. See https://github.com/lbryio/lbry-app/commit/2b77c16cdd581069b10f67737e5648fd23e67ff7

The biggest change that ought to have happen in the original PR is that we wouldn't want to store a single instance of `publishedDate`. Even if you didn't want to store the full blocks, mapping block heights and/or claim IDs to dates would have been better.

PR #526, opened 2017-09-04
Improve editing UX (fix)

btzr-io:
### Added
Add an edit button inside `fileActions` menu ->   https://github.com/lbryio/lbry-app/issues/415

### Fixes
 Fix for https://github.com/lbryio/lbry-app/issues/467

### Todo
- [x] Test it
- [x] Set label button to `update` instead of `publish`.
- [x] Update changelog.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,9 @@ Web UI version numbers should always match the corresponding version of LBRY App
 
 
 ### Changed
+  * Improved UX for editing claims.
   *
-  *
+
 
 ### Fixed
   * Some CSS changes to prevent the card row from clipping the scroll arrows after the window width is reduced below a certain point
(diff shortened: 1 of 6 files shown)


kauffj:
This is also merged into branch `v16` and will go out in the next update. This is a solid upgrade so I'd be okay releasing as-is, especially given the design sloppiness in publishing at the moment. 

But the following design changes ought to happen if you'd be interested to try them:

- Publish form accepts pre-filled state from outside
- Publish page handles the edit logic and passes pre-fill state to the form
- Publish form drops the logic handling loading values. The link on a name clash simply becomes a navigate to the proper URL.
- Navigation logic is claim ID based rather than name based (we want to support multiple publishes to the same name, though we don't atm).
- Update the title displayed in the address bar based on the action being an edit.

#540

PR #527, opened 2017-09-04
Better Deposit Help Description

MSFTserver:
better description of the deposit amount had a couple people asking what its for and why it doesnt work when they set it to 0


MSFTserver:
https://github.com/lbryio/lbry-app/issues/528

kauffj:
Tweaked your suggestion slightly in b8dd1fc, but I don't think this fully addresses #528. Closing PR, leaving issue open.

PR #533, opened 2017-09-06
Fixed menu issue when shrunk to smallest size

MSFTserver:
## Gif of the fix:
https://spee.ch/5/resize-bug.gif

## Before:
image

## After:
image


btzr-io:
This won't work for new features,  but I guess it's ok as a temp fix.

kauffj:
I think this fix is too crude and am going to reject it.

Another crude but more palatable fix would be to just increase the min width (are people really using it at such small resolutions? if so we can consider this unsupported for now).

A real solution is presumably to improve the responsive behavior here, but that seems like a really low priority.

PR #535, opened 2017-09-06
Fixed unwanted modal showing up in Publish Page (#534)

### LINKED ISSUE #534: Developer Reload on Publish Page Forces Pop-up
## The Issue
when on the Publish page and you proceed to hit developer -> reload it will refresh the page with a More Credits Required Pop-up
### Steps to reproduce
1. Go to Publish Page
2. Developer -> Reload
3. More Credits Required Pop-up

### Expected behaviour
i expect the page to normally refresh
### Actual behaviour
the page refreshes to More Credits Required Pop-up

## System Configuration

 

- LBRY Daemon version: 0.15.1
- LBRY App version: 0.15.0
- LBRY Installation ID: 4xUNfNEDfK5irh7vPgMgvBUoxjAosAmJK4m5n5idPiw6hMGmj8feMYKWWvcfAokPM8
- Operating system: Windows (Windows-2012ServerR2-6.3.9600)

## Screenshots

https://spee.ch/0/Publish-Page-Refresh-Error.gif

hackrush01:
Fixes #534

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
 
 ### Fixed
   * Some CSS changes to prevent the card row from clipping the scroll arrows after the window width is reduced below a certain point
-  *
+  * Fixed unwanted modal showing up in Publish Page (#534)
 
 ### Deprecated
   *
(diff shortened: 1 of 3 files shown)


kauffj:
Using https://github.com/lbryio/lbry-app/commit/b32f3cbafee12ca7cb29b5fab4e115e5b8767801 insead.

1. No need to `|| undefined`, plus this turns 0 into undefined.
2. Avoid changes like mapping `undefined` to 0 in `formatCredits`. Ought `formatCredits` actually support taking undefined? Probably not, and instead we should be rendering an undefined balance as something other than 0.

(For now, we will still render as 0 in the header, but we adjust this elsewhere, rather than in the format function itself.)

Issue #497, opened 2017-08-25
Allow purchase confirmations to be bypassed


kauffj:
As a user, I want to be able to not be asked to confirm extremely cheap purchases. Especially after #490 removes this check for prices below 0.01.

This PR would involve:
- Adding a setting to track a price below which the user is not shown a confirmation dialog on purchase
- Re-factor settings page to group this setting with maximum purchase price.
- Update `doPurchaseUri` action to not display the modal when making a purchase below the price

My first proposed name for this setting is "Single-Click Purchase Price" or "Automatic Purchase Price" but I do not really like either of these.

btzr-io:
I think amazon have something similar...

btzr-io:
![amazon](https://user-images.githubusercontent.com/14793624/29743814-217c9602-8a56-11e7-9375-f457f88c14e5.jpeg)

PR #541, opened 2017-09-08
Instant Purchase

alexliebowitz:
Solves #497.

### DIFF
--- a/ui/js/actions/content.js
+++ b/ui/js/actions/content.js
@@ -1,4 +1,5 @@
 import * as types from "constants/action_types";
+import * as settings from "constants/settings";
 import lbry from "lbry";
 import lbryio from "lbryio";
 import lbryuri from "lbryuri";
@@ -322,26 +323,55 @@ export function doPurchaseUri(uri) {
     const downloadingByOutpoint = selectDownloadingByOutpoint(state);
     const alreadyDownloading =
       fileInfo && !!downloadingByOutpoint[fileInfo.outpoint];
-    const costInfo = makeSelectCostInfoForUri(uri)(state);
-    const { cost } = costInfo;
 
-    if (
-      alreadyDownloading ||
-      (fileInfo && fileInfo.completed && fileInfo.written_bytes > 0)
-    ) {
-      return;
+    function attemptPlay(cost, instantPurchaseMax = null) {
+      if (!instantPurchaseMax || cost > instantPurchaseMax) {
+        dispatch(doOpenModal(modals.AFFIRM_PURCHASE, { uri }));
+      } else {
+        dispatch(doLoadVideo(uri));
+      }
     }
 
     // we already fully downloaded the file.
-    if (
-      cost === 0 ||
-      (fileInfo && (fileInfo.completed || fileInfo.download_directory))
-    ) {
-      return dispatch(doLoadVideo(uri));
+    if (fileInfo && fileInfo.completed) {
+      // If written_bytes is false that means the user has deleted/moved the
+      // file manually on their file system, so we need to dispatch a
+      // doLoadVideo action to reconstruct the file from the blobs
+      if (!fileInfo.written_bytes) dispatch(doLoadVideo(uri));
+
+      return Promise.resolve();
     }
 
+    // we are already downloading the file
+    if (alreadyDownloading) {
+      return Promise.resolve();
+    }
+
+    const costInfo = makeSelectCostInfoForUri(uri)(state);
+    const { cost } = costInfo;
+
     if (cost > balance) {
-      return dispatch(doOpenModal(modals.INSUFFICIENT_CREDITS));
+      dispatch(doOpenModal(modals.INSUFFICIENT_CREDITS));
+      return Promise.resolve();
+    }
+
+    if (
+      cost == 0 ||
+      !lbry.getClientSetting(settings.INSTANT_PURCHASE_ENABLED)
+    ) {
+      attemptPlay(cost);
+    } else {
+      const instantPurchaseMax = lbry.getClientSetting(
+        settings.INSTANT_PURCHASE_MAX
+      );
+      if (instantPurchaseMax.currency == "LBC") {
+        attemptPlay(cost, instantPurchaseMax.amount);
+      } else {
+        // Need to convert currency of instant purchase maximum before trying to play
+        lbryio.getExchangeRates().then(({ lbc_usd }) => {
+          attemptPlay(cost, instantPurchaseMax.amount / lbc_usd);
+        });
+      }
     }
 
     return dispatch(doOpenModal(modals.AFFIRM_PURCHASE, { uri }));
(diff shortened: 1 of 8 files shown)


alexliebowitz:
@kauffj 

The entry in the settings page is mostly done. Still needs some state management fixes, and will need feedback on the UI as well:
screen shot 2017-09-08 at 1 09 26 am

screen shot 2017-09-08 at 1 09 21 am

- Any thoughts/corrections on the wording?
- Do we want the default currency to be LBC or dollars? Should dollars even be an option? We seem to prefer LBC for tiny amounts like this.
- You said this and the max key fee should be grouped together in one section. If we do that we're going to need to figure out how we want it to look, and probably add new styles to deal with the concept of sub-sections within tiles.

ui/js/actions/content.js
```diff
@@ -342,8 +342,23 @@ export function doPurchaseUri(uri, purchaseModalName) {
...
+    /**
+     * Play video without a price warning dialog if:
+     *  - The file is free
+     *  - Instant purchase is enabled and the file is cheap enough
+     *  - The file is already downloading
+     */
+    const instantPurchaseEnabled = lbry.getClientSetting(
+      "instantPurchaseEnabled"
+    );
+    const instantPurchaseMax = lbry.getClientSetting("instantPurchaseMax")[
```
alexliebowitz: If we want to allow the max instant purchase value to be denominated in USD, then we'll have to do a currency conversion here. If not, then we'll have to extend FormFieldPrice to disable the currency selector and always use LBC (very easy).

alexliebowitz:
@kauffj: Once we deal with the above UI issues and the LBC vs. USD issue, this is ready to go. I added an inline comment about the currency stuff.

ui/js/actions/content.js
alexliebowitz: If we want to make it possible to have the max instant purchase price be denominated in USD, then we'll have to do a currency conversion here. If not, then we'll have to add an option to FormFieldPrice to disable the currency selector and always use LBC (very easy).

kauffj: I believe the app already has the ability to do client side currency conversion, so it shouldn't be too bad.

kauffj: Also, make sure to put these in `const/settings.js`

kauffj:
- USD should be an option.
- Yes, group them. The headings become the labels. The new heading is "Purchase Settings". No new styles are necessary, these are just two form rows with labels, the styles that already exist work fine.
- Proposed radio choices:
- [] "Ask for confirmation of all purchases"
- [] "Single-click purchasing of content less than..."

ui/js/actions/content.js
alexliebowitz: Oh right, you'd think I'd remember that we can do client side currency conversion, given that I wrote the code for it :P Will do.

alexliebowitz: Oh, right. OK, I'll make it handle USD. You'd think I would remember that we have client-side currency conversion, given that I wrote the code for it...

alexliebowitz:
So this creates a few more wording issues. The Max Purchase Price section will need to be reworded as well since the options don't make sense without a header:

screen shot 2017-09-16 at 10 03 08 pm

Also, currently the max purchase price section has a totally separate wording when the option is not chosen ("Choose limit"). It sounds like you're saying the new section would have the same wording either way, but the "..." changes into a price field when you choose the option. I guess we would want to keep that consistent betweeen both.


Here's one possibility just as a starting point:

**Purchase Settings**
- No maximum price when downloading content
- Don't download content more than... / Don't download content more than [price field]

 --

- Ask for confirmation of all purchases
- Single-click purchasing of content less than... / Single-click purchasing of content less than [price field]

An upshot of this wording is that neither of them would need helper text to make sense.

ui/js/actions/content.js
```diff
@@ -342,16 +351,28 @@ export function doPurchaseUri(uri, purchaseModalName) {
...
-    if (cost > balance) {
-      dispatch(doOpenModal(modals.INSUFFICIENT_CREDITS));
+    if (
+      cost == 0 ||
+      !lbry.getClientSetting(settings.INSTANT_PURCHASE_ENABLED)
+    ) {
+      attemptPlay(cost);
     } else {
-      dispatch(doOpenModal(purchaseModalName));
+      const instantPurchaseMax = lbry.getClientSetting(
```
kauffj: Settings should be read `selector/settings.js`.

ui/js/actions/content.js
```diff
@@ -342,16 +351,28 @@ export function doPurchaseUri(uri, purchaseModalName) {
...
+        settings.INSTANT_PURCHASE_MAX
+      );
+      if (instantPurchaseMax.currency == "LBC") {
+        attemptPlay(cost, instantPurchaseMax.amount);
+      } else {
+        // Need to convert currency of instant purchase maximum before trying to play
+        lbryio.getExchangeRates().then(({ lbc_usd }) => {
```
kauffj: Should this be it's own conversion promise? Wonder if there is a clean way to do this without a promise.

ui/js/page/settings/view.jsx
```diff
@@ -18,6 +18,8 @@ class SettingsPage extends React.PureComponent {
     this.state = {
       // isMaxUpload: daemonSettings && daemonSettings.max_upload != 0,
       // isMaxDownload: daemonSettings && daemonSettings.max_download != 0,
+      instantPurchaseEnabled: lbry.getClientSetting("instantPurchaseEnabled"),
+      instantPurchaseMax: lbry.getClientSetting("instantPurchaseMax"),
```
kauffj: Make constants.

ui/js/page/settings/view.jsx
```diff
@@ -199,6 +217,41 @@ class SettingsPage extends React.PureComponent {
               )}
             </div>
           </div>
+          <div className="card__content">
+            <div className="form-row">
```
kauffj: These should probably be `FormRow` components, and definitely need labels.

ui/js/actions/content.js
alexliebowitz: It seems like there's some kind of caching going on in `lbryio.getExchangeRates()`, but even then, there's no guarantee that the exchange rates were refreshed recently. We could have it automatically update into the cache at intervals; then we could change `lbryio.getExchangeRates()` to be synchronous and directly return a result from the cache (and/or add a function like `lbryio.convert()` that does the same). Otherwise there's no guarantee that we won't have to go out to the server for this call, so there needs to be a promise or callback.

ui/js/page/settings/view.jsx
alexliebowitz: I converted the "disabled" option to a `FormRow`. For the second option (where you choose the actual setting), I left it as a div for now. If made it a `FormRow`, we would have to stufff the price selector inside the label prop, which feels awkward. The Max Key Fee section also uses a div, I'm guessing for the same reason. Let me know if you'd rather it be inside the label, it's very quick fix.

I added helper text in the most recent commit; let me know if you have any adjustments on the wording.

kauffj:
@alexliebowitz made very few additional changes to this, basically just adding labels. You can see here https://github.com/lbryio/lbry-app/commit/b24b6e2e8e4b7ab7750f82776cb2f5cb641201be

Man do we need better form components!

PR #545, opened 2017-09-09
added color to echo's to stand out more(build.sh)

MSFTserver:
[How To Use Colors In Bash](https://docs.google.com/document/d/145mpTLHAQW2RgoWkLflyF6fYdgLxpYwPcdMD7SOwIJA/edit?usp=sharing)
also is there point to having the set -e at the top it defeats the purpose of the rest of the set -e wrappers you guys put together

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -14,7 +14,7 @@ if [ "$(uname)" == "Darwin" ]; then
 elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
   LINUX=true
 else
-  echo "Platform detection failed"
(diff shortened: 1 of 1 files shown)


MSFTserver:
@lyoshenka may i ask why we use set -x? that jsut print everything to the console before it executes would it be better to wrap what you want to print not everything ` set -x` ... `set +x` work like the `set -e`

lyoshenka:
I think it was easier for debugging to just show everything. Personally I like seeing the whole thing. We could definitely have the script give better output, but someone would have to go through it and make sure it's printing all the right things.

Issue #516, opened 2017-08-30
Cyan Screen on Launch


gyroninja:
If you launch LBRY when your working directory isn't in lbry-app then the application says at a blank cyan screen.
My current work around is just creating a start script which changes the working directory to there and starts it up.

tzarebczan:
What working directory are you referring to? Can you be more specific? I can provide a different solution tomorrow when I'm at my PC

gyroninja:
If you are running it via `node_modules/.bin/electron app`, you have to have your working directory be the root directory of the git repo. If you have packaged it up, you have to have your working directory be `/dist/linux-unpacked`.
This is likely due to it just not being able to find assets that it assume will be located in the working directory.
I don't really need a different solution, but would rather see this bug fixed.

lyoshenka:
@gyroninja how are you packaging it? for the unpackaged version, it makes sense that this is a bug. for the packaged version, it should create an installable file (a .deb or .dmg or similar) that gets installed by the system. I'm not sure where /dist/linux-unpacked is coming from.

If you run `node_modules/.bin/electron app` but with full paths to the binary and the app dir, does this bug still happen? If it's just about relative paths in the command, its easy to fix. If the paths are absolute and it still errors, then its something we need to fix in electron.

gyroninja:
I don't know how many times I need to explain it when it is so simple.
No, I'm not installing it from a dpkg.
dist/linux_unpacked is created when it creates the package
Whether the command uses full or relative paths doesn't matter.

**Like I have said in every post so far, the bug is from having the wrong working directory. The bug very likely stems from using a resource that it assumes exists it the working directory. Since the working directory is not where it is expected, it can not find the resource and hence fails to load.**

lyoshenka:
I was asking because I'm having trouble reproducing this. I've never seen the `linux_unpacked` dir and I don't know how it is created. It would be helpful if we knew what OS you're on and what commands you ran. It would be even more helpful if you submit a PR fixing this issue. We're happy to tip LBC for dev work. :smile:

gyroninja:
Okay, I went ahead and diagnosed and resolved the issue I was getting. You were probably having issues reproducing this because the code which caused this error was being avoided. (There was a return inserted to the beginning of the function which errors to temporarily disable it)

PR #547, opened 2017-09-10
Get the app directory via an absolute path instead of a relative one.

gyroninja:
In the internationalization code it is assumed that the app directory is in the current working directory.
This causes problems when the current working directory is not where it is expected. Before some of the internationalization code was temporarily disabled (a return was inserted at the begging of the doDownloadLanguages function), on startup it would get stuck and display a pure cyan screen. In the background it would start the daemon and everything, but it would not display anything but the cyan background. This was reported by me in #516.
This PR proposes that instead of assuming that the app folder for the application is in our current working directory, we grab the path of the app folder directly from Electron itself. We grab this directory by using Electron's remote module, which is made to be used by the render thread, to grab an instance of the Electron app object to get the path that we need.

### DIFF
--- a/ui/js/actions/settings.js
+++ b/ui/js/actions/settings.js
@@ -4,7 +4,6 @@ import batchActions from "util/batchActions";
 import lbry from "lbry";
 import fs from "fs";
 import http from "http";
-import { remote } from "electron";
 
 export function doFetchDaemonSettings() {
   return function(dispatch, getState) {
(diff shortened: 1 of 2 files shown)


kauffj:
Thanks @gyroninja! Please DM @kauffj in our Slack or email jeremy@lbry.io if you'd like a tip for this PR.

PR #549, opened 2017-09-10
Publish Drag-and-Drop

### LINKED ISSUE #350: Support drag-and-drop file publishing
## The Issue
On Publish, dragging a file you want opens the file inside the electron app rather than into the file selection box, subsequently requiring app closure to reset.
### Steps to reproduce
1. Go to Publish
2. Drag/Drop the file you're going to publish from desktop into "Choose a file" field
3. View result

### Expected behaviour
Queue up file selection for upload

### Actual behaviour
File playback

## System Configuration

 

- LBRY Daemon version:
- LBRY App version: 13.0
- LBRY Installation ID:
- Operating system: Mac OS

hackrush01:
@kauffj 
- This is able to get the file's path(in the `source` state), but not able to display it in the `FormRow` compnent
- This also computes a name for the file based on path
- An overlay is shown when a file is dragged on the form window.(CSS is just a place-holder, since I was not sure about it)
- This also fixes the bug due to which dragging a file open it in the app.

Closes #350

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
 
 ## [Unreleased]
 ### Added
+  * Publishing now supports drag-and-drop file selection.
   * Add setting to automatically purchase low-cost content without a confirmation dialog
   *
 
--- a/ui/js/component/app/view.jsx
+++ b/ui/js/component/app/view.jsx
@@ -17,6 +17,16 @@ class App extends React.PureComponent {
       alertError(event.detail);
     });
 
+    document.addEventListener("drop", event => {
+      event.preventDefault();
+      event.stopPropagation();
+    });
+
+    document.addEventListener("dragover", event => {
+      event.preventDefault();
+      event.stopPropagation();
+    });
+
     if (!this.props.upgradeSkipped) {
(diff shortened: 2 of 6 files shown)


kauffj:
This likely isn't going to make it in until the new app devs start. It'll require a decent refactor and re-design.

Removing myself as a reviewer for now.

liamcardenas:
@hackrush01 what is the status of this?

kauffj:
This isn't merge-able as-is. It possibly should block fully on https://github.com/lbryio/lbry-app/issues/540

Otherwise, at a minimum, it needs to be implemented as a `<FormField>` or `<FormRow>`.

tzarebczan:
@hackrush01 what's the status on this? The problem right now is that if you drag and drop something into the app, you need to restart it because it opens the file inside the app window. If this won't get done for a while, we should at least disable that somehow.

hackrush01:
@tzarebczan sorry for the late response. This feature will be completely re-done. Until then the default behaviour of electron will remain as is, i.e. open the open image(in a way similar to browser behaviour). This will likely be a while. I agree this should be disabled in-app for the time being.

seanyesmunt:
Closing this since it doesn't seem like it will be finished.

PR #550, opened 2017-09-11
Form component

### LINKED ISSUE #542: Bind enter key press to form submission as appropriate
## The Issue
If you try to use the Enter key to send credits on the Send screen or send an invite on the Invites screen, the page reloads and nothing happens

### Steps to reproduce
1. Go to Send screen, enter amount and address, press Enter
2. Go to invite screen, enter email to invite, press Enter
3.

### Expected behaviour
Enter key should be allowed to confirm action

### Actual behaviour
Enter key refreshes page. 

## System Configuration

 

- LBRY Daemon version: 0.15.1
- LBRY App version: 0.15.1
- LBRY Installation ID:
- Operating system: Windows

btzr-io:
- Implement form component.
- Implement submit component.
- Fix weird behavior in wallet send.
- Prevent refresh on enter key.
- Submit form on enter key in `Send`, and `Invite` page.

### DIFF
--- a/ui/js/component/inviteNew/view.jsx
+++ b/ui/js/component/inviteNew/view.jsx
@@ -1,7 +1,6 @@
 import React from "react";
 import { BusyMessage, CreditAmount } from "component/common";
-import Link from "component/link";
-import { FormRow } from "component/form.js";
+import { Form, FormRow, Submit } from "component/form.js";
 
 class FormInviteNew extends React.PureComponent {
   constructor(props) {
@@ -18,16 +17,16 @@ class FormInviteNew extends React.PureComponent {
     });
   }
 
-  handleSubmit(event) {
-    event.preventDefault();
-    this.props.inviteNew(this.state.email);
+  handleSubmit() {
+    const { email } = this.state;
+    this.props.inviteNew(email);
   }
 
   render() {
     const { errorMessage, isPending } = this.props;
 
     return (
-      <form>
+      <Form onSubmit={this.handleSubmit.bind(this)}>
         <FormRow
           type="text"
           label="Email"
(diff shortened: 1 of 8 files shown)


btzr-io:
@kauffj any thoughts on this ? :upside_down_face:

ui/js/component/inviteNew/view.jsx
```diff
@@ -27,7 +27,7 @@ class FormInviteNew extends React.PureComponent {
     const { errorMessage, isPending } = this.props;
 
     return (
-      <form>
+      <form onSubmit={event => this.handleSubmit(event)}>
```
kauffj: If every form needs to call `event.preventDefault()` to prevent enter based submission from reloading the page, maybe we should just make a `Form` component that takes a required submit property? 

Then this component can handle preventing the event, without every individual form being required to remember to add this.

ui/scss/component/_button.scss
```diff
@@ -84,4 +84,10 @@ $button-focus-shift: 12%;
 .button--flat
 {
   box-shadow: none !important;
-}
\ No newline at end of file
+}
+
+.button-submit {
```
kauffj: Why do we need this? And it should probably be `button--submit` if this is a specific state or type of a button.

btzr-io: the current styles are for `<a></a>` and don't work with an actual `<button></button>` ^^

ui/js/component/inviteNew/view.jsx
btzr-io: Yeah, I should have done that... :stuck_out_tongue: 
 here?  -> https://github.com/lbryio/lbry-app/blob/master/ui/js/component/form.js

btzr-io: maybe move that in to `components/forms` ?

PR #559, opened 2017-09-13
compiling dependencies installer script (prebuild-dep.sh)

MSFTserver:


lyoshenka:
There's also already a [prebuild.sh](https://github.com/lbryio/lbry-app/blob/master/build/prebuild.sh) script. This looks like its 80% a duplicate of that. Can the two be combined?

MSFTserver:
yea i guess the two can be combined i just know the pre-build didnt get me everything i needed, thats more if you already have most of the common tools there

MSFTserver:
ill see about touching up the pre-build with missing deps that we should check for

PR #576, opened 2017-09-18
Improve tabs ( UI / UX )

btzr-io:
### Changes
- Improve tabs style.
- Add animation for active tab.

### Before
old-tabs

### After
 new-tabs


btzr-io:
I think this can still make it to v16 :upside_down_face: ^

btzr-io:
@kauffj  should I close this ? ^

PR #577, opened 2017-09-18
V16

kauffj:
This is @btzr-io's theme work and edit fix, @hackrush01's transaction and tipping changes, and then a whole bunch of refactoring, fixes, and overall changes by me.

Known issues:
- Transaction list is really slow.
- It's weird that you get a bunch of information about transactions, but not purchases.
- Editing is still pretty confusing. You don't receive feedback on changes, and there's issues with several fields.


kauffj:
The feedback above was for Amit, so he understood what I changed and why with regards to his transaction list.

Issue #579, opened 2017-09-18
Keep video playing when navigating from a playing video


kauffj:
When the user navigates from a page with a playing video, this video can be displayed in a picture-in-picture-esque popup, ala Twitch.

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

btzr-io:
I don't think this is gonna be that easy but I'll try :stuck_out_tongue:

kauffj:
1. Put component `VideoOverlay` (or similar) in `App` inside of `"main-content"` div.
1. `VideoOverlay` renders a video iff `selectPlayingUri` and not on a show/file page.
1. Remove/modify cancelPlay in unmount of `Video`, so that the playing URI is not cleared when the user navigates away from a video.
1. Add cancel/x/close option to `VideoOverlay` that clears the playing URI (and thus closes the overlay)
1. Add expand/maximize option that either takes the user to show, expands the video to full screen, or possibly both.

btzr-io:
Also needs an autoplay feature and let's not forget about this one: https://github.com/lbryio/lbry-app/issues/491

skhameneh:
I'm on it, maybe I'll have something this weekend. 😄

skhameneh:
Notes

1. Inserting another video player with a resumed timestamp, outside of the page's DOM context, will introduce an interruption of video playback.
2. Storing DOM nodes in Redux is ill-advised.  All content within the Redux store should be fully JSON compatible.
3. Moving playback DOM nodes outside of the embedded video's containers will introduce layout complexities. (The playback elements must scroll with content while on the video's source page)
4. Some interactions may require manual HTML entry for manual DOM manipulation, I will need to see is node placement changes can be done without interrupting playback. See https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

tzarebczan:
@dan1d was taking a look at this one too... Not sure if he started yet.

Old pr: https://github.com/lbryio/lbry-app/pull/585

dan1d:
[...]

PR #585, opened 2017-09-19
Overlay component

btzr-io:
### Changes
Keep video playing when navigating from a playing video #579
Play / stream videos while searching https://github.com/lbryio/lbry-app/issues/393

### Todo
- [X] Basic ui / animations.
- [X] Fix animations.
- [X] Basic logic.
- [X] Fix scrollbar issue.
- [x] Fix history duplication.
- [x] Cleanup logic.
- [x] Close button that clears the playing URI and hide / close the component.
- [ ] Remove / modify `cancelPlay` in unmount of Video, to keep the playing URI.
- [ ] Improve / fix structure layout and css animations.
- [ ] Swipe to remove ( youtube style )
- [ ] Animate components on `unmount` and `mount`.
- [ ] tidy up / optimize.

~Minor tweaks for video component ( no-controls )~
~Remove / modify `cancelPlay` in unmount of Video, to keep the playing URI~

Check this: https://github.com/lbryio/lbry-app/issues/579#issuecomment-330282722


ui/js/actions/app.js
```diff
@@ -219,3 +219,15 @@ export function doChangeVolume(volume) {
     });
   };
 }
+
+export function doOpenOverlayMedia() {
```
kauffj: I suspect this code can be written without this action. We can know whether or not it should be open solely from existing app state.

ui/js/actions/app.js
```diff
@@ -219,3 +219,15 @@ export function doChangeVolume(volume) {
...
+  return {
+    type: types.OPEN_OVERLAY_MEDIA,
+  };
+}
+
+export function doCloseOverlayMedia() {
```
kauffj: I also suspect this action can be avoided, and instead just be `doSetPlayingUri(null)`.

ui/js/component/app/view.jsx
```diff
@@ -53,6 +54,7 @@ class App extends React.PureComponent {
         <Theme />
         <Header />
         <div id="main-content">
+          <OverlayRouter />
```
kauffj: This should not be a router. This should just be `VideoOverlay`.

ui/js/component/overlayRouter/view.jsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  render() {
+    const { currentPage, params, playingUri, keepPlaying } = this.props;
+
+    const uri = params.uri || playingUri;
+    const isPage = currentPage === "show";
+
+    const page = uri && (keepPlaying || isPage)
+      ? <ShowPage expanded={isPage} uri={uri} />
```
kauffj: The `ShowPage` should not be doing double duty of being both a full page and the overlay component. Allowing them to each be their own function will be a much cleaner design.

ui/scss/component/_overlay-media.scss
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  &.minimized {
+    width: 100% !important;
+    position: fixed;
+    top: 0;
+    left: 0;
+    /*bottom: $inner-height + $inner-margin;*/
+    transform: translateY(calc(100vh - 248px));
+    z-index: 99;
```
kauffj: Right now our z-index is very clean:

- 3 layers (1, 2, 3) in the app.
- 2 layers (9999, 10000) for modals and snackbars, respectively.

I suspect z-index: 3 here would be fine, or at most 4.

btzr-io: Ok, I need rewrite styles anyway :+1:

kauffj:
Noooooooooooooooooo it was so close!

PR #646, opened 2017-10-02
update disabled reward text

tzarebczan:
This message appears when a user's account is manually disabled (rewards abuse, etc).  We don't reach out to them via email, so we should ask them to reach out if they are actually legitimate users.


kauffj:
Great change, love that you are looking out for this stuff!

Issue #613, opened 2017-09-25
Unresponsive header


btzr-io:
## The Issue
As pointed out by some users at slack:

![image](https://user-images.githubusercontent.com/14793624/30791119-a83e7a38-a16d-11e7-8a4a-a7284c642c2f.png)
> When it's shrunk the text pops out and you get, well, that

### Steps to reproduce
1. Resize window to `min-width`.

### Expected behaviour
The header should be responsive and don't have a weird behavior 
when you resize the window.

btzr-io:
Fixed ^

tzarebczan:
Thanks @btzr-io !

PR #574, opened 2017-09-17
Improve scrollbar ( UI / UX ) 

btzr-io:
### Changes
You can check the `slack` app ( one of the best apps out there using electron ) for reference:
- The scrollbar don't overlap the header, and have a simple style ( less annoying )
- Add more focus on the main content ( less distractions )
- Customization for different themes ( eg Light, Dark ... )

### Before  /  After
 old new-scrollbar

### Todo
- [x] Update Changelog

PR #647, opened 2017-10-02
Minor redesign (css)

btzr-io:
> ### A user interface is like a joke. If you have to explain it, it’s not that good.

### Changes
A bunch of minor `CSS` fixes and additional changes.
- [x] Update themes.
- [x] Tabs (redesign) #576
- [x] Custom styled scrollbar #574 
- [x] Add more focus on search.

### Fixes
- [x] Fix: unresponsive header #613 
- [x] Fix: Dark theme issues
- [x] Minor Fix: https://github.com/lbryio/lbry-app/issues/620
- [x] Minor issue with card-row header.
- [x] Minor css fixes.

### Additional changes
- [x] Header redesign (css).
- [x] Search bar UX / UI enhancement (css).
- [ ] Checkbox redesign (css).
- [ ] Radio button redesign (css).
- [x] Minor update for tile component (css)
- [x] Minor update for report button. (icon)
- [x] Minor update for support button. (css)
- [x] Minor update for edit button. (css)
- [x] Minor update for file actions. (css)
- [x] File input (css).
- [x] Cleanup Wallet page (css).

### Todo
- [ ] Update changelog.


btzr-io:
@kauffj I re-open this because -> #639 would take a while to get merged,
You can merge this right now :upside_down_face:

btzr-io:
### Additional changes
- [ ] Search bar UX / UI enhancement.
- [ ] Checkbox redesign (css).
- [ ] Radio button redesign (css).
- [ ] File input (css).
- [ ] Update changelog.

ui/js/component/file-selector.js
```diff
@@ -47,16 +52,31 @@ class FileSelector extends React.PureComponent {
...
+                ? __("Choose File")
+                : __("Choose Directory")}
+            </span>
+          </span>
         </button>
         {" "}
         <span className="file-selector__path">
-          {this.state.path ? this.state.path : __("No File Chosen")}
+          <input
+            className="input-copyable"
```
kauffj: I understand the text is selected to make this copyable, but it also gives the user an implication that they can type in this field. Can we somehow express/provide both?

CHANGELOG.md
```diff
@@ -9,10 +9,15 @@ Web UI version numbers should always match the corresponding version of LBRY App
 ## [Unreleased]
 ### Added
   * Add setting to automatically purchase low-cost content without a confirmation dialog
+  * New custom styled scrollbar (#574)
```
kauffj: I like overall style and that it doesn't overlap header, but can it be wider and/or get wider on hover?

ui/js/component/fileTile/view.jsx
```diff
@@ -126,17 +126,22 @@ class FileTile extends React.PureComponent {
...
                 </h3>
               </div>
               {description &&
                 <div className="card__content card__subtext">
-                  <TruncatedText lines={!showActions ? 4 : 2}>
+                  <TruncatedText lines={2}>
                     {description}
                   </TruncatedText>
                 </div>}
+              <div className="card__content card__subtext">
```
kauffj: I agree we ought to make a change here to give a stronger indication how to open these.

But what if the change should just be to match card behavior of discover page? These card elements are really quite similar, so we should just be consistent in the effect/UI we provide.

ui/scss/_gui.scss
```diff
@@ -7,7 +7,7 @@ html
 body
 {
   color: var(--text-color);
-  font-family: 'Source Sans Pro', sans-serif;
+  font-family: 'Roboto', sans-serif;
```
kauffj: Bold move. Is this to move us inline with MD? Or do you just think this is appropriate font?

ui/scss/component/_file-tile.scss
```diff
@@ -19,4 +19,20 @@ $height-file-tile: $spacing-vertical * 6;
   .card__title-primary {
     margin-top: 0;
   }
-}
\ No newline at end of file
+  .meta {
```
kauffj: I'm not sure we need to retain this if we match the behavior of discover, but if there is a new styling here specific to `file-tile__row`, we ought to make it a BEM class name and drop `meta`.

ui/scss/component/_header.scss
```diff
@@ -14,11 +16,14 @@
   box-sizing: border-box;
 }
 .header__item {
-  flex: 0 0 content;
   padding-left: $spacing-vertical / 4;
   padding-right: $spacing-vertical / 4;
   .button-alt {
       background: var(--header-button-bg) !important;
+      font-size: 17px;
```
kauffj: Does this need to be abstracted?

ui/scss/component/_tabs.scss
```diff
@@ -0,0 +1,57 @@
+/* Tabs */
+
+nav.sub-header
```
kauffj: Like this element!

ui/scss/component/_tabs.scss
```diff
@@ -0,0 +1,57 @@
...
+  > a
+  {
+    height: 38px;
+    line-height: 38px;
+    text-align: center;
+    font-weight: 500;
+    text-transform: uppercase;
+    display: inline-block;
+    vertical-align: baseline;
+    margin: 0 8px;
```
kauffj: Would consider making this a little larger?

ui/scss/mixin/link.scss
```diff
@@ -9,8 +9,8 @@
     }
   }
 
-  &:not(.no-underline) {
-    text-decoration: underline;
+  &:not(.underline) {
+    text-decoration: none;
```
kauffj: I think this change added underlining to text button icons (e.g. the ones in `FileActions`). I think this can look poor with some icons, so should just be avoided.

ui/scss/_gui.scss
btzr-io: lbry it's already usign `MD` so yes I think it's  appropriate.

ui/js/component/fileTile/view.jsx
btzr-io: > add experimental fileTile styles 87c4bd2 
Ok I was just testing / experimenting with some stuff but yeah I'll remove it

ui/scss/component/_file-tile.scss
btzr-io: > add experimental fileTile styles 87c4bd2
Ok, I'll remove this, was just a simple experiment :+1:

ui/scss/component/_tabs.scss
btzr-io: Ok ^

ui/scss/mixin/link.scss
btzr-io: ok, I'll revert this, but at least header can't have underline text :upside_down_face:

ui/scss/_gui.scss
btzr-io: This PR isn't dropping `MD`, is doing some enhancement,
this would probably change when we implement the @nizuka01 `design`

ui/js/component/fileTile/view.jsx
btzr-io: Agree, but a tile and a card are two different things...

btzr-io: >  we should just be consistent in the effect/UI

btzr-io: Would you considering removing the `uri` from the tile component ?

btzr-io: we could add a menu / or a set of actions to the tile component...

btzr-io: a small button `Get uri` or `Share`

btzr-io:
dark-uri
@kauffj  ? ^

btzr-io:
### More notes
- Currently the report button seems hard to noticed, and looks like don't belong there...

ui/js/component/file-selector.js
btzr-io: I don't think the user should be allowed to type a directory / file, it's better letting the file dialog handle this to ensure that the file / directory exists, considering the current validaton system :stuck_out_tongue:

CHANGELOG.md
btzr-io: Ok, increased the size, if users start to report some problems with this 
I'll make it more wider, but I think it's ok for now

ui/js/component/file-selector.js
kauffj: I was unclear; I agree users should not be allowed to type in this field.

Is there a style we can apply that indicates BOTH copyable AND readonly?

kauffj:
@btzr-io I am fine with dropping the URI from `FileTile`. It would actually probably be pretty easy to add in `FileActions` to those cards... not sure if that's actually a good idea or not.

btzr-io:
> don't support your own claims :(  dd8ab64 

@kauffj I'm still confused about the support behavior:
Do we actually need to support our own claims ?

### Note
There are too many buttons in `fileActions`,I guess it's time to bring back the menu component or do something smart :stuck_out_tongue:

btzr-io:
### Duplicated content 
duplicated

btzr-io:
cleanup

kauffj:
@btzr-io I think dd8ab64 could cause user confusion ("why is the support button missing on my content?")

kauffj:
And yes... let's fix FileActions!

kauffj:
I merged this, since it's a clear step up from the current app. I opened two issues regarding outstanding changes that I think are still worth making:

https://github.com/lbryio/lbry-app/issues/656
https://github.com/lbryio/lbry-app/issues/657

PR #648, opened 2017-10-02
add changelog link

tzarebczan:
Add hyperlink to releases page which has the changelog. Had a complaint recently that we don't show what changed when prompting to update - kind of makes sense to have it here.


kauffj:
This was a good idea, but I wanted to change the message a bit, so I just did it in a separate commit here: b8980c00

Also, I noticed this PR added a `Link` component to the pop-up, but did not import the `Link` component. Please test your code locally now that you have a working environment.

tzarebczan:
Thanks and will be more thorough next time!

Issue #600, opened 2017-09-21
Cannot edit claim 2nd time around via Edit Claim button


tzarebczan:
## The Issue
The Edit Claim button doesn't seem to re-populate any claim data when trying to edit a claim that was recently edited.  Developer>Reload or clearing the cache don't help, need to restart the app again to be able to edit via the button. You are able to edit via the Publish screen after typing the URL > Copy existing claim data. 

Note- after the first edit, updated claim info is not shown on the claim page until the app is reloaded (via Developer Reload) - this issue is waiting on https://github.com/lbryio/lbry-app/issues/392 

### Steps to reproduce
1. Edit an existing claim, wait for blockchain confirmation
2. Try to edit the claim again using the Edit claim button
3. It goes to Publish page, but no data is populated

### Expected behaviour
Should be able to edit the claim

### Actual behaviour
Tell us what happens instead

## System Configuration

 

- LBRY Daemon version: 0.16.1
- LBRY App version: 0.16.0
- LBRY Installation ID:
- Operating system: Windows/Linux

btzr-io:
@tzarebczan I can't recreate this in last master, can you confirm this ?

btzr-io:
> Never mind, after the third edit I was able to recreate this issue

tzarebczan:
Per @btzr-io, if it does work and you get it to edit a 2nd time, it shows old claim data which in part may be related to https://github.com/lbryio/lbry-app/issues/392.  The real question is why is fileinfo is blank after an update sometimes. I'll see if I can reproduce with the debugging info. 

Some notes from Slack chat: 

not that simple, I found the main issue which is also causing other issues...
[12:16] 
 ```Object
fileInfo : null
outpoint: "024839f1be1f118de633361e894faf1cfff25d507160f58023bbdb7df8644592:0"
Object
type : "FETCH_FILE_INFO_COMPLETED"
```
[12:17] 
to edit a claim you need to pass `fileInfo.claimId`
[12:17] 
but as you can see is returning null after the `re-publish` :confused:
[12:18] 
this also prevents me to remove or unclaim it

https://github.com/lbryio/lbry-app/blob/master/ui/js/actions/file_info.js#L44
btzr [6 days ago] 
did some more deep debugging ^^
btzr [6 days ago] 
it fails when you search for a `file_list` and pass the `outpoint` of the updated claim

tzarebczan:
[...]

PR #649, opened 2017-10-04
Improve edit ux flow

btzr-io:
### Changes
- Disable edit button if no `fileInfo` exists.
- Display `updating...` state for edit button.
- Pass claim uri as fallback for `fileInfo.claimId`

### Fixes
- Fixes https://github.com/lbryio/lbry-app/issues/600 (more or less)

### DIFF
--- a/ui/js/component/publishForm/view.jsx
+++ b/ui/js/component/publishForm/view.jsx
@@ -187,7 +187,7 @@ class PublishForm extends React.PureComponent {
   }
 
   handleEditClaim() {
-    const claimInfo = this.claim() || this.myClaimInfo();
+    const claimInfo = this.myClaimInfo() || this.claim();
 
     if (claimInfo) {
       this.handlePrefillClaim(claimInfo);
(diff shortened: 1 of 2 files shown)


btzr-io:
I want a free t-shirt :stuck_out_tongue:

tzarebczan:
@btzr-io I assume you have the link already, but just in case: https://hacktoberfest.digitalocean.com/sign_up/register

ui/js/component/publishForm/view.jsx
```diff
@@ -166,8 +167,8 @@ class PublishForm extends React.PureComponent {
 
   claim() {
     const { claimsByUri } = this.props;
-    const { uri } = this.state;
-    return claimsByUri[uri];
+    const { uri, publishedUri } = this.state;
+    return claimsByUri[uri] || claimsByUri[publishedUri];
```
kauffj: Can you help me understand what the underlying reason we need this additional value is? This seems like a potentially dangerous approach, and ideally edit would be initiated with just a single value (presumably the claim id).

btzr-io:
@kauffj ok I'll close this since the real bug is coming from the daemon...

ui/js/component/publishForm/view.jsx
btzr-io: it's a fallback for `#600`, after re-publish you can't pass the `claim_id`

tzarebczan:
Sorry, I still need to look into this one to see why fileinfo is blank after edit.

PR #652, opened 2017-10-09
ShapeShift integration

seanyesmunt:
For https://github.com/lbryio/lbry-app/issues/609

Looks like this
screen shot 2017-12-04 at 12 25 40 pm
screen shot 2017-12-04 at 12 25 49 pm


kauffj:
My feedback:

- This component should be above "Where to Find Credits" on the "Receive" page.
- The "Receive" page currently shows as "Wallet Address" in the Wunderbar. It should probably just say Receive, and the icon should possibly change.
- Possibly the word "Receive" should change entirely.
- **This component should not be presented as ShapeShift**. This is a component for buying credits. Shapeshift is a detail. IMO the only place the word ShapeShift should appear is in help text. 
- This component should likely have a corresponding FAQ article. You can either create this yourself, or create an Issue to create the article and assign to @tzarebczan.
- The button should have the disabled style when clicked until it is done performing it's action (see "Generate New Address" for example of this)
- If a wallet address is for a user to copy, it should use the same input styling as we do for other addresses that allows easy copying.
- "Send up to XXX BTC to" after creating the shift is larger than it needs to be. This is just normal `<p>` text IMO.
- It's mildly confusing that one would need to click "New" if they wanted to cancel.
- **The component introduces more new styles than it needs to**. While we're not 100% at following Material Design, I think we ought to conform pretty closely to it at least until we have a redesign. Non-exhaustive list:
  - New inline text next to the header
  - "New" button in the top right
  - Most of the deposit component
- **I'm worried about the need to know the send address up front.** AFAIK CoinBase does not support this. Not sure about other exchanges.
- Form validation should not happen until a user types. Right now, it tells me invalid address if I focus the input and then unfocus.
- (Possibly not introduced in this PR: the styling on readonly inputs for copying text, like the wallet address, are broken.)

liamcardenas:
Looking at this now. By the way, you dont need to use your own forked repo for branches in github anymore. You can just branch off of this repo and push your branch to create a pull request.

I'll let you know my thoughts.

ui/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+    getActiveShift(shiftDepositAddress);
+    this.continousFetch = setInterval(() => {
+      getActiveShift(shiftDepositAddress);
+    }, 10000);
+  }
+
+  componentWillUpdate(nextProps) {
+    const { shiftState } = nextProps;
+    if (shiftState === "complete") {
```
kauffj: Shift states should probably be constants.

ui/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,107 @@
...
+  constructor() {
+    super();
+    this.continousFetch = undefined;
+  }
+
+  componentDidMount() {
+    const { getActiveShift, shiftDepositAddress } = this.props;
+
+    getActiveShift(shiftDepositAddress);
+    this.continousFetch = setInterval(() => {
```
kauffj: Should this be in the action?

ui/js/component/shapeShift/view.jsx
```diff
@@ -0,0 +1,123 @@
...
+                className="button-primary button-block"
+              >
+                New ShapeShift
+              </button>
+            </div>}
+        </div>
+
+        <div className="card__content">
+          {error && <div className="error-text">{error}</div>}
+          {loading && <div>Loading...</div>}
```
kauffj: Consider if `<BusyMessage>` is appropriate here

ui/js/util/form-validation.js
```diff
@@ -0,0 +1,35 @@
...
+export const coinRegexPatterns = {
+  BTC: /^[13][a-km-zA-HJ-NP-Z1-9]{25,34}$/,
+  BCH: /^[13][a-km-zA-HJ-NP-Z1-9]{25,34}$/,
+  ETH: /^(0x)?[0-9a-fA-F]{40}$/,
+  DASH: /^X([0-9a-zA-Z]){33}/,
+  LTC: /^[a-zA-Z0-9]{26,33}$/,
+  XMR: /^4[0-9ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{94}$/,
+};
+
+const validateAddress = (coinType, address) => {
```
kauffj: Great work starting a validation library! Let's determine a consistent signature for validation functions.

One that I've seen and like is that validators return sanitized values and except if they encounter an error.

So far, this is two inconsistent signatures, one that returns true/false, and one that returns an error set.

ui/scss/_gui.scss
```diff
@@ -188,3 +188,31 @@ section.section-spaced {
 {
   text-align: center;
 }
+
+.error-text {
+  color: var(--color-error);
+}
+
+.display-inline-block {
```
kauffj: I think these additions are likely not necessary. Just put on the components as relevant.

ui/js/util/redux-actions.js
```diff
@@ -0,0 +1,23 @@
+export const handleActions = (actionMap, defaultState) => {
+  return (state = defaultState, action) => {
+    const handler = actionMap[action.type];
+    if (handler) return handler(state, action);
+    return state;
+  };
+};
+
+export const createAction = type => {
+  return payload => {
```
kauffj: Let's make this data to be consistent.

ui/js/util/redux-actions.js
```diff
@@ -0,0 +1,23 @@
+export const handleActions = (actionMap, defaultState) => {
```
kauffj: What do you think of allowing this wrapper to handle the `Object.assign` that every reducer does? Something like...
```
export const handleActions = (actionMap, defaultState) => {
  return (state = defaultState, action) => {
    const handler = actionMap[action.type];
    const newState = handler ? handler(state, action) : {}
    return Object.assign({}, state, newState);
  };
};
```

In writing this, I noticed that the reducers you added in `shape_shift.js` do not include the `Object.assign` calls. Aren't those always necessary?

seanyesmunt: Not if using the spread operator `...`.
https://twitter.com/dan_abramov/status/688087202312491008

seanyesmunt: We could still add that to the `handleActions` function anyway. So we don't have to add `...state` in every reducer handler

kauffj: `Object.assign` isn't a deep clone either. `Object.assign({}, state, { foo: "bar"})` and `{ ...state, { foo: "bar" }` are functionally equivalent. 

(Correct me if above seems wrong to you.)

seanyesmunt: Yeah they are the same. I believe babel transpiles the spread operator into object.assign

ui/js/component/shapeShift/internal/active-shift.jsx
seanyesmunt: I think it should stay here

seanyesmunt:
Alright @kauffj do your worst. This code is starting to look a little wonky after looking it at so much so I'm sure there are a few things I missed/should fix. Working good though.

src/renderer/js/component/common/spinner.jsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import classnames from "classnames";
+
+export default ({ dark, className }) => {
+  return (
+    <div
+      className={classnames(
```
kauffj: @IGassmann @liamcardenas please be aware of this pattern/design for determining class names that @seanyesmunt has added.

src/renderer/js/component/link/view.jsx
```diff
@@ -41,16 +44,22 @@ const Link = props => {
     );
   }
 
+  // An acutal button is needed to use type="submit" for form submission
+  // this should probably just go off an "href" prop
+  // if we do that we will need to check all of the buttons that aren't supposed to look like buttons
+  const WrapperEl = actualButton ? "button" : "a";
```
kauffj: We have a `<SubmitButton>` component used elsewhere. Please use that instead.

(This code also has a styling error: the text weight when rendered as a button is less than that when rendered as an anchor.)

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,139 @@
...
+      doShowSnackBar,
+    } = this.props;
+
+    return (
+      <div className="shapeshift__active-shift">
+        {shiftState === statuses.NO_DEPOSITS && (
+          <div>
+            <p>
+              Send up to{" "}
+              <span className="font-bold">
```
kauffj: Can the existing `credit-amount` classes be used for this? I'm generally opposed to this type of classname-as-styling, though you can certainly find other examples in the code. If `credit-amount` does not work, then I would instead propose new CSS classes specifically for styling financial values, not `font-bold`.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,139 @@
...
+      shiftReturnAddress,
+      shiftOrderId,
+      shiftState,
+      shiftDepositLimit,
+      clearShapeShift,
+      doShowSnackBar,
+    } = this.props;
+
+    return (
+      <div className="shapeshift__active-shift">
```
kauffj: Unused class name.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,139 @@
...
+                <span className="shapeshift__success">{shiftCoinType}</span>
+              </span>{" "}
+              to the address below
+            </p>
+
+            <div className="shapeshift__deposit-address-wrapper">
+              <Address address={shiftDepositAddress} />
+              {/* the header__item class should be a generic button class */}
+              <span className="header__item">
+                <Link
```
kauffj: Can this behavior be added to the existing `Address` component? If not all `<Address>` components should have a button, then this can be an optional property.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,139 @@
...
+              shiftState === statuses.RECEIVED
+                ? __("New")
+                : __("Cancel")
+            }
+          />
+          <span className="shapeshift__link">
+            <Link
+              label={__("View the status on Shapeshift.io")}
+              onClick={() =>
+                shell.openExternal(
```
kauffj: There is already code to detect external addresses on links and call `shell.OpenExternal`. Just a simple `href` property is sufficient here.

src/renderer/scss/component/_spinner.scss
```diff
@@ -0,0 +1,58 @@
+.spinner {
```
kauffj: :clap: good refactor

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,63 @@
...
+  // don't use a variable here. adds a white border for easier reading in dark mode
+  // needs to stay the same no matter what theme is present
+  background-color: #fff;
+  padding: 2px;
+  margin-left: 40px;
+}
+
+.shapeshift__received,
+.shapeshift__complete {
+  padding: 20px 0;
```
kauffj: Probably `$spacing-vertical` for consistency.

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,63 @@
...
+}
+
+.shapeshift__actions {
+  padding-top: 20px;
```
kauffj: Most of these paddings should be `$spacing-vertical` or 1/2 `$spacing-vertical`.

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,63 @@
...
+    margin-top: 75px;
+  }
+}
+
+.shapeshift__success {
+  color: var(--color-primary);
+}
+
+.shapeshift__error {
+  color: var(--color-error);
```
kauffj: Is a `shapeshift__error` sufficiently different from other errors that it needs it's own styling? Same for success.

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,63 @@
...
+  flex-direction: row;
+
+  * {
+    align-self: center;
+  }
+}
+
+.shapeshift__deposit-address {
+  max-width: 500px;
+  font-size: 22px;
```
kauffj: Does this actually require it's own styling? Is there no other styling that already exists that this should match?

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,63 @@
+.shapeshift__wrapper.shapeshift__initial-wrapper {
+  min-height: 311px;
```
kauffj: This scares me.

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,63 @@
...
+}
+
+.shapeshift__content {
+  .spinner {
+    margin-top: 75px;
```
kauffj: `$spacing-vertical` * 3

src/renderer/scss/component/_button.scss
```diff
@@ -99,3 +99,8 @@ $button-focus-shift: 12%;
     font-family: inherit;
     line-height: 0;
 }
+
+button:disabled {
```
kauffj: Isn't there already a disabled style? Can these share definitions?

src/renderer/scss/_gui.scss
```diff
@@ -188,3 +188,23 @@ section.section-spaced {
 {
   text-align: center;
 }
+
+.error-text {
+  color: var(--color-error);
+}
+
+.display-inline-block {
```
kauffj: IMO this and all of the definitions below it probably shouldn't exist.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
seanyesmunt: Good idea

src/renderer/scss/component/_shapeshift.scss
seanyesmunt: Don't need this class. The success class is just to add the green color on the text. There isn't any styles that only do that, maybe it should be a more generic className?

src/renderer/scss/component/_shapeshift.scss
seanyesmunt: Yeah...

Can' think of another way to do it. Nothing bad could really happen, at least nothing worse than not having this. If someone changes the form content to be a different height, it will just "jump" to that height after it's rendered. The same thing would happen if we don't have this style, it will "jump" and increase the height when the form is rendered.

Happy to change it if we can think of a better way. I just hate having content jumping when stuff is loading

src/renderer/scss/_gui.scss
seanyesmunt: Not a fan of these? I think they can be pretty helpful to avoid creating a brand new class just to add a simple display style.

Totally down to get rid of them if you don't want them.

src/renderer/js/component/common/spinner.jsx
liamcardenas: Interesting, why was it done this way? (no need to respond here, I'll ask about it tomorrow)

src/renderer/scss/component/_shapeshift.scss
liamcardenas: I don't see a better way of doing this either. As you said, worst case scenario, it will just do what it otherwise would have done had you not put this in. Best case scenario, it looks great.

But yeah, absolute px values are never great for dimensions :P

src/renderer/js/component/address/view.jsx
```diff
@@ -0,0 +1,51 @@
+import React from "react";
+import { clipboard } from "electron";
+import Link from "component/link";
+import classnames from "classnames";
+
+export default class Address extends React.PureComponent {
+  static propTypes = {
+    address: React.PropTypes.string,
```
liamcardenas: just curious, we just merged the react upgrade. I saw you changed all the React.PropTypes.etc... to just PropTypes.etc...

I believe this will break (there probably won't be a merge conflict because this is a new file. Maybe merge with master and re-check that this works after making necessary changes?

seanyesmunt: Whoops. You're right. I'll update

seanyesmunt: all good now 👍

src/renderer/js/redux/actions/shape_shift.js
```diff
@@ -0,0 +1,86 @@
+import { createAction } from "util/redux-utils";
+import Promise from "bluebird";
```
liamcardenas: please save to package.json

seanyesmunt: whoops. added

seanyesmunt:
@liamcardenas I think we should merge this separate from adding analytics just because the PR is so big. Then we can also do some testing in an RC. We can hold off on doing an actual release until we add the analytics.

liamcardenas:
@seanyesmunt you're right, I removed the dependency

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,57 @@
...
+  // don't use a variable here. adds a white border for easier reading in dark mode
+  // needs to stay the same no matter what theme is present
+  background-color: #fff;
+  padding: 2px;
+  margin-left: 40px;
+}
+
+.shapeshift__received,
+.shapeshift__complete {
+  padding: $spacing-vertical 0;
```
kauffj: All of the CSS from here and below this can likely be done via existing card and form CSS classes and styles.

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,57 @@
...
+    margin-top: $spacing-vertical * 3;
+  }
+}
+
+.shapeshift__success {
+  color: var(--color-primary);
+}
+
+.shapeshift__error,
+.shapeshift__maxdeposit {
```
kauffj: This CSS should be able to go.

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,57 @@
...
+  min-height: 311px;
+}
+
+.shapeshift__content {
+  .spinner {
+    margin-top: $spacing-vertical * 3;
+  }
+}
+
+.shapeshift__success {
```
kauffj: I think you can use `credit-amount--indicator` instead. If you dislike this, you can componentize the rendering of this.

src/renderer/scss/component/_shapeshift.scss
kauffj: I'm okay shipping this as-is.

However, if we wanted to avoid it, I think there's a potential solution where we render the component before shapeshift is ready, but render it with a very low opacity or `visibility: hidden`. In addition, either turn off interaction or render a blocking element on top of entire component. Then we can absolutely center a loading icon on top of this, and have it match the height precisely.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,123 @@
...
+            label={
+              shiftState === statuses.COMPLETE ||
+              shiftState === statuses.RECEIVED
+                ? __("New")
+                : __("Cancel")
+            }
+          />
+          <span className="shapeshift__link">
+            <Link
+              label={__("View the status on Shapeshift.io")}
```
kauffj: Try `button="text"` here.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,123 @@
...
+      <div>
+        {shiftState === statuses.NO_DEPOSITS && (
+          <div>
+            <p>
+              Send up to{" "}
+              <span className="credit-amount--bold">
+                {shiftDepositLimit}{" "}
+                <span className="shapeshift__success">{shiftCoinType}</span>
+              </span>{" "}
+              to the address below
```
kauffj: Add "."

src/renderer/js/component/shapeShift/view.jsx
```diff
@@ -0,0 +1,127 @@
...
+          "shapeshift__initial-wrapper": loading,
+        })}
+      >
+        <div className="card__title-primary">
+          <div className="display-inline-block">
+            <h3 className="display-inline">{__("Convert Crypto to LBC")}</h3>
+            {!hasActiveShift && (
+              <p className="help">
+                {__("Powered by ShapeShift. Read our FAQ")}{" "}
+                <Link onClick={() => {}}>{__("here")}</Link>.
```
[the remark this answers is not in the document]
seanyesmunt: I was thinking this would be a link to the FAQ Tom writes up

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,123 @@
...
+          </div>
+        )}
+
+        {shiftState === statuses.RECEIVED && (
+          <div className="shapeshift__received">
+            <p>
+              ShapeShift has received your payment! Sending the funds to your
+              LBRY wallet.<br />
+              <span className="help">
+                {__("This can take a while, especially with BTC")}
```
kauffj: Add ".". Also, let's link to FAQ here again.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,123 @@
...
+          </div>
+        )}
+        <div className="shapeshift__actions">
+          <Link
+            button="alt"
+            onClick={clearShapeShift}
+            label={
+              shiftState === statuses.COMPLETE ||
+              shiftState === statuses.RECEIVED
+                ? __("New")
```
kauffj: I'm not in love with "New" here - maybe just "Reset"? Though I don't like this either. As a (potentially weird) user, I have a desire for things to be back in their default state when "Done". But I don't necessary want to do a new one. I just want it back to where it was. Not sure on best name for this.

kauffj: Also this should probably be primary, whatever the label is.

src/renderer/js/component/shapeShift/internal/form.jsx
```diff
@@ -0,0 +1,83 @@
...
+          {__("Return address")} ({__("optional but recommended")})
+        </label>
+        <input
+          type="text"
+          name="returnAddress"
+          placeholder={getExampleAddress(originCoin)}
+          onChange={handleChange}
+          onBlur={handleBlur}
+          value={values.returnAddress}
+          className="wunderbar__input wunderbar__input--shapeshift"
```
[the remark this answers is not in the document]
seanyesmunt: Yeah definitely not a fan of the current input styling. I was thinking this was fine for now since the PR is getting bigger and bigger. But might as well do it now and avoid more tech debt. I will update the generic `input` style to be more inline with the wunderbar

src/renderer/js/component/shapeShift/internal/form.jsx
```diff
@@ -0,0 +1,83 @@
...
+  shiftSupportedCoins,
+  originCoin,
+  originCoinDepositLimit,
+  updating,
+  getCoinStats,
+  receiveAddress,
+}) => {
+  return (
+    <form onSubmit={handleSubmit}>
+      <div className="form-field">
```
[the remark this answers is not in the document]
seanyesmunt: I think this should stay as just classes with regular html elements. I can make my case more tomorrow, but my main idea is this:

Shared components are good for elements that can very a lot, but only with small differences. Buttons have many different possible styles (different background colors, borders, labels, etc). A lot of different variation, but the component is generally the same, a box with some text in the middle.

For this element (`select`), there can be many different ways to present it. I think it definitely makes sense to use a shared components if there might be inline errors or something we want to keep the same across the app, but there won't be any errors on this `select` element. I just want the styling with text on the left and right, which is very easy just by using regular html elements (and the same classes that the `FormRow` component uses).

If we start building out the shared components so much that they meet every possible use case, they can quickly become huge, un-maintainable files that really only end up serving one use-case.

Would love to hear everyone else's thoughts. @liamcardenas @IGassmann

src/renderer/scss/component/_shapeshift.scss
seanyesmunt: I'll keep the min-height on `shapeshift__maxdeposit` to avoid content jumping when we fetch new max limits after the user changes the origin coin

src/renderer/scss/component/_shapeshift.scss
seanyesmunt: I don't think it's necessary to use the `card-xx` styles for this. It doesn't have anything to do with the `card` style. It's just a small bit of markup that needs some padding. I think ideally we could just use a `s-vertical-padding` helper class or something similar. We can discuss this more along side the `display-xx` helper classes.

src/renderer/scss/component/_shapeshift.scss
seanyesmunt: 👍

src/renderer/scss/component/_shapeshift.scss
kauffj: `card__content` is the class I'd propose using instead.

seanyesmunt: 👍

seanyesmunt:
@kauffj only thing left (I think) is adding a link to the FAQ on lbry.io, and flow.

kauffj:
@seanyesmunt no need to block app PRs on FAQ content. Just put the link in the code, get the ticket open  (I know this happened), and you can forget about it.

seanyesmunt:
@kauffj sounds good. What should the url be?

src/renderer/flow-typed/bluebird.js
```diff
@@ -0,0 +1,3 @@
+declare module 'bluebird' {
```
seanyesmunt: There's got to be a better way to do this. Can these be added to the `.flowconfig`?

src/renderer/webpack.common.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
   },
   resolve: {
     modules: [appPath, "node_modules"],
-    extensions: [".js", ".jsx", ".css"]
+    extensions: [".js", ".jsx", ".css", ".json"]
```
seanyesmunt: This is required because the shapeshift.io module has an import like `const package = require('../package')

It causes the bundle to increase quite a bit. I will make a pr to that project to change it to `const package = require('../package.json')` which allow use to remove this

seanyesmunt:
@kauffj @liamcardenas 

I could probably spend another week just on cleaning up the flow stuff. It is integrated into the entire shapeshift flow and already caught a few issues. I think it is good for now, but could certainly be presented a little better. I will come back to this as we add flow to more files and determine how we want to structure it.

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,48 @@
...
+// this should be pulled out into it's own styling when we add more qr codes
+.shapeshift__qrcode {
+  // don't use a variable here. adds a white border for easier reading in dark mode
+  // needs to stay the same no matter what theme is present
+  background-color: #fff;
+  padding: 2px;
+  margin-left: 40px;
+}
+
+.shapeshift__actions {
```
kauffj: I still dislike both this rule and the rule below. Nothing about Shapeshift seems special enough that it needs it's own spacing rules vs. other cards.

src/renderer/scss/component/_shapeshift.scss
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.shapeshift__content {
+  .spinner {
+    margin-top: $spacing-vertical * 3;
+  }
+}
+
+.shapeshift__tx-info {
+  min-height: 55px;
```
kauffj: If this is to avoid jumping, I'd rather just put placeholder text like "Calculating rates..."

src/renderer/js/component/shapeShift/internal/form.jsx
```diff
@@ -0,0 +1,106 @@
...
+        >
+          {shiftSupportedCoins.map(coin => (
+            <option key={coin} value={coin}>
+              {coin}
+            </option>
+          ))}
+        </select>
+        <span> {__("for LBC")}</span>
+        <p className="shapeshift__tx-info help">
+          {!updating &&
```
kauffj: 1) Is it possible to display the actual rate here?
2) This can be more compact. If we have the rate, proposed text is "Receive X LBC per Y BTC, less Z LBC fee. Max: W BTC, Min: V BTC"
3) Typo in minimum

src/renderer/js/component/shapeShift/internal/form.jsx
```diff
@@ -0,0 +1,106 @@
...
+      <FormRow
+        type="text"
+        name="returnAddress"
+        placeholder={getExampleAddress(originCoin)}
+        label={__("Return address")}
+        onChange={handleChange}
+        onBlur={handleBlur}
+        value={values.returnAddress}
+        errorMessage={errors.returnAddress}
+        hasError={touched.returnAddress && errors.returnAddress}
```
kauffj: This needs to be cast as a boolean.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,147 @@
...
+            <p>
+              {__(
+                "Transaction complete! You should see the new LBC in your wallet."
+              )}
+            </p>
+          </div>
+        )}
+        <div className="shapeshift__actions">
+          <Link
+            button="primary"
```
kauffj: Should be "alt" for Cancel.

kauffj: What happened to the idea of a "Sent" button that displays the shift code for tracking purposes?

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,147 @@
...
+      shiftOrderId,
+      shiftState,
+      originCoinDepositMax,
+      clearShapeShift,
+      doShowSnackBar,
+    } = this.props;
+
+    return (
+      <div>
+        {shiftState === statuses.NO_DEPOSITS && (
```
kauffj: The exchange rate and fees should still be displayed at this step.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,147 @@
...
+          <div>
+            <p>
+              Send up to{" "}
+              <span className="credit-amount--bold">
```
kauffj: This whole element should be the same style IMO. I'd be fine with just `credit-amount--bold` for the whole thing.

src/renderer/js/component/shapeShift/internal/form.jsx
seanyesmunt: Yes we can display the actual rate. I'll add that

src/renderer/scss/component/_shapeshift.scss
seanyesmunt: We would still need some sort of min height on that unless the heights are the same. I would rather not since this check is usually very quick and you would only see that for a small time before it goes back to the info text.

src/renderer/js/component/shapeShift/internal/active-shift.jsx
seanyesmunt: I'm not sure how much I like that button. What if they click it before actually sending? We would need some new bit of UI to say "We will keep pinging shapeshit to see if you actually sent it" (obviously worded better).

We say "this will update automatically" and have a link to the shapeshift status page, I don't see any reason to create another shift status "user says they sent it but shapeshift hasn't received it"

src/renderer/flow-typed/bluebird.js
liamcardenas: Have you checked if flow-typed has any of the type definitions?

https://github.com/flowtype/flow-typed

`yarn run flow-typed install bluebird`

src/renderer/js/component/shapeShift/internal/active-shift.jsx
```diff
@@ -0,0 +1,161 @@
...
+  originCoinDepositMax: ?number,
+  originCoinDepositFee: ?number,
+  originCoinDepositMin: ?string,
+};
+
+class ActiveShapeShift extends React.PureComponent<Props> {
+  constructor() {
+    super();
+    // $FlowFixMe
+    this.continousFetch = undefined;
```
liamcardenas: outside of the constructor, I think you need to add

`continuousFetch: ?number;`

as per https://flow.org/en/docs/types/classes/

liamcardenas: this should get rid of all the flow fix mes

src/renderer/js/redux/actions/shape_shift.js
```diff
@@ -0,0 +1,137 @@
...
+} from "redux/reducers/shape_shift";
+import type { FormikActions } from "types/common";
+
+// use promise chains instead of callbacks for shapeshift api
+const shapeShift = Promise.promisifyAll(require("shapeshift.io"));
+
+// All ShapeShift actions
+// Action types defined in the reducer will contain some payload
+export type Action =
+  | { type: types.GET_SUPPORTED_COINS_START }
```
liamcardenas: I get that you are defining these here so as not to create extra boilerplate work, but I wonder if all of the types should be defined in the same place? I think this is probably best, but I just wanted to throw this question out there in case anyone has any thoughts on it

src/renderer/js/redux/actions/shape_shift.js
```diff
@@ -0,0 +1,137 @@
...
+  | PrepareShapeShiftFail
+  | PrepareShapeShiftSuccess
+  | { type: types.GET_ACTIVE_SHIFT_START }
+  | GetActiveShiftFail
+  | GetActiveShiftSuccess;
+
+// Basic thunk types
+// It would be nice to import these from types/common
+// Not sure how that would work since they rely on the Action type
+type PromiseAction = Promise<Action>;
```
liamcardenas: can you make these generic and move them to a different folder as you alluded to in the comment? I also am not a huge fan of using the any type here but for now I think this is good.

i.e.

```type PromiseAction<T> = Promise<T>;
 type ThunkAction<T> = (dispatch: Dispatch<T>) => any;
 export type Dispatch<T> = (
   action: T | ThunkAction<T> | PromiseAction<T> | Array<T>
 ) => any;

Comment (liamcardenas) [reply src/renderer/js/redux/actions/shape_shift.js]: I don't want you to get bogged down with this if for some reason its really difficult

File: src/renderer/scss/_gui.scss [hunk 47]
```diff
@@ -34,6 +34,11 @@ body
   color: var(--color-meta-light);
 }
 
+.credit-amount--bold
```
liamcardenas: is this used elsewhere in the app or is it just for shapeshift? I don't have a criticism here, I'm just wondering what we decided about using reusable vs case specific classes?

src/renderer/js/redux/actions/shape_shift.js
kauffj: My proposed design is:

1) Add a `types` folder inside of `js`. `types` will eventually swallow `constants`.
2) An `Action` type is defined in `Action.js` inside of this folder.
3) This file also includes all action constant definitions.
4) The `Action` type explicitly enumerates all possible action type values.

My biggest question about this design is whether there should only be one action type, or whether we should use subtypes for each possible action and payload signature.

kauffj: Possibly better to discuss this here https://github.com/lbryio/lbry-app/issues/810

kauffj: Let's not block this PR on getting action definition exactly correct.

PR #653, opened 2017-10-09
Resolve cleanup

alexliebowitz:
- Updates Discover page to pass all URLs to resolve at once, rather than one at a time.
- Changes FileCard, FileTile and ShowPage to always call resolve, even when claim info is provided
- Removes XHR caching used for cancelling requests


kauffj:
@alexliebowitz merged, but I restored the resolve behaviors on `FileCard` and `FileTile`. I believe those checks are still necessary to prevent unnecessary resolves (however, I let the re-resolve happen now on show pages).

PR #654, opened 2017-10-09
[WIP] Added subscriptions

liamcardenas:
Done:
Added subscribe button to file and channel pages
Saves channels that have been subscribed to
Once you have subscribed to one or more channels, it presents 10 recent videos on the discover page under the "Subscription" section

Not Done:
I made the subscribe button on the file page work, but not on the channel page.
Ideally, the subscribe button would say unsubscribe after you click it (and you would be able to unsubscribe).
It does not save the state in local storage -- subscriptions go away after a refresh
There is nothing to prevent storing duplicate subscriptions
There appears to be some performance issues when generating the list of subscription videos to display on the discover page.

I would be more than happy to continue working on any of these problems, but I thought the code is to a point where I touched all key functionality, thus demonstrating my abilities.

Let me know if you have any questions or comments.


liamcardenas:
This needs further thought, I'm closing this pull request for now.

PR #661, opened 2017-10-10
Fixes #614, Makes things consistent to allow sorting by title

### LINKED ISSUE #614: Published files Sort by Title not working correctly
## The Issue
Published files list correctly by Date, but not by Title

### Expected behavior
When I go into my Published files, I should be able to view the files alphabetically when I select the Title option.

### Actual behaviour
Please see my screen shot. When I display my videos by Title, they are not in alphabetical order. I'm not sure how they're being displayed. There are two videos that start with "Betty" and then the video that starts with "A Little" is displayed between them. I thought maybe it was skipping the A, but even so, it still isn't in any kind of alphabetical order.

## System Configuration

 

- LBRY Daemon version: 0.16.1
- LBRY App version: App | 0.16.0
- LBRY Installation ID: 5eFxsAfhBFTKA5byNFVkGL5Lu5owWzQFthzKjRkxN2iUzWh2XNKLJAiv82mxAcq3AU
- Operating system: Mac OS X 15.6.0 

## Screenshots

screen shot 2017-09-24 at 10 07 38 pm

sovanna:
[#614](https://github.com/lbryio/lbry-app/issues/614)

Basically, the component `FileTile` displays *title* using
`lbryuri.parse(uri).contentName;` and not `metadata.title` cause it's
never passed to the component and `title` prop does not exist anyway
(see *FileList* with *fileInfos* object).

`.contentName` is filled with `fileInfos.name` value.

So to make all things **consistent**, we simply need to use
`fileInfo.name` in our comparaison function.

### DIFF
--- a/ui/js/component/fileList/view.jsx
+++ b/ui/js/component/fileList/view.jsx
@@ -18,12 +18,8 @@ class FileList extends React.PureComponent {
       },
       title: function(fileInfos) {
         return fileInfos.slice().sort(function(fileInfo1, fileInfo2) {
-          const title1 = fileInfo1.metadata
-            ? fileInfo1.metadata.stream.metadata.title.toLowerCase()
-            : fileInfo1.name;
-          const title2 = fileInfo2.metadata
-            ? fileInfo2.metadata.stream.metadata.title.toLowerCase()
-            : fileInfo2.name;
+          const title1 = fileInfo1.name.toLowerCase();
+          const title2 = fileInfo2.name.toLowerCase();
           if (title1 < title2) {
             return -1;
           } else if (title1 > title2) {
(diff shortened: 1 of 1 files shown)


tzarebczan:
@sovanna - thanks for the PR! Please take a look here https://lbry.io/faq/tips if you'd like a small reward for your work!

kauffj:
@sovanna This is not a proper fix. The metadata absolutely is passed in to `FileTile` via the Redux (see the index.js file in the same folder).

After digging into this briefly, it appears that the signature of `fileInfo` in the sort function is different between the Downloaded and Published tags. To solve #614 would require further investigation into the root cause of this.

PR #662, opened 2017-10-11
Better title for menu bar

btzr-io:
If no title is set, it would always show the current file path (`./index.html`)

### Changes
Display the application name as default title.


btzr-io:
Only home page or any page that don't have a title ^^

PR #686, opened 2017-10-18
Minimize LBRY app to tray.

hackrush01:


kauffj:
@hackrush01 this has a significant bug: closing the window will close the daemon, even though the app stays open. I believe this will require changes to how `before-quit` and or `window-all-closed` is currently being handled.

I also changed the behavior so that only close moves to tray and that minimize is just minimize.

kauffj:
@hackrush01 at least on Linux, I get an error now every time I close and re-open the app. Please test in Ubuntu VM if you are not seeing this behavior.

Additionally, I do not want "Minimize if closed" to be a context menu in the tray. We can consider adding this in settings, but I don't think this option is necessary currently (and if we add an option for minimize, we should also turn close behavior into an option between tray or quit).

kauffj:
https://spee.ch/8/tray-error.png

tzarebczan:
I've seen that as a context menu setting on other Windows apps I've used, but I also agree that the settings would make more sense. Users with BW caps most likely want to close as a default behavior.

hackrush01:
Known bug: error on start. Working on it

hackrush01:
@kauff I've added `Alt+F4` as an accelerator to quit the app. Though I suggest using a different combination(such as `Alt+X` or other). As we could want the default actions for all default quitting techniques to minimize the app to tray, and a special, different action for quitting the app.

app/main.js
```diff
@@ -167,17 +174,84 @@ function createWindow () {
...
+  return [
+    getToggleItem(),
+    {
+      label: "Quit",
+      click: () => safeQuit(),
+    },
+  ]
+
+  function getToggleItem () {
+    if (win.isVisible()) {
```
kauffj: I think this needs to be a check if the window is focused, not just if it is visible. If I Alt+Tab from an open app, I am shown "Hide" in the context menu, when it should be "Show".

app/main.js
```diff
@@ -167,17 +174,84 @@ function createWindow () {
...
+  win.on("show", () => {
+    if (minimize) updateTray();
+  });
+
+  win.on("blur", () => {
+    globalShortcut.unregister('Alt+F4');
+  });
+
+  win.on("focus", () => {
+    globalShortcut.register('Alt+F4', () => safeQuit());
```
kauffj: This should be Ctrl/Cmd + Q as well.

kauffj: (Or if possible, detect OS setting for this, if one exists)

app/main.js
```diff
@@ -167,17 +174,84 @@ function createWindow () {
...
+  });
+
   // Menu bar
   win.setAutoHideMenuBar(true);
   win.setMenuBarVisibility(isDebug);
   setMenu();
 
 };
 
+function createTray () {
```
kauffj: Can you add a behavior that double clicking the tray icon also triggers the show behavior?

app/main.js
kauffj: Also needs to be available via a File menu in the app, I think.

tzarebczan:
@kauffj - running LBRY again when its minimized to tray does not bring the app to the foreground. Neither does clicking lbry:// hyperlinks, but the link is opened when the app is manually shown again.  This was on Windows, not sure about other OS but I can try later if needed. 

When LBRY is running in a maximized window - I also noticed that if you minimize LBRY, then close the app (right click taskbar icon > close) and finally bring it back from the tray, the Window size goes back to the previous unmaximized size.

kauffj:
@hackrush01 at a minimum the app will need to be brought to the foreground when the URI handler is initiated. Ideally, it'd come to the foreground whenever triggered. And the sizing issue ought to be fixed as well.

app/main.js
hackrush01: There is a predefined accelerator for this `CommandOrControl+Q`, which works on all OSes

And regarding the File Menu in the App, I already gave it a shot, and ran into some issues, I'll work on it. I would probably need to look at `module.exports`, as i need to "export" the `safeQuit()` function from `main.js` to `menu/main-menu.js`

app/main.js
hackrush01: Done.

app/main.js
hackrush01: Done

hackrush01:
@tzarebczan @kauffj I think the other 2 issues are also solved. I could test it as clicking links doesn't open the app in my laptop, and I couldn't reproduce the maximization issue.

hackrush01:
Update: @kauffj I added the quit option to file menu in the menu bar too :-)

tzarebczan:
Hyperlinks bring the app to focus now from the tray, woohoo. 

Launching LBRY again while it's in the tray does not bring it up. 

If the LBRY app is not maximized and closed to tray, it comes back maximized. 

When the app is in focus, the right click still show "show app" - should it say hide app and have the same function as close?

PR #687, opened 2017-10-18
Quick fix for "checked propType" errors spewing in console, for settings

hackrush01:


kauffj:
This is now in 8f1e2ea7

Issue #581, opened 2017-09-18
Allow claims to be revoked from TransactionList


kauffj:
As a user, I may want to use TransactionList to see past supports, publishes, claims, etc. It would be convenient to have the ability to reclaim credits from this screen.

hackrush01:
I will get to this once I figure the rest of abandon related stuff. One of them is directly related to abandoning claims(with supports and tips).

tzarebczan:
Good progress so far, but I'm sure there will be some confusion going forward. When an action is taken, the icon disappears and a new transaction is created (which is not apparent immediately). 

Other areas of possible confusion:
1) users won't understand why they can take action in some cases but not others - i.e. a publish and then an update - the publish has no action available, but the update does. 

2) Previously revoked/claimed items won't show an action - the only history is the resulting transaction. Maybe some indication an action was already taken?

kauffj:
@hackrush01 shouldn't this make publishes revokeable?

tzarebczan:
It does... Maybe you misunderstood what I was saying? But if you publish and then update, the update is now revokeable but the publish isn't (since the bid for the claim is now at a different address)

kauffj:
Yep, nm.

lyoshenka:
Merged

PR #697, opened 2017-10-24
Revoke claims from Txn list

hackrush01:
Closes (#581)

### DIFF
--- a/ui/js/actions/content.js
+++ b/ui/js/actions/content.js
@@ -505,3 +505,34 @@ export function doPublish(params) {
     });
   };
 }
+
+export function doAbandonClaim(claimId, txid, nout) {
+  return function(dispatch, getState) {
+    const state = getState();
+
+    dispatch({
+      type: types.ABANDON_CLAIM_STARTED,
+      data: {
+        claimId: claimId,
+        txid: txid,
+        nout: nout,
+      },
+    });
+
+    const success = dispatch({
+      type: types.ABANDON_CLAIM_SUCCEEDED,
+      data: {
+        claimId: claimId,
+        txid: txid,
+        nout: nout,
+      },
+    });
+
+    lbry
+      .claim_abandon({
+        txid: txid,
+        nout: nout,
+      })
+      .then(success);
+  };
+}
--- a/ui/js/actions/content.js
+++ b/ui/js/actions/content.js
@@ -11,7 +11,7 @@ import {
 } from "selectors/file_info";
 import { selectResolvingUris } from "selectors/content";
 import { makeSelectCostInfoForUri } from "selectors/cost_info";
-import { doAlertError, doOpenModal } from "actions/app";
+import { doAlertError, doOpenModal, doCloseModal } from "actions/app";
 import { doClaimEligiblePurchaseRewards } from "actions/rewards";
 import { selectBadgeNumber } from "selectors/app";
 import { selectTotalDownloadProgress } from "selectors/file_info";
@@ -509,6 +509,8 @@ export function doPublish(params) {
 export function doAbandonClaim(claimId, txid, nout) {
   return function(dispatch, getState) {
     const state = getState();
+
+    dispatch(doCloseModal());
 
     dispatch({
       type: types.ABANDON_CLAIM_STARTED,
--- a/ui/js/actions/content.js
+++ b/ui/js/actions/content.js
@@ -4,14 +4,14 @@ import lbry from "lbry";
 import lbryio from "lbryio";
 import lbryuri from "lbryuri";
 import { makeSelectClientSetting } from "selectors/settings";
-import { selectBalance } from "selectors/wallet";
+import { selectBalance, selectTransactionItems } from "selectors/wallet";
 import {
   makeSelectFileInfoForUri,
   selectDownloadingByOutpoint,
 } from "selectors/file_info";
 import { selectResolvingUris } from "selectors/content";
 import { makeSelectCostInfoForUri } from "selectors/cost_info";
-import { doAlertError, doOpenModal, doCloseModal } from "actions/app";
+import { doAlertError, doOpenModal } from "actions/app";
 import { doClaimEligiblePurchaseRewards } from "actions/rewards";
 import { selectBadgeNumber } from "selectors/app";
 import { selectTotalDownloadProgress } from "selectors/file_info";
@@ -506,9 +506,13 @@ export function doPublish(params) {
   };
 }
 
-export function doAbandonClaim(claimId, name, txid, nout) {
+export function doAbandonClaim(txid, nout) {
   return function(dispatch, getState) {
     const state = getState();
+    const transactionItems = selectTransactionItems(state);
+    const { claim_id: claimId, claim_name: name } = transactionItems.find(
+      claim => claim.txid == txid && claim.nout == nout
+    );
 
     dispatch({
       type: types.ABANDON_CLAIM_STARTED,
--- a/ui/js/actions/content.js
+++ b/ui/js/actions/content.js
@@ -506,35 +506,41 @@ export function doPublish(params) {
   };
 }
 
-export function doAbandonClaim(claimId, txid, nout) {
+export function doAbandonClaim(claimId, name, txid, nout) {
   return function(dispatch, getState) {
     const state = getState();
-
-    dispatch(doCloseModal());
 
     dispatch({
       type: types.ABANDON_CLAIM_STARTED,
       data: {
         claimId: claimId,
-        txid: txid,
-        nout: nout,
       },
     });
 
-    const success = dispatch({
-      type: types.ABANDON_CLAIM_SUCCEEDED,
-      data: {
-        claimId: claimId,
-        txid: txid,
-        nout: nout,
-      },
-    });
+    const errorCallback = error => {
+      dispatch(doOpenModal(modals.TRANSACTION_FAILED));
+    };
+
+    const successCallback = results => {
+      if (results.txid) {
+        dispatch({
+          type: types.ABANDON_CLAIM_SUCCEEDED,
+          data: {
+            claimId: claimId,
+          },
+        });
+        dispatch(doResolveUri(lbryuri.build({ name, claimId })));
+        dispatch(doFetchClaimListMine());
+      } else {
+        dispatch(doOpenModal(modals.TRANSACTION_FAILED));
+      }
+    };
 
     lbry
       .claim_abandon({
         txid: txid,
         nout: nout,
       })
-      .then(success);
+      .then(successCallback, errorCallback);
   };
 }
--- a/ui/js/actions/file_info.js
+++ b/ui/js/actions/file_info.js
@@ -1,6 +1,6 @@
 import * as types from "constants/action_types";
 import lbry from "lbry";
-import { doFetchClaimListMine } from "actions/content";
+import { doFetchClaimListMine, doAbandonClaim } from "actions/content";
 import {
   selectClaimsByUri,
   selectIsFetchingClaimListMine,
@@ -105,29 +105,7 @@ export function doDeleteFile(outpoint, deleteFromComputer, abandonClaim) {
         txid = fileInfo.outpoint.slice(0, -2);
         nout = fileInfo.outpoint.slice(-1);
 
-        dispatch({
-          type: types.ABANDON_CLAIM_STARTED,
-          data: {
-            claimId: fileInfo.claim_id,
-            txid: txid,
-            nout: nout,
-          },
-        });
-
-        const success = dispatch({
-          type: types.ABANDON_CLAIM_SUCCEEDED,
-          data: {
-            claimId: fileInfo.claim_id,
-            txid: txid,
-            nout: nout,
-          },
-        });
-        lbry
-          .claim_abandon({
-            txid: txid,
-            nout: nout,
-          })
-          .then(success);
+        dispatch(doAbandonClaim(txid, nout));
       }
     }
 
--- a/ui/js/actions/file_info.js
+++ b/ui/js/actions/file_info.js
@@ -102,10 +102,15 @@ export function doDeleteFile(outpoint, deleteFromComputer, abandonClaim) {
       const fileInfo = byOutpoint[outpoint];
 
       if (fileInfo) {
+        txid = fileInfo.outpoint.slice(0, -2);
+        nout = fileInfo.outpoint.slice(-1);
+
         dispatch({
           type: types.ABANDON_CLAIM_STARTED,
           data: {
             claimId: fileInfo.claim_id,
+            txid: txid,
+            nout: nout,
           },
         });
 
@@ -113,9 +118,16 @@ export function doDeleteFile(outpoint, deleteFromComputer, abandonClaim) {
           type: types.ABANDON_CLAIM_SUCCEEDED,
           data: {
             claimId: fileInfo.claim_id,
+            txid: txid,
+            nout: nout,
           },
         });
-        lbry.claim_abandon({ claim_id: fileInfo.claim_id }).then(success);
+        lbry
+          .claim_abandon({
+            txid: txid,
+            nout: nout,
+          })
+          .then(success);
       }
     }
 
--- a/ui/js/component/transactionList/internal/TransactionListItem.jsx
+++ b/ui/js/component/transactionList/internal/TransactionListItem.jsx
@@ -4,42 +4,22 @@ import { CreditAmount } from "component/common";
 import DateTime from "component/dateTime";
 import Link from "component/link";
 import lbryuri from "lbryuri";
+import * as txnTypes from "constants/transaction_types";
 
 class TransactionListItem extends React.PureComponent {
   abandonClaim() {
-    const {
-      claim_id: claimId,
-      claim_name: name,
-      txid,
-      type,
-      nout,
-    } = this.props.transaction;
-    let msg;
+    const { txid, nout } = this.props.transaction;
 
-    if (type == "tip") {
-      msg = "This will reduce the committed credits to the URL";
-    } else {
-      msg = "This will reclaim you lbc, back to your account";
-    }
-
-    const abandonData = {
-      name: name,
-      claimId: claimId,
-      txid: txid,
-      nout: nout,
-      msg: msg,
-    };
-
-    this.props.revokeClaim(abandonData);
+    this.props.revokeClaim(txid, nout);
   }
 
   getLink(type) {
-    if (type == "tip") {
+    if (type == txnTypes.TIP) {
       return (
         <Link
           onClick={this.abandonClaim.bind(this)}
           icon="icon-unlock-alt"
-          title={__("Unlock Tip")}
+          title={__("Unlock")}
         />
       );
     } else {
@@ -47,7 +27,7 @@ class TransactionListItem extends React.PureComponent {
         <Link
           onClick={this.abandonClaim.bind(this)}
           icon="icon-trash"
-          title={__("Delete")}
+          title={__("Revoke")}
         />
       );
     }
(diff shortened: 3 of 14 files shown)


hackrush01:
Currently, there is no pop up confirming whether to revoke the claim or not. I just revokes it, do we want a pop up to be displayed or it should be done in a single click(as the button clearly indicates the action)?

kauffj:
@hackrush01 yes, revoking a claim should require a confirmation modal.

hackrush01:
@kauffj Added the confirmation dialog. It is mergable(not hard blocked by anything, that is to say).
Soft blocked by sequential execution of dispatch calls.

ui/js/actions/content.js
```diff
@@ -505,3 +505,36 @@ export function doPublish(params) {
     });
   };
 }
+
+export function doAbandonClaim(claimId, txid, nout) {
+  return function(dispatch, getState) {
+    const state = getState();
+
+    dispatch(doCloseModal());
```
kauffj: This should probably be in the modal not in the action.

ui/js/actions/content.js
```diff
@@ -505,3 +505,36 @@ export function doPublish(params) {
...
+        nout: nout,
+      },
+    });
+
+    lbry
+      .claim_abandon({
+        txid: txid,
+        nout: nout,
+      })
+      .then(success);
```
kauffj: What if the call fails?

kauffj: We should at least alert an error if we don't want a whole event chain. As is, this will trigger the success event on a failure.

ui/js/component/transactionList/internal/TransactionListItem.jsx
```diff
@@ -80,6 +92,12 @@ class TransactionListItem extends React.PureComponent {
         <td>
           <LinkTransaction id={txid} />
         </td>
+        <td>
```
kauffj: I'm inclined to make this an icon next to either the amount or the type.

Also, since this logic makes both received tips, supports, and publishes revokable, I think we need to consider the way we present this in each case.

For a tip someone else sent me, this should be presented as something like an "unlock", with a warning that it will reduce the committed credits to that URL.

For a claim I made, or a support I made to someone else's content, this is more of a delete action.

ui/js/component/transactionList/view.jsx
```diff
@@ -23,6 +25,25 @@ class TransactionList extends React.PureComponent {
...
+  revokeClaim(abandonData) {
+    const {
+      name: name,
+      claimId: claimId,
+      txid: txid,
+      nout: nout,
+    } = abandonData;
+
+    const uri = lbryuri.build({ name, claimId });
+    this.props.resolveUri(uri);
```
kauffj: Why does this happen here? If the modal needs it, it should probably be in the modal.

ui/js/modal/modalRevokeClaim/view.jsx
```diff
@@ -0,0 +1,34 @@
...
+        isOpen={true}
+        contentLabel={__("Confirm Claim Revoke")}
+        type="confirm"
+        confirmButtonLabel={__("Yes, Revoke")}
+        // onConfirmed={this.s.bind(this)}
+        onConfirmed={() => abandonClaim(claimId, txid, nout)}
+        onAborted={closeModal}
+      >
+        <p>
+          {__("Are you sure you want to revoke the claim?")}
```
kauffj: As indicated previously, I think this question/presentation needs to be more conditional to what is being abandoned.

ui/js/reducers/claims.js
```diff
@@ -78,6 +81,7 @@ reducers[types.FETCH_CLAIM_LIST_MINE_COMPLETED] = function(state, action) {
   return Object.assign({}, state, {
     isFetchingClaimListMine: false,
     myClaims: myClaims,
+    allMyClaimsByTxidNout: allMyClaimsByTxidNout,
```
kauffj: This set doesn't need to be stored in state. It can be generated from `myClaims` via a selector.

ui/js/reducers/claims.js
```diff
@@ -157,21 +161,35 @@ reducers[types.ABANDON_CLAIM_STARTED] = function(state, action) {
...
+  // so we need to remove the support from the state
+  // but this is not working, even after calling resolve on the uri.
+  if (supports && supports.length > 0) {
+    const indexToDelete = supports.findIndex(support => {
+      return support.txid === txid && support.nout === nout;
+    });
 
-  delete byId[claimId];
-  myClaims.delete(claimId);
+    supports.splice(indexToDelete, 1);
```
kauffj: I think it is fine/right to update the state in this way, but it would also be an option to just trigger a re-fetch these after an abandon.

ui/js/reducers/claims.js
hackrush01: I think `myClaims` only filter out claims which are strictly claims(not updates, supports and tips), so I had to make this as a separate state.

kauffj: So then perhaps the filtering should happen in the other direction, but it still seems like we're storing redundant data and this could be avoided.

Maybe it'd make sense that `myClaims` include _all_ of your claims, and instead the relevant `selectXXX` calls in `selectors/claims.js` be updated and renamed as appropriate.

ui/js/reducers/claims.js
hackrush01: I'm removing this logic. And taking your suggestion of re-fetching.

ui/js/component/transactionList/view.jsx
hackrush01: I am getting rid of this resolve before revoking all together, and trigger it after revoking. That would eliminate the need to make the synchronous dispatch calls too.

ui/js/component/transactionList/internal/TransactionListItem.jsx
hackrush01: I've wrote some placeholder texts in the new commit, as draft. I was hoping you could assist me in what will be the final modal say.

ui/js/actions/content.js
hackrush01: Done.

ui/js/actions/content.js
hackrush01: Yes, I did it.

ui/js/reducers/claims.js
hackrush01: I agree. I'd have to look deeper into it, as there is code dependent on this.

hackrush01:
@kauffj This https://github.com/lbryio/lbry-app/pull/697/files#diff-30fd434d64af092be888fabf8953dfdf
is the file which contains the message displayed in the confirmation modal and the icon displayed next to the Tip/Update etc. The message right now is placeholder-ish. I couldn't figure out what to write. I was hoping you could assist.

ui/js/component/transactionList/internal/TransactionListItem.jsx
kauffj: I'll make any copy changes before I merge.

ui/js/selectors/claims.js
```diff
@@ -121,15 +121,26 @@ export const selectIsFetchingClaimListMine = createSelector(
   state => !!state.isFetchingClaimListMine
 );
 
-export const selectMyClaimsRaw = createSelector(
+export const cantFigureOutVarName = createSelector(
```
hackrush01: @kauffj I couldn't figure out the variable name(it is basically `selectMyClaimsRaw` but that is already used. Two things could be done, either rename `selectMyClaimsRaw` at every place or find a good/better a name for `cantFigureOutVarName`)

ui/js/actions/content.js
```diff
@@ -11,7 +11,7 @@ import {
 } from "selectors/file_info";
 import { selectResolvingUris } from "selectors/content";
 import { makeSelectCostInfoForUri } from "selectors/cost_info";
-import { doAlertError, doOpenModal } from "actions/app";
+import { doAlertError, doOpenModal, doCloseModal } from "actions/app";
```
kauffj: No longer needed

ui/js/actions/file_info.js
```diff
@@ -102,20 +102,32 @@ export function doDeleteFile(outpoint, deleteFromComputer, abandonClaim) {
       const fileInfo = byOutpoint[outpoint];
```
kauffj: Can all or most of the code inside of this if statement simply be a call to `doAbandonClaim`? Seems like some redundant logic.

ui/js/actions/content.js
```diff
@@ -505,3 +505,42 @@ export function doPublish(params) {
...
+        dispatch({
+          type: types.ABANDON_CLAIM_SUCCEEDED,
+          data: {
+            claimId: claimId,
+          },
+        });
+        dispatch(doResolveUri(lbryuri.build({ name, claimId })));
+        dispatch(doFetchClaimListMine());
+      } else {
+        dispatch(doOpenModal(modals.TRANSACTION_FAILED));
```
kauffj: Probably better as something like `errorCallback(new Error("Claim abandon received unknown response signature"))`. If we made it to this code, it's quite possible the transaction succeeded, right?

ui/js/actions/content.js
```diff
@@ -505,3 +505,42 @@ export function doPublish(params) {
...
+
+    dispatch({
+      type: types.ABANDON_CLAIM_STARTED,
+      data: {
+        claimId: claimId,
+      },
+    });
+
+    const errorCallback = error => {
+      dispatch(doOpenModal(modals.TRANSACTION_FAILED));
```
kauffj: Is it safe to show this modal for all failures?

ui/js/component/transactionList/internal/TransactionListItem.jsx
```diff
@@ -6,12 +6,59 @@ import Link from "component/link";
...
+    const {
+      claim_id: claimId,
+      claim_name: name,
+      txid,
+      type,
+      nout,
+    } = this.props.transaction;
+    let msg;
+
+    if (type == "tip") {
```
kauffj: "tip" ought to be a constant so that it can be clear where it is being referenced. Probably via a new file in the `constants` folder, since it doesn't fit in clearly anywhere.

ui/js/component/transactionList/internal/TransactionListItem.jsx
```diff
@@ -6,12 +6,59 @@ import Link from "component/link";
...
+      msg = "This will reduce the committed credits to the URL";
```
kauffj: The determination of the message can be put in the modal itself.

ui/js/component/transactionList/internal/TransactionListItem.jsx
```diff
@@ -6,12 +6,59 @@ import Link from "component/link";
...
+    } else {
+      msg = "This will reclaim you lbc, back to your account";
+    }
+
+    const abandonData = {
```
kauffj: The only input to revoking a claim should be the transaction id (or the outpoint(s)?). The rest of this data can be reselected rather than passed.

More generally, it is usually a bad idea to create arbitrary data structures to pass multiple values. Even if 4 or 5 values were required here, I'd prefer them be passed individually.

ui/js/component/transactionList/view.jsx
```diff
@@ -23,6 +24,16 @@ class TransactionList extends React.PureComponent {
     return !filter || filter == transaction.type;
   }
 
+  isRevokeable(txid, nout) {
+    // a claim/support/update is revokable if it
+    // is in my claim list(claim_list_mine)
+    return this.props.myClaims.has(`${txid}:${nout}`);
```
kauffj: This is fine, but if we find ourselves doing this repeatedly (which I think we are), it may make sense to have outpoint be a first-class property so it's not rebuilt in 16 million different places.

ui/js/component/transactionList/internal/TransactionListItem.jsx
```diff
@@ -6,12 +6,59 @@ import Link from "component/link";
...
+    this.props.revokeClaim(abandonData);
+  }
+
+  getLink(type) {
+    if (type == "tip") {
+      return (
+        <Link
+          onClick={this.abandonClaim.bind(this)}
+          icon="icon-unlock-alt"
+          title={__("Unlock Tip")}
```
kauffj: Just "Unlock"

ui/js/component/transactionList/internal/TransactionListItem.jsx
```diff
@@ -6,12 +6,59 @@ import Link from "component/link";
...
+        />
+      );
+    } else {
+      return (
+        <Link
+          onClick={this.abandonClaim.bind(this)}
+          icon="icon-trash"
+          title={__("Delete")}
```
kauffj: Let's go with "Revoke"

ui/js/component/transactionList/internal/TransactionListItem.jsx
kauffj: Modal body:

Are you sure you want to unlock these credits?

These credits are permanently yours and can be unlocked at any time. Unlocking them allows you to spend them, but can hurt the performance of your content in lookups and search results. It is recommended you leave tips locked until you need or want to spend them.

ui/js/component/transactionList/internal/TransactionListItem.jsx
```diff
@@ -6,12 +6,59 @@ import Link from "component/link";
...
+      txid,
+      type,
+      nout,
+    } = this.props.transaction;
+    let msg;
+
+    if (type == "tip") {
+      msg = "This will reduce the committed credits to the URL";
+    } else {
+      msg = "This will reclaim you lbc, back to your account";
```
kauffj: Modal body:

Are you sure want to revoke this claim?

This will prevent others from resolving and accessing the content you published. It will return the LBC to your spendable balance, less a small transaction fee.

ui/js/modal/modalRevokeClaim/view.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+  render() {
+    const { msg, closeModal } = this.props;
+
+    return (
+      <Modal
+        isOpen={true}
+        contentLabel={__("Confirm Claim Revoke")}
+        type="confirm"
+        confirmButtonLabel={__("Yes, Revoke")}
```
kauffj: "Yes, Unlock" or "Yes, Revoke"

ui/js/modal/modalRevokeClaim/view.jsx
```diff
@@ -0,0 +1,36 @@
...
+        onConfirmed={this.revokeClaim.bind(this)}
+        onAborted={closeModal}
+      >
+        <p>
```
kauffj: In additional to the copy I outlined above, please add an h3 here with either "Confirm Tip Unlock" or "Confirm Claim Revoke" as appropriate.

ui/js/selectors/claims.js
kauffj: My proposal:

`cantFigureOutVarName` -> `selectMyClaims`
`selectMyClaimsRaw` -> `selectMyActiveClaims` (or another word that says the difference between this and `selectMyClaims` is the ones pending abandon are filtered)
`selectMyClaims` -> `selectMyPublishClaims`
`selectAllMyClaimsByTxidNout` -> `selectMyClaimsByOutpoint`

ui/js/modal/modalRevokeClaim/view.jsx
hackrush01: Done.

ui/js/modal/modalRevokeClaim/view.jsx
hackrush01: Done.

ui/js/component/transactionList/view.jsx
hackrush01: Not doing this, as per the discussion in the meeting.

ui/js/component/transactionList/internal/TransactionListItem.jsx
hackrush01: Done.

ui/js/component/transactionList/internal/TransactionListItem.jsx
hackrush01: Done.

ui/js/component/transactionList/internal/TransactionListItem.jsx
hackrush01: Done.

ui/js/component/transactionList/internal/TransactionListItem.jsx
hackrush01: Done.

ui/js/component/transactionList/internal/TransactionListItem.jsx
hackrush01: Done.

ui/js/actions/content.js
hackrush01: If the txn succeeds then `txid` _will be_ returned in the `results`. So if we make into the `else` code then, for some reason the transaction must have failed.

ui/js/actions/content.js
hackrush01: Since an abandoning txn is essentially a txn, so I figured that a `TRANSACTION_FAILED` modal would be apt. We could maybe go for a better error message altogether for a failed txn though. Or did you have something else that could be better in this case?

ui/js/actions/content.js
hackrush01: Done.

ui/js/selectors/claims.js
hackrush01: Counter Proposal:
`cantFigureOutVarName` -> `selectMyClaimRaw`
`selectMyClaimsRaw` -> `selectMyActiveClaims`
`selectMyClaims` -> No Change
`selectAllMyClaimsByTxidNout` -> `selectAllMyClaimsByOutpoint`

ui/js/actions/file_info.js
hackrush01: Done.

hackrush01:
@kauff We also have a `support` type txn, under present circumstances, it will show the message and button prompt of a claim/update. Is that okay or do we need to add different message for that?

ui/js/component/transactionList/internal/TransactionListItem.jsx
hackrush01: Done.

hackrush01:
@kauffj After making the above changes, this line: https://github.com/lbryio/lbry-app/pull/697/files#diff-313e02014d06fe554d4dd5ea2c2e7139R536 and the next one are not getting executed, instead, there are some errors in the console

```
Uncaught (in promise) TypeError: claims.map is not a function
    at eval (claims.js:159)
    at eval (index.js:76)
    at eval (index.js:36)
    at eval (index.js:90)
    at eval (index.js:36)
    at Function.select [as mapToProps] (index.js:30)
    at mapToPropsProxy (wrapMapToProps.js:47)
    at handleNewState (selectorFactory.js:56)
    at handleSubsequentCalls (selectorFactory.js:73)
    at pureFinalPropsSelector (selectorFactory.js:78)
```
This was not happening during the previous implementation.

The line `selectors/claims.js:156` is https://github.com/lbryio/lbry-app/pull/697/files#diff-ad91716cbe6f8ec74f35446514bdf1a3R173

hackrush01:
The error is still not fixed and it is coming up during reloading of a file page. Removing you as a reviewer for now.

ui/js/actions/content.js
kauffj: I think this is mildly incorrect, but I'm going to merge it anyway. If say, the API signatures of the daemon change such that `txid` is no longer confirmed (suppose it is renamed to `transaction_id`), this would still be a successful transaction, but now report as a failed transaction.

In general, we ought to trust that if the daemon says it is a success in terms of the JSONRPC response, then it is a success.

If the success doesn't match the signature we expect, it's fine and good to panic/error, but the most likely explanation of this would be a change in protocol API signatures without a corresponding application update.

PR #713, opened 2017-11-05
Simplify webpack

seanyesmunt:
Simplifying by adding more webpack config files?! Yep.

There are now three webpack files for the `ui`, but only because there is now a common config with the webpack essentials (entry, output, babel) and two other configs with extend the base config.

#### Before
`webpack.config.js` was the prod config
`webpack.dev.config.js` was the dev config
A lot of duplication of basic webpack setup

#### After
`webpack.common.js` is the base webpack config
`webpack.prod.js` extends the base config
`webpack.dev.js` extends the base config
Also updated to `webpack@3`

Note: Most of the additions are from yarn
cc @liamcardenas

### DIFF
--- a/ui/webpack.prod.js
+++ b/ui/webpack.prod.js
@@ -1,0 +1,11 @@
+const webpack = require("webpack")
+const merge = require('webpack-merge');
+const common = require('./webpack.common.js');
+
+module.exports = merge(common, {
+  plugins: [
+    new webpack.DefinePlugin({
+      ENV: JSON.stringify("production")
+    })
+   ]
+});
--- a/ui/package.json
+++ b/ui/package.json
@@ -70,8 +70,9 @@
     "node-loader": "^0.6.0",
     "prettier": "^1.4.2",
     "style-loader": "^0.18.2",
-    "webpack": "^2.6.1",
+    "webpack": "^3.0.0",
     "webpack-dev-server": "^2.4.4",
+    "webpack-merge": "^4.1.1",
     "webpack-notifier": "^1.5.0",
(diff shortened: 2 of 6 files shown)


ui/webpack.prod.js
```diff
@@ -0,0 +1,13 @@
+const webpack = require("webpack")
+const merge = require('webpack-merge');
+const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
```
kauffj: Does this actually add any significant performance benefit when the files are being loaded locally? I suppose marginally, since the file still needs to be read from disk and interpreted, but my guess is it is very minor. 

At the same time, we fairly frequently ask users to give console output related to users.

If there's no real performance win here, I'd suggest this is not worth the added difficulty in debugging user issues, but I could possibly be persuaded otherwise.

seanyesmunt: Good point. I'll take it out for now until I can get some real performance numbers. Then we can decide.

kauffj: @seanyesmunt once this is removed, go ahead and merge this into master.

Recommended merging practices are here: https://github.com/lbryio/lbry/wiki/Branching-and-Merging

I'd say I follow these about 80% of the time :

kauffj: @liamcardenas you should check out above link as well

seanyesmunt:
@kauffj Removed the prod minification. I don't have merge access.

lyoshenka:
@seanyesmunt accept your @lbryio invite :-)

seanyesmunt:
Thanks! :)

Issue #616, opened 2017-09-25
Kill npm and other build clean up


kauffj:
2 package managers is pretty dumb.

In addition, our set up steps require manual installation of modules that ought to be installable via configuration files.

There was a time when the setup process was:

- Install package manager
- Checkout project
- Run `./build.sh`

We ought to get back to this standard.

lyoshenka:
npm is only used for `npm rebuild node-sass` ([1](https://github.com/lbryio/lbry-app/blob/dafe9bb87d6dc3f2438aca22f8a47643dee8fc4f/build/build.sh#L69), [2](https://github.com/lbryio/lbry-app/blob/ce0dd5ae62e5daab09523aa49831ca707d521c34/build/build.ps1#L24)) and for [installing yarn on windows](https://github.com/lbryio/lbry-app/blob/ce0dd5ae62e5daab09523aa49831ca707d521c34/build/build.ps1#L7). If we can replace both and the build script still works, we will have killed npm.

Yarn does not support `rebuild` yet. See https://github.com/yarnpkg/yarn/issues/756

PR #716, opened 2017-11-07
Build improvements (WIP)

alexliebowitz:
Addresses the build issues raised in #616 as well as some others.

This streamlines our build process on macOS and Linux down to just this:
```INSTALL_DEPENDENCIES=true ./build.sh```

I added the `INSTALL_DEPENDENCIES` param, which causes `prebuild.sh` to run without building the whole app. `prebuild.sh` is already automating all of the installation steps suggested in the instructions, so why not use it?
 - I wasn't able to remove npm from the process, but I don't think it's important. As Grin says, it's only being used for `electron-rebuild` and for installing yarn on Windows. Redoing our `electron-rebuild` setup to avoid npm seems to be a can of worms and may not even be possible. There's no official command line way to install yarn on Windows (nothing like `get-pip.py`)
   - We might be able to use `msiexec` to run the Yarn Windows installer instead, but clearly not a priority.
- This takes some steps toward making it possible to use a path with spaces. This helps a lot when working with test virtual machines, as both Parallels and VMWare can only share files with the host OS through a "Shared Files" directory that contains a space.
    - The keytar install still chokes on the spaces (seems to be related to https://github.com/nodejs/node-gyp/issues/65) so this isn't actually working yet
 
There are 2 fairly small issues outstanding:
 - `prebuild.sh` is probably installing some stuff that is only needed for building a full app, not just a normal build. If needed, I will add conditionals so it's not installing unnecessary stuff.
 - `prebuild.sh` uses `apt-get`, so this won't work on Fedora. Hopefully, I can just have it use `yum` on Fedora and it will be able to install the same packages. It has been heavily tested on fresh VMs for Ubuntu and Windows.

### CODE REVIEW

File: build/build.sh [hunk 1]
```diff
@@ -90,7 +96,7 @@ DAEMON_URL=$(echo ${DAEMON_URL_TEMPLATE//DAEMONVER/$DAEMON_VER} | sed "s/OSNAME/
 DAEMON_VER_PATH="$BUILD_DIR/daemon.ver"
 echo "$DAEMON_VER_PATH"
 if [[ ! -f $DAEMON_VER_PATH || ! -f $ROOT/app/dist/lbrynet-daemon || "$(< "$DAEMON_VER_PATH")" != "$DAEMON_VER" ]]; then
-    wget --quiet "$DAEMON_URL" -O "$BUILD_DIR/daemon.zip"
+    curl -sL -o "$BUILD_DIR/daemon.zip" "$DAEMON_URL"
```
lyoshenka: why is curl better?

kauffj:
LGTM. I agree there should possibly be a larger disclaimer about what this command should do, but otherwise this seems a lot nicer.

@lyoshenka is this one for you to approve and merge, or me?

alexliebowitz:
@kauffj @lyoshenka 

It seems like everything in `prebuild.sh` is necessary even for a normal build, so we're all set there.

If we can think of a better name for the variable, that would fix both issues that we are talking about (shorter variable name, clearer indication about what the command does). `PREPARE_BUILD`?

Once we finalize an interface, this can be merged, as long as we're OK with the fact that it doesn't support Fedora yet. I'm going to work on adding `yum` and hoping to have it done by the end of tomorrow (11/8). If it's not that important, we can merge this and do Fedora in a separate PR.

kauffj:
@lyoshenka you were going to merge this?

PR #732, opened 2017-11-12
Fix scroll bar issues

btzr-io:


seanyesmunt:
Hey @btzr-io, sorry I forgot to message you, but I decided to go a different route and do some refactoring for this fix. Still a work in progress but my changes are here: https://github.com/lbryio/lbry-app/pull/750

This way we will be able to get rid of the `setTimeout` for handling the scroll restore

PR #737, opened 2017-11-13
Move all redux code into /redux folder

seanyesmunt:
So all the redux stuff is in one place.


liamcardenas:
@seanyesmunt lets not merge this until after my patch goes out

kauffj:
:+1: this is good with me.

It's possibly worth hoisting `ui` to the top-level entirely, but I haven't thought about this too much.

Issue #721, opened 2017-11-09
Inconsistent thumbnail sizes in 'Trending' section


wilsonk:
## The Issue
One can game the trending section somewhat by using a very long channel name to make the size of the thumbnail bigger than others in that section.
### Steps to reproduce
1.View trending section today (Nov 9, 2017)

### Expected behaviour
All thumbnails should be the same size

### Actual behaviour
Thumbnails with large channel names are larger than the rest

## System Configuration

 

- LBRY Daemon version: 0.17.1
- LBRY App version:0.17.1
- LBRY Installation ID: 26bru4Kjp3i3VM5KTFVV3ESEkDWEb9CCA2ate4L9XJYTr8PHUK4jPDQqz4JsG73Wfr

- Operating system: Linux (Linux-4.4.0-97-generic-x86_64-with-Ubuntu-16.04-xenial)

## Screenshots

![screenshot_2017-11-09_11-45-19](https://user-images.githubusercontent.com/38802/32623539-bd36c6e4-c543-11e7-845e-38d2c4c83e4b.png)

wilsonk:
Sorry, maybe that isn't the channel name being too long that is the problem, but the actual size of the file name. Perhaps that forces the 'red ship/free' icon to be pushed onto the next line which then makes the thumbnail longer not wider.

Easier to game the system this way, but with the ellipses eliding the file name size it might be easier to fix?

wilsonk:
Hmm, maybe I will eventually get this right!  It actually looks like a combination of both the long file name and channel name may be the problem...??

seanyesmunt:
I think we should remove the card design. It can be tough to deal with when trying to keep boxes the same height.

If we remove the card we could increase the visible text limit and not have to worry about different container heights. Something like this redesign http://lab.zzk.fr/lbry/

oakey22:
@seanyesmunt i absolutely love that design, so professional looking

akinwale:
Created a PR to fix this since it kept bothering me.

Also, @seanyesmunt that new design is amazing! You should show the entire team.

liamcardenas:
@tzarebczan can you verify this is working?

PR #744, opened 2017-11-14
Issue #721 fix - Inconsistent thumbnail sizes

akinwale:
Fix for long channel names resulting in the inconsistent thumbnail sizes.

### DIFF
--- a/ui/scss/component/_channel-indicator.scss
+++ b/ui/scss/component/_channel-indicator.scss
@@ -1,4 +1,12 @@
 
+.channel-name {
+  width: calc(var(--card-small-width) * 2 / 3);
+  display: inline-block;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis
+}
+
 .channel-indicator__icon--invalid {
   color: var(--color-error);
 }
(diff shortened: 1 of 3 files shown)


ui/scss/component/_channel-indicator.scss
```diff
@@ -1,4 +1,12 @@
 
+.channel-name {
+  width: calc(var(--card-small-width) * 2 / 3);
```
seanyesmunt: Is this needed? I think it should be fine with just the `nowrap` and `ellipsis`. If it's extending to the edge, we should add a padding to the card.

akinwale: Without specifying the width, long text will push the price indicator to the next line. I suppose we could simplify this using a percentage width instead of the calculation. 60% or 70% should be fine.

seanyesmunt: Gotcha. This should be fine for now, we will be moving to the new design sometime soon.

kauffj: (Proper solution is probably a flex layout rather than `calc`, but not demanding propriety. Especially since we're likely to redesign soon anyway.)

PR #747, opened 2017-11-15
Add prominent upgrade button when upgrade is pending

IGassmann:
I implemented a timer for checking for new upgrades every 10 minutes in a redux action.


kauffj:
@liamcardenas would like to see your review on this before mine.

ui/js/component/app/view.jsx
```diff
@@ -21,6 +22,8 @@ class App extends React.PureComponent {
       checkUpgradeAvailable();
     }
 
+    initCheckUpgradeTimer();
```
liamcardenas: doesn't this check for an upgrade in the if statement above? maybe replace checkUpgradeAvailable on line 22 with this line?

ui/js/component/app/view.jsx
```diff
@@ -30,6 +33,8 @@ class App extends React.PureComponent {
     this.setTitleFromProps(this.props);
   }
 
+  componentDidMount() {}
```
liamcardenas: please remove unused functions, thanks!

ui/js/component/app/index.js
```diff
@@ -2,10 +2,15 @@ import React from "react";
 import { connect } from "react-redux";
 import { selectPageTitle } from "selectors/navigation";
 import { selectUser } from "selectors/user";
-import { doCheckUpgradeAvailable, doAlertError } from "actions/app";
+import {
+  doCheckUpgradeAvailable,
+  doStartUpgradeCheckTimer,
```
liamcardenas: I dont think we ever use doStartUpgradeCheckTimer and also given some of my comments below we also might not need doCheckUpgradeAvailable

ui/js/component/app/view.jsx
IGassmann: The timer doesn't trigger a check for an upgrade immediately. That's why I kept the ``checkUpgradeAvailable();``, so the app does check for an upgrade when it starts.

ui/js/component/app/index.js
IGassmann: Yep, my bad. I missed this one probably because the imports were folded in my IDE. I should've been more precise.

ui/js/component/app/view.jsx
IGassmann: Again. Should've been more precise.

ui/js/actions/app.js
```diff
@@ -5,12 +5,15 @@ import {
...
   selectUpgradeFilename,
+  selectIsUpgradeSkipped,
+  selectRemoteVersion,
 } from "selectors/app";
 import { doFetchDaemonSettings } from "actions/settings";
 import { doBalanceSubscribe } from "actions/wallet";
 import { doAuthenticate } from "actions/user";
 import { doFetchFileInfosAndPublishedClaims } from "actions/file_info";
 import * as modals from "constants/modal_types";
+import { selectBalance } from "../selectors/wallet";
```
kauffj: Probably better to omit `../` for consistency.

ui/js/actions/app.js
```diff
@@ -63,33 +66,31 @@ export function doDownloadUpgrade() {
...
       onProgress: p => dispatch(doUpdateDownloadProgress(Math.round(p * 100))),
       directory: dir,
     };
-    download(
-      remote.getCurrentWindow(),
-      selectUpdateUrl(state),
-      options
-    ).then(downloadItem => {
-      /**
+    download(remote.getCurrentWindow(), selectUpdateUrl(state), options).then(
```
kauffj: Did you make these layout changes or was this Prettier? If it wasn't Prettier, please don't change code formatting and leave it to Prettier.

(Prettier: https://github.com/prettier/prettier)

ui/js/actions/app.js
```diff
@@ -129,22 +130,56 @@ export function doCancelUpgrade() {
...
+          remoteVersion,
+        },
+      });
+
+      // If there's an available upgrade and the user hasn't skipped it or if there's a newer one, show un upgrade modal
+      if (
+        upgradeAvailable &&
+        (!selectIsUpgradeSkipped(state) ||
+          remoteVersion !== selectRemoteVersion(state))
+      ) {
```
kauffj: Since this function is now firing repeatedly rather than just on first-run, it is worth some consideration as to whether this can fire at an inappropriate time.

For example: what if a modal is already open? Should this overwrite it?

ui/js/component/app/index.js
```diff
@@ -2,10 +2,15 @@ import React from "react";
...
+  doAlertError,
+} from "actions/app";
 import { doRecordScroll } from "actions/navigation";
 import { doFetchRewardedContent } from "actions/content";
 import App from "./view";
+import { doInitCheckUpgradeTimer } from "../../actions/app";
```
kauffj: Same comment as previously, let's keep one method for writing paths.

ui/js/component/app/view.jsx
kauffj: If this can (or should) wait until the daemon is ready, we have other actions that are first-run actions here:

https://github.com/lbryio/lbry-app/blob/master/ui/js/redux/actions/app.js#L177

If this should happen before the daemon is ready, then this is an appropriate place.

ui/js/component/header/index.js
```diff
@@ -12,18 +12,23 @@ import {
...
+import { selectIsUpgradeAvailable } from "../../selectors/app";
+import { doDownloadUpgrade } from "../../actions/app";
 
 const select = state => ({
   isBackDisabled: selectIsBackDisabled(state),
   isForwardDisabled: selectIsForwardDisabled(state),
+  isUpgradeAvailable: selectIsUpgradeAvailable(state),
   balance: formatCredits(selectBalance(state) || 0, 1),
   publish: __("Publish"),
+  upgradeLabel: __("Upgrade App"),
```
kauffj: I see you followed the pattern established above, but I'm guessing this is legacy behavior from a time when that label could switch somehow. Both `upgradeLabel` and `publish` can just be put directly in `view.jsx`.

ui/js/reducers/app.js
```diff
@@ -64,7 +64,7 @@ reducers[types.SKIP_UPGRADE] = function(state, action) {
   sessionStorage.setItem("upgradeSkipped", true);
 
   return Object.assign({}, state, {
-    upgradeSkipped: true,
+    isUpgradeSkipped: true,
```
kauffj: I like this change :)

ui/js/actions/app.js
IGassmann: Good point! We shouldn't restart the upgrade process if one is already being run.

ui/js/actions/app.js
IGassmann: I haven't touched this from what I remember. On my side, there's a TODO there. It may be Prettier.

ui/js/actions/app.js
kauffj: Not just the upgrade process - what if a modal was open displaying an important error message or point of feedback? Only one modal can be open at a time - we don't have a modal stack.

ui/js/actions/app.js
kauffj: Ok, I'm guessing something just changed such that Prettier reformatted it. You're good here.

ui/js/actions/app.js
IGassmann: Yep! I thought about that when I fixed it.

liamcardenas:
@IGassmann can you fix these merge conflicts?

PR #750, opened 2017-11-15
refactor scroll navigation/restore

seanyesmunt:
Doing some refactoring to how the scroll restore works. This fixes the issue where after navigating, pages were scrolled down a bit.

This gets ride of the need to use a `setTimeout` which was required to let the new page load before scrolling. We can handle that inside the app component which I think makes more sense because it is interacting with the DOM. I also added throttling to the scroll listener, no need to constantly fire events when scrolling.

#### After (not sure why my mouse disappeared)
scroll

#### Some things to think about/future ideas
Do we want to restore for all pages? Forwards _and_ back? Not really sure.
If users click the `settings` icon while on the settings page, maybe scroll to the top instead of doing nothing?

### DIFF
--- a/ui/js/component/app/view.jsx
+++ b/ui/js/component/app/view.jsx
@@ -4,8 +4,14 @@ import Header from "component/header";
 import Theme from "component/theme";
 import ModalRouter from "modal/modalRouter";
 import lbry from "lbry";
+import throttle from "util/throttle";
 
 class App extends React.PureComponent {
+  constructor() {
+    super();
+    this.mainContent = undefined;
+  }
+
   componentWillMount() {
     const {
       alertError,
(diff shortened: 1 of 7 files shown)


ui/js/component/app/view.jsx
```diff
@@ -4,8 +4,14 @@ import Header from "component/header";
 import Theme from "component/theme";
 import ModalRouter from "modal/modalRouter";
 import lbry from "lbry";
+import _ from "lodash";
```
kauffj: This is easy, but it is not simple.

(This is a distinction you will see me make often. Please watch "Simple Made Easy" by Rich Hickey if you have never seen it. It's relatively short and funny and covers one of the most important lessons in programming. You too @liamcardenas)

If all we need is a throttle method, rather than bringing in an entire library, I'd propose simply bringing that method or an analogous one in directly and making it a core part of our code.

From a quick Google, there's no shortage of implementations for throttling javascript events. Here are several in one thread: https://stackoverflow.com/questions/27078285/simple-throttle-in-js

seanyesmunt:
Added the redux to this changelog because I didn't add it to the other PR

seanyesmunt:
@kauffj pinging for re-review

ui/js/redux/selectors/navigation.js
```diff
@@ -146,3 +146,9 @@ export const selectHistoryStack = createSelector(
   _selectState,
   state => state.stack
 );
+
+// returns current page attributes (scrollY, path)
+export const selectCurrentPageAttributes = createSelector(
```
kauffj: This function returns a single entry in the history stack. I would propose a name that better reflects this, perhaps something like `selectCurrentHistoryEntry` (or use "peek" somehow if we want to keep stack terminology consistent).

PR #756, opened 2017-11-17
Fixed regression in #697

hackrush01:


ui/js/redux/selectors/claims.js
```diff
@@ -171,7 +171,8 @@ export const selectMyClaimsWithoutChannels = createSelector(
 
 export const selectAllMyClaimsByOutpoint = createSelector(
   selectMyClaimsRaw,
-  claims => new Set(claims.map(claim => `${claim.txid}:${claim.nout}`))
+  claims =>
+    new Set(claims && claims.map(claim => `${claim.txid}:${claim.nout}`))
```
kauffj: Won't this be `new Set(false)` when claims is false? This may work but is probably slightly wrong.

kauffj:
@hackrush01 I wanted to switch the check slightly and it was easier to just introduce it separately than correct this PR. So it is now merged via 3c941043

Thanks!

PR #767, opened 2017-11-21
Fix menubar icon for macOS

seanyesmunt:
Somewhat better. I was just playing around with resizing. We might have to rework an icon specifically for mac displays if we want it more precise than this (larger space above/below?). Should be fine for now, definitely looks better than before.

test

screen shot 2017-11-21 at 1 13 09 am

Related issue https://github.com/electron/electron/issues/417


seanyesmunt:
@liamcardenas can you test this on your mac? It _should_ be the same (I think), but not sure since we have different screen sizes.

app/main.js
```diff
@@ -226,7 +226,15 @@ function createWindow () {
 function createTray () {
   // Minimize to tray logic follows:
   // Set the tray icon
-  const iconPath = path.join(app.getAppPath(), "/dist/img/fav/32x32.png");
+  let iconPath;
+  if (process.platform === 'darwin') {
```
kauffj: https://gist.github.com/lyoshenka/0a43205aa9a072b196ff87e2c689a8b9

liamcardenas:
@seanyesmunt how can we check and fix this issue on Windows and Debian if needed?

seanyesmunt:
@liamcardenas I'll spend some time once I get home updating windows/linux icons. From what I remember I don't think the windows icon looked that bad.

PR #776, opened 2017-11-22
fix channel name alignment on file page

seanyesmunt:
Not super happy with this fix, but good to get it in sooner since it looks so bad right now.


seanyesmunt:
I don't know why Prettier is changing parts of files I didn't touch. The entire file _should_ be "prettified", maybe we should do another global prettier command to prevent this in the future. It can be confusing and hard to tell the actual changes I made.

PR #778, opened 2017-11-22
Added Flow Typing

liamcardenas:
Not ready for review yet

- [x] Add flow to project
- [x] Integrate with traditional webpack flow
- [x] ~~Allow background flow daemon to run w/o webpack integration (crossing out because I am waiting until watch.sh refactor is completed)~~
- [ ] Add standards and documentation for using flow
- [x] Add example typed file/component (maybe this will be in conjunction with the immutable js integration)


src/renderer/js/redux/reducers/app.js
```diff
@@ -1,22 +1,45 @@
...
+
+// $FlowIssue: Flow cannot find election
 const { remote } = require("electron");
+
 const application = remote.app;
 const win = remote.BrowserWindow.getFocusedWindow();
 
 const reducers = {};
-const defaultState = {
+type appState = {
```
kauffj: thoughts on all upper case for types?

src/renderer/js/redux/reducers/app.js
```diff
@@ -1,22 +1,45 @@
...
+  modalProps: mixed,
+  platform: string,
+  upgradeSkipped: boolean,
+  daemonVersionMatched: ?boolean,
+  daemonReady: boolean,
+  hasSignature: boolean,
+  badgeNumber: number,
+  volume: number,
+  downloadProgress?: number,
+  upgradeDownloading?: boolean,
```
kauffj: Any particular reason are these optional properties that we don't set on `defaultState`? IMO optional properties should generally be avoided.

src/renderer/js/redux/reducers/app.js
kauffj: (also, state types will generally need to be exported so they can be used by selectors)

src/renderer/js/redux/reducers/app.js
```diff
@@ -1,22 +1,45 @@
...
+  isUpgradeSkipped?: boolean,
+  snackBar?: mixed,
+};
+const defaultState: appState = {
   isLoaded: false,
   modal: null,
   modalProps: {},
   platform: process.platform,
-  upgradeSkipped: sessionStorage.getItem("upgradeSkipped"),
+  upgradeSkipped: sessionStorage.getItem("upgradeSkipped") === "true",
```
kauffj: Can you clarify this change?

(And btw, we want to be killing this storage pattern)

src/renderer/js/redux/reducers/app.js
```diff
@@ -1,22 +1,45 @@
...
 const { remote } = require("electron");
+
 const application = remote.app;
 const win = remote.BrowserWindow.getFocusedWindow();
 
 const reducers = {};
-const defaultState = {
+type appState = {
+  isLoaded: boolean,
+  modal: mixed,
```
kauffj: Isn't this just `?string`

src/renderer/js/redux/reducers/app.js
```diff
@@ -1,22 +1,45 @@
...
+  modalProps: mixed,
```
kauffj: Isn't this an unsealed `Object` or `null`? (not clear on Flow syntax for this: `?{}`?)

src/renderer/flowtype-plugin.js
```diff
@@ -0,0 +1,116 @@
...
+    formatter: function (errorCode, errorDetails) {
+      return 'Flow: ' + errorCode + '\n\n' + errorDetails;
+    },
+  },
+};
+
+
+function flowErrorCode(status) {
+  var error;
+  switch (status) {
```
seanyesmunt: Maybe add a link to flow docs for each of these cases?

src/renderer/js/redux/reducers/app.js
```diff
@@ -1,22 +1,45 @@
...
+  upgradeSkipped: boolean,
+  daemonVersionMatched: ?boolean,
+  daemonReady: boolean,
+  hasSignature: boolean,
+  badgeNumber: number,
+  volume: number,
+  downloadProgress?: number,
+  upgradeDownloading?: boolean,
+  upgradeDownloadComplete?: boolean,
+  checkUpgradeTimer?: mixed,
```
kauffj: If this is the return value for an interval call, I believe it is `?number`

src/renderer/flowtype-plugin.js
```diff
@@ -0,0 +1,116 @@
...
+      error = 'Out of Shared Memory';
+      break;
+  }
+
+  return error;
+}
+
+
+function checkFlowStatus(compiler, next) {
+  var res = spawnSync(flow, store.flowOptions);
```
seanyesmunt: I think we should try to use `let` and `const` to show if a variable will change or not

src/renderer/js/redux/reducers/app.js
```diff
@@ -1,22 +1,45 @@
...
+  isUpgradeAvailable?: boolean,
+  isUpgradeSkipped?: boolean,
+  snackBar?: mixed,
```
kauffj: I don't remember what's put in this but I'm just hating on all of your usage of `mixed` at this point.

src/renderer/js/redux/reducers/app.js
```diff
@@ -164,7 +187,7 @@ reducers[types.VOLUME_CHANGED] = function(state, action) {
   });
 };
 
-export default function reducer(state = defaultState, action) {
+export default function reducer(state: appState = defaultState, action: any) {
```
seanyesmunt: We should have an `action` type

src/renderer/js/redux/reducers/app.js
liamcardenas: I believe mixed is the correct way for to indicate Object

src/renderer/js/redux/reducers/app.js
kauffj: Also also, this should be a sealed object.

src/renderer/js/redux/reducers/app.js
liamcardenas: optional properties can be thought of as implicitly there, but set to "undefined". I am more than happy to add them, though.

The one edge case I could see, and I will double check this, is, as I was saying in slack:

downloadProgress?: number could be a number or undefined
downloadProgress: ?number could be a number or undefined or null

In order to remove optional fields, we would have to either allow null or determine a non-null default value for each field.

src/renderer/js/redux/reducers/app.js
liamcardenas: me too, i'll figure it out

src/renderer/js/redux/reducers/app.js
liamcardenas: I agree, I meant to bring this up, we also need to figure out how to properly do naming.

src/renderer/flowtype-plugin.js
liamcardenas: this file contains code from another repo, I don't want to change it too much because, once they update their npm package, we will be able to use that instead of this file.

src/renderer/js/redux/reducers/app.js
liamcardenas: ah i didn't realize that, thanks.

src/renderer/js/redux/reducers/app.js
liamcardenas: `sessionStorage.getItem("upgradeSkipped")` returns the string: `"true"` so this is, essentailly, casting it to a boolean.

src/renderer/js/redux/reducers/app.js
liamcardenas: yes, thanks

src/renderer/js/redux/reducers/app.js
liamcardenas: completely agreed, I'm trying to figure that out now

There is some cool stuff we can do with flow and actions, but I also don't want to introduce too much unnecessary boilerplate if its not going to practically help us catch errors

src/renderer/js/redux/reducers/app.js
```diff
@@ -1,22 +1,60 @@
...
   platform: process.platform,
-  upgradeSkipped: sessionStorage.getItem("upgradeSkipped"),
+  upgradeSkipped: sessionStorage.getItem("upgradeSkipped") === "true",
   daemonVersionMatched: null,
   daemonReady: false,
   hasSignature: false,
   badgeNumber: 0,
-  volume: sessionStorage.getItem("volume") || 1,
+  volume: Number(sessionStorage.getItem("volume")) || 1,
+  downloadProgress: undefined,
```
liamcardenas: I am going to go through each of these one by one to determine whether or not I can add null, or some other initial value to them instead of undefined, however, there inevitably will be some fields that are initialized as undefined (given what we decided in terms of having no optional types + using undefined as a distinct value). Does this seem alright to you guys? @kauffj @seanyesmunt

kauffj: Yes, given that the existing behavior a property not explicitly set on `defaultState` is to default to `undefined`, it makes sense to set values to `undefined` rather than `null` to match existing behavior.

liamcardenas: Got it. Makes sense. I'm cool with this.

src/renderer/js/redux/reducers/app.js
kauffj: @liamcardenas but this changeset also changes this value from `true` to `"true"`.

(If you just want to cast, I'd propose either a typical cast of `Boolean(foo)` or `!!foo`)

src/renderer/flowtype-plugin.js
seanyesmunt: What is the other module? Seems simple enough, do we need to use it?

src/renderer/js/redux/reducers/app.js
seanyesmunt: I think he's saying no matter how it was stored in `sessionStorage` it was still being returned as `"true"`. 

Using the `!!` operator or `Boolean` will still return `true` for `"false"` strings because strings are truthy.

@liamcardenas is that right?

liamcardenas: that doesn't work, because `Boolean('false')` and `!!'false'` evaluate to `true`

This is the best way to get a true value that is recorded as a string. The alternative would be `JSON.parse(foo)`, but I wouldn't want to be able to introduce arbitrary, non-boolean values.

src/renderer/js/redux/reducers/app.js
kauffj: @liamcardenas I do not think that is correct https://flow.org/en/docs/types/mixed/

src/renderer/js/redux/reducers/app.js
kauffj: Wow, sure enough, the Storage API changes the type! 
```
window.sessionStorage.setItem("foo", true);
undefined
window.sessionStorage.getItem("foo");
"true"
```

Absolute insanity!

liamcardenas: I know, I wouldn't have caught this if flow didn't tell give me a type error for trying to input/expect booleans where it is supposed to be a string.

That's the beauty of type checking, this could've led to some pretty nasty bugs.

src/renderer/flowtype-plugin.js
liamcardenas: this is a pretty decently maintained package: https://github.com/zhirzh/flow-babel-webpack-plugin

I'd like to be able to include other changes made by that developer if possible. I left an issue about updating the NPM package. If it looks abandoned after enough time then we should look at bringing this in-house. I expect to deprecate this file relatively soon though.

liamcardenas: but your right, if need be it is definitely simple

PR #781, opened 2017-11-24
prettier

seanyesmunt:
Running Prettier again, I think some of the lint rules changed so we were seeing a lot of updates to code that wasn't being touched. After this PR's should be looking slimmer again.


kauffj:
@seanyesmunt I don't mind merging this, but I suspect it will just cause merge headaches with #783. Probably easier to merge that then re-run this.

seanyesmunt:
Sure. I'll keep and eye on PRs and try and get it in when it doesn't cause
a lot of issues.

seanyesmunt:
Don't need this, prettier ran on the precommit hook for #783

PR #783, opened 2017-11-24
Update file structure to follow electron-webpack guidelines

IGassmann:
This change the file structure to follow the [webpack-electron](https://webpack.electron.build/project-structure)'s  default one. The [dirs under `src/`](https://github.com/lbryio/lbry-app/tree/restructuring/src) are named after electron's processes: Main (previously, under `app/`) and Renderer (previously, under `ui/`).


seanyesmunt:
👍

kauffj:
I most agree with this. The only result I don't like is that the most used folder actually being a level deeper than it was previously, which kind of seems like a step backwards. We do 90+% of our work inside of `src/renderer/js` and it's 3 levels deep.

IGassmann:
A sort of solution for project files nested too deep into directories is config your IDE or text editor to create "modules". On JetBrains' IDE, you can create them for a specific dir and then solely display it.

Module config file (should work with WebStorm) to place in ´src/renderer/js´:
[react.iml.zip](https://github.com/lbryio/lbry-app/files/1502426/react.iml.zip)

On IntelliJ IDEA:
screen shot 2017-11-24 at 16 31 25

kauffj:
@IGassmann thanks, but it's not really about usability in navigation. I rarely navigate the tree and instead use `Go to File` and `Go to Type` to open everything anyway. It's more about having a hierarchy that breaks things down in a way that is logical, consistent, and useful, particularly for new contributors. 

Since this is evidently the convention, I'm good with it.

liamcardenas:
I agree with @kauffj, adding layers of directories is typically undesirable but is worth doing so since it is the convention and provides substantial benefits..

I looked through the code and it looks like you covered all your bases. Good work! Will a lot of these formatting changes stop showing up in PRs when we use prettier on the whole app?

I’m fine with merging ASAP, I will be back on my laptop later today and I can further test this if need be.

IGassmann:
The formatting changes came from Prettier, like in #781 . Since I moved files, Prettier is probably being applied to all /ui files.

PR #785, opened 2017-11-26
Change header balance to 2 decimal places

kauffj:
Several places in the app default to credit values of 0.01. If we are suggesting hundredths based usage of LBC, the balance in the header should also have two decimal places.


kauffj:
@seanyesmunt if you approve a PR and it doesn't need multiple reviewers (most shouldn't), you can just merge.

PR #794, opened 2017-11-28
Check the users balance before trying to publish/create a channel

seanyesmunt:
Also added a start script just for fun.

Mostly a test to see what it would look like. Not sure if we want to go this route, or go through the daemon for everything

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,9 @@
       "perMachine": true
     }
   },
+  "scripts": {
+    "start": "./node_modules/.bin/electron src/main"
+  },
   "devDependencies": {
     "devtron": "^1.4.0",
     "electron": "^1.7.9",
--- a/src/renderer/js/component/publishForm/index.js
+++ b/src/renderer/js/component/publishForm/index.js
@@ -1,5 +1,10 @@
 import React from "react";
 import { connect } from "react-redux";
 import PublishForm from "./view";
+import { selectBalance } from "redux/selectors/wallet";
 
-export default connect(null, null)(PublishForm);
+const select = state => ({
+  balance: selectBalance(state),
+});
+
+export default connect(select, null)(PublishForm);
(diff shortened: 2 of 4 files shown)


src/renderer/js/component/publishForm/index.js
```diff
@@ -1,5 +1,11 @@
 import React from "react";
 import { connect } from "react-redux";
 import PublishForm from "./view";
+import { selectBalance } from "redux/selectors/wallet";
+import { formatFullPrice } from "util/formatCredits";
 
-export default connect(null, null)(PublishForm);
+const select = state => ({
+  balance: formatFullPrice(selectBalance(state) || 0, 10),
```
kauffj: Why format this? Also we should probably trust selector to fallback to 0?

kauffj:
@seanyesmunt does the start script just make it so you can run the app with `npm start`?

src/renderer/js/component/publishForm/index.js
seanyesmunt: Ignore this for now. Just copy/pasted from somewhere else

seanyesmunt:
Yeah. Or "yarn start". Still need to build before running it

package.json
```diff
@@ -66,6 +66,9 @@
       "perMachine": true
     }
   },
+  "scripts": {
+    "start": "./node_modules/.bin/electron src/main"
```
liamcardenas: does it work to just say "electron src/main" since electron is in the bin?

seanyesmunt: It does, I like to show the full path so others know exactly whats going on

seanyesmunt: I can change it to `electron src/main` if you think that looks better

liamcardenas: hmm no need, lets keep it as is then

PR #808, opened 2017-12-04
 Auto Update with electron-updater (WIP)

alexliebowitz:
Build is fixed now and the UI is practically done. The main issue is that I'm getting bug at the very end of the process: when I call `appUpdater.quitAndRestart()` to trigger the install, it closes the window but doesn't exit the app or do the install. Hopefully it won't be too hard to figure out.


src/main/main.js
```diff
@@ -554,4 +559,21 @@ ipcMain.on('get-auth-token', (event) => {
 
 ipcMain.on('set-auth-token', (event, token) => {
   keytar.setPassword("LBRY", "auth_token", token ? token.toString().trim() : null);
-});
\ No newline at end of file
+});
+
+autoUpdater.logger = log;
+autoUpdater.on('checking-for-update', () => {
```
kauffj: It might make sense to split the update logic into it's own file if it's not interacting with `main.js` much. `main.js` is already a little big and cluttered.

src/main/index.js
```diff
@@ -429,8 +450,26 @@ app.on('before-quit', event => {
...
+      autoUpdater.quitAndInstall();
+    } else if (process.platform == 'win32' && !showingAutoUpdateCloseAlert) {
+      // We have an update downloaded, but the user declined it (or closed the app without
+      // accepting it). Now the user is closing the app, so the new update will install.
+      // On Mac this is silent, but on Windows they get a confusing permission escalation
+      // dialog, so we show Windows users a warning dialog first.
+      event.preventDefault();
+      showingAutoUpdateCloseAlert = true;
+      dialog.showMessageBox({
+        title: "LBRY will upgrade",
```
kauffj: "LBRY Will Upgrade"

src/main/index.js
```diff
@@ -429,8 +450,26 @@ app.on('before-quit', event => {
...
+        message: "Please select \"Yes\" at the upgrade prompt shown after the app closes.",
```
kauffj: "LBRY has a pending upgrade. Please select "Yes" to install it on the prompt shown after this one."

(Assuming this is correct, and that prompt isn't shown until I ACK this one).

src/renderer/modal/modalAutoUpdateDownloaded/view.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+    return (
+      <Modal
+        isOpen={true}
+        type="confirm"
+        contentLabel={__("Update downloaded")}
+        confirmButtonLabel={__("Update and Restart")}
+        abortButtonLabel={__("Don't Update")}
+        onConfirmed={() => {
+          ipcRenderer.send("autoUpdateAccepted");
```
kauffj: This isn't necessarily a blocker, but is there a strong reason we have to expose `ipcRenderer` inside of a component, as well as put all of the handling of these events inside of `main/index.js`? Could this component instead dispatch an `APP_QUIT` action, and a reducer could close the app? (Or the top-level component?)

src/renderer/modal/modalAutoUpdateDownloaded/view.jsx
```diff
@@ -0,0 +1,40 @@
...
+        }}
+        onAborted={() => {
+          ipcRenderer.send("autoUpdateDeclined");
+          closeModal();
+        }}
+      >
+        <section>
+          <h3 className="text-center">{__("LBRY Leveled Up")}</h3>
```
kauffj: I noticed a `<br/>` was dropped here.

src/renderer/modal/modalAutoUpdateDownloaded/view.jsx
```diff
@@ -0,0 +1,40 @@
...
+class ModalAutoUpdateDownloaded extends React.PureComponent {
+  render() {
+    const { closeModal } = this.props;
+
+    return (
+      <Modal
+        isOpen={true}
+        type="confirm"
+        contentLabel={__("Update downloaded")}
+        confirmButtonLabel={__("Update and Restart")}
```
kauffj: "Use It Now"

src/renderer/modal/modalAutoUpdateDownloaded/view.jsx
```diff
@@ -0,0 +1,40 @@
...
+        onAborted={() => {
+          ipcRenderer.send("autoUpdateDeclined");
+          closeModal();
+        }}
+      >
+        <section>
+          <h3 className="text-center">{__("LBRY Leveled Up")}</h3>
+          <p>
+            {__(
+              "A new version of LBRY has been downloaded and is ready to install."
```
kauffj: A new version of LBRY has been released, downloaded, and is ready for you to use pending a restart.

src/renderer/modal/modalAutoUpdateDownloaded/view.jsx
```diff
@@ -0,0 +1,40 @@
...
+  render() {
+    const { closeModal } = this.props;
+
+    return (
+      <Modal
+        isOpen={true}
+        type="confirm"
+        contentLabel={__("Update downloaded")}
+        confirmButtonLabel={__("Update and Restart")}
+        abortButtonLabel={__("Don't Update")}
```
kauffj: "Not Now" or "Upgrade on Restart"

src/renderer/modal/modalAutoUpdateDownloaded/view.jsx
alexliebowitz: There are other modals (most?) with no `<br />`, and it's not semantic, and seemed to look fine. What did you think of the look?

src/main/index.js
alexliebowitz: You are correct, and I'll update the wording (along with all of the other wording changes you mentioned)

src/renderer/modal/modalAutoUpdateDownloaded/view.jsx
alexliebowitz: I was going back and forth on this. Technically, neither of these IPC calls have anything to do with the store, and they do not lead to DOM updates. The "autoUpdateAccepted" IPC call essentially tells the main app that it's time to shut down (not UI state) and the "autoUpdateDeclined" IPC call just tells the main process to remember that the update was declined so that it can show the extra dialog on Windows (again, not UI state; the dialog only happens after the UI is closed).

Is it typical to move code into Redux land when it is only used for its side effects, and doesn't affect anything going on in the browser window?

kauffj: I think this is fine to leave as-is, at least for now.

build/build.sh
```diff
@@ -79,6 +79,12 @@ if [ "$FULL_BUILD" == "true" ]; then
 
   yarn build
 
+  # Workaround: TeamCity expects the dmg to be in dist/mac, but in the new electron-builder
+  # it's put directly in dist/ (the right way to solve this is to update the TeamCity config)
+  if $OSX; then
+    cp dist/*.dmg dist/mac
```
IGassmann: Wouldn't be better to change TeamCity config directly?

package.json
```diff
@@ -20,7 +20,6 @@
     "dev": "electron-webpack dev",
     "compile": "electron-webpack && yarn extract-langs",
     "build": "yarn compile && electron-builder build",
-    "build:dir": "yarn build -- --dir -c.compression=store -c.mac.identity=null",
```
IGassmann: We should keep this. It was requested here https://github.com/lbryio/lbry-app/pull/933.

src/main/index.js
```diff
@@ -5,10 +5,28 @@ import SemVer from 'semver';
...
+
+// This is set to true if an auto update has been downloaded through the Electron
+// auto-update system and is ready to install. If the user declined an update earlier,
+// it will still install on shutdown.
+let autoUpdateDownloaded = false;
+
+// Keeps track of whether the user has accepted an auto-update through the interface.
+let autoUpdateAccepted = false;
+
+// This is used to keep track of whether we are showing he special dialog
```
IGassmann: nit: `he` => `the`.

src/main/index.js
```diff
@@ -68,7 +86,26 @@ app.on('activate', () => {
   if (!rendererWindow) rendererWindow = createWindow();
 });
 
-app.on('will-quit', () => {
+app.on('will-quit', (e) => {
```
IGassmann: nit: use a more descriptive name (http://airbnb.io/javascript/#naming--descriptive)

src/main/index.js
```diff
@@ -5,10 +5,28 @@ import SemVer from 'semver';
 import url from 'url';
 import https from 'https';
 import { shell, app, ipcMain, dialog } from 'electron';
+import { autoUpdater } from 'electron-updater';
+import log from 'electron-log';
 import Daemon from './Daemon';
 import Tray from './Tray';
 import createWindow from './createWindow';
 
+// For now, log info messages in production for easier debugging
```
IGassmann: Will `electron-log` be removed before merge?

src/main/index.js
alexliebowitz: Does this apply even to `e`? That has a standard meaning of "event," similar to `x`, `y`, `i`, etc.

If we want to be even more specific than "event," it would have to be `willQuitEvent` or something, which sounds redundant

package.json
alexliebowitz: Just a merge error; will fix.

build/build.sh
alexliebowitz: It would, but I figured we would worry about it later. It's already going to be a little complicated to roll out auto update.

src/main/index.js
IGassmann: I think `event` would fit best here. We don't have multiple events in this callback function so no need to specify precisely the event, and it would also be redundant as you pointed it out. Anyway, this is a nit.

PR #811, opened 2017-12-05
Subscriptions

seanyesmunt:
There are a few notable issues (mostly dealing with the `FeaturedCategory` component), and it could be presented much nicer as subscriptions are being loaded. I think we should try to get this in now, to try and catch any major bugs. I will continue to clean it up and improve the styling over the next couple days and fix any major bugs we find during RC testing.

Notable issues:
~Subscriptions don't load when refreshing the page~
(hacked together a fix, we will need to rework the initial load to wait for the rehydrate action)

Can't subscribe to pages with no content
Can only view the first 10 files per channel on the subscriptions page (this should be easy to fix)

The actual saving of subscriptions seems to work pretty well though


src/renderer/js/redux/reducers/subscriptions.js
```diff
@@ -0,0 +1,67 @@
+// @flow
```
seanyesmunt: I wasn't seeing `Flow` catch state errors when I moved all of the types into `types/subscriptions.js`, but it is working when kept in the reducer file.

Needs more exploring.

liamcardenas: interesting :/

src/renderer/js/page/file/view.jsx
```diff
@@ -62,6 +66,35 @@ class FilePage extends React.PureComponent {
       Object.values(player.mime).indexOf(contentType) !== -1 ||
       mediaType === "audio";
 
+    const channelClaimId =
+      value &&
+      value.publisherSignature &&
+      value.publisherSignature.certificateId;
+
+    const canSubscribe = !!channelName && !!channelClaimId;
```
liamcardenas: when would this happen?

src/renderer/js/redux/actions/subscriptions.js
```diff
@@ -0,0 +1,21 @@
+import * as actions from "constants/action_types";
```
liamcardenas: no flow on this page?

src/renderer/js/page/file/view.jsx
seanyesmunt: If someone posts an anonymous video. It might be fine with just the `!!channel` check. If there is no `channelName` there will be no `channelClaimId` (i think?)

src/renderer/js/redux/actions/subscriptions.js
seanyesmunt: whoops. I added the types, but just forgot to re-add the flow comment

src/renderer/js/page/file/view.jsx
seanyesmunt: @kauffj any problems with only checking for `channelName`?

liamcardenas: FWIW, that is what I would do ^

liamcardenas:
small nitpick, does anyone else agree that this spacing is too much between the Home / Subscriptions menu and Trending?
screen shot 2017-12-08 at 8 48 17 am

seanyesmunt:
@liamcardenas agreed. dropped it down to 24px

src/renderer/js/component/subHeader/view.jsx
```diff
@@ -1,8 +1,9 @@
 import React from "react";
 import Link from "component/link";
+import classnames from "classnames";
 
 const SubHeader = props => {
-  const { subLinks, currentPage, navigate, modifier } = props;
```
seanyesmunt: the `modifier` prop wasn't being used anywhere

src/renderer/js/page/subscriptions/view.jsx
```diff
@@ -0,0 +1,106 @@
...
+  doFetchClaimsByChannel: (string, number) => any,
+  savedSubscriptions: SavedSubscriptions,
+  // TODO build out claim types
+  subscriptions: Array<any>,
+  setHasFetchedSubscriptions: () => void,
+  hasFetchedSubscriptions: boolean
+};
+
+export default class extends React.PureComponent<Props> {
+  // setHasFetchedSubscriptions is a terrible hack
```
kauffj: I'm okay shipping it like this, but this could be an example where the empty vs. undefined is a useful distinction. If the subscription array starts undefined, then we can differentiate between never fetched (undefined) and fetched but empty (empty array).

src/renderer/js/page/subscriptions/view.jsx
```diff
@@ -0,0 +1,106 @@
...
+        doFetchClaimsByChannel(sub.uri, 1);
+      });
+    }
+  }
+
+  render() {
+    const { subscriptions, savedSubscriptions } = this.props;
+
+    let someClaimsNotLoaded;
+    for (var i = 0; i < subscriptions.length; i++) {
```
kauffj: `const someClaimsNotLoaded = Boolean(subscriptions.find(subscription => subscription.claims.length))`

src/renderer/js/page/subscriptions/view.jsx
```diff
@@ -0,0 +1,106 @@
...
+              subscriptions.map(subscription => {
+                if (!subscription.claims.length) {
+                  // will need to update when you can subscribe to empty channels
+                  // for now this prevents issues with FeaturedCategory being rendered
+                  // before the names (claim uris) are populated
+                  return "";
+                }
+
+                // creating uris for each subscription file
+                const names = subscription.claims.slice().map(claim => {
```
kauffj: This could conceivably be moved into the/a selector. I get the impression that most data wrangling should happen inside of Redux rather than React.

src/renderer/js/redux/selectors/navigation.js
```diff
@@ -61,6 +61,12 @@ export const selectHeaderLinks = createSelector(selectCurrentPage, page => {
         settings: __("Settings"),
         help: __("Help"),
       };
+    case "discover":
+    case "subscriptions":
+      return {
+        discover: __("Home"),
```
kauffj: Let's commit to naming this "Discover"

src/renderer/scss/component/_tabs.scss
```diff
@@ -50,8 +50,13 @@ nav.sub-header
...
+
+  &.sub-header--full-width {
+    max-width: 100%;
+  }
 }
 
+
 @keyframes activeTab {
     from {width: 0;}
     to {width: 100%;}
```
kauffj: There appear to be two binary files below that shouldn't have been committed.

src/renderer/scss/component/_tabs.scss
```diff
@@ -50,8 +50,13 @@ nav.sub-header
       color: var(--tab-active-color);
     }
   }
+
+  &.sub-header--full-width {
```
kauffj: I have some non-blocking concerns about this, but will wait to discuss them on a meeting.

src/renderer/js/page/subscriptions/view.jsx
seanyesmunt: 👍 always forget about Array.find

seanyesmunt:
@kauffj We can probably move the subscribe button inside `FileActions`, I think we should keep it as is for now just to get it in now

PR #815, opened 2017-12-05
Channel names in pages appear highlighted

hackrush01:
Changed overflow to `inline-flex` as `inline-block` was not aligning the channel name properly in the Tipping page.

### DIFF
--- a/src/renderer/js/component/uriIndicator/view.jsx
+++ b/src/renderer/js/component/uriIndicator/view.jsx
@@ -1,7 +1,8 @@
 import React from "react";
 import { Icon } from "component/common";
 import Link from "component/link";
-import lbryuri from "lbryuri.js";
+import lbryuri from "lbryuri";
+import classnames from "classnames";
 
 class UriIndicator extends React.PureComponent {
   componentWillMount() {
@@ -60,7 +61,12 @@ class UriIndicator extends React.PureComponent {
 
     const inner = (
       <span>
-        <span className={`channel-name ${smallCard && "channel-name--small"}`}>
+        <span
+          className={classnames("channel-name", {
+            "channel-name--small": smallCard,
+            "button-text no-underline": link,
+          })}
+        >
           {channelName}
         </span>{" "}
         {!signatureIsValid ? (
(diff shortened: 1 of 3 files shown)


hackrush01:
Also, for line https://github.com/lbryio/lbry-app/pull/815/files#diff-e47e29116e67be706ea589295bc76bdcR76
No css styles exist for the different values of `modifier`. I was not sure what to do with...

src/renderer/js/component/uriIndicator/view.jsx
```diff
@@ -60,7 +60,13 @@ class UriIndicator extends React.PureComponent {
 
     const inner = (
       <span>
-        <span className={`channel-name ${smallCard && "channel-name--small"}`}>
+        <span
+          className={
+            "channel-name" +
```
seanyesmunt: can you use `classnames` for this? https://github.com/lbryio/lbry-app/blob/610814160572a51840366eff7e604e418c58b353/src/renderer/js/component/address/view.jsx#L24

src/renderer/scss/component/_channel-indicator.scss
```diff
@@ -1,9 +1,13 @@
 
 .channel-name {
-  display: inline-block;
+  display: inline-flex;
```
seanyesmunt: I don't think we need this, unless you added it for a specific reason

seanyesmunt: Oh just saw your comment

seanyesmunt: 👍

seanyesmunt:
We can keep the `modifier` css attribute for now. I think I will be using it in my subscriptions pull request

src/renderer/js/component/uriIndicator/view.jsx
hackrush01: Yes, this seems awesome.

seanyesmunt:
Thinking about this some more I think we should just use the generic `<Link />` styling instead of introducing more style rules.

PR #819, opened 2017-12-06
Amplitude

liamcardenas:


src/renderer/js/main.js
```diff
@@ -66,18 +89,29 @@ var init = function() {
   app.store.dispatch(doDownloadLanguages());
 
   function onDaemonReady() {
-    window.sessionStorage.setItem("loaded", "y"); //once we've made it here once per session, we don't need to show splash again
-    app.store.dispatch(doDaemonReady());
+    lbry.status().then(info => {
+      amplitude
+        .getInstance()
+        .init(
+          "0b130efdcbdbf86ec2f7f9eff354033e",
```
seanyesmunt: what is this value?

src/renderer/js/main.js
```diff
@@ -48,6 +50,27 @@ ipcRenderer.on("window-is-focused", (event, data) => {
...
+          .getInstance()
+          .logEvent("CLICK", {
+            target: element,
+            location:
+              hrefParts.length > 1 ? hrefParts[hrefParts.length - 1] : "/",
+          });
+      } else {
+        amplitude
+          .getInstance()
+          .logEvent("UNMARKED_CLICK", {
```
seanyesmunt: Do you have an example of when this would fire? Just curious

liamcardenas: For instance, if there is an icon with no title attribute. I put this in there so, if we see it, we can go and add a title.

src/renderer/js/main.js
liamcardenas: this is our (public) api key

src/renderer/js/main.js
seanyesmunt: 👍

PR #822, opened 2017-12-06
Added more understandable error message

liamcardenas:
The old error message was very confusing to users. Users don't need to know about sd hashes and blobs, they just want to understand what the problem pertained to -- and what steps they can take to fix it.

There is no reason to show daemon errors to the user in this instance.

Before
screen shot 2017-12-06 at 8 59 45 am

After
screen shot 2017-12-06 at 12 36 23 pm

### DIFF
--- a/src/renderer/js/redux/actions/content.js
+++ b/src/renderer/js/redux/actions/content.js
@@ -280,13 +280,17 @@ export function doLoadVideo(uri) {
           dispatch(doDownloadFile(uri, streamInfo));
         }
       })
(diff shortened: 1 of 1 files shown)


MSFTserver:
maybe add reference to check log file for full error report?

tzarebczan:
@MSFTserver We want to make a quick change here and keep it as simple as possible (regular users don't need to go digging through logs). Our plan is to create a better timeout specific modal in the future.

PR #825, opened 2017-12-07
Video overlay

hackrush01:


src/renderer/js/component/common.js
```diff
@@ -194,3 +194,22 @@ export class Thumbnail extends React.PureComponent {
...
+  static defaultProps = {
+    spinner: true,
+  };
+
+  render() {
+    const { status, spinner } = this.props;
+    return (
+      <div className="video__loading-screen">
+        <div>
+          {spinner && <div className="video__loading-spinner" />}
```
seanyesmunt: This should use the `<Spinner />` component

hackrush01: I noticed one thing. We already have a `common.js` file [here(file 1)](https://github.com/lbryio/lbry-app/blob/master/src/renderer/js/component/common.js), but the `<Spinner />` component comes from [here(folder 1)](https://github.com/lbryio/lbry-app/blob/master/src/renderer/js/component/common), a new common folder with just that component.
Are we breaking down file 1 to individual components and putting them in folder 1? Or should I include `<Spinner />` component in file 1. The latter option is better IMO as then we don't have to include different files for each "common" component, unless former is better for some use cases.

kauffj: @hackrush01 @seanyesmunt @liamcardenas I think `common.js` should die and these should all just be individual components.

liamcardenas: I agree

lyoshenka:
This is on hold. When we're ready  to work on this, we should re-evaluate and figure out the right approach.

PR #826, opened 2017-12-07
First run email

kauffj:
- New first run allows claiming credit for verifying email
- "Send" and "Receive" combined into one page
- "Get Credits" re-oriented towards users adding credits


src/renderer/js/modal/modalCreditIntro/view.jsx
```diff
@@ -4,55 +4,42 @@ import { CreditAmount, CurrencySymbol } from "component/common";
...
   return (
     <Modal type="custom" isOpen={true} contentLabel="Welcome to LBRY">
       <section>
-        <h3 className="modal__header">{__("Blockchain 101")}</h3>
+        <h3 className="modal__header">
+          {__("Computer Wizard Needs Tokens Badly")}
+        </h3>
         <p>
-          LBRY is controlled and powered by a blockchain asset called{" "}
+          Some actions require{" "}
```
liamcardenas: as far as the i18n goes, why is some text not inside __(...)?

kauffj: Because our i18n system doesn't allow me to embed other elements. It's debatable whether a translation fragment is better or worse than not translating at all.

liamcardenas: Ah makes sense, yeah I think translation fragments aren't great.

src/renderer/js/component/userEmailNew/view.jsx
```diff
@@ -23,35 +23,41 @@ class UserEmailNew extends React.PureComponent {
...
-          onChange={event => {
-            this.handleEmailChanged(event);
-          }}
-        />
-        <div className="form-row-submit">
-          <Submit label="Next" disabled={isPending} />
-        </div>
-      </Form>
+        <p>
+          {__("We'll never sell your email, and you unsubscribe at any time.")}
```
liamcardenas: you *can* unsubscribe at any time

src/renderer/js/modal/modalCreditIntro/view.jsx
liamcardenas: this is slightly confusing because it first says: "Some actions require LBC, the blockchain token that powers the LBRY network."

and then it refers to them as credits in the rest of the modal:

"You currently have 0 credits, so the actions you can take are limited.

There are a variety of ways to get credits, including more than 20 credits in free rewards for participating in the LBRY beta."

It makes sense for us because we understand it, but I don't think people who are not familiar with crypto will catch on. I would suggest at least rewording it as follows:

""Some actions require LBRY credits (LBC), the blockchain token that powers the LBRY network."

tzarebczan:
Yea just tried with existing email - it just pops back up the email request screen again. It does send a verification email though.

kauffj:
@liamcardenas @tzarebczan fixed and merged

PR #827, opened 2017-12-07
Set app-wide user agent

liamcardenas:
removing the user-agent from other request caused errors, not a priority to figure out at this point in my opinion


kauffj:
@liamcardenas I merged this but neglected to update CHANGELOG - can you?

PR #837, opened 2017-12-10
Another fix for abandoning claims

hackrush01:


src/renderer/redux/actions/app.js
```diff
@@ -222,6 +222,7 @@ export function doDaemonReady() {
     dispatch(doBalanceSubscribe());
     dispatch(doFetchFileInfosAndPublishedClaims());
     dispatch(doFetchRewardedContent());
+    dispatch(doFetchTransactions(false));
```
kauffj: Why this change?

hackrush01: Without fetch tip info the command is and offline command and it populates the field `claim_id` and `claim_name` field required for abandon's reducers. I put it there to have a less costly API call.
(Is this an example of premature optimization{and thereby bad coding practice}?).

PR #847, opened 2017-12-11
show empty rewards message on overview page

seanyesmunt:


seanyesmunt:
whoops. took it out when i was testing. all good now

PR #865, opened 2017-12-12
Fix error on settings page on development

IGassmann:
Fix settings page's error and apply some minor changes:
- Enforce engine versions
- Update path for prettier
- Rename electron processes' entry points to avoid confusion with the main process.
- Change node version to 8.x for supporting Ubuntu 17.10


seanyesmunt:
Why the changes to prettier? I updated the path last night. We should be using it in the /main process too. I also took it out of the "scripts" block because it was causing issues on `precommit`. 

I think you accidentally moved the `precommit: lint-staged` script under `bugs`.

IGassmann:
@seanyesmunt I messed up that on the merge conflict. I'm rolling back the merge commit and doing it again.

PR #867, opened 2017-12-13
Reworked abandon to not be depenedent on txn list

hackrush01:
@kauffj, so this PR removes the dependency of abandon_claim on transaction list. Now it fetches the data from myClaims instead. @tzarebczan and I have given this a whirl and so far the edge cases work too.

### DIFF
--- a/src/renderer/redux/actions/content.js
+++ b/src/renderer/redux/actions/content.js
@@ -4,7 +4,8 @@ import lbry from "lbry";
 import lbryio from "lbryio";
 import lbryuri from "lbryuri";
 import { makeSelectClientSetting } from "redux/selectors/settings";
-import { selectBalance, selectTransactionItems } from "redux/selectors/wallet";
+import { selectMyClaimsRaw } from "redux/selectors/claims";
+import { selectBalance } from "redux/selectors/wallet";
 import {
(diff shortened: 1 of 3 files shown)


src/renderer/redux/actions/content.js
```diff
@@ -503,8 +506,8 @@ export function doPublish(params) {
 export function doAbandonClaim(txid, nout) {
   return function(dispatch, getState) {
     const state = getState();
-    const transactionItems = selectTransactionItems(state);
-    const { claim_id: claimId, claim_name: name } = transactionItems.find(
+    const myClaims = selectMyClaimsRaw(state);
```
kauffj: And we know this is always populated because `doFetchFileInfosAndPublishedClaims()` is called in `onDaemonReady()`?

Assuming yes, LGTM and can be merged.

liamcardenas:
Code looks good, I have one question, I sent you a message on slack @hackrush01

src/renderer/redux/actions/content.js
hackrush01: Yes, also `doFetchClaimListMine()` is also called at multiple important places such as after successful publish, visiting Published/Downloaded page etc. So it seems to keep updated information about claims.

PR #869, opened 2017-12-13
make channel names on discover page links

daovist:


daovist:
lbry_pr_869

liamcardenas:
Thanks @daovist! Sean can you review this to make sure it is up to your UI standards? also, the `@stealthishow` is now green because it is a link, do you think we should keep it that way or change it back to gray?

seanyesmunt:
Thanks for the contribution @daovist!

@liamcardenas I think the green color is good, it looks the same as on the file page.

src/renderer/component/fileCard/view.jsx
```diff
@@ -89,25 +89,37 @@ class FileCard extends React.PureComponent {
             className="card__link"
           >
             <CardMedia title={title} thumbnail={thumbnail} />
-            <div className="card__title-identity">
+          </Link>
```
seanyesmunt: Instead of closing the link here, does it work if you keep it at the bottom and encapsulate the new link? You should still be able to click on the file if your mouse is below the channel name.

Hope that makes sense

daovist: That was my first thought too. In that case it takes you to the file page and not the channel page. Though I think it actually takes you to the channel page and then the file page (calling the navigate function twice?) because when I click the channel link and then click Back it does take me to the channel page.

daovist: I ended up wrapping links around the image, title, and price to avoid this.

tzarebczan:
@daovist - also be sure to check out https://lbry.io/faq/tips  

Thanks again for your contribution!

src/renderer/component/fileCard/view.jsx
seanyesmunt: You should be able to use `e.stopPropagation` to fix that. We still want to keep just one link for the whole card. You could update the `<Link />` component to take a prop `noPropagation` which would call `e.stopPropagation()` before doing the navigate.

I think that would work.

seanyesmunt:
Actually I am seeing that nested links are illegal in HTML5. I just tested and my idea works, but I see an ugly error.
```
<a> cannot appear as a descendant of <a>
```

@kauffj thoughts?

kauffj:
@seanyesmunt Not sure of another solution to the nested anchor issue other than just attaching an `onClick` to a non-anchor element like a `<div>`. I believe we ran into this on `<FileTile>` and that's what it does.

seanyesmunt:
I think that would be fine for now. @daovist could you try replacing the wrapper `<Link />` to a div and moving the `onClick` to that? Then you would be able to add the link to the channel no problem. (you would still need to add the `e.stopPropagation`

daovist:
I'm trying. The URI it gives me is for the video and I haven't figured out how to get the channel link the way the UriIndicator component does. I'm messing around with lbryuri.js but not having any luck so far.

seanyesmunt:
You should still use the `UriIndicator` component. Just change the highest level wrapper `<Link` to a `div`.

If you still the the `navigate` action being fired twice you can change the `<Link` component to use `event.stopPropagation()`

Maybe something like this
```
// in UriIndicator/view.jsx

<Link noPropagation />

// inside link/view.jsx
const onClick = (e) => {
  e.stopPropagation();
  props.onClick();
}
```

seanyesmunt:
You can message me on discord if that doesn't work, a little easier to chat there.

daovist:
Nested <a>nchors work but are not legit HTML. When I replace <a> with a <span> or <div> it messes up a lot of things throughout LBRY, such as FAQ and TOS links, not to mention navigation. So I added a span value to props which if true returns a <span> instead of an <a> which I am using to link to the channel pages of videos on the Discover page.

src/renderer/component/link/view.jsx
```diff
@@ -44,6 +46,20 @@ const Link = props => {
     );
   }
 
+  if (span === true) {
```
liamcardenas: this can just be `if(span) {`

src/renderer/component/link/view.jsx
```diff
@@ -44,6 +46,20 @@ const Link = props => {
...
+        onClick={onClick}
+        {...("style" in props ? { style: style } : {})}
+      >
+        {content}
+      </span>
+    );
+  }
+
   return (
     <a
```
liamcardenas: this whole thing should probably be structured like this

`let props = { className: combinedClassName, href: href || "javascript:;" etc}`

and then the if statement shold be something like
`span ? <span {...props}> {content} </span> : <a {...props}> {content} </a>`

as per https://zhenyong.github.io/react/docs/jsx-spread.html

let me know if you have any questions!

src/renderer/component/uriIndicator/view.jsx
```diff
@@ -22,7 +22,9 @@ class UriIndicator extends React.PureComponent {
   }
 
   render() {
-    const { claim, link, uri, isResolvingUri, smallCard } = this.props;
+    const { claim, link, uri, isResolvingUri, smallCard, span } = this.props;
+
+    console.log("URIINDICATOR SPAN:", span);
```
liamcardenas: whenever you are ready to merge, be sure to remove this ;)

src/renderer/component/uriIndicator/view.jsx
```diff
@@ -91,6 +91,7 @@ class UriIndicator extends React.PureComponent {
         navigate="/show"
         navigateParams={{ uri: channelLink }}
         className="no-underline"
+        span={span ? span : {}}
```
seanyesmunt: You can just pass this in as is
`span={span}`

If it is `undefined` that is ok

seanyesmunt:
Could you also squash these commits into one commit?

PR #880, opened 2017-12-18
Pushing for help with Redux selectors

daovist:
I watched a Redux video and read some of the docs. I've created my action (please ignore the do and confirm actions, I will delete those, sorry) and reducer and they seem to be working but my selector is not. I could use help on this. I think once I manage to get the videoPause value from the store into the VideoPlayer component I will be able to wrap this up.


liamcardenas:
I'm looking into this

src/renderer/component/video/internal/player.jsx
```diff
@@ -121,6 +124,11 @@ class VideoPlayer extends React.PureComponent {
     }
   }
 
+  pauseVideo() {
```
liamcardenas: what if you change this to `componentWillReceiveProps` instead of `pauseVideo`?

liamcardenas: then of course you would need to filter out the prop changes that don't lead to a pause, but does that make sense?

daovist: I tried that and the method(?) was never called.

I think the issue is that the selector isn't working, or that the reducer isn't actually changing the state. My focus is on line 63 of components/video.view.jsx where it console logs the value of videoPause. It's always false, even after I've clicked open, which should update valuePause to true through the reducer.

kauffj: @daovist try using the console to help debug what unexpected behavior is happening here. The console will show all action dispatches as well as all state changes (expand the console messages that are printed on action dispatches). This should tell you whether it's an issue with actions or the reducer. If both the action is fired and the Redux state looks correct, you can add a console statement to see if the selector is firing when you think it ought to.

src/renderer/component/video/internal/player.jsx
```diff
@@ -157,28 +166,39 @@ class VideoPlayer extends React.PureComponent {
...
+          )}
+          <div ref="media" className="media" />
+        </div>
       </div>
     );
   }
 }
 
-export default VideoPlayer;
+const select = (state, props) => ({
```
liamcardenas: this should go in videos/index.js

src/renderer/component/video/view.jsx
```diff
@@ -56,8 +57,11 @@ class Video extends React.PureComponent {
       changeVolume,
       volume,
       uri,
+      // videoPause,
```
liamcardenas: you want this in here, and then you want to pass it into VideoPlayer

src/renderer/redux/selectors/video.js
```diff
@@ -0,0 +1,9 @@
+import * as settings from "constants/settings";
+import { createSelector } from "reselect";
+
+const _selectState = state => state.videoPause || false;
```
liamcardenas: this should be

`const _selectState = state => state.video || {};`

src/renderer/component/video/internal/player.jsx
```diff
@@ -20,6 +20,13 @@ class VideoPlayer extends React.PureComponent {
     this.togglePlayListener = this.togglePlay.bind(this);
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.videoPause) {
+      this.refs.media.children[0].pause();
+      this.props.setVideoPause(false);
```
liamcardenas: is this (`this.props.setVideoPause(false);`) necessary? can we trigger this, instead, when they push "play" in app?

liamcardenas:
Making a follow up issue to this since this is work.

PR #890, opened 2017-12-21
track and manage video state

daovist:
This feature tracks play and pause actions on a video element. Below is a model of the parts.

blue = state
red = actions
purple = reducers
light blue = selectors
green = component lifecycle hooks
orange = distant component which can alter state
black = not React/Redux

lbry_video_state_model


daovist:
Are we using the time for something other than starting back where the user left off?

liamcardenas:
@daovist yes, exactly. There are going to be several features that depend on this

src/renderer/redux/selectors/media.js
```diff
@@ -0,0 +1,15 @@
+import * as settings from "constants/settings";
```
liamcardenas: I noticed that `/selectors/video.js` has not been deleted. Are we still using it?

src/renderer/store.js
```diff
@@ -71,6 +72,7 @@ const reducers = combineReducers({
   shapeShift: shapeShiftReducer,
   subscriptions: subscriptionsReducer,
   video: videoReducer,
```
liamcardenas: do we still need `video: videoReducer,`?

src/renderer/redux/selectors/media.js
```diff
@@ -0,0 +1,15 @@
...
+const _selectState = state => state.media || {};
+
+export const selectMediaPaused = createSelector(
+  _selectState,
+  state => state.paused
+);
+
+export const makeSelectMediaPositionForUri = uri =>
+  createSelector(_selectState, state => {
+    const id = uri.split("#")[1];
```
liamcardenas: will this work when the uri does not contain a `#` such as in the case of `lbry://one`?

src/renderer/redux/reducers/media.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+import * as actions from "constants/action_types";
+import { handleActions } from "util/redux-utils";
+
+export type MediaState = {
+  paused: Boolean,
+  positions: Object,
```
liamcardenas: can you define a type more specifically than `Object`? Maybe `{ [string]: number }`?

https://flow.org/en/docs/types/objects/#toc-objects-as-maps

src/renderer/component/video/view.jsx
```diff
@@ -93,6 +96,9 @@ class Video extends React.PureComponent {
     }
     const poster = metadata.thumbnail;
 
+    const mediaId = uri.split("#")[1];
```
liamcardenas: as I said in a comment below, does this work for `lbry://one`?

src/renderer/component/video/view.jsx
```diff
@@ -93,6 +96,9 @@ class Video extends React.PureComponent {
...
+    console.log("mediaId:", mediaId);
```
liamcardenas: please remove when ready to merge :)

src/renderer/component/video/index.js
```diff
@@ -29,14 +34,19 @@ const select = (state, props) => ({
...
+  mediaPaused: selectMediaPaused(state),
+  mediaPosition: makeSelectMediaPositionForUri(props.uri)(state),
 });
 
 const perform = dispatch => ({
   play: uri => dispatch(doPlayUri(uri)),
   cancelPlay: () => dispatch(doSetPlayingUri(null)),
   changeVolume: volume => dispatch(doChangeVolume(volume)),
-  setVideoPause: val => dispatch(setVideoPause(val)),
+  // setVideoPause: val => dispatch(setVideoPause(val)),
```
liamcardenas: please remove this comment when ready to merge

src/renderer/component/video/index.js
```diff
@@ -15,7 +16,11 @@ import {
 } from "redux/selectors/file_info";
 import { makeSelectCostInfoForUri } from "redux/selectors/cost_info";
 import { selectShowNsfw } from "redux/selectors/settings";
-import { selectVideoPause } from "redux/selectors/video";
+// import { selectVideoPause } from "redux/selectors/video";
```
liamcardenas: please remove this comment when ready to merge

src/renderer/component/video/index.js
```diff
@@ -3,7 +3,8 @@ import { connect } from "react-redux";
 import { doChangeVolume } from "redux/actions/app";
 import { selectVolume } from "redux/selectors/app";
 import { doPlayUri, doSetPlayingUri } from "redux/actions/content";
-import { setVideoPause } from "redux/actions/video";
+import { doPlay, doPause, savePosition } from "redux/actions/media";
+// import { setVideoPause } from "redux/actions/video";
```
liamcardenas: please remove this comment when ready to merge :)

kauffj:
@daovist outpoint not claimid a8988d9

src/renderer/component/video/index.js
```diff
@@ -29,14 +34,17 @@ const select = (state, props) => ({
...
 });
 
 const perform = dispatch => ({
   play: uri => dispatch(doPlayUri(uri)),
   cancelPlay: () => dispatch(doSetPlayingUri(null)),
   changeVolume: volume => dispatch(doChangeVolume(volume)),
-  setVideoPause: val => dispatch(setVideoPause(val)),
+  doPlay: () => dispatch(doPlay()),
+  doPause: () => dispatch(doPause()),
+  savePosition: (id, position) => dispatch(savePosition(id, position)),
```
kauffj: what kind of id? I assume this is a `claimId`, let's call it that

src/renderer/component/video/view.jsx
```diff
@@ -55,9 +55,13 @@ class Video extends React.PureComponent {
...
       volume,
+      claim,
       uri,
-      videoPause,
-      setVideoPause,
+      doPlay,
+      doPause,
+      savePosition,
+      mediaPaused,
+      mediaPosition,
```
kauffj: The fact that `Video` is receiving so many properties simply to pass them on to `VideoPlayer` makes me think `VideoPlayer` should just be a first-class component. Although perhaps the names should be clarified if this is the case.

daovist: I was also thinking it should be its own component. Maybe as a separate issue along with displaying a paused video instead of the image wrapper when position has a value?

src/renderer/component/video/internal/player.jsx
```diff
@@ -20,11 +20,10 @@ class VideoPlayer extends React.PureComponent {
...
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.videoPause) {
-      this.refs.media.children[0].pause();
-      this.props.setVideoPause(false);
-    }
+  componentWillReceiveProps(next) {
+    const el = this.refs.media.children[0];
+    if (!this.props.paused && next.paused && !el.paused) el.pause();
+    // if (this.props.paused && !next.paused && el.paused) el.play();
```
liamcardenas: remove comment please

src/renderer/component/video/view.jsx
hackrush01: I agree. #825 can work as a reference as the media component has been separated there. Also I think that Video and VideoPlayer names are very specific, we use the same component to render all other types of media as well, so IMO the variables and components should be refactored to reflect the same(also in the PR).
Thoughts?

PR #899, opened 2017-12-23
fix snackbar text

tzarebczan:
Snackbar text was blank  - looks like the type was mismatched.


tzarebczan:
I tried http://releases.lbry.io/app/20171227-090900-v0.19.2-3-ge7f35caa/ which should have the fix but it didn't fix the snackbar message. 

I also noticed I wasn't able to claim the email reward without being verified...looking into that next.

PR #921, opened 2018-01-03
update wunderbar text

tzarebczan:
Change "movies" to "videos" (more generic)
See "Manchester first run experiments" doc on Google drive


seanyesmunt:
Nice. I updated the header text with the new search bar. Good idea to get it in now

PR #929, opened 2018-01-08
Fix Electron linting errors

### LINKED ISSUE #397: Fix linting errors
errors

IGassmann:
This pull request fixes #397 and:

- Adds `brew update` to build script for getting latest dependencies versions
- Cleans [`.gitignore`](https://github.com/lbryio/lbry-app/commit/7068ebaf57ed00f316b313b26574f788f513d280#diff-a084b794bc0759e7a6b77810e01874f2) to remove IDE specific ignores and old rules.
- Renames `package-json` name field to be compliant with npm and yarn rules, and adds productName to build
- Replaces dependency `keytar` with `keytar-prebuild` until https://github.com/atom/node-keytar/pull/67 is not merged. This simplifies the build process on Windows. There's no need anymore for Python and Windows building tools.


build/install_deps.sh
```diff
@@ -42,6 +42,7 @@ if $LINUX; then
   $INSTALL build-essential libssl-dev libffi-dev libgmp3-dev python2.7-dev libsecret-1-dev curl
 elif $OSX && ! cmd_exists brew ; then
   /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+  brew update
```
lyoshenka: this only gets run if `brew` command does not exist. so it will only update when it is installed. is that what you intended?

README.md
```diff
@@ -31,12 +31,11 @@ development and testing purposes.
 * [Git](https://git-scm.com/downloads)
 * [Node.js](https://nodejs.org/en/download/)
 * [Yarn](https://yarnpkg.com/en/docs/install)
-* `yarn --add-python-to-path install --global --production windows-build-tools` (Windows only)
 
 ### One-time Setup
 
 1. Clone this repo
-2. `DEPS=true ./build.sh`
+2. `./build.sh`
```
lyoshenka: why did you get rid of DEPS=true?

.gitignore
```diff
@@ -1,30 +1,5 @@
 /node_modules
-/LBRY-darwin-x64
 /dist
-
-/src/main/dist
-/src/main/locales
-/src/main/node_modules
-/src/renderer/dist
-/build/venv
```
lyoshenka: i believe build/venv is still used (by build.sh)

.gitignore
```diff
@@ -1,30 +1,5 @@
...
 /build/daemon.ver
-/lbry-app-venv
-/lbry-app
-/lbry-venv
 /static/daemon/lbrynet*
-/static/locales
-/daemon/build
-/daemon/venv
-/daemon/requirements.txt
-/.idea
```
lyoshenka: this is here because many of us use JetBrains, which creates that dir

.gitignore
```diff
@@ -1,30 +1,5 @@
...
-
-*.pyc
```
lyoshenka: this has to be here while upload_assets.py is still in use. if you want to rewrite that in JS instead, then we can drop python dependency completely, which would be sweet!

.gitignore
IGassmann: I think it's better to keep project related ignore rules into the project `.gitignore`and user-specific rules into the user's global `.gitignore`.

README.md
IGassmann: There's no need anymore for installing dependencies from the build script. The user should install the prerequisites listed in `README.md`. That is Git, Node, and Yarn. No need anymore for Python for developing.

IGassmann:
`build.sh` is currently only necessary for downloading and extracting the daemon. I'm planning to try to get the daemon as a `package.json` dependency. It should be done with `node-pre-gyp` as far as I know.

lyoshenka:
is the app no longer built/released by teamcity? teamcity runs build.sh

README.md
lyoshenka: it would be much better if users ran a script to install dependencies instead of following a README. that way if the dependencies change, no one has to remember to update the README

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "LBRY",
+  "name": "lbry-app",
```
lyoshenka: why did this change? will this affect the .deb package that's built? if the name changes, i think ubuntu will think its a different package...

IGassmann: I changed the name to follow the [norms](https://yarnpkg.com/lang/en/docs/package-json/#toc-name).  I'm not sure how exactly it'll behave in Ubuntu since I didn't test on it, but it should normally use the [`productName` field](https://yarnpkg.com/lang/en/docs/package-json/#toc-name) declared in the `electron-builder.json` file and not the `package.json`'s `name`.

I'll test it to be sure.

IGassmann: Debian does identify the app as a new package and doesn't replace the one with the previous name.

I think we should wait for the AppImage switch, which should come with the auto-update PR, before applying this change and then request Linux users that they uninstall their previous install.

PR #936, opened 2018-01-09
App Navigation and Wallet pages groundwork

seanyesmunt:
A lot of files but most are small JSX/css changes and Flow additions

Focus:
Add side nav/improve header (search)
Use Feather Icons (needs standarization)
Sub-link navigation (subheader reworked)
Home page cleanup
Wallet Pages (Basic stuff. I'll go more in depth later, just getting stuff looking kinda ok)
Added flow to a lot of components (going to add it to more redux stuff in another pr)


src/main/index.js
```diff
@@ -143,6 +143,7 @@ function createWindow() {
     minWidth: 800,
     minHeight: 600,
     autoHideMenuBar: true,
+    titleBarStyle: 'hidden',
```
seanyesmunt: This is preventing a draggable window. Should be a simple fix but definitely getting rid of the title bar. Need to explore how it looks on window/linux though

src/renderer/component/shapeShift/internal/form.jsx
```diff
@@ -49,23 +49,27 @@ export default (props: Props) => {
...
-          }}
-        >
-          {shiftSupportedCoins.map(coin => (
-            <option key={coin} value={coin}>
-              {coin}
-            </option>
-          ))}
-        </select>
-        <span> {__('for LBC')}</span>
+      <FormRow
```
seanyesmunt: I think I'm going to change this to `FormElement` and keep `FormRow` which just allows several `FormElement`s side by side

src/renderer/component/address/view.jsx
```diff
@@ -16,37 +16,35 @@ export default class Address extends React.PureComponent {
...
-          onFocus={() => {
-            this._inputElem.select();
-          }}
-          readOnly="readonly"
-          value={address || ''}
-        />
-        {showCopyButton && (
-          <span className="header__item">
+      <FormField
+        render={() => (
```
seanyesmunt: Using render props makes it really simple add additional items to the form field like this button. Instead of having a huge `FormField` element that can do everything, allowing the user of the component to do whatever they want can be  really nice.

src/renderer/component/common/form.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
```
seanyesmunt: Using `render` and `children` are really similar. Maybe we should just choose one and use it everywhere. The additional feature that `render` gives you is that you can pass state/computed values into the render function and allow consuming components to interact with it in a more customizable way.

src/renderer/component/form/index.js
```diff
@@ -0,0 +1,12 @@
+import { connect } from 'react-redux';
```
seanyesmunt: This will be removed

src/renderer/component/menu.js
```diff
@@ -1,111 +0,0 @@
-import React from 'react';
```
seanyesmunt: This wasn't being used anywhere

src/renderer/constants/icons.js
```diff
@@ -1,5 +1,5 @@
-export const FEATURED = 'rocket';
-export const LOCAL = 'folder';
+export const FEATURED = 'Award';
```
seanyesmunt: Going to add every icon to this file and only change every Icon use to constants

seanyesmunt: * at a later time

npm-debug.log.3540228334
```diff
@@ -0,0 +1,25 @@
+0 info it worked if it ends with ok
```
liamcardenas: should this be git ignored?

src/renderer/component/uriIndicator/view.jsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import Button from 'component/link';
 import { buildURI } from 'lbryURI';
 import classnames from 'classnames';
-import Icon from 'component/common/icon';
+// import Icon from 'component/common/icon';
```
liamcardenas: i assume you will remove the commented code here and elsewhere before we merge

src/renderer/constants/icons.js
liamcardenas: is there a reason why the ones you just added are upper case but the others are lower?

seanyesmunt:
@liamcardenas Not sure why I can't respond to your comment about icons. The old icons used lowercase. This new module `react-feather` uses uppercase. One of the reasons I want to use constants for all of them.

npm-debug.log.3540228334
seanyesmunt: yes

src/renderer/component/uriIndicator/view.jsx
seanyesmunt: I wasn't planning on it. Mostly they are just reminders to come back to something. I'm not sure if I need to keep this, but I will know when I start on the File Page (next)

liamcardenas:
@seanyesmunt makes sense!

src/renderer/component/uriIndicator/view.jsx
liamcardenas: sounds good

npm-debug.log.3540228334
seanyesmunt: Just removed these and added to the .gitignore

seanyesmunt:
Gonna merge this. I will come back to every file touched here at least once more. Please let me know if you see something that looks funky.

Issue #540, opened 2017-09-07
Publishing Logic Cleanup


kauffj:
Publishing could use the following design changes, as inspired by #526:

- Publish form accepts pre-filled state from outside
- Publish page handles the edit logic and passes pre-fill state to the form
- Publish form drops the logic handling loading values. The link on a name clash simply becomes a navigate to the proper URL.
- Navigation logic is claim ID based rather than name based (we want to support multiple publishes to the same name, though we don't atm).
- Update the title displayed in the address bar based on the action being an edit.

kauffj:
Related for the ambitious: #373 #528 #493

btzr-io:
All that could be handle more eassy if we split: `publish / edit / new channel `
https://github.com/lbryio/lbry-app/issues/528#issuecomment-327068051

PR #939, opened 2018-01-09
Thumbnail upload

daovist:
This is coming along but I could use some help.

I have the file upload component calling the confirmation modal upon selecting a file which calls beginUpload which takes the correct data, the file path. I'm looking for the optimal way to upload a file with fetch. I'm seeing a bunch of different options with sync, async, stream, formdata.

The relevant code is in `src/renderer/redux/actions/upload.js`

Note: I am currently uploading to httpbin for development based on the assumption that the fetch mechanics will be the same so I can switch over to spee.ch easily enough later.


src/renderer/component/publishForm/view.jsx
```diff
@@ -442,6 +466,38 @@ class PublishForm extends React.PureComponent {
     this.setState({ id });
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.uploadStatus === "error") {
+      alert("UPLOAD FAILED, PLEASE TRY AGAIN.");
```
liamcardenas: we should display this in a better way (although, makes sense to have that in there for testing/dev purposes)

src/renderer/component/publishForm/view.jsx
```diff
@@ -442,6 +466,38 @@ class PublishForm extends React.PureComponent {
...
+    // }
+
+    // if (nextProps.uploadStatus === "sending") {
+    //   this.setState({
+    //     thumbnailUploadStatus: "sending",
+    //   });
+    // }
+
+    if (nextProps.uploadStatus === "complete") {
+      this.setState({
```
liamcardenas: can you do this without setting state?

src/renderer/component/publishForm/view.jsx
```diff
@@ -526,6 +582,14 @@ class PublishForm extends React.PureComponent {
     });
   }
 
+  upload(path) {
+    this.closeModal();
+    this.setState({
+      thumbnailUploadStatus: "sending",
```
liamcardenas: sending status would be better in redux than local state... but it does make sense that you would have it like this for quickly developing the other parts

src/renderer/redux/actions/upload.js
```diff
@@ -0,0 +1,59 @@
...
+        .catch(err => dispatch({ type: actions.SPEECH_UPLOAD_ERROR }));
+    })
+    .catch(err => dispatch({ type: actions.SPEECH_UPLOAD_ERROR }));
+
+  function makeid() {
+    var text = "";
+    var possible =
+      "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+    for (var i = 0; i < 7; i++)
+      text += possible.charAt(Math.floor(Math.random() * possible.length));
```
liamcardenas: @seanyesmunt maybe we use uuids here?

src/renderer/redux/reducers/upload.js
```diff
@@ -0,0 +1,79 @@
+// @flow
+import * as actions from "constants/action_types";
+import { handleActions } from "util/redux-utils";
+
+export type UploadState = {
+  status: string | null,
```
liamcardenas: you can us the type `?string` instead of `string | null`

src/renderer/component/publishForm/view.jsx
kauffj: `doAlertError` exists as an action, and can be an appropriate way of handling generic errors

kauffj:
@daovist I did not do a deep code dive, but I would much prefer an approach that cleanly separates the upload logic into an individual component. Implementing this directly inside of `PublishForm` adds to the existing technical debt of this component (see #540).

liamcardenas:
@kauffj just to clarify, you mean pulling out the jsx into its own component and NOT that the upload logic should be in the component, correct?

If so, I agree.

kauffj:
@liamcardenas correct, upload logic should presumably be in an action. But the form element itself should be it's own component. Possibly extending `FormField`.

src/renderer/component/speechUpload/view.jsx
```diff
@@ -152,9 +82,7 @@ class SpeechUpload extends React.PureComponent {
         <div
           className="card__content"
           style={
-            this.state.thumbnailUploadStatus === "sending"
-              ? null
-              : { display: "none" }
+            this.props.uploadStatus === "sending" ? null : { display: "none" }
```
kauffj: You should probably just not render this element at all, rather than rendering it and hiding it.

src/renderer/component/speechUpload/view.jsx
```diff
@@ -85,15 +24,12 @@ class SpeechUpload extends React.PureComponent {
         <div
           className="card__content"
           style={
-            this.state.thumbnailUploadStatus !== "manual"
-              ? null
-              : { display: "none" }
+            this.props.uploadStatus !== "manual" ? null : { display: "none" }
```
kauffj: Upload status states should either be locally declared variables (if used only in one file) or exportable constant definitions if used in multiple files.

https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants

src/renderer/component/publishForm/view.jsx
```diff
@@ -651,7 +595,10 @@ class PublishForm extends React.PureComponent {
                   />
                 </div>
 
-                <SpeechUpload />
+                <SpeechUpload
+                  uploadStatus={uploadStatus}
```
kauffj: I think this can be dropped.

src/renderer/component/publishForm/view.jsx
```diff
@@ -442,6 +444,22 @@ class PublishForm extends React.PureComponent {
...
+      this.props.alertError("Upload failed. Please try again.");
+      this.setState({
+        thumbnailUploadStatus: "upload",
+      });
+    }
+
+    if (nextProps.uploadStatus === "complete") {
+      this.setState({
+        thumbnailUploadStatus: "complete",
+        meta_thumbnail: nextProps.uploadUrl,
```
liamcardenas: where do you use thumbnailUploadStatus and meta_thumbnail? I don't see why this is necessary

liamcardenas:
Please fix these 
flowerrorscreenshot

daovist:
My recent push fixes the Magic Number issue and removes the local state in publishForm. I kept meta_thumbnail because that's where the thumbnail url has been stored.

src/renderer/redux/actions/upload.js
seanyesmunt: Guess I never saw this. Yeah that would be preferrable. There is some chance this could produce two of the same ids

kauffj:
@daovist @seanyesmunt any other potential solutions than using a random ID for the claim name? Not in love with this as a solution.

What about using a sanitized file name or delaying upload until they enter a claim name (then you could do `claimname-thumb` or similar).

Also, regarding spee.ch rejecting duplicates, you could append a number when a name is rejected. Alternatively, lbryio/spee.ch could implement an API endpoint that returns the name available closest to a desired one.

seanyesmunt:
IMO users shouldn't have to worry about naming their thumbnail. Going off your idea, maybe spee.ch could offer a "quick publish" feature. Where it tries the claim name you give it (sanitized file name) and if that isn't available it chooses a close match.

Then the file is somewhat accessible via search, and the user doesn't have to worry about an additional step in the publish process.

kauffj:
@seanyesmunt 100% agreed, do not ask the user. Since this only happens during publish flow, I was suggesting using the name that they are already selecting for the content that they are publishing.

seanyesmunt:
Gotcha. I think it could be confusing for a publish to fail just because the thumbnail upload failed. Or were you thinking start the thumbnail upload in the background?

I think I would rather have some indicator like "uploading..." appear as soon as you choose the file. We populate the `name` field from the content file, so you could just disable the thumbnail input until there is a file chosen to publish.

kauffj:
@seanyesmunt @daovist fuck, I think I did not read this code carefully enough at first and misunderstood it. Apologies for the noise.

daovist:
My master got too far behind so here is the new PR:

`[https://github.com/lbryio/lbry-app/pull/1047](https://github.com/lbryio/lbry-app/pull/1047)`

kauffj:
@daovist this is an okay solution, but if you want to learn "proper" practices, it'd be to `git rebase`.

PR #945, opened 2018-01-13
Fix CONTRIBUTING.md level issues links

Arrowana:
I wanted to get started with contributing, however a column character is missing on the level links resulting on the filter matching no issue.


kauffj:
Ty @Arrowana! Please see https://lbry.io/faq/tips if you'd like some LBC for this.

PR #946, opened 2018-01-15
Phone verification

liamcardenas:
not quite ready for review yet, still ironing out some things


src/renderer/component/userFieldNew/view.jsx
```diff
@@ -0,0 +1,129 @@
...
+      this.props.addUserPhone(phone.replace(/\D/g, ''), country_code.replace(/\D/g, ''));
+    } else {
+      this.props.addUserEmail(email);
+    }
+  }
+
+  render() {
+    const { cancelButton, emailErrorMessage, phoneErrorMessage, isPending, fieldType } = this.props;
+
+    return fieldType === 'phone' ? (
```
kauffj: Is there a particular reason we use a single component to do two (seemingly) distinct duties, as opposed to `UserEmailNew` and `UserPhoneNew` or equivalent?

src/renderer/component/userFieldNew/view.jsx
```diff
@@ -0,0 +1,129 @@
...
+      <div>
+        <p>
+          {__(
+            'Enter your phone number and we will send you a verification code. We will not share your phone number with third parties.'
+          )}
+        </p>
+        <Form onSubmit={this.handleSubmit.bind(this)}>
+          <FormRow
```
kauffj: Does this have to be collected separately as opposed to letting them type a full number? If it is separate field, could it at least be an inline select next to the full number entry?

src/renderer/modal/modalPhoneCollection/index.js
```diff
@@ -3,19 +3,20 @@ import * as settings from 'constants/settings';
...
-  email: selectEmailToVerify(state),
+  phone: selectPhoneToVerify(state),
   user: selectUser(state),
 });
 
 const perform = dispatch => () => ({
   closeModal: () => {
-    dispatch(doSetClientSetting(settings.EMAIL_COLLECTION_ACKNOWLEDGED, true));
     dispatch(doCloseModal());
+    dispatch(doNavigate('/rewards'));
```
kauffj: Is it correct to always bump the user to rewards from here?

kauffj:
@liamcardenas Also, I seem to no longer be prompted for my email immediately after the welcome popup.

src/renderer/component/userFieldNew/view.jsx
liamcardenas: this was in the spirit of DRY a la our coding maxims.

There is less of a case for userFieldNew to be used for two field types, but userFieldVerify is basically the same for both types. I made userFieldNew to be consistent with that.

src/renderer/component/userFieldNew/view.jsx
liamcardenas: I suppose it could be a select. Twilio wants a separate country code. I tried to make this as fool proof and intuitive as possible. I think the way it is makes a lot of sense (it was based on a discussion in tech-app), but I am completely open to changing it.

src/renderer/modal/modalPhoneCollection/index.js
liamcardenas: I believe so, since the only place you can access this is from the verify identity page.

liamcardenas:
@kauffj oh, for some reason, I thought we were taking that out. Should I add it back in?

src/renderer/modal/modalPhoneCollection/index.js
kauffj: :+1:

src/renderer/component/userFieldNew/view.jsx
kauffj: I don't see very much savings here aside from maybe some boilerplate in index.js, and it adds complexity to declaration and handling changes. I'd split these.

src/renderer/component/userFieldNew/view.jsx
kauffj: @liamcardenas yes, please make the country code a select displayed before the phone number entry.

Issue #920, opened 2018-01-02
LBRY app quits with daemon already running error 


tzarebczan:
## The Issue
This happens mostly between upgrades when the LBRY app is updated but the lbrynet-daemon process continues to run/get stuck in the background. When attempting to run LBRY, it looks like it's starting and then will close out with an "daemon unexpectly quit" error. In the logs, you see:

`2017-12-30 17:47:39,213 INFO     lbrynet.daemon.DaemonControl:65: lbrynet-daemon is already running`

This usually requires the daemon to be killed manually and LBRY restarted. Sometimes you can restart just the app and it picks it up the 2nd time. 

Sort of related to https://github.com/lbryio/lbry-app/issues/298 as that should shut down any running daemons. 

### Steps to reproduce
1. Get daemon stuck in background when closing LBRY (not easily reproducible)
2. Start LBRY
3. Notice error in log / LBRY app shuts down

## System Configuration

 

- LBRY Daemon version: 0.18.0
- LBRY App version: 0.19.3
- LBRY Installation ID:
- Operating system:

tzarebczan:
@IGassmann is this why you ended up killing the daemon? I should have been more specific in my original issue, but if you ran the daemon as a separate process before, the app would pick it up. This would work a majority of the time.  I think it only ran into problems between app upgrades - but if the daemon version is the same between different app versions, it should still use the running daemon (could the app check the version before killing it?)

george-mcintyre:
I get this when installing LBRY for the first time.  I am a developer and I suspect there is some interaction with some of the SDKs I develop with.  

Error message "Daemon has Exited" "The daemon may have encountered an unexpected error, or another daemon instance is already running.  For more information please visit: 
https://lbry.com/faq/startup-troubleshooting"

- LBRY Version: 0.33.2
- OS: MacOS Mojave: 10.14.5
- Console output:
`Current system appearance, (HLTB: 2), (SLS: 1 
Post-registration system appearance: (HLTB: 2) 
NSApp cache appearance:
-NSRequiresAquaSystemAppearance: 1
-appearance: (null)
-effectiveAppearance: <NSCompositeAppearance: 0x600003feb840
 (
    "<NSAquaAppearance: 0x6000009f9fc0>",
    "<NSSystemAppearance: 0x6000009fa1c0>"
)>
nw_path_close_fd Failed to close guarded necp fd 37 [9: Bad file descriptor]
GVA info: preferred scaler idx 1
[...]

PR #951, opened 2018-01-18
Refactor Electron's main process

### LINKED ISSUE #298: LBRY installer must detect running daemon
## The Issue
Under certain conditions (have not reproduced successfully), the daemon is not updated on an existing Windows installation.  I was on 0.12.2RC2 with 0.13RC1 daemon.  I then ran the RC7 exe file. After running LBRY, the daemon stayed at 0.13RC1 - I confirmed that the daemon wasn't updated in the Program Files directory.  In this particular instance, 

Do we check if the daemon is running during install, or only lbry.exe? 

### Steps to reproduce
1. Install 0.12.2RC2, run lbry
2. Download 0.12.2RC7, install update. Lbry may or may not have been closed at this point. 
3. Lbry updates, but daemon still at old version. Not sure if daemon was running during update. 

### Expected behaviour
Daemon should be updated

### Actual behaviour
Daemon exe file was left untouched

## System Configuration

 

- LBRY Daemon version: 0.13.1RC1
- LBRY App version: 0.12.2RC2/0.12.2.RC7
- LBRY Installation ID:
- Operating system: Windows 7

### LINKED ISSUE #938: Refactor Electron's main process

IGassmann:
This closes #938, closes #298 and fixes #298.


kauffj:
@alexliebowitz I think a decent portion of this code was originally yours, it'd be good to get your eyes on this changeset.

liamcardenas:
I agree, everything looks good here to me but it would be gret to get Alex's take. I am ready to merge if Alex gives the thumbs up

IGassmann:
This also closes issue #920 and #298.

src/main/index.js
```diff
@@ -549,8 +166,33 @@ ipcMain.on('set-auth-token', (event, token) => {
...
+
+// Force single instance application
+const isSecondInstance = app.makeSingleInstance(argv => {
+  // Protocol handler for win32
+  // argv: An array of the second instance’s (command line / deep linked) arguments
+
+  let URI;
+  if (process.platform === 'win32' && String(argv[1]).startsWith('lbry')) {
+    // Keep only command line / deep linked arguments
+    URI = argv[1].replace(/\/$/, '').replace('/#', '#');
```
alexliebowitz: Hmm, now we're doing this same substitution in two places instead of in one function. That feels brittle (what if we find another problem with Windows URI handling and don't realize it's there in two places), and also

We might want to move this back into its own function, or at least add another comment here. Just a thought.

src/main/createWindow.js
```diff
@@ -0,0 +1,100 @@
...
+
+  window.webContents.on('did-finish-load', () => {
+    window.webContents.send('open-uri-requested', deepLinkingURI, true);
+    window.webContents.session.setUserAgent(`LBRY/${app.getVersion()}`);
+    if (process.env.NODE_ENV === 'development') {
+      window.webContents.openDevTools();
+    }
+  });
+
+  window.webContents.on('crashed', () => {
```
alexliebowitz: It doesn't seem like this event is actually sent on the renderer side. Did you mean to build that, or is it something we're going to add later?

IGassmann: It's necessary to handle it: https://electronjs.org/docs/api/web-contents#event-crashed

src/main/Tray.js
```diff
@@ -0,0 +1,63 @@
+import { app, Menu, Tray as ElectronTray } from 'electron';
+import path from 'path';
+import createWindow from './createWindow';
+
+export default class Tray {
```
alexliebowitz: Does this need to be a whole object? `create()` is only called once and it doesn't need any state, so maybe just a `createTray` function that takes an `updateAttachedWindow` callback?

IGassmann: I went for a exported `createTray()` function initially. However, since it also needs to manage the state of the window, I finally went for using a class.

PR #963, opened 2018-01-26
[redesign] File page

seanyesmunt:
What I did:
Basic styling/layout for file page components

What I need to come back to:
The video player doesn't scale properly. I spent some time on it and was having issues, just going to leave it for now and come back later
Send tip modal (style)
Download progress/open/remove/report buttons inside of the file viewer

For the buttons inside the file viewer I didn't spend much time on them. I think more discussion is needed on if we actually want them there, and if so what the exact behavior should be.


src/renderer/component/common/spinner.jsx
```diff
@@ -1,14 +1,14 @@
+// @flow
```
seanyesmunt: Changed the spinner to not actually be a spinner, but might as well keep the name as is

src/renderer/component/video/internal/play-button.jsx
```diff
@@ -1,47 +1,42 @@
+// @flow
 import React from 'react';
-import Link from 'component/link';
+import Button from 'component/link';
 
-class VideoPlayButton extends React.PureComponent {
-  componentDidMount() {
```
seanyesmunt: Play/pause from the spacebar seems to work fine without this. Maybe it's need for other OS's?

src/renderer/page/file/view.jsx
```diff
@@ -43,74 +75,85 @@ class FilePage extends React.PureComponent {
...
+      obscureNsfw,
+      playingUri,
+      isPaused,
+      openModal,
     } = this.props;
 
-    const showTipBox = tab == 'tip';
-
+    // This should be included below in the page
+    // Come back to me
```
seanyesmunt: I'll come back to this later

src/renderer/scss/_gui.scss
```diff
@@ -259,16 +295,16 @@ input {
...
-  text-align: right;
-  line-height: 1;
-  font-size: 0.85em;
-  color: var(--color-help);
-}
-
-section.section-spaced {
-  margin-bottom: $spacing-vertical;
-}
+// .sort-section {
```
seanyesmunt: Not sure if we need this, I'll come back to it

src/renderer/component/video/internal/play-button.jsx
liamcardenas: yeah it should be fine without this... ill check linux

liamcardenas: yeah it should be fine without this, I will check linux

liamcardenas: ok i figured our why it was there.

When you load the page, don't click anything, and press the space bar, this starts the video / initiates the download. When you take this out, it no longer works -- at least on linux.

That said, I think its better to take out this functionality, since we are going to implement autoplay soon anyway.

src/renderer/component/walletSendTip/view.jsx
```diff
@@ -1,66 +1,92 @@
...
+class WalletSendTip extends React.PureComponent<Props, State> {
+  constructor(props: Props) {
     super(props);
 
     this.state = {
-      amount: 0.0,
+      amount: 0,
     };
+
+    (this: any).handleSendButtonClicked = this.handleSendButtonClicked.bind(this);
```
liamcardenas: I dont think it really matters, but isn't the better way to do this to make a type for WalletSendTip?

liamcardenas:
@seanyesmunt I think the video player is too small, and the font "Published on ..." is too big and far away from the video title if that makes sense. Also maybe the line spacing in the title is too large.

filepageredesignscreenshot

seanyesmunt:
@liamcardenas Agreed the video player is too small. That is pretty much the min height. When you shrink the window down all the way that is what it will be. I spent some time trying to get it to scale/grow nicely as the width increases, but got made and gave up.

Valid concerns, that will be something we can discuss with nicholas on a group call when I come back around to this.

src/renderer/component/walletSendTip/view.jsx
seanyesmunt: Not sure I follow

PR #973, opened 2018-01-29
Do not kill an existing daemon, but check if one exists

### LINKED ISSUE #298: LBRY installer must detect running daemon

IGassmann:
This removes the functionality of killing an existing daemon and instead checks for an existing one and connect to it.

This will also close #298.


IGassmann:
This will wait for https://github.com/electron-userland/electron-builder/commit/ad199c45525814aba4c457254a2a259c44383486 to be released.

kauffj:
@IGassmann are you sure daemon killing actually happens? Is it default behavior? I'm relatively confident that if I launch `lbrynet-daemon` from the command line and then the app, that it will connect to that running instance already.

IGassmann:
@kauffj that was the behavior before the refactoring. However, during the refactoring I made the app to kill any existing daemon before starting in order to fix #298. With the coming electron-builder release it won't be necessary anymore.

kauffj:
:+1:

tzarebczan:
@IGassmann does electron builder need to be updated again after the lastest commit in: https://github.com/electron-userland/electron-builder/issues/2516

IGassmann:
@tzarebczan Yes, I'm waiting for 20.5.0 to be released for upgrading. It'll also solve this: https://github.com/electron-userland/electron-builder/issues/2670.

Thanks for notifying me.

PR #981, opened 2018-01-31
Fix validity of Linux .desktop file

nedrichards:
Per the menu spec https://standards.freedesktop.org/menu-spec/latest/apa.html if you specify Video you also need to specify the AudioVideo category.

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -39,7 +39,7 @@
   ],
   "linux": {
     "target": "deb",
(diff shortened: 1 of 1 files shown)


kauffj:
Thanks @nedrichards! If you send an LBC address to hello@lbry.io and link this PR, we'll send a tip (more here: lbry.io/faq/tips).

PR #987, opened 2018-02-04
Enable play button after clicking download

### LINKED ISSUE #906: Cannot click "play" immediately after clicking Download
## The Issue
If a user first clicks Download and then decides they want to stream the file, there is a delay where the play button cannot be clicked until the file has started downloading.

### Steps to reproduce
1. Click download
2. Try to click play (on thumbnail right after)

### Expected behaviour
Should be able to click play so that the player immediately starts playing the file when it comes in.

### Actual behaviour
Have to wait until file starts downloading

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

miikkatu:
Fixes #906 by ignoring the isLoading prop when determining the disabled state of the button.

### DIFF
--- a/src/renderer/component/video/internal/play-button.jsx
+++ b/src/renderer/component/video/internal/play-button.jsx
@@ -23,7 +23,7 @@ class VideoPlayButton extends React.PureComponent {
   }
 
   render() {
-    const { button, label, isLoading, fileInfo, mediaType } = this.props;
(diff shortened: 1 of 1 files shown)


tzarebczan:
@miikkatu thanks for the PR. We'll have our devs review it for correctness.  

Please see https://lbry.io/faq/tips if you'd like to be compensated for your contribution.

PR #1017, opened 2018-02-14
[redesign] downloaded/published/transactions page

seanyesmunt:
What I did:
Strip out some unnecessary logic in the published/downloaded pages and change those to use FileCard
Greatly cleanup the transactions page and improve the table styles based off of nizukas design.

What I will do later:
The published/downloaded pages are not responsive at all. Right now the cards are static width. I will come back to this when I work on app wide media queries.


src/main/createWindow.js
```diff
@@ -5,7 +5,7 @@ import setupContextMenu from './menu/setupContextMenu';
 export default deepLinkingURIArg => {
   let windowConfiguration = {
     backgroundColor: '#44b098',
-    minWidth: 800,
+    minWidth: 950,
```
seanyesmunt: Upped this to give us some more space. This is similar to Discords min width

src/renderer/page/fileListDownloaded/view.jsx
```diff
@@ -1,36 +1,32 @@
...
 import Link from 'component/link';
 import { FileTile } from 'component/fileTile';
-import { BusyMessage, Thumbnail } from 'component/common.js';
 import FileList from 'component/fileList';
+import Page from 'component/page';
 
 class FileListDownloaded extends React.PureComponent {
-  componentWillMount() {
-    if (!this.props.isFetchingClaims) this.props.fetchClaims();
-    if (!this.props.isFetching) this.props.fetchFileInfosDownloaded();
```
seanyesmunt: We fetch this on app startup, no need to fetch them again

seanyesmunt:
This is for your downloads/publishes under "My LBRY". I'm working on the "publish" page next. Currently everything is just commented out, that's why it is a blank screen.

PR #1018, opened 2018-02-15
Fix #986 published claims out of order

### LINKED ISSUE #986: Published claims are shown out of order
## The Issue
Because of https://github.com/lbryio/lbry/issues/1098, newly published content does not show at the top of the Published list. This causes confusion because the user may think that their item was not published correctly.  

Currently, the app takes the output of claim_list_mine and reverses it, which produces incorrect results. We should order the published claims by height when using the "date" ordering mechanism. 

### Steps to reproduce
1. Publish a few items
2. Notice that new publishes may not show at the top of the list

### Expected behaviour
Newly published items should be at the top of the Published list

### Actual behaviour
Newly published items show randomly

## System Configuration

 

- LBRY Daemon version: 0.18.2
- LBRY App version: 0.20.0
- LBRY Installation ID:
- Operating system: Windows

Jeremy1026:


src/renderer/component/fileList/view.jsx
```diff
@@ -95,7 +116,8 @@ class FileList extends React.PureComponent {
         <span className="sort-section">
           {__('Sort by')}{' '}
           <FormField type="select" onChange={this.handleSortChanged.bind(this)}>
-            <option value="date">{__('Date')}</option>
+            <option value="dateNew">{__('Date (Newest First)')}</option>
```
kauffj: Let's just label these "Newest First" and "Oldest First"

tzarebczan:
Thanks @Jeremy1026 , sent you LBC on Discord for this PR!.

Issue #1021, opened 2018-02-15
Unable to reopen closed LBRY App window in macOS


Jeremy1026:
## The Issue
After closing the LBRY App window by clicking the red circle in macOS It isn't possible to reopen the window without quitting the application, either by right clicking in the dock and choosing quit, or by giving the LBRY App focus and using Command+Q. Then relaunching the application.

### Steps to reproduce
1. Close the application window.
2. Attempt to reopen the application window.

### Expected behaviour
There should be a way to reload the application window without having to fully quit the application.

### Actual behaviour
The application has to be fully quit and relaunched to have the application become visible again.

## System Configuration
- Operating system: macOS High Sierra (10.13.3)

seanyesmunt:
Fixed by https://github.com/lbryio/lbry-app/pull/1042

PR #1022, opened 2018-02-15
Nullify the window when its closed

Jeremy1026:


tzarebczan:
@Jeremy1026 can you be a little more descriptive in terms of what this fixes? I linked the issue to the PR (with the Zenhub plugin), but would be good to address that in your comments

Jeremy1026:
Sure. The window object currently remains instantiated even after closing it. In macOS clicking the dock icon brings forward all windows of that application. Since the window has been closed, but not removed from memory, macOS still thinks it exists. This fix removes references to the window when it is closed, so that when the dock icon is clicked it will recreate the application window.

This should carry over to Windows/Linux users as well, but needs to be tested on those OS builds to confirm. I don’t currently have those VMs set up to fully test them though.

tzarebczan:
@Jeremy1026 don't have issues bringing LBRY back up after closing to tray on Windows/Ubuntu.

tzarebczan:
This might be fixed by https://github.com/lbryio/lbry-app/pull/1042 also, @Jeremy1026 can you confirm?

seanyesmunt:
Yeah just tested #1042 and confirmed it fixes #1021

Jeremy1026:
Looks good to me as well.

PR #1023, opened 2018-02-16
Fix for loss of data when pasting via context menu

Jeremy1026:


tzarebczan:
Thanks @Jeremy1026 , sent you LBC on Discord for this!

seanyesmunt:
This doesn't fix the issue.

I don't think it's an issue with the context. It actually looks like we don't even need this event listener for the context menu to work. I just deleted it and can still paste stuff.

If I had to guess I think it's an issue with the publish component.

src/renderer/index.js
```diff
@@ -23,8 +23,10 @@ const { autoUpdater } = remote.require('electron-updater');
 autoUpdater.logger = remote.require("electron-log");
 
 window.addEventListener('contextmenu', event => {
```
seanyesmunt: The check you added gets rid of the error that was showing up, but now it just never reaches the `contextMenu()` line. This entire piece can just be removed

seanyesmunt: You can remove it. It currently doesn't do anything. `contextMenu` isn't imported into that file, which is why the error is showing up. The actual context menu is handled in a different place

seanyesmunt:
I've done a little more digging and it seems the  issue is with `react-simplemde-editor`. When pasting with the mouse the `handleChange` is never called.

Not really sure if there is a way around it or if we will need to make a PR to that project

Jeremy1026:
I'll give that a look-see.

lyoshenka:
closing for now. please reopen if you're ready to work on it.

PR #1024, opened 2018-02-16
Analytics overhaul

liamcardenas:
Unfortunately, due to changes in how we route URLs as well as increasing the power, scope, and privacy of analytics, I had to introduce analytics calls outside of renderer/index.js. That said, all the code is contained to one separate file, so its not bad at all.

### DIFF
--- a/src/renderer/analytics.js
+++ b/src/renderer/analytics.js
@@ -1,0 +1,31 @@
+// @flow
+import mixpanel from 'mixpanel-browser';
+
+mixpanel.init('691723e855cabb9d27a7a79002216967');
+
+type Analytics = {
+  track: (string, ?Object) => void,
+  setUser: (Object) => void,
+}
+
+const analytics: Analytics = {
+  track: (name: string, payload: ?Object): void => {
+    if(payload) {
+      mixpanel.track(name, payload);
+    } else {
+      mixpanel.track(name);
+    }
+  },
+  setUser: (user: Object): void => {
+    if(user.id) {
+      mixpanel.identify(user.id);
+    }
+    if(user.primary_email) {
+      mixpanel.people.set({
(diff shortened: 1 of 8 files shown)


src/renderer/analytics.js
```diff
@@ -0,0 +1,42 @@
...
+      } else {
+        mixpanel.track(name);
+      }
+    }
+  },
+  setUser: (user: Object): void => {
+    if(user.id) {
+      mixpanel.identify(user.id);
+    }
+    if(user.primary_email) {
```
seanyesmunt: Should we be using emails in tracking? I thought it would be only user id to keep it anonymized.

src/renderer/analytics.js
```diff
@@ -0,0 +1,42 @@
...
+      mixpanel.people.set({
+        "$email": user.primary_email
+      });
+    }
+  },
+  toggle: (enabled: boolean, logDisabled: ?boolean): void => {
+    if(!enabled && logDisabled) {
+      mixpanel.track('DISABLED');
```
seanyesmunt: What does this do?

seanyesmunt: Oh I'm guessing it's so we can know how many people disable analytics?

src/renderer/analytics.js
liamcardenas: this is so it only tracks people who have an account, which is reasonable

src/renderer/analytics.js
liamcardenas: yes, so we don't get bad data. For instance, we don't think people drop off and we know, instead, they just opted-out of analytics. Also, the reason for the second argument is so that we don't get multiple DISABLED's logged, thereby telling us whenever the open the app.

PR #1032, opened 2018-02-21
Upgrade daemon to 0.19.0

liamcardenas:


tzarebczan:
@liamcardenas adding my comments from Slack:
1) my channels are showing up in Published area, they shouldn't be. 

2) Clicking on items in Downloads/Published area go to an "undefined" claim name, but has correct claim_id 

3) clicking on claims from the transaction screen also goes to undefined claims as #2

Still waiting to hear back from Jack about some file_list changes, but it may be worth looking at https://github.com/lbryio/lbry/issues/1127 to see how it currently works. My main concern will be the editing of already edited claims - I think we use the file_list outpoint when trying to edit, which may not be up to date on the file_list side but should be on claim_list_mine.

PR #1038, opened 2018-02-21
FIX #1027 added link to error box

### LINKED ISSUE #1027: Add startup-troubleshooting FAQ URL to daemon error screen
## The Issue
I think it would be helpful to add a hyperlink to https://lbry.io/faq/startup-troubleshooting (or just the URL if hyperlink not possible) to the error screen when the daemon can't start up properly:
daemonexited

This would be helpful to new app users who have trouble starting up and to existing users who may be having issues connecting to the LBRY network. If we have wallet or server issues that prevent users from starting up, we can put a notice up at the top of the [FAQ](https://lbry.io/faq/startup-troubleshooting) with the latest status. We could also possibly share the https://status.lbry.io on the FAQ.  

### Expected behaviour
Users should be pointed to a resource where they can find more information. If it's a network issue on our end, they would now be notified. 

### Actual behaviour
User is told an error was encountered 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

ykris45:
i think it can solve this issues.. @tzarebczan

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -68,6 +68,7 @@ app.on('ready', async () => {
         dialog.showErrorBox(
           'Daemon has Exited',
           'The daemon may have encountered an unexpected error, or another daemon instance is already running.'
+          <link rel="Read our FAQ" href="https://lbry.io/faq/startup-troubleshooting " />
         );
         app.quit();
       }


src/main/index.js
```diff
@@ -68,6 +68,7 @@ app.on('ready', async () => {
         dialog.showErrorBox(
           'Daemon has Exited',
           'The daemon may have encountered an unexpected error, or another daemon instance is already running.'
+          <link rel="Read our FAQ" href="https://lbry.io/faq/startup-troubleshooting " />
```
btzr-io: You should replace this with the actual url:
 `'https://lbry.io/faq/startup-troubleshooting'`

tzarebczan:
Closing for https://github.com/lbryio/lbry-app/pull/1039 instead.

PR #1039, opened 2018-02-22
Improve daemon error message 

btzr-io:
### Changes
- Add startup-troubleshooting FAQ URL to daemon error

### Preview 
url_fix


tzarebczan:
Should we shorten the URL down? I can add a re-direct on lbry.io/startup if that makes it easier for someone seeing it to access.

btzr-io:
Fix conflicts f6b1e80 ^^

tzarebczan:
I'm OK with the current URL too then.

PR #1040, opened 2018-02-23
[redesign] Publish improvements

seanyesmunt:
#### Changes
Moved publish form values to redux
Move pending publish logic to redux

#### Issues
There are a few minor bugs but I would rather file those as existing issues and keep moving forward.

- I am having issues with publishing under a channel but Tom can successfully do it (just my machine? I'm seeing some certificate error) 
- A few times Tom and I have seen "insufficient funds" errors from the daemon even if you have enough funds (possibly an issue with the number input or the value being coerced to a string)

#### Additional comments
Not super happy with the types. It feels like a lot of duplication. Additionally I'm just keeping the types in the reducer file, until Akinwale's PR to see how he goes about that.


src/renderer/page/publish/index.js
```diff
@@ -4,40 +4,82 @@ import { doNavigate, doHistoryBack } from 'redux/actions/navigation';
...
-  resolvingUris: selectResolvingUris(state),
-});
+const select = (state, props) => {
+  const publishState = state.publish;
+  const { uri, name } = publishState;
+  const { params: { id } } = props;
+
+  let prefillClaim;
+  if (id) {
+    prefillClaim = selectClaimById(id)(state);
```
seanyesmunt: I think I'm probably going to change this and have whoever links to the publish page be responsible for populating the form before navigating. Currently there can be issues with this if you navigate away from the form and come back to an edit.

possibly
```
// file page
populatePublishForm(fileInfo);
doNavigate('/publish')
```

src/renderer/component/address/view.jsx
```diff
@@ -22,36 +22,30 @@ export default class Address extends React.PureComponent<Props> {
...
+            }}
+            onFocus={() => {
+              if (this.input) {
+                this.input.select();
+              }
+            }}
+          />
+          <Button
+            alt
+            icon="Clipboard"
```
kauffj: Use `constants/icons` or decide we don't deem this necessary

src/renderer/component/common/form-components/form-field.jsx
```diff
@@ -15,29 +19,30 @@ type Props = {
...
-  stretch?: boolean
+  stretch?: boolean,
 };
 
 export class FormField extends React.PureComponent<Props> {
   render() {
     const { render, label, prefix, postfix, error, helper, name, type, children, stretch, ...inputProps } = this.props;
 
-    // Allow a type prop to determine the input or more customizability with a render prop
     let Input;
```
kauffj: Why capital? To represent a JSX element?

src/renderer/component/fileCard/index.js
```diff
@@ -8,14 +8,29 @@ import { makeSelectFileInfoForUri } from 'redux/selectors/file_info';
...
-  isResolvingUri: makeSelectIsUriResolving(props.uri)(state),
-});
+const select = (state, props) => {
+  let claim;
+  let fileInfo;
+  let metadata;
+  let isResolvingUri;
+  
+  // If a publish is pending, don't bother with this because props.uri isn't the complete uri at that moment
+  if (!props.pending) {
```
kauffj: Why pass pending in rather than select it from Redux?

src/renderer/component/fileList/view.jsx
```diff
@@ -30,12 +30,13 @@ const sortFunctions = {
...
   name: string,
   channelName: ?string,
+  pending?: boolean,
   value?: {
     publisherSignature: {
       certificateId: string,
     }
   },
-  metadata: {
+  metadata?: {
```
kauffj: Is this `metadata?` because the API is inconsistent about returning metadata? In general, I'm in favor of consistent signatures that include `null` or otherwise empty values (e.g. `[]`), rather than omitting fields entirely.

src/renderer/component/fileList/view.jsx
```diff
@@ -70,8 +71,9 @@ class FileList extends React.PureComponent<Props, State> {
   getChannelSignature = (fileInfo: FileInfo) => {
     if (fileInfo.value) {
       return fileInfo.value.publisherSignature.certificateId;
+    } else if (fileInfo.metadata) {
```
kauffj: What's going on here that this can have two different data structures?

If the daemon is inconsistent with it's responses, I'd like to get issues opened to clean this up.

src/renderer/component/link/view.jsx
```diff
@@ -26,85 +26,94 @@ type Props = {
...
+      onClick,
+      href,
+      title,
+      label,
+      icon,
+      iconRight,
+      disabled,
+      children,
+      navigate,
+      navigateParams,
```
kauffj: Just your normal 22 parameter function. Likely out of scope for this PR, but is it possible some `Button` usage should simply be another component?

src/renderer/component/publishForm/internal/license-type.jsx
```diff
@@ -0,0 +1,145 @@
...
+  copyrightNotice: ?string,
+  licenseUrl: ?string,
+  otherLicenseDescription: ?string,
+  handleLicenseChange: (string, string) => void,
+  handleLicenseDescriptionChange: (SyntheticInputEvent<*>) => void,
+  handleLicenseUrlChange: (SyntheticInputEvent<*>) => void,
+  handleCopyrightNoticeChange: (SyntheticInputEvent<*>) => void,
+}
+
+class LiscenseType extends React.PureComponent<Props> {
```
kauffj: Extra `s`

src/renderer/component/publishForm/internal/license-type.jsx
```diff
@@ -0,0 +1,145 @@
...
+      <div className="card__content">
+        <FormField
+          label={__("License (Optional)")}
+          type="select"
+          value={licenseType}
+          onChange={this.handleLicenseOnChange}
+        >
+          <option>{__('None')}</option>
+          <option value="Public Domain">{__('Public Domain')}</option>
+          <option
```
kauffj: This could probably be done by iterating over a data structure rather than enumerating all of them

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,884 +1,574 @@
...
+import ChannelSection from 'component/selectChannel';
+import Icon from 'component/common/icon';
+import classnames from 'classnames';
+import path from 'path';
+import type { PublishParams, UpdatePublishFormData } from 'redux/actions/publish';
+import FileSelector from 'component/common/file-selector';
+import BidHelpText from './internal/bid-help-text';
+import LicenseType from './internal/license-type';
+
+type Props = {
```
kauffj: This still seems like a giant collection of `Props`. Is it possible to subdivide this into more components to facilitate easier understanding?

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,884 +1,574 @@
...
+    } = this.props;
 
-    for (const metaField of ['title', 'description', 'thumbnail', 'language']) {
-      const value = this.state[`meta_${metaField}`];
-      if (value) {
-        metadata[metaField] = value;
-      }
+    let publishingLicense;
+    switch (licenseType) {
+      case 'copyright':
```
kauffj: If these string values are shared across components, they should be defined as constant types similar to icons or actions.

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,884 +1,574 @@
...
-  handleEditClaim() {
-    const claimInfo = this.claim() || this.myClaimInfo();
+  handleCancelPublish() {
+    const { clearPublish } = this.props;
 
-    if (claimInfo) {
-      this.handlePrefillClaim(claimInfo);
+    // #content wraps every page and lives in component/page
+    const mainContent = document.getElementById("content");
+    if (mainContent) {
```
kauffj: This creates a non-obvious dependency. I'd suggest making `scrollToTop` an action, at a minimum.

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,884 +1,574 @@
...
-
-  handleOtherLicenseUrlChange(event) {
-    this.setState({
-      otherLicenseUrl: event.target.value,
-    });
-  }
+    let bidError;
+    if (balance <= bid) {
+      bidError = __('Not enough credits');
+    } else if (bid <= 0.00000001) {
```
kauffj: If this is the minimum bid amount, it should be defined somewhere.

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,884 +1,574 @@
...
-    return '';
-  }
 
-  closeModal() {
-    this.setState({
-      modal: null,
-    });
+    return (
+      <div className="card__subtitle form-field__error">
+        {!title && <p>A title is required</p>}
```
kauffj: `__()`

kauffj: Also, should these be inside `<p>`?

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,884 +1,574 @@
...
+                  label={__('Language')}
+                  type="select"
+                  name="content_language"
+                  value={language}
+                  onChange={event => updatePublishForm({ language: event.target.value })}
+                  >
+                  <option value="en">{__('English')}</option>
+                  <option value="zh">{__('Chinese')}</option>
+                  <option value="fr">{__('French')}</option>
+                  <option value="de">{__('German')}</option>
```
kauffj: Can we add more languages than this? (Possibly out of scope, file separately if so)

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,884 +1,574 @@
...
+                  step="any"
+                  label={__('Deposit')}
+                  postfix="LBC"
+                  value={bid}
+                  error={bidError}
+                  min="0"
+                  disabled={!name}
+                  onChange={event => this.handleBidChange(event.target.value)}
+                  helper={__('This LBC remains yours and the deposit can be undone at any time.')}
+                  placeholder={winningBidForClaimUri ? winningBidForClaimUri + 10 : 10}
```
kauffj: I thought we shifted from 10 as the placeholder to a smaller amount, like 0.01.

I know some people were confused that placeholders were default or suggested values.

And even if they're not confused, the value here will act as a psychological anchor. We should consider where we want to anchor people.

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,884 +1,574 @@
...
+                  formDisabled ||
+                  !formValid ||
+                  publishing
+                }/>
+                <Button alt onClick={this.handleCancelPublish} label={__('Cancel')} />
+            </div>
+            {!formDisabled && !formValid && this.renderFormErrors()}
+          </div>
+        </Form>
+        <Modal
```
kauffj: Modals should no longer exist directly in views.

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,884 +1,574 @@
...
+            navigate('/published');
+          }}
+        >
+          <p>
+            {__('Your file has been published to LBRY at the address')}{' '}
+            <code>{uri}</code>!
+          </p>
+          <p>
+            {__(
+              'The file will take a few minutes to appear for other LBRY users. Until then it will be listed as "pending" under your published files.'
```
kauffj: Probably out of scope, but this is something we could allow a user to acknowledge and not see again.

src/renderer/component/selectChannel/view.jsx
```diff
@@ -0,0 +1,221 @@
...
+      fetchChannelListMine();
+    }
+  }
+
+  handleChannelChange(event: SyntheticInputEvent<*>) {
+    const { onChannelChange } = this.props;
+    const channel = event.target.value;
+
+    if (channel === 'new') {
+      this.setState({ addingChannel: true });
```
kauffj: Now that the publish page generally retains state via Redux, will this internal component state be a potential source of error or confusion?

src/renderer/component/selectChannel/view.jsx
```diff
@@ -0,0 +1,221 @@
...
+                key="channel"
+                type="select"
+                tabIndex="1"
+                onChange={this.handleChannelChange}
+                value={channel}
+              >
+                <option key="anonymous" value="anonymous">
+                  {__('Anonymous')}
+                </option>
+                {channels.map(({ name }) => (
```
kauffj: It'd be cool if this defaulted to the last channel you published with.

src/renderer/lbry.js
```diff
@@ -35,77 +35,6 @@ function setLocal(key, value) {
   localStorage.setItem(key, JSON.stringify(value));
 }
 
-/**
- * Records a publish attempt in local storage. Returns a dictionary with all the data needed to
- * needed to make a dummy claim or file info object.
- */
-let pendingId = 0;
-function savePendingPublish({ name, channelName }) {
```
kauffj: :clap:

src/renderer/page/publish/index.js
kauffj: :+1:

src/renderer/lbry.js
```diff
@@ -137,39 +66,10 @@ Lbry.connect = () => {
...
-    result => {
-      if (returnPendingTimeout) clearTimeout(returnPendingTimeout);
-      publishedCallback(result);
-    },
-    err => {
-      if (returnPendingTimeout) clearTimeout(returnPendingTimeout);
-      errorCallback(err);
-    }
-  );
+Lbry.publishContent = (params, publishedCallback, errorCallback) => {
```
kauffj: I think you can drop this and just use `lbry.publish(...).then(...)` in code.

src/renderer/component/selectChannel/view.jsx
seanyesmunt: That's a good idea

src/renderer/component/fileList/view.jsx
seanyesmunt: That part specifically is for pending publishes. A pending publish won't have the `metadata` object. Maybe the issue is not keeping it the same. I know there was a lot of code to build an actual dummy claim in `lbry.js`, currently `pendingPublishes[0]` is the payload we send to `lbry.publish`

seanyesmunt: Yeah, I'm going to add the dummy claim stuff back inside the pending publish. I ran into a couple issues with it, and I'm sure I will only run into more

src/renderer/component/fileCard/index.js
seanyesmunt: 👍

src/renderer/component/common/form-components/form-field.jsx
seanyesmunt: It was capital when I had the render function. Now that I removed it I'll set it back to lowercase.

src/renderer/component/link/view.jsx
seanyesmunt: Agreed. Thinking about what can be removed

src/renderer/component/publishForm/view.jsx
seanyesmunt: I guess a `<div>` is probably better since this isn't a paragraph.

src/renderer/component/selectChannel/view.jsx
seanyesmunt: This might be a good component to test out your idea, just to see how it works in practice. I'll keep it as is for now, but look into refactoring when we add it to the "Your publishes" page, or somewhere else.

kauffj: :+1:

seanyesmunt:
@kauffj Realized I didn't add contstants for 'anonymous' and 'new' for channel creation. And also forgot to fix the price placeholder after your comments. I'll add those.

PR #1042, opened 2018-02-23
Save app state when closing to tray

IGassmann:


tzarebczan:
@IGassmann working as expected here, tested on Windows and Linux. 

Would be nice to have a build because lbry:// hyperlinking doesn't seem to work in dev mode.  Do you think that's a quick fix, I can open an issue for it.

IGassmann:
@tzarebczan it's actually expected that `lbry://` deep linking URLs don't work on dev mode. Deep link URLs only work after they have been "announced" to the OS. For example on macOS, this is done with the App descriptor file `Info.plist` which is only generated on a packaged build and not on a dev build.

I've looked into the double-clicking for the tray icon. The feature is not yet supported on Linux: https://electronjs.org/docs/all#event-double-click-macos-windows

Issue #976, opened 2018-01-29
Ability to export wallet transactions to CSV format


tzarebczan:
## The Issue
As an app user, I'd like a way to export all my wallet transactions and details to a csv file. This would allow users to perform an external analysis for various purposes.

btzr-io:
@tzarebczan feature merged in https://github.com/lbryio/lbry-app/commit/c6daff53c5b3281db5892ae176c9aa93662763b4,
are you going to close this issue after the release :upside_down_face: ?

tzarebczan:
Thanks @btzr-io , closing now. @liamcardenas , can you get some credits over to Baltazar for this PR? He's available on Discord.

PR #1043, opened 2018-02-24
Feature: export  wallet transactions

btzr-io:
## Changes
- Added Component `FileExporter`.
- Ability to export wallet transactions to `JSON` and `CSV` format #976
- Added `util/parseTransactions.js` : 
Parse relevant data to be exported (Wallet transactions, bookmarks, playlist...)
- WIP...


tzarebczan:
Was able to export to both csv and json.  Can it default to csv by default? Looks like it adheres to the filter chosen too, great! 

I'm not sure if I'd put the export button on the Overview page, maybe just history? Maybe put it on the same line as Transaction History, but right justified? @kauffj, thoughts?

btzr-io:
Fix conflicts: 3505e43

PR #1046, opened 2018-02-26
Fix publish errors show [object Object] error instead of actual error

IGassmann:


tzarebczan:
@IGassmann When I tested this, I noticed that the publish success message pops up very briefly right before the error - any way to prevent that?

IGassmann:
@tzarebczan If I'm not mistaken, that was already happening before this PR. Would you be able to open an issue for it?

PR #1047, opened 2018-02-26
Speech Thumbnail Upload

daovist:
NEW PULL REQUEST; PLEASE IGNORE #939 

Speech Thumbnail Upload allows a user to upload a thumbnail to spee.ch through a file input element. Everything works through Redux, the upload fetch is in an action. The publishForm component listens for changes through a selector and updates meta_thumbnail accordingly.


src/renderer/component/speechUpload/view.jsx
```diff
@@ -0,0 +1,86 @@
...
+      this.refs.thumbnail.val = null;
+    }
+  }
+
+  render() {
+    const { openModal } = this.props;
+
+    return (
+      <div>
+        {this.props.uploadStatus === status.MANUAL ? null : (
```
liamcardenas: can't you combine this and the below into one turnary operation?

i.e.
`this.props.uploadStatus === status.MANUAL ? <text input goes here> : <file select input goes here>`

src/renderer/component/speechUpload/view.jsx
```diff
@@ -0,0 +1,86 @@
+import React from "react";
```
liamcardenas: I think this and other files should be called something along the lines of `thumbnailSelect` and not `speechUpload`

tzarebczan:
@liamcardenas added you back as a reviewer, looks like @daovist made some commits.

PR #1051, opened 2018-02-27
Autoplay

daovist:
Add autoplay client setting

Video component calls doPlayUri at componentWillMount if selectAutoplay is true

### DIFF
--- a/src/renderer/component/video/index.js
+++ b/src/renderer/component/video/index.js
@@ -11,7 +11,7 @@ import {
   makeSelectDownloadingForUri,
 } from 'redux/selectors/file_info';
 import { makeSelectCostInfoForUri } from 'redux/selectors/cost_info';
-import { selectShowNsfw } from 'redux/selectors/settings';
+import { selectShowNsfw, selectAutoplay } from 'redux/selectors/settings';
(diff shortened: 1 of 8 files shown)


seanyesmunt:
It should set it on `componentDidMount` to be safe.

daovist:
What I have here works but I am getting an error in the console:

`Uncaught (in promise) DOMException: The play() request was interrupted by a new load request.`

Should I try to [fix this with a promise](https://developers.google.com/web/updates/2017/06/play-request-was-interrupted) or something?

seanyesmunt:
@daovist Seems like that link offers a safe way to fix the error you are seeing.

daovist:
@seanyesmunt I am trying to resolve the error but since it fails silently outside the console I have had another issue that I have been more focused on here.

I am unable to purchase videos, so I can't figure out what I need to make this work for purchased videos. I have tried many times, and even reinstalled LBRY, but I have been unable to download a single non-free video. Several times I have been charged though.

tzarebczan:
@liamcardenas , what's the next step here? I can give this a test later on today also.

tzarebczan:
@daovist, I'll reach out to you on Discord regarding purchasing issues.

PR #1045, opened 2018-02-25
 [FIX #988] Disable drag and drop

ykris45:
## Preview

https://spee.ch/a/ScreenCapture2018-2-26-042956.mp4

PR #1056, opened 2018-02-28
[FIX #988] Disable drag and drop 

### LINKED ISSUE #988: Disable drag and drop of files into the app
## The Issue
Based on the discussion in https://github.com/lbryio/lbry-app/pull/549#issuecomment-362819852, we should disable file drag & drop into the app until it can support publishing, otherwise, it causes confusion because the file is simply shown in the app. 

### Steps to reproduce
1. Drag a picture or video file into the LBRY app
2. LBRY displays the video/image
3. No way to exit, need to close the app and reopen (completely or from tray)

### Expected behaviour
Prompt publish feature or do nothing

### Actual behaviour
Opens the file

## System Configuration

 

- LBRY Daemon version: 0.18.2
- LBRY App version: 0.20.0
- LBRY Installation ID:
- Operating system: Windows

ykris45:
## Preview

https://spee.ch/a/ScreenCapture2018-2-26-042956.mp4

#1045 ( my old PR i close because i cannot squash github)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Web UI version numbers should always match the corresponding version of LBRY App
   * Fixed sort by date of published content ([#986](https://github.com/lbryio/lbry-app/issues/986))
   * Fix night mode start time, set to 9PM (#1050)
   * Fix night mode start time, set to 9PM ([#1050](https://github.com/lbryio/lbry-app/issues/1050))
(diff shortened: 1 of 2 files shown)


seanyesmunt:
This will affect https://github.com/lbryio/lbry-app/pull/549. That won't be in anytime soon so this is fine as is.

Issue #338, opened 2017-07-07
Retain previous screen sizing on startup (LBRY starts maximized)


tzarebczan:
## The Issue
When LBRY is run, the app always goes into full screen mode.  We've had a user that mentioned this should not happen or should be configurable.  Is it possible to remember the previous window size upon next start up?  

If this is the intended functionality to maximize Home page content viewing, so feel free to close. 

### Steps to reproduce
1. Launch LBRY
2. Goes full screen

### Expected behaviour
LBRY should either not go in full screen mode or remember previous app close window size

### Actual behaviour
Always goes into full screen mode

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system: Windows

randulo:
Yes, this is annoying when you have a large screen. Recalling window size is an excellent solution. Thanks for entering this.

tzarebczan:
Actually, on Windows it does seem to keep the window size to the setting which LBRY was closed. Can someone confirm on Mac? I'll check Ubuntu later.

kauffj:
Maximize is called on every run. It's called here: https://github.com/lbryio/lbry-app/blob/fce702f737909d21e9b7581051690f356a82bd56/app/main.js#L155

To address this ticket, size should be saved locally on close and set in the above linked function.

akinwale:
Should the saved width and height override the minWidth and minHeight properties respectively?

randulo:
If I were coding this, I'd make sure there was a minimum, because I have seen app windows that became so small by accident, they couldn't be made larger. The minimum could be 800x600 if the remembered size was too small. That size window is actually usable, if not optimal.

kauffj:
Agreed, please retain the minimum @akinwale.

tzarebczan:
Per suggestion from a community member: https://github.com/mawie81/electron-window-state

PR #1057, opened 2018-03-01
Retain previous screen sizing on startup

miikkatu:
This change fixes issue #338.

### DIFF
--- a/src/main/createWindow.js
+++ b/src/main/createWindow.js
@@ -1,15 +1,32 @@
-import { app, BrowserWindow, dialog } from 'electron';
+import { app, BrowserWindow, dialog, screen } from 'electron';
 import isDev from 'electron-is-dev';
+import windowStateKeeper from 'electron-window-state';
+
 import setupBarMenu from './menu/setupBarMenu';
 import setupContextMenu from './menu/setupContextMenu';
 
 export default appState => {
+  // Get primary display dimensions from Electron.
+  const { width, height } = screen.getPrimaryDisplay().workAreaSize;
+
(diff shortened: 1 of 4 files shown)


src/main/createWindow.js
```diff
@@ -1,14 +1,27 @@
 import { app, BrowserWindow, dialog } from 'electron';
+import windowStateKeeper from 'electron-window-state';
+
 import setupBarMenu from './menu/setupBarMenu';
 import setupContextMenu from './menu/setupContextMenu';
 
 export default appState => {
+  // Load the previous state with fallback to defaults.
```
liamcardenas: is there any way that we can have a fully maximized default?

miikkatu: I updated this PR so that if the window state is not saved yet (the default situation with new app installs), the display resolution is asked from Electron and that's used as the default.

tzarebczan:
@seanyesmunt do you think you can review this?

seanyesmunt:
@miikkatu Looks great. Could you fix the conflicts? I'll merge after that

miikkatu:
@seanyesmunt PR updated.

tzarebczan:
@IGassmann can you review when you get a moment. This touches similar files to the URL issue we've discussed.

IGassmann:
@miikkatu Please add a new changelog entry in the CHANGELOG.md file. After that, this should be ready to merge.

miikkatu:
@IGassmann Ok, change log is now updated.

tzarebczan:
Thanks again @miikkatu , sending over a tip shortly :)

Issue #1052, opened 2018-02-27
Eliminate or severely reduce typing timeout to trigger search


kauffj:
Obviously, many people are still confused about how to search.

I'm still not ready to fully cave on this, but it's clear there is a problem.

As a starting place, let's try eliminating the timeout to transition the user to the search page. As soon as they type, take them there. We may want to still delay subsequent searches for performance reasons (sending a search on every keypress will trigger a lot of XHR requests).

tzarebczan:
Think this is addressed in the redesign.

PR #1060, opened 2018-03-02
Reduce typing timeout to trigger search

miikkatu:
This would solve #1052.

There are two commits. The first one fixes various linter issues in the files related to the actual fix, which is in the second commit.

The fix is simply to navigate immediately to the search page, if the user is not there yet. This does not change the existing 800ms typing delay between searches, which remains tweakable in the code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "mixpanel-browser": "^2.17.1",
     "moment": "^2.20.1",
     "npm": "^5.5.1",
+    "prop-types": "^15.6.1",
     "qrcode.react": "^0.7.2",
     "rc-progress": "^2.0.6",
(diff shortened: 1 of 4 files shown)


seanyesmunt:
@miikkatu Thanks for your work. I was hoping to merge this after app v21 was released and get it in for one release before the redesign.

The redesign completely overhauls the search functionality so this won't be needed anymore. I am merging that soon. Sorry we can't merge this. We will still pay you for this. @tzarebczan have you sent @miikkatu any LBC for this?

tzarebczan:
@seanyesmunt nope, not yet, but will do. @miikkatu reach out to me on hello@lbry.io like last time :)

Issue #1049, opened 2018-02-26
Support channel definitions from /file/list_homepage


kauffj:
Modify `doFetchFeaturedUris` to support working for channels.

Currently, the following URL returns the content to be displayed on the app homepage:

https://api.lbry.io/file/list_homepage

If the object key begins with an `@` (indicating a channel) and the listed content array is empty, `doFetchFeaturedUris` should call `channel_list` to return the 10 most recent items in that channel.

Additionally, if the object key begins with an `@`, the subscribe button should be shown next to the channel name.

kauffj:
@reillysmith

daovist:
So I dug into `doFetchFeaturedUris` and got down to `lbry.js` and `jsonrpc.js` and now understand much better what all is going on. It took me a while to grok that Proxy mechanics are handling methods not explicitly stated in `lbry.js`. I think @kauffj meant `claim_list_by_channel` rather than `channel_list` as I was able to hit that endpoint on the daemon with Postman. I see two ways of doing this.

1)
A) modify [doFetchFeaturedUris](https://github.com/lbryio/lbry-app/blob/master/src/renderer/redux/actions/content.js#L72) to also include `channelsToResolve`; use an if statement to update URI and channel lists accordingly
B) use `Lbry.claim_list_by_channel` to get URIs for each channel name
C) put relevant data into `urisToResolve`
D) and let `doResolveUris` get the claim info for channel content along with the rest of the featured content it's already grabbing

or 2) call `doFetchClaimsByChannel` in the 1A if statement when the category is a channel and sort it all out in the view

I think 1 is simpler from a data model perspective but would be slower because there may be multiple channels each of which would make its own RPC before making the one big RPC for all the info. I think 2 would be faster but more complex putting the data back together.

I also think I may be overthinking this. Please advise.

@seanyesmunt @liamcardenas

kauffj:
@daovist yes, I did mean `claim_list_by_channel`.

I agree these are the two approaches. Approach 2 is probably more "proper" or "React-like".

That is:
1. Refactor `FeaturedCategory` to be a standalone component.
2. If `FeaturedCategory` receives a channel name but no URIs, call out to `doFetchClaimsByChannel` to receive them.

It may make sense to refactor other names here a bit as well (e.g. `FeaturedCategory` takes "names" currently, but "uris" is a more appropriate term).

You can look at `page/channel` for inspiration.

tzarebczan:
[...]

PR #1061, opened 2018-03-02
Support channel definitions #1049

daovist:
This is not finished, submitting for review.

I updated `doFetchFeaturedUris` to call `doFetchClaimsByChannel` for Uris beginning with "@" and hard coded 2 channels to be added to `Uris` for development

I copied the logic from the subscriptions page to the discover page. The subscriptions display is beneath the normal discover content.

When I view Subscriptions and then go back to Discover, I see the channels I'm subscribed to displayed correctly (in terms of UI) but not the channels in the `Uris` list which is what we actually want. Basically, the UI part of this is good or close to good but the data flow is still coming along.


kauffj:
Codacy bot angry! Codacy bot smash!

PR #1066, opened 2018-03-06
Subscribe notify

liamcardenas:
This notifies users whenever their is new content available for them on a channel they have subscribed to:
- it checks for new content
- it notifies the user that 1 or more (up to "9+") items are available per channel
- it starts downloading the most recent newly available video from each channel (if it is free)
- its behavior was modeled after the iPhone podcast app

TODO: badges, I could add them into this PR, or maybe make a follow up


src/renderer/redux/actions/subscriptions.js
```diff
@@ -14,5 +21,98 @@ export const doChannelUnsubscribe = (subscription: Subscription) => (dispatch: D
...
+          (prev, cur, index) =>
+            buildURI({ contentName: cur.name, claimId: cur.claim_id}, false) === subscription.latest ? index : prev,
+          -1
+        )
+      : 1;
+
+    if (count !== 0) {
+      if (!claimsInChannel[0].value.stream.metadata.fee) {
+        dispatch(
+          doPurchaseUri(buildURI({ contentName: claimsInChannel[0].name, claimId: claimsInChannel[0].claim_id }, false), { cost: 0 })
```
seanyesmunt: Why do you need to pass in the new cost object. Shouldn't `doPurchaseUri` be able to select that data with the uri?

src/renderer/page/file/view.jsx
```diff
@@ -17,6 +17,7 @@ class FilePage extends React.PureComponent {
   componentDidMount() {
     this.fetchFileInfo(this.props);
     this.fetchCostInfo(this.props);
+    this.checkSubscriptionLatest(this.props);
```
seanyesmunt: You can avoid a lot of unnecessary calls by only calling `checkSubscriptionLatest` if you are subscribed to that channel

src/renderer/page/file/view.jsx
```diff
@@ -35,6 +36,16 @@ class FilePage extends React.PureComponent {
     }
   }
 
+  checkSubscriptionLatest(props) {
+    props.checkSubscriptionLatest(
+      {
+        channelName: props.claim.channel_name,
+        uri: buildURI({ contentName: props.claim.channel_name, claimId: props.claim.value.publisherSignature.certificateId }, false),
```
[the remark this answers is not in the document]
kauffj: @IGassmann this is just a formatting suggestion right? Can we have `Prettier` just do this rather than receive this message?

src/renderer/redux/actions/subscriptions.js
```diff
@@ -14,5 +21,98 @@ export const doChannelUnsubscribe = (subscription: Subscription) => (dispatch: D
...
+
+    //$FlowIssue
+    dispatch({
+      type: ACTIONS.CHECK_SUBSCRIPTION_COMPLETED,
+      data: subscription,
+    });
+  });
+};
+
+export const checkSubscriptionLatest = (channel: Subscription, uri: string) => (
```
kauffj: Is this function actually necessary? Could `doCheckSubscription` just call `setSubscriptionLatest`?

src/renderer/redux/actions/subscriptions.js
```diff
@@ -14,5 +21,98 @@ export const doChannelUnsubscribe = (subscription: Subscription) => (dispatch: D
...
+      : 1;
+
+    if (count !== 0) {
+      if (!claimsInChannel[0].value.stream.metadata.fee) {
+        dispatch(
+          doPurchaseUri(buildURI({ contentName: claimsInChannel[0].name, claimId: claimsInChannel[0].claim_id }, false), { cost: 0 })
+        );
+      }
+
+      const notif = new window.Notification(subscription.channelName, {
```
kauffj: Can we wait to show this notification until after content is ready to watch? I.e. after `get` has successfully started? That way when the user clicks the notification they get something immediately.

src/renderer/page/file/view.jsx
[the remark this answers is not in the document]
IGassmann: Exactly. Prettier was previously run with every commit thanks to Husky. Since Husky is currently disabled due to a problem it was causing with the publishing the app, it doesn't automatically run. However, Prettier can still be run on all source files by entering the command `yarn format`. https://github.com/lbryio/lbry-app/blob/master/CONTRIBUTING.md#code-formatting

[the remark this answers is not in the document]
kauffj: @IGassmann the feedback from codacy seems largely useful so far, but the ones labelled `prettier/prettier` feel like noise. They're difficult to read and also all fixed in the same way - run `Prettier`.

Unless others disagree and assuming it's possible I'd disable just these for now.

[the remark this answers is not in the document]
IGassmann: Done. Prettier errors are disabled until Husky come back.

src/renderer/redux/actions/subscriptions.js
```diff
@@ -14,5 +21,98 @@ export const doChannelUnsubscribe = (subscription: Subscription) => (dispatch: D
...
+        body: `Posted ${claimsInChannel[0].value.stream.metadata.title}${
+          count > 1 ? ` and ${count - 1} other new items` : ''
+        }${count < 0 ? ' and 9+ other new items' : ''}`,
```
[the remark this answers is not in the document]
liamcardenas: it does, it just says
```"Sean"
"Posted myVideo"```

instead of, for examplem
```"Sean"
"Posted myVideo and 9+ other new items"```

src/renderer/redux/actions/subscriptions.js
liamcardenas: its is more efficient this way because I already have the cost information so we don't need to fetch it again and

src/renderer/page/file/view.jsx
liamcardenas: good call

src/renderer/redux/actions/subscriptions.js
liamcardenas: check and set are different, set simply sets the value in the redux store. Check does an asynchronous call and sets the value accordingly by dispatching set, if that makes sense.

src/renderer/redux/actions/subscriptions.js
liamcardenas: Yes, @seanyesmunt brought this up as well. I have reasons for why I didn't make it this way, which I think we should discuss -- although I think you are right that we probably want to download it first and I probably should have implemented it this way. I will schedule this in the next sprint.

kauffj: I checked out the code here:
1) `doLoadVideo` and `doPurchaseUri` are confusingly -- almost backwardly -- named. It may make sense to tweak these.
2) Since you are already checking that the price is free, it seems like you could potentially skip `doPurchaseUri` and go right to `doLoadVideo`.
3) Both `doLoadVideo` and `doPurchaseUri` can trigger modal popups. It is weird that asynchronous background loading of content can trigger a modal. Fairly guaranteed that this will cause confusion at some point.
4) It doesn't seem particularly tricky to attach a notification to the success of the call to `get`. I think a light modification of `doLoadVideo` to take a notification or callback would get you there.

src/renderer/redux/actions/subscriptions.js
kauffj: I'm asking two things in my question:
1) If we're already fetching the latest subscription data at time interval `n`, is it actually necessary to check for the latest for this same information when accessing a file page?
2) If it is necessary (for staleness reasons, presumably), why can't more of the code be shared between the two? Isn't `doCheckSubscription` a superset of the functionality provided by `checkSubscriptionLatest`? and why wouldn't you want `checkSubscriptionLatest` to also begin downloads? Furthermore, why do _both_ of these call `Lbry.claim_list_by_channel` when they could call `doFetchClaimsByChannel` instead?

kauffj: @liamcardenas did you see above?

PR #1077, opened 2018-03-08
start search quickly if search dispatch was pending

daovist:
#1052 

This adds a new constant `START_SEARCH` to WunderBar. Before there was only `TYPING_TIMEOUT`. Now the SetTimeout call uses `TYPING_TIMEOUT` for delay time (800ms) if there was a search dispatch pending, so it won't load more requests than it already was, but if there was no search dispatch pending it uses `START_SEARCH` for delay time (100ms).

It works like I think it should based on reading the issue


seanyesmunt:
1052 probably shouldn't have been created. The redesign will overhaul the search process so this won't be needed any more.

PR #1081, opened 2018-03-11
Update view.jsx

LavRadis:
A small idea on how the "Help" section could be improved with a link to status.lbry.io. It could help the users to identify any current slow-down or problems. I was not sure what would be the appropriate button icon so I just left the "question mark" one. Feel free to dismiss this idea as not important or change any wording. It was not tested locally.

### DIFF
--- a/src/renderer/page/help/view.jsx
+++ b/src/renderer/page/help/view.jsx
@@ -124,7 +124,23 @@ class HelpPage extends React.PureComponent {
             <div className="meta">{__('Thanks! LBRY is made by its users.')}</div>
           </div>
         </section>
-
+        <section className="card">
+          <div className="card__title-primary">
+            <h3>{__('LBRY Services status')}</h3>
+          </div>
+          <div className="card__content">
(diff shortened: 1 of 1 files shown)


tzarebczan:
@kauffj do we want to expose status.lbry.io in the LBRY app?

kauffj:
@tzarebczan this page doesn't need to a secret, but I think I'm inclined to agree we don't want to drive extra traffic to it. Additionally:

- I think that page could cause confusion among users about what all those services are and what they do
- It's not like people will know to go to the Help page and click a link to learn about status issues.

LavRadis:
I'm always for giving users more options and information. We could make a description somewhere about what those services are. we can also assume that a high percentage will understand that if something is red or "down" that something is not working on the server side and their app is not broken.

tzarebczan:
@LavRadis will close this for now and revisit in the future. Can you open an issue for something along the lines of "Show simple LBRY network external network status" and reference this PR. Goal is not to be confused with an existing issue we have where we want to show users their connection status to the network (i.e. are my ports open, am I connected properly).

Issue #1048, opened 2018-02-26
Re-poll doFetchFeaturedUris every so often


lyoshenka:
`doFetchFeaturedUris` should be re-called every 1-2 hours (?)

tzarebczan:
Fixed in https://github.com/lbryio/lbry-app/pull/1646

PR #1090, opened 2018-03-12
Re-poll doFetchFeaturedUris every so often

daovist:
issue #1048 

I copied the way `doBalanceSubscribe` and `doUpdateBalance` work together in the form of `doFetchFeaturedUris` and `doFetchFeaturedUrisNow`. Every 40 seconds (will be 1 hour once it works properly) this calls the action and fetches the correct data, but if you leave the Discover page and come back, it starts a new loop of calling the action every 40 seconds. I've a tried a few ways of naming the setInterval so I can clearInterval but haven't got it working properly yet. I can think of how I'd do it with new state properties and reducers but am trying to avoid that.

I am digging deeper into this and submitting the PR to discuss the code

### DIFF
--- a/src/renderer/redux/actions/content.js
+++ b/src/renderer/redux/actions/content.js
@@ -71,6 +71,13 @@ export function doResolveUri(uri) {
 
 export function doFetchFeaturedUris() {
   return dispatch => {
+    dispatch(doFetchFeaturedUrisNow());
+    setInterval(() => dispatch(doFetchFeaturedUrisNow()), 40000);
+  }
+}
+
+export function doFetchFeaturedUrisNow() {
+  return dispatch => {
     dispatch({
       type: ACTIONS.FETCH_FEATURED_CONTENT_STARTED,
     });


src/renderer/redux/actions/content.js
```diff
@@ -70,6 +70,13 @@ export function doResolveUri(uri) {
 }
 
 export function doFetchFeaturedUris() {
+  return dispatch => {
+    dispatch(doFetchFeaturedUrisNow());
+    setInterval(() => dispatch(doFetchFeaturedUrisNow()), 40000);
+  }
+}
+
+export function doFetchFeaturedUrisNow() {
```
liamcardenas: Setting an interval like this is absolutely correct, but it would be better if you found the place that originally triggers `doFetchFeaturedUris` and set the interval there instead of creating two actions.

seanyesmunt:
@daovist It would be great to get this in. I agree with Liam's comments, you should be able to move the `setInterval` to wherever the `doFetchFeaturedUris` is originally called

tzarebczan:
@daovist do you plan on picking this back up?

PR #1100, opened 2018-03-14
Fix sort

liamcardenas:


tzarebczan:
@liamcardenas per our discussion, we were only supposed to revert the downloads side to use the default sort from file_list, but the publish side should still be using claim height.

PR #1105, opened 2018-03-15
Remove app side balance checks

tzarebczan:
Adding 2 checks to channel bid (equal to balance, and equal to 0) based on https://github.com/lbryio/lbry-app/pull/794 because the error messages don't make it from the daemon, so will these in place.

Removed publish side check, error message from the daemon are now more accurate. 

Tested this out locally.


seanyesmunt:
Looks good. I would just switch to `===` instead of `==`

tzarebczan:
@seanyesmunt / @liamcardenas this can actually be merged now, doesn't need the latest daemon RC.

PR #1116, opened 2018-03-16
Disable rewards when diagnostics are off

liamcardenas:


tzarebczan:
@liamcardenas tested this out on master and it LGTM!

PR #1144, opened 2018-03-20
Allow channels on Discover Page

daovist:
#1049 the way @kauffj explained in PR #1064 

sorry to make another PR but this one is entirely different

this still has hard-coded values and the UI needs a little work

### DIFF
--- a/src/renderer/component/featuredCategory/view.jsx
+++ b/src/renderer/component/featuredCategory/view.jsx
@@ -244,6 +244,9 @@ class FeaturedCategory extends React.PureComponent {
                 <FileCard key={name} displayStyle="card" uri={normalizeURI(name)} />
               ))
             }
+            {isChannel && !channelLoaded && (
+              <div style={{height: "224px"}} />
+            )}
             {channelLoaded &&
               claimsByChannel[category][1].map(claim => (
                 <FileCard key={claim} displayStyle="card" uri={normalizeURI(`${claimsById[claim].name}#${claim}`)} />
--- a/src/renderer/component/featuredCategory/view.jsx
+++ b/src/renderer/component/featuredCategory/view.jsx
@@ -194,6 +194,9 @@ class FeaturedCategory extends React.PureComponent {
       canScrollNext,
       isChannel
     } = this.state;
+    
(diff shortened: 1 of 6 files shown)


daovist:
This issue should be just about ready now. The Discover page shows the correct data and the scrollers function properly. Just one thing...

Since the channels data comes in several seconds later than `featuredUris`, if a `featuredUri` comes before a channel in the object, then when the data comes in it will move the UI, so a user could click on the wrong target. I threw in a blank `<div>` to avoid this but it looks kinda bad. I am at a loss for the best way to deal with this.

loading channel buffer space

src/renderer/component/featuredCategory/view.jsx
```diff
@@ -0,0 +1,263 @@
...
+            </div>
+          )}
+          <div ref="rowitems" className="card-row__items">
+            {!isChannel && names.length &&
+              names.map(name => (
+                <FileCard key={name} displayStyle="card" uri={normalizeURI(name)} />
+              ))
+            }
+            {isChannel && !channelLoaded && (
+              <div style={{height: "224px"}} />
```
liamcardenas: lets find a way to not use inline styles. Is this even necessary?

src/renderer/component/featuredCategory/view.jsx
```diff
@@ -0,0 +1,263 @@
...
+            )}
+            {channelLoaded &&
+              claimsByChannel[category][1].map(claim => (
+                <FileCard key={claim} displayStyle="card" uri={normalizeURI(`${claimsById[claim].name}#${claim}`)} />
```
liamcardenas: please use `buildURI`

src/renderer/redux/actions/content.js
```diff
@@ -78,6 +78,20 @@ export function doFetchFeaturedUris() {
     });
 
     const success = ({ Uris }) => {
+      Uris["@LupoTV#5e07439df5f75de0449d7af1e115c8952d5f6dd2"] = [];
```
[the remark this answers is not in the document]
liamcardenas: obviously remove these when ready to merge

PR #1166, opened 2018-03-22
Add go to page navigation to channel pagination options

### LINKED ISSUE #711: Add "go to page" navigation to channel pagination options
## The Issue
For channels with many pages of content, the current pagination option allows you to skip forward one or two pages or click the last 2.  There is no option to go directly to a particular page. The current workaround would be to change the page number in the address bar. 

Would be a good feature to add - possibly by clicking the ellipses (need to indicate this somehow) or having a separate "Go to Page" option altogether. 

### Steps to reproduce
1. Go to lbry://@NileRed
2. Can go forward a couple pages or to the end

### Expected behaviour
Allow users to skip to any page they want

### Actual behaviour
Can skip page by page, or use address bar

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

miikkatu:
This would close #711.

The component used for pagination (react-navigate) doesn't offer customisation options, so a separate control for choosing the target page is needed. The implementation is such that the user would enter a page number and presses the enter key.

unfocused

focused

### DIFF
--- a/src/renderer/component/common/form-components/form-row.jsx
+++ b/src/renderer/component/common/form-components/form-row.jsx
@@ -4,6 +4,7 @@ import * as React from 'react';
 import classnames from 'classnames';
 
 type Props = {
+  centered?: boolean,
   children: React.Node,
   padded?: boolean,
   verticallyCentered?: boolean,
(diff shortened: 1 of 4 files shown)


seanyesmunt:
Please feel free to reach out on discord if you have any questions!

miikkatu:
I updated the PR. I agree that the suggested solution is definitely better.

I also fixed some unrelated issues in the same file that prevented committing, because of the new rules introduced after the initial PR.

The new version looks like this:

screen shot 2018-03-30 at 19 34 20

seanyesmunt:
@miikkatu Awesome! Looking a lot better. Could you just change the `paginate-channel` id to a class?

miikkatu:
That id is now a class, done!

seanyesmunt:
Sweet! I'll merge once the tests pass. @tzarebczan Will reach out with LBC payment

PR #1197, opened 2018-03-27
fix: re-enable flow type checking for the project

IGassmann:


tzarebczan:
@liamcardenas please review when you get a moment, thanks! Will follow up via DM on Slack too.

PR #1233, opened 2018-04-01
Obscure NSFW content title

miikkatu:
If NSFW content is not enabled in settings, thumbnails for NSFW content are obscured, but the file titles are not. This change hides the title information, so no (likely) NSFW text (in the title or username) is shown either. The implementation before redesign actually blurred both the image and the text.

It looks like this now:

screen shot 2018-04-01 at 13 33 56

### DIFF
--- a/src/renderer/component/fileCard/view.jsx
+++ b/src/renderer/component/fileCard/view.jsx
@@ -1,19 +1,17 @@
 // @flow
 import * as React from 'react';
 import { normalizeURI } from 'lbryURI';
+import Button from 'component/button';
 import CardMedia from 'component/cardMedia';
 import TruncatedText from 'component/common/truncated-text';
 import Icon from 'component/common/icon';
 import FilePrice from 'component/filePrice';
 import UriIndicator from 'component/uriIndicator';
-import NsfwOverlay from 'component/nsfwOverlay';
 import * as icons from 'constants/icons';
 import classnames from 'classnames';
 
 // TODO: iron these out
 type Props = {
-  isResolvingUri: boolean,
-  resolveUri: string => void,
   uri: string,
   claim: ?{ claim_id: string },
   fileInfo: ?{},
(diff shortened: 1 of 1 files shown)


Smurphonomics:
Personally, I think all **NSFW** content should be **omitted** from the home screen is disabled, rather than having a home screen full of pointless blurred spaces.

seanyesmunt:
@Smurphonomics I agree. I'm pretty sure there is an issue to add an additonal setting which would remove NSFW content from the app.

There will be two settings (worded better than what I have):

- Obscure NSFW files 
- Hide all NSFW content

Maybe instead of hiding the title all together, we can show a message? Saying this is hidden because you have a setting checked.

Also should we hide the channel name too?

@liamcardenas @tzarebczan @kauffj

seanyesmunt:
I am very much in favor of allowing you to navigate to the file page and showing the full title there (still obscuring the thumbnail). I do not like how you have to turn off the setting to view NSFW content.

miikkatu:
@kauffj Yes, the point 4. was why I made this PR. If only to prevent this happening while a more permanent solution is implemented.

kauffj:
@miikkatu Ah, I'm a bit slow sometimes (actually I was racing through email and should have taken longer to get the full context). Good and thoughtful change!

tzarebczan:
@seanyesmunt I like how it was done in the previous version of the app - I believe the channel name + entire card was blurred which alerted the user and gave them an option to go into settings. 

What do we want to go forward with in the redesign release?

seanyesmunt:
Lets hide the channel name and title, and show a message:

"This content is obscured because it is NSFW. You can change this in your settings"

There is probably a better way to word that.

miikkatu:
@seanyesmunt Your wording sounds good to me. I updated the code, and it looks like this now:

screen shot 2018-04-03 at 13 18 27

tzarebczan:
@miikkatu can we make "settings" a link to the settings area? Make it a different color so it stands out a bit.
Previous version for reference: 
nsfw

miikkatu:
@tzarebczan Yes we can 🙂

Updated screenshot:

screen shot 2018-04-03 at 21 21 53

m160:
How about having a separate NSFW tab?

When clicking the NSFW tab everything is blurred with a message saying that it can be enabled in the settings, maybe the settings can even have an option to hide the NSFW tab altogether with an option to put a password on re-enabling it.

Now the homepage is completely family friendly, in fact, the whole app is family friendly. The NSFW tab can have separate discover, trending, top bids etc.

I think this solution is beneficial for both those who are interested in NSFW and those who are not.

tzarebczan:
@m160 not the right place to discuss, but we'll enable something similar to that down the road where users will need to opt into certain types of content. For now, this behavior is staying as such.

miikkatu:
@tzarebczan Should we go with the original wording then?

seanyesmunt:
@miikkatu I think the original wording is fine for now. There is a better description in the settings page.

PR #1234, opened 2018-04-01
Unset form field input height to make it scale

### LINKED ISSUE #1228: share diagnostic data doesnt wrap to small screen
## The Issue
some word doesnt wrap on small screen , for example : if user have 1024*768 screen resulution . 

### Expected Behaviour
would be nice its supported many monitor resulution

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

## Screenshots

image

miikkatu:
This would close #1228.

The problem is that the input div height was fixed to 36px, which prevented it to wrap the containing text to additional lines. This simple change enables wrapping in practice.

It looks like this:

screen shot 2018-04-01 at 13 49 17

### DIFF
--- a/src/renderer/scss/component/_form-field.scss
+++ b/src/renderer/scss/component/_form-field.scss
@@ -41,17 +41,19 @@
 .form-field__input {
   display: flex;
   padding-top: $spacing-vertical / 3;
-  height: 36px;
 
-  input[type='checkbox'],
+  input[type='checkbox'] {
+    margin-top: 4px;
+  }
+
   input[type='radio'] {
-    margin-top: 5px;
+    margin-top: 2px;
   }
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Looks like some of the inputs are misaligned with this. (It's still better than before) I think there is a padding on the radio/checkbox that maybe isn't needed after removing the height?

miikkatu:
That's right, the alignment of some labels and checkboxes or radio buttons was a bit off. A common margin setting didn't help, so I tweaked them separately.

tzarebczan:
Thanks for your contribution @miikkatu ! Got your email for the tip, will review and follow up.

PR #1235, opened 2018-04-01
Fix black screen on macOS after maximizing LBRY and then closing

### LINKED ISSUE #1200: Mac: Black screen after maximizing LBRY and then closing
## The Issue
On Mac, if you maximize the LBRY app window and then click the close button, you end up with a black screen with no way of bringing LBRY back up. On 0.20.0, the window would just disappear without a way to bring it back up. 

Possibly related: https://github.com/electron/electron/issues/6033

### Steps to reproduce
1. Maximize Window
2. Close the window
3. Blackscreen

### Expected behaviour
Window show close to tray

### Actual behaviour
Black screen

## System Configuration

 

- LBRY Daemon version: 0.19.1
- LBRY App version: 0.21.2
- LBRY Installation ID:
- Operating system: Mac

miikkatu:
This would close #1200.

What this does is, when closing a window, wait for a full screen window to leave full screen mode first, before actually attempting t close it.

This works on macOS, but the code _has not_ been tested on other platforms.

### DIFF
--- a/src/main/createWindow.js
+++ b/src/main/createWindow.js
@@ -78,7 +78,14 @@ export default appState => {
   window.on('close', event => {
     if (!appState.isQuitting && !appState.autoUpdateAccepted) {
       event.preventDefault();
-      window.hide();
+      if (window.isFullScreen()) {
+        window.once('leave-full-screen', () => {
+          window.hide();
+        });
+        window.setFullScreen(false);
+      } else {
+        window.hide();
+      }
     }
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks for yet another contribution @miikkatu ! Got your email for the tip, will review and follow up. 

Not sure if @IGassmann tested on other OSes, but I'll make sure to give it a shot later today.

PR #1266, opened 2018-04-03
round out the amount of credit

ykris45:
image
round out the amount of credit like the top of the app


tzarebczan:
Chris, we don't want to do this. This area should show the exact amount of credits.

Issue #1034, opened 2018-02-21
[redesign] implement dark mode


seanyesmunt:
This will include cleaning up all the color variables/removing what we don't need

btzr-io:
@seanyesmunt Do you need any help with this ? 
I'm currently fixing the dark theme on `redesign` just to see how will look :eyes: 

BTW `redesign` is looking awesome :+1:

seanyesmunt:
Yeah if you are interested I would love some help with this! Colors across the `redesign` branch are missing a lot of stuff. I was planning on working on this last so feel free to make a PR to `redesign` if you would like!

You can check out http://lbry.zzk.fr/explore to get an idea of the new dark mode colors.

seanyesmunt:
@btzr-io I think currently there are a fair bit of color variables that aren't being used. Feel free to delete those/rename any that are confusing.

I know I've used color variables instead of a specific element variables as I've been working on this just to save some time. Ex:
```
.btn {
  background-color: var(--color-primary);
}
```
which should be something like
```
.btn {
  background-color: var(--button-background);
}
```

Same for `color: var(--color-black)` instead of `color: var(--text-color)`

Let me know if you have any questions!

PR #1269, opened 2018-04-03
Dark

btzr-io:
## Added
New theme: `Blueberry`

## Fixes
- Download percentage indicator overlay #1271
- Fix dark theme: #1034

### DIFF
--- a/src/renderer/scss/_gui.scss
+++ b/src/renderer/scss/_gui.scss
@@ -95,7 +95,10 @@ input {
     border: 1px dashed var(--input-copyable-border);
   }
 
-  &:not(.input-copyable):not(.wunderbar__input):not(:placeholder-shown) {
+  &:not(.input-copyable):not(.wunderbar__input):not(:placeholder-shown):not(:disabled) {
+    &:hover {
+      border-color: var(--input-hover-border-color);
+    }
     border-bottom: var(--input-border-size) solid var(--input-border-color);
   }
 
@@ -107,6 +110,7 @@ input {
 }
 
 input::placeholder {
+  color: var(--input-placeholder-color);
   opacity: 0.5;
 }
 
--- a/src/renderer/scss/component/_modal.scss
+++ b/src/renderer/scss/component/_modal.scss
@@ -22,7 +22,6 @@
   flex-direction: column;
   justify-content: center;
   align-items: center;
-
   border: var(--modal-border);
   background: var(--modal-bg);
   overflow: auto;
@@ -31,6 +30,11 @@
   box-shadow: var(--box-shadow-layer);
   max-width: var(--modal-width);
   word-break: break-word;
+
+  .btn.btn--alt {
+    // Set modal buttons bg color
+    background-color: var(--modal-btn-bg-color);
+  }
 }
 
 .modal--fullscreen {
@@ -40,7 +44,7 @@
   right: 0;
   bottom: 0;
   padding: $spacing-vertical;
-  background: var(--modal-bg);
+  background: var(--modal-fullscreen-bg);
   overflow-y: scroll;
 
   .main {
@@ -54,7 +58,7 @@
 // Slide down isn't possible without doing a lot of re-work to the modal component
 .ReactModal__Overlay {
   .modal--fullscreen {
-    transition: height var(--animation-style) var(--animation-style);
+    transition: height var(--animation-duration) var(--animation-style);
     height: 0;
   }
 
--- a/src/renderer/scss/component/_search.scss
+++ b/src/renderer/scss/component/_search.scss
@@ -31,21 +31,24 @@
   border-bottom: none;
 
   &:focus {
-    background-color: var(--color-bg);
+    background-color: var(--search-active-bg-color);
     border-radius: 0;
-    border-bottom: 1px solid var(--color-grey);
+    border-bottom: 1px solid var(--color-divider);
     box-shadow: var(--box-shadow-button);
   }
 }
 
 .wunderbar__menu {
   max-width: 100px;
+  border-radius: 0 0 3px 3px !important;
+  padding: 0 !important;
+  background: transparent !important;
   overflow-x: hidden;
 }
 
 .wunderbar__suggestion {
   padding: 5px 10px;
-  background-color: var(--header-bg);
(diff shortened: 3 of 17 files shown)


btzr-io:
Fix dark theme: https://github.com/lbryio/lbry-app/issues/1034

src/renderer/scss/_gui.scss
```diff
@@ -69,7 +69,7 @@ ul {
 }
 
 input {
-  border-bottom: var(--input-border-size) dotted var(--input-border-color);
+  border-bottom: var(--input-border-size) solid var(--input-border-color);
```
seanyesmunt: The `dashed`/`dotted` should stay the same

btzr-io:
@seanyesmunt I Added the new colors as an additional theme ->  `blueberry`, 
so now we have three themes: `light, dark and blueberry`.

btzr-io:
@seanyesmunt  Ready for the next review

tzarebczan:
@btzr-io not sure if we want to maintain 3 themes going forward, but I'll let @seanyesmunt make the call. 

Will give this a shot later tonight/tomorrow.

btzr-io:
@seanyesmunt  Ok, it's done.

src/renderer/scss/_gui.scss
```diff
@@ -92,10 +92,17 @@ input {
     background: var(--input-copyable-bg);
     color: var(--input-copyable-color);
     padding: 10px 16px;
-    border: 1px dashed var(--input-copyable-border);
+    border-bottom: 1px dashed var(--input-copyable-border);
```
seanyesmunt: This shouldn't change. It should be a border on every side

src/renderer/scss/_gui.scss
```diff
@@ -92,10 +92,17 @@ input {
...
   }
 
-  &:not(.input-copyable):not(.wunderbar__input):not(:placeholder-shown) {
+  &:not(.input-copyable):not(.wunderbar__input):not(:placeholder-shown):not(:disabled) {
+    &:hover {
+      border-color: var(--input-hover-border-color);
+    }
+    &:active,
+    &:focus {
+      border-color: var(--color-primary) !important;
```
seanyesmunt: This isn't the correct style. We shouldn't need to change anything here.

src/renderer/scss/_gui.scss
```diff
@@ -107,6 +114,7 @@ input {
 }
 
 input::placeholder {
+  color: var(--input-placeholder-color) !important;
```
seanyesmunt: We shouldn't need an `!important` here. If there is something that is overriding this style, we should remove that, or make this a more specific rule with classnames

src/renderer/scss/component/_modal.scss
```diff
@@ -18,11 +18,13 @@
 }
 
 .modal {
+  /* Set modal buttons background color */
+  --btn-bg-alt: var(--modal-btn-bg-color);
```
seanyesmunt: Do we need this? It doesn't look like it's doing anything. We should set any variables in `vars.scss`

src/renderer/scss/component/_search.scss
```diff
@@ -31,29 +31,37 @@
...
   &:focus {
-    background-color: var(--color-bg);
+    background-color: var(--search-active-bg-color);
     border-radius: 0;
-    border-bottom: 1px solid var(--color-grey);
+    border-bottom: 1px solid var(--color-divider);
     box-shadow: var(--box-shadow-button);
   }
+
+  &::-webkit-input-placeholder {
```
seanyesmunt: Can we not use `::placeholder` here?

src/renderer/scss/component/_modal.scss
```diff
@@ -31,6 +30,11 @@
   box-shadow: var(--box-shadow-layer);
   max-width: var(--modal-width);
   word-break: break-word;
+
+  .btn.btn--alt {
+    // Set modal buttons bg color
+    background-color: var(--modal-btn-bg-color);
+  }
```
btzr-io: @seanyesmunt Not sure if this is the right way to do this, but I don't have time to add a new class to all the `btn-alt` used inside the modal component :stuck_out_tongue:

src/renderer/scss/_gui.scss
btzr-io: Not sure why I did that :upside_down_face:

src/renderer/scss/component/_modal.scss
seanyesmunt: I think this is ok for now.

seanyesmunt: Well, do modals in dark mode not use the regular alt button style?

seanyesmunt:
I totally messed up the number of commits. Could you squash?

I think the only two issues I still see are that we lost the background-color for the `text-copyable` used in the address and file selector. And the "You have {lbc}" doesn't change color on hover.

src/renderer/scss/component/_modal.scss
btzr-io: @seanyesmunt yes but it's the same color as the background of the modal :upside_down_face:,
see https://design.lbry.io

seanyesmunt:
Ahh gotcha

On Apr 16, 2018 9:45 PM, "Baltazar Gomez" <notifications@github.com> wrote:

*@btzr-io* commented on this pull request.
------------------------------

In src/renderer/scss/component/_modal.scss
<https://github.com/lbryio/lbry-app/pull/1269#discussion_r181932250>:

> @@ -31,6 +30,11 @@
   box-shadow: var(--box-shadow-layer);
   max-width: var(--modal-width);
   word-break: break-word;
+
+  .btn.btn--alt {
+    // Set modal buttons bg color
+    background-color: var(--modal-btn-bg-color);
+  }

@seanyesmunt <https://github.com/seanyesmunt> yes but it's the same color
as the background of the modal 🙃,
see https://design.lbry.io

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/lbryio/lbry-app/pull/1269#discussion_r181932250>, or mute
the thread

btzr-io:
Fix conflicts: a00f4eb

PR #1286, opened 2018-04-05
add FAQ to Referral page

ykris45:
Based on @tzarebczan ideas
image


tzarebczan:
Thanks @ykris45 ! Not sure if Referrals should be capitalized or not. We should also just say "Read our FAQ to learn more".

ykris45:
image

Issue #1293, opened 2018-04-05
description field form doesn't exist on report bug section


ykris45:
## The Issue
Would be nice if have a form to submiting details of bug or issues
### Steps to Reproduce
1. Go to Help
2. click submit a bug report
3. see what happend instead my screenshot

## Screenshots

![image](https://user-images.githubusercontent.com/26609573/38384254-d82b0b18-3938-11e8-855c-8241c5d70bb1.png)
compare old design
![image](https://user-images.githubusercontent.com/26609573/38384495-808a9d14-3939-11e8-84de-b4aad30e3981.png)

miikkatu:
Working on this.

miikkatu:
A pull request was added to address this issue.

tzarebczan:
@miikatu, Sean already had this fixed in a  PR. We'll need to do a better job at marking these in the future.

PR #1301, opened 2018-04-08
Fix report page

miikkatu:
This addresses #1293.

In addition to fixing the layout, the file structure for report page was refactored to be similar to the other pages.

screen shot 2018-04-08 at 11 20 58


tzarebczan:
This is fixed in another PR already, thanks!

Issue #1284, opened 2018-04-04
"Put something here" links to claim instead of Publish page


tzarebczan:
## The Issue
When searching a URL that is unused, you are told "put something here" which links to the URL. This should go to the Publish page with the URL filled in. The old app would go to Publish, but not fill in the URL. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

miikkatu:
A pull request was added to address this issue.

tzarebczan:
Fixed in #1303

PR #1303, opened 2018-04-08
Pre-fill publish URL after clicking "Put something here" link

miikkatu:
This addresses #1284.

When "put something here" link is clicked, the URL here is first stripped of the lbry:// prefix, and then given as a navigation parameter. Then, the publish page uses this parameter to pre-fill the URL field in the form. Also, if the parameter exists, adding the actual file to be published will not overwrite the URL with the chosen file name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
    * Add referral FAQ to Invites screen([#1314](https://github.com/lbryio/lbry-app/pull/1314))
    * Show exact wallet balance on mouse hover over ([#1305](https://github.com/lbryio/lbry-app/pull/1305))
    * New dark mode ([#1269](https://github.com/lbryio/lbry-app/pull/1269))
+   * Pre-fill publish URL after clicking "Put something here" link ([#1303](https://github.com/lbryio/lbry-app/pull/1303))
 
 ### Changed
 * Add flair to snackbar ([#1313](https://github.com/lbryio/lbry-app/pull/1313))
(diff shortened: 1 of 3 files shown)


src/renderer/component/fileTile/view.jsx
```diff
@@ -101,8 +103,10 @@ class FileTile extends React.PureComponent<Props> {
                     label={__('Put something here!')}
                     onClick={e => {
                       // avoid navigating to /show from clicking on the section
-                      e.preventDefault();
-                      navigate('/publish');
+                      e.stopPropagation();
+                      // Strip prefix from the Uri and use that as navigation parameter
+                      const nameFromUri = uri.replace(/lbry:\/\//g, '').replace(/-/g, ' ');
+                      navigate('/publish', { name: nameFromUri });
```
seanyesmunt: I'm hoping to avoid using navigation params for the publish section. Instead you can use something like:
```
doClearPublish() // to remove any existing publish data
doUpdatePublishForm({ name: nameFromUri }); // to populate the name
navigate('/publish')
```

Then nothing should need to be changed inside the `publishForm/view.jsx`

miikkatu:
PR updated per review!

CHANGELOG.md
```diff
@@ -7,23 +7,23 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Added
    * Add keyboard shortcut to quit the app on Windows ([#1202](https://github.com/lbryio/lbry-app/pull/1202))
-   * Build for both architectures (x86 and x64) for Windows ([#1262](https://github.com/lbryio/lbry-app/pull/1262))
+   * Build for both architectures (x86 and x64) for Windows 
+([#1262](https://github.com/lbryio/lbry-app/pull/1262))
```
seanyesmunt: Guessing you accidentally added this

seanyesmunt:
Also looks like there are some conflicts after merging the `lbry-redux` pr's

miikkatu:
@seanyesmunt I fixed the conflicts and removed that extra line from changelog.

miikkatu:
Maybe the changelog is now ok...

PR #1304, opened 2018-04-08
Simplify Media Identification Function

ProfessorDey:
This just makes the code for finding documents a lot easier. the regex patterns and the media type are converted into keypairs, [regex, mediatype] which is then reduced to a single value, which will either be the original filename if there are no matching extensions, or the mediatype. So long as the mediatype does not contain an extension at the end of the string, this system should work far more simply.

This will also allow for easier addition of new extensions to the function, so long as they all follow the same pattern for the regex:
^ = Start of string
.+ = 1 or more characters of any type
\. = A literal dot
(a|b|c) = A list of ORed extensions within a capture group '( )'
$ = End of string, so it is certain this is the actual extension of the filename

Won't be hard to add things like Archives, Executables, etc with just a single extra line now.

### DIFF
--- a/src/renderer/lbry.js
+++ b/src/renderer/lbry.js
@@ -78,20 +78,15 @@ Lbry.getMediaType = (contentType, fileName) => {
   if (contentType) {
     return /^[^/]+/.exec(contentType)[0];
   } else if (fileName) {
-    const dotIndex = fileName.lastIndexOf('.');
-    if (dotIndex === -1) {
-      return 'unknown';
-    }
-
-    const ext = fileName.substr(dotIndex + 1);
-    if (/^mp4|m4v|webm|flv|f4v|ogv$/i.test(ext)) {
-      return 'video';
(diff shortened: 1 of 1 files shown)


tzarebczan:
@liamcardenas assigning this to you to review, but if you don't feel comfortable, let's discuss with @kauffj. 

The issue originally started as https://github.com/lbryio/lbry-app/issues/863 but was moved to the daemon: https://github.com/lbryio/lbry/issues/1028. 

It looks this PR addresses it on the app level, so I'm inclined to make it work better before a deamon solution is delivered. Also, not sure if we want any changes from: https://github.com/lbryio/lbry-app/pull/1168/files which was closed (something about showing a full content type)

tzarebczan:
@ProfessorDey can you re-base your PR? We had a large redux change go in yesterday.

ProfessorDey:
@tzarebczan See lbryio/lbry-redux PR# 7 for rebased version with codacy corrections.

Closing this now since this is moved over.

Issue #1279, opened 2018-04-04
Add flair to the snackbar to go along with the redesign


tzarebczan:
## The Issue
Snackbar stayed the same, we should add some flair to it! 

![snackbar](https://user-images.githubusercontent.com/8120721/38335588-c6f9deee-382c-11e8-9f50-8b0e6282bef5.jpg)

@nizuka3 thoughts? 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

miikkatu:
PR #1313 has a suggested implementation.

tzarebczan:
Closed via https://github.com/lbryio/lbry-app/pull/1313

PR #1313, opened 2018-04-10
Add flair to snackbar

miikkatu:
This is for #1279.

screen shot 2018-04-10 at 14 06 56

Unlike modal dialogs in the redesign, this is with a dark background to make it pop a bit more. Functionality remains the same as with the original snackbar.

The prop-types package was added as a dependency.

### DIFF
--- a/src/renderer/scss/component/_snack-bar.scss
+++ b/src/renderer/scss/component/_snack-bar.scss
@@ -1,30 +1,51 @@
 .snack-bar {
-  padding: $spacing-vertical;
-  position: fixed;
-  top: $spacing-vertical;
-  left: 0;
-  right: 0;
-  margin-left: auto;
-  margin-right: auto;
-  min-width: 300px;
-  max-width: var(--snack-bar-width);
-  background: var(--color-dark-overlay);
-  color: #f0f0f0;
+  background-color: var(--snackbar-bg-primary);
+  border-radius: 10px;
+  box-shadow: var(--box-shadow-layer);
+  color: var(--snackbar-color-primary);
   display: flex;
   justify-content: space-between;
-  align-items: center;
-  border-radius: 2px;
+  left: 0;
+  margin-left: auto;
+  margin-right: auto;
+  max-width: var(--snack-bar-width);
+  min-width: 300px;
+  opacity: 0.95;
+  padding: 14px 20px 10px 20px;
+  position: fixed;
+  right: 0;
+  top: $spacing-vertical;
   transition: all var(--transition-duration) var(--transition-type);
+  width: 100%;
   z-index: 10000; /*hack to get it over react modal */
 }
 
 .snack-bar__action {
   display: inline-block;
-  text-transform: uppercase;
-  color: var(--color-primary-light);
-  margin: 0px 0px 0px $spacing-vertical;
+  margin-bottom: 4px;
   min-width: min-content;
-  &:hover {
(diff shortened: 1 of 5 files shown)


seanyesmunt:
Nice. I will check out the code more later but the picture looks good. We are using `flow` instead of `prop-types` so your types should be moved to that.

src/renderer/scss/component/_snack-bar.scss
```diff
@@ -1,30 +1,54 @@
 .snack-bar {
-  padding: $spacing-vertical;
-  position: fixed;
-  top: $spacing-vertical;
+  background-color: var(--color-primary);
+  border-radius: 10px;
+  box-shadow: var(--box-shadow-layer);
+  color: #f0f0f0;
```
seanyesmunt: This should use a text color variable. Probably just `--text-color` or whatever we call it.

src/renderer/scss/component/_snack-bar.scss
```diff
@@ -1,30 +1,54 @@
 .snack-bar {
-  padding: $spacing-vertical;
-  position: fixed;
-  top: $spacing-vertical;
+  background-color: var(--color-primary);
```
seanyesmunt: You should create a variable for this `--snack-bar-bg` `--snack-bar-color`

So we can change them for dark mode. Similar to how we do it with buttons:
https://github.com/lbryio/lbry-app/blob/master/src/renderer/scss/_vars.scss#L93

src/renderer/scss/component/_snack-bar.scss
```diff
@@ -1,30 +1,54 @@
...
+
+  .snack-bar__action {
+    color: var(--color-primary-light);
+    display: inline-block;
+    margin-bottom: 4px;
+    min-width: min-content;
+    text-transform: uppercase;
+
+    span {
+      color: #f0f0f0;
```
seanyesmunt: This should just be a color variable. It's just the same as the color you have above so maybe this isn't needed?

miikkatu:
@seanyesmunt I updated the code as suggested in the comments, which consequently were cleared because I updated the commit and force pushed here again. 😅

src/renderer/scss/component/_snack-bar.scss
miikkatu: There's a snackbar specific variable now.

src/renderer/scss/component/_snack-bar.scss
```diff
@@ -1,30 +1,51 @@
...
 
-.snack-bar__action {
-  display: inline-block;
-  text-transform: uppercase;
-  color: var(--color-primary-light);
-  margin: 0px 0px 0px $spacing-vertical;
-  min-width: min-content;
-  &:hover {
-    text-decoration: underline;
+  .message {
```
seanyesmunt: Looks good. The only thing I would do is change this to `snack-bar__message` and move it to the top level. Then I think it's ok to merge.

miikkatu:
PR changed per review. Also, `snack-bar__action` was moved to the top level along with `snack-bar__message`.

seanyesmunt:
Awesome! Just add a changelog commit and I will merge

miikkatu:
Changelog updated.

tzarebczan:
Thanks @miikkatu! Have your contribution email, will follow up soon!

PR #1353, opened 2018-04-18
General bug fixes

seanyesmunt:
Since we moved away from keeping your address in localStorage, we need to populate one if it isn't in redux


.gitignore
```diff
@@ -3,3 +3,4 @@
 /static/daemon/lbrynet*
 /static/locales
 yarn-error.log
+package-lock.json
```
IGassmann: We're solely using `yarn` for the project since most of our building dependencies use it. Using `npm` instead can break the build. Thus, `package-lock.json` doesn't need to be ignored.

seanyesmunt: :+1: wasn't sure

PR #1361, opened 2018-04-19
feat: block black listed content from the app

IGassmann:


src/renderer/constants/action_types.js
```diff
@@ -52,6 +52,10 @@ export const SUPPORT_TRANSACTION_FAILED = 'SUPPORT_TRANSACTION_FAILED';
 // Claims
 export const FETCH_FEATURED_CONTENT_STARTED = 'FETCH_FEATURED_CONTENT_STARTED';
 export const FETCH_FEATURED_CONTENT_COMPLETED = 'FETCH_FEATURED_CONTENT_COMPLETED';
+export const FETCH_BLACK_LISTED_CONTENT_STARTED = 'FETCH_BLACK_LISTED_CONTENT_STARTED';
```
seanyesmunt: Just noting to remind myself: these need to be added to lbry-redux

src/renderer/redux/reducers/app.js
```diff
@@ -90,7 +120,7 @@ reducers[ACTIONS.AUTO_UPDATE_DOWNLOADED] = state =>
   });
 
 reducers[ACTIONS.AUTO_UPDATE_DECLINED] = state => {
-  return Object.assign({}, state, {
+  Object.assign({}, state, {
```
seanyesmunt: You need to return these since there is an opening `{`

src/renderer/redux/actions/app.js
```diff
@@ -285,6 +287,57 @@ export function doAlertError(errorList) {
...
+
+export function doBlackListedOutpointsSubscribe() {
+  return dispatch => {
+    dispatch(doFetchBlackListedOutpoints());
+    const checkUpgradeTimer = setInterval(
+      () => dispatch(doFetchBlackListedOutpoints()),
+      CHECK_BLACK_LISTED_CONTENT_INTERVAL
+    );
+    dispatch({
+      type: ACTIONS.CHECK_UPGRADE_SUBSCRIBE,
```
seanyesmunt: This isn't used anywhere

PR #1383, opened 2018-04-24
Merge dmca changes into master

seanyesmunt:
Creating a new branch because we need to move some of these changes to `lbry-redux`


tzarebczan:
@seanyesmunt can we push this to get it into the RC?

PR #1306, opened 2018-04-09
Refactor file page controls

miikkatu:
This is a suggested solution for #1272.

The controls were moved below the video horizontally. Also their style was altered to reflect the other buttons under the video.

There's an "eslint-disable react/prop-types" rule in one of the files, mainly because the object structure of some of the props was a bit complex to figure out for this PR.

Screenshots:

screen shot 2018-04-09 at 10 42 51

screen shot 2018-04-09 at 10 43 10

screen shot 2018-04-09 at 10 43 25

screen shot 2018-04-09 at 10 44 21

PR #1418, opened 2018-04-30
change that button color on content click

ykris45:
based on @tzarebczan task to me. 
Here The Interface
image
thanks


tzarebczan:
@seanyesmunt how do you feel about this color scheme for the white theme? Or do we fix the white primary button?

seanyesmunt:
I think we are going to keep the colors as is for now. If we do change them they will probably be different colors. It will need to be something that works well with the changes from #1306

Issue #1392, opened 2018-04-26
price should not wrap inside pricebox 


tzarebczan:
## The Issue
See below
![price](https://user-images.githubusercontent.com/8120721/39306236-90f0fdde-492d-11e8-968c-8412f82d9f6f.jpg)

This is on lbry://two right now

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

tzarebczan:
closed by https://github.com/lbryio/lbry-app/pull/1420

Issue #1193, opened 2018-03-27
Increase the font size of text in the price badges on the discover page


akinwale:
The text displayed in the price badges for content on the discover page appears to be too tiny and I find myself straining to read, even with glasses. It would make sense to increase the font size a bit.

tzarebczan:
Closed by https://github.com/lbryio/lbry-app/pull/1420

PR #1420, opened 2018-05-01
Price badge improvements

miikkatu:
This fixes both #1193 and #1392.

### DIFF
--- a/src/renderer/scss/_gui.scss
+++ b/src/renderer/scss/_gui.scss
@@ -252,10 +252,11 @@ p {
 }
 
 .credit-amount {
-  padding: 5px;
   border-radius: 5px;
   font-family: 'metropolis-bold';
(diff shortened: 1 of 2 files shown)


seanyesmunt:
Why the weird padding sizes? Lets just keep it at 5px

src/renderer/scss/_gui.scss
```diff
@@ -252,10 +252,11 @@ p {
 }
 
 .credit-amount {
-  padding: 5px;
```
seanyesmunt: Just keep this and change the font-size. It seems to look fine.

miikkatu: PR is updated now.

tzarebczan:
Thanks @miikkatu ! Got your email, will follow up with a tip shortly.

PR #1424, opened 2018-05-02
Internal subscriptions

seanyesmunt:
#### Notes
Ideally this should live in `lbry-redux`. It seems like there is still some work moving `lbry.call` over, mostly with auth stuff. And we need to figure out how we want to interact between app code and lbry-redux code. Currently the the subscriptions actions we are using `doPurchaseUri` which doesn't exist in `lbry-redux`. It shouldn't be too difficult to move over, but will just take some planning. I thought it was out of the scope for this PR.

#### Implementation
There are 3 main scenarios that cause some extra logic re: respecting users privacy. If a user decides not to share app usage, we will not make any calls to the db. If they are sharing data with us, this is what happens:

```
0 subs in db, but x subs in redux:
- This will happen on the initial run, or if users decide to start sharing data after they have subscribed to a channel
- In this case we will populate every sub in redux into the db


x subs in db, but 0 subs in redux:
- This will happen if a user clears there cache
- Populate all subs in db into redux
- This will cause issues with users who have subscribed, then stopped sharing data, and unsubscribed from everyone, then started sharing data again. They will be re-subscribed (I think this is ok for now)

x subs in db, y subs in redux:
- There is some mismatch from failing api calls or stopping/starting to share app usage
- Default to whats in redux
- Subscribe to any missing channels in the db
- Unsubscribe to any channels in the db that aren't in redux
```


seanyesmunt:
Currently having an issue with not displaying a notification when a subscription is downloaded, I will add that before merging once I find it, but I'm guessing it's something really small.

src/renderer/component/page/view.jsx
```diff
@@ -1,33 +1,96 @@
...
-      )}
-      {children}
-    </main>
-  );
+type State = {
+  showLoader: ?boolean,
 };
 
+class Page extends React.PureComponent<Props, State> {
+  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
```
seanyesmunt: React recommends `getDerivedStateFromProps` moving forward instead of `componentWillReceiveProps`

.eslintrc.json
```diff
@@ -34,6 +34,7 @@
       "singleQuote": true
     }],
     "func-names": ["warn", "as-needed"],
-    "jsx-a11y/label-has-for": 0
+    "jsx-a11y/label-has-for": 0,
+    "import/prefer-default-export": 0
```
seanyesmunt: IMO this isn't needed. Mostly for util files, I don't want to add the first one as a default export if I think there may be more functions added later.

src/renderer/redux/actions/subscriptions.js
```diff
@@ -10,48 +10,162 @@ import type {
...
+    })
+    .catch(() => {
+      dispatch({
+        type: ACTIONS.FETCH_MY_SUBSCRIPTIONS_FAIL,
+      });
+    });
 };
 
-export const doChannelUnsubscribe = (subscription: Subscription) => (dispatch: Dispatch) => {
+export const setSubscriptionLatest = (subscription: Subscription, uri: string) => (
```
seanyesmunt: Didn't touch this, lint was telling me to move it

src/renderer/redux/actions/subscriptions.js
```diff
@@ -114,40 +228,80 @@ export const doCheckSubscription = (subscription: Subscription, notify?: boolean
   });
 };
 
-export const setSubscriptionLatest = (subscription: Subscription, uri: string) => (
+export const setSubscriptionNotifications = (notifications: SubscriptionNotifications) => (
```
seanyesmunt: Same here, just re-ordering

src/renderer/redux/actions/subscriptions.js
```diff
@@ -114,40 +228,80 @@ export const doCheckSubscription = (subscription: Subscription, notify?: boolean
...
+    // If they aren't subscribed, we still want to remove it from the db
+    // If it doesn't exist, it will still return a success
+    const claimId = getClaimId(subscription.uri);
+    Lbryio.call('subscription', 'delete', {
+      claim_id: claimId,
+    });
+  }
+};
+
+export const doCheckSubscriptions = () => (
```
seanyesmunt: also this, just moved it

src/renderer/util/dom.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import * as React from 'react';
+
+// is a child component being rendered?
+export const isShowingChildren = (children: React.Node) => {
```
kauffj: Should indicate boolean return type

src/renderer/util/dom.js
```diff
@@ -0,0 +1,16 @@
...
+  if (Array.isArray(children)) {
+    for (let i = 0; i < children.length; i += 1) {
```
kauffj: I'm guessing this code was copy/pasted, but `children.findIndex()` would make this code both more succinct and reduce state.

src/renderer/constants/action_types.js
```diff
@@ -166,6 +166,9 @@ export const SET_SUBSCRIPTION_NOTIFICATIONS = 'SET_SUBSCRIPTION_NOTIFICATIONS';
 export const CHECK_SUBSCRIPTION_STARTED = 'CHECK_SUBSCRIPTION_STARTED';
 export const CHECK_SUBSCRIPTION_COMPLETED = 'CHECK_SUBSCRIPTION_COMPLETED';
 export const CHECK_SUBSCRIPTIONS_SUBSCRIBE = 'CHECK_SUBSCRIPTIONS_SUBSCRIBE';
+export const FETCH_MY_SUBSCRIPTIONS_START = 'FETCH_MY_SUBSCRIPTIONS_START';
```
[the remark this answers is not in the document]
seanyesmunt: Will add the comment.

Yes This should be in lbry-redux, currently there some parts of subscriptions use code that isn't in lbry-redux. We will need to decouple that, or just also move that to lbry-redux first. It's mostly dealing with starting downloads automatically.

src/renderer/page/subscriptions/view.jsx
```diff
@@ -43,69 +32,45 @@ export default class extends React.PureComponent<Props> {
...
-    const someClaimsNotLoaded = Boolean(
-      subscriptions.find(subscription => !subscription.claims.length)
-    );
+    const { subscriptions, subscriptionClaims, isFetchingSubscriptions } = this.props;
 
-    const fetchingSubscriptions =
-      !!savedSubscriptions.length &&
-      (subscriptions.length !== savedSubscriptions.length || someClaimsNotLoaded);
+    let claimList = [];
+    subscriptionClaims.forEach(claimData => {
```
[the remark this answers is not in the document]
seanyesmunt: Good point, it should just be in a selector

src/renderer/redux/actions/subscriptions.js
```diff
@@ -10,48 +10,162 @@ import type {
...
+        const newSubscriptions = dbSubscriptions.map(sub => ({
+          channelName: sub.channel_name,
+          uri: `lbry://${sub.channel_name}#${sub.claim_id}`,
+        }));
+
+        return newSubscriptions;
+      }
+
+      // There is some mismatch between redux state and db state
+      // If something is in the db, but not in redux, remove it from the db
```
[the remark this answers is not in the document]
seanyesmunt: We were thinking about if users stop and start sharing their data. But this is a much more valid case. We can just be cautious and always keep subs from the db.

src/renderer/redux/selectors/subscriptions.js
```diff
@@ -22,7 +28,7 @@ export const selectSubscriptionsFromClaims = createSelector(
     const fetchedSubscriptions = [];
 
     savedSubscriptions.forEach(subscription => {
-      let channelClaims = [];
+      const channelClaims = [];
```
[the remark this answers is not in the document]
seanyesmunt: Pretty sure lint was catching this. We can remove that rule if you think it makes sense.

src/renderer/page/subscriptions/view.jsx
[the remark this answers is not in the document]
seanyesmunt: Actually upon closer look I think it's ok if it's inside `render()`. I kept it inside this in this format so we can check if a subscription uri has any fetched claims in the `componentDidUpdate`

Since we are already passing them in this format, I don't think it makes sense to create a separate selctor to pass in the same data in a different format (I think)

src/renderer/component/fileList/view.jsx
```diff
@@ -145,6 +144,10 @@ class FileList extends React.PureComponent<Props, State> {
     const { sortBy } = this.state;
     const content = [];
 
+    if (!fileInfos) {
+      return null;
```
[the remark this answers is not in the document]
kauffj: :+1:

src/renderer/page/subscriptions/view.jsx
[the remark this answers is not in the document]
kauffj: :+1:

seanyesmunt:
@kauffj ready for more feedback

src/renderer/constants/action_types.js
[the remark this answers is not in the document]
kauffj: @seanyesmunt this is fine for now. My suggestion for that case would be to de-couple noticing new subscriptions from downloading or to add an auto-download setting that can default to a different value for desktop vs. mobile.

(@akinwale )

src/renderer/constants/settings.js
```diff
@@ -11,4 +11,6 @@ export const INSTANT_PURCHASE_ENABLED = 'instantPurchaseEnabled';
 export const INSTANT_PURCHASE_MAX = 'instantPurchaseMax';
 export const THEME = 'theme';
 export const THEMES = 'themes';
+export const DARK_THEME = 'dark';
+export const LIGHT_THEME = 'light';
```
kauffj: Currently these values are setting _keys_ not setting _values_. It's a good idea to const these theme values, but we ought to do this in a way that retains that distinction.

src/renderer/redux/actions/subscriptions.js
```diff
@@ -10,48 +10,125 @@ import type {
...
 
 const CHECK_SUBSCRIPTIONS_INTERVAL = 60 * 60 * 1000;
 const SUBSCRIPTION_DOWNLOAD_LIMIT = 1;
 
-export const doChannelSubscribe = (subscription: Subscription) => (dispatch: Dispatch) => {
-  dispatch({
-    type: ACTIONS.CHANNEL_SUBSCRIBE,
-    data: subscription,
-  });
+const getClaimId = uri => {
```
kauffj: Use `parseURI`?

src/renderer/types/claim.js
```diff
@@ -0,0 +1,30 @@
+// @flow
+
+// Actual claim type has more values than this
+// Add them as they are used
+export type Claim = {
```
kauffj: Should `Subscription` also be a standalone type file like `Claim`?

src/renderer/redux/actions/subscriptions.js
```diff
@@ -10,48 +10,125 @@ import type {
...
+              claim_id: claimId,
+              channel_name: sub.channelName,
+            });
+          }
+        });
+
+        storedSubscriptions.forEach(sub => {
+          if (!reduxSubMap[sub.claim_id]) {
+            const uri = `lbry://${sub.channel_name}#${sub.claim_id}`;
+            subscriptionsToReturn.push({ uri, channelName: sub.channel_name });
```
kauffj: I'm mildly surprised this works. If `subscriptionsToReturn` is an array of `Subscription`, doesn't this need to include the `latest` property?

src/renderer/constants/action_types.js
```diff
@@ -166,6 +176,9 @@ export const SET_SUBSCRIPTION_NOTIFICATIONS = 'SET_SUBSCRIPTION_NOTIFICATIONS';
...
+export const FETCH_MY_SUBSCRIPTIONS_FAIL = 'FETCH_MY_SUBSCRIPTIONS_FAIL';
+export const FETCH_MY_SUBSCRIPTIONS_SUCCESS = 'FETCH_MY_SUBSCRIPTIONS_SUCCESS';
```
kauffj: Can we drop `MY` from these? Or let me know what additional clarity it is providing if you think it should remain.

src/renderer/redux/actions/subscriptions.js
```diff
@@ -10,48 +10,125 @@ import type {
...
+      const storedSubscriptions = dbSubscriptions || [];
+
+      // User has no subscriptions in db or redux
+      if (!storedSubscriptions.length && (!reduxSubscriptions || !reduxSubscriptions.length)) {
+        return [];
+      }
+
+      // There is some mismatch between redux state and db state
+      // If something is in the db, but not in redux, add it to redux
+      // If something is in redux, but not in the db, add it to the db
```
kauffj: I think this is the most reasonable choice and that we can't (or shouldn't) do anything differently immediately, but I also think it's guaranteed that this will cause a bug or confusion in the future.

src/renderer/redux/actions/subscriptions.js
seanyesmunt: `latest` isn't required. It will populate it in `doCheckSubscriptions`. We are just adding them to redux, which is where it looks when it's checking for the latest.

src/renderer/types/claim.js
seanyesmunt: Yes

src/renderer/redux/actions/subscriptions.js
kauffj: I'm fairly sure this is my misunderstanding, but I still don't understand why Flow type checking wouldn't be forcing `storedSubscriptions` to be `Array<Subscription>` and thus require the `latest` property.

(Going to merge anyway, but if you know the answer would appreciate the education :) )

Issue #909, opened 2017-12-28
Move "Report" feature to inside of app?


tzarebczan:
## The Issue
The Report feature, which is used to report content for infringement, opens in a separate web page outside of the app.  

Should this be moved to inside the app eventually?

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

seanyesmunt:
I think this should definitely live inside the app. Shouldn't be too difficult.

miikkatu:
Working on this.

miikkatu:
@seanyesmunt @tzarebczan I have an implementation ready, but it's missing the actual submitting of the report. I couldn't figure out how the current web form does the submitting, and there's nothing in the app for this yet.

How this implementation works is:

1. Click the report button on a file page
2. App shows report page and prefills the uri of the file
3. Fill in the form and submit
4. App shows a confirmation with snackbar and navigates back to the file page

This can also be used by itself, for example from the sidebar. Only then the infringing file uri would not be prefilled.

<img width="514" alt="screen shot 2018-04-11 at 12 35 58" src="https://user-images.githubusercontent.com/16205520/38609736-86df3e2a-3d87-11e8-9ff0-2e9a7641a11e.png">

kauffj:
@miikkatu thanks for working on this. Unfortunately, the current end behavior of a DMCA report is just trigger an email (see [here](https://github.com/lbryio/lbry.io/blob/90eae9cbfc7a66dba9b59c51652adf06a63f0718/controller/action/ReportActions.class.php)) and this is not easy to just move into the app.

I filed an issue in our internal-apis project to build an endpoint for receiving these, but I'm afraid this one is blocked until that happens.

Although I suppose you could post to lbry.io/dmca in the interim... that would allow this to be shipped.

miikkatu:
> Although I suppose you could post to lbry.io/dmca in the interim

I was thinking about this too. For this, we would need an URL where to POST, and an object format that it would accept. I couldn't figure those out in the lbry.io/dcma code. Then, a team member could verify if the posted data is proper or not.

Otherwise, we could wait with this until there's an endpoint available. 🙂

tzarebczan:
![post](https://user-images.githubusercontent.com/8120721/38889548-0e68598e-424d-11e8-9735-e7662e5ddedf.jpg)

[...]

PR #1425, opened 2018-05-02
Report infringing content within app

miikkatu:
This is a suggested implementation for #909.

This is work-in-progress, as there is a CORS issue when posting the report, as discussed with @maximest-pierre.

### DIFF
--- a/src/renderer/component/fileActions/index.js
+++ b/src/renderer/component/fileActions/index.js
@@ -5,6 +5,7 @@ import {
   makeSelectClaimIsMine,
   doNotify,
 } from 'lbry-redux';
+import { doNavigate } from 'redux/actions/navigation';
 import FileActions from './view';
 
 const select = (state, props) => ({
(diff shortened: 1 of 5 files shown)


seanyesmunt:
This is looking good, a couple comments.

Most of this should live in redux. We should be doing api calls in action creators to keep components simple. You should also avoid mixing props and internal state like you are doing in this line:
`identifier: this.props.params.uri !== undefined ? this.props.params.uri : '',`

I propse creating a `report` reducer. Which will handle fetch errors, and loading while the api call is being made. The form values don't need to be stored in it, so `Formik` would be a good solution for this, similar to how we do it with the shapeshift component.

miikkatu:
> I propose creating a report reducer. Which will handle fetch errors, and loading while the api call is being made. The form values don't need to be stored in it, so Formik would be a good solution for this, similar to how we do it with the shapeshift component.

Looks like this would require more time than I currently have, unfortunately. Hopefully someone else can continue from here.

alyssaoc:
closing, looking at a new direction/solution

seanyesmunt:
@miikkatu We may come back to this. I will re-open if we decide to go with this solution.

PR #1431, opened 2018-05-02
Remove price placeholder during thumbnail loading

### LINKED ISSUE #1427: "PRICE" text is visible when content thumbnail has not loaded yet
## The Issue

When the app is loading and/or content thumbnails with actual price information have not loaded yet you can see the "PRICE" text in the top right corner of each content field. It just looks bad when you see a bunch of empty squares and PRICE displayed all over the screen.    

### Steps to Reproduce
1. You can see it while the thumbnails are loading in any app section 
2.
3.

### Expected Behaviour
The PRICE text should not be displayed. 

## System Configuration

- LBRY Daemon version: 0.19.1
- LBRY App version: 0.22.0-rc.3
- LBRY Installation ID: 7tYN144rHXrspzJG9xLP28fi1JztXnbpKro6DBJEQyzBmXTyDgqRLfYsUvHYFVmMe2
- Operating system: win 10

Here is the screenshot, I was to slow to catch all the fields empty. It takes some 10 seconds on my side to load up.

## Screenshots

price text

miikkatu:
This would fix #1427.

Note: the actual fix is just to return null in the `if (!costInfo)` block.

However, committing that was prevented by linter error `PropType is defined but prop is never used` for multiple props, even if those props are being used.

As a workaround and for less code, I think that it's enough to check props and call `fetchCostInfo` only in `render` function. Linter is happy this way. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Changed
    * Add flair to snackbar ([#1313](https://github.com/lbryio/lbry-app/pull/1313))
+   * Remove price placeholder during thumbnail loading ([#1431](https://github.com/lbryio/lbry-app/pull/1431))
 
 ### Fixed
    * Black screen on macOS after maximizing LBRY and then closing ([#1235](https://github.com/lbryio/lbry-app/pull/1235))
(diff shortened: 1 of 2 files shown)


miikkatu:
@seanyesmunt This was not a good solution to the problem. I noticed afterwards that calling `fetchCostInfo` in `render`causes the component state to change, and this warning to the console:

`Warning: Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.`

I tried moving the call to `componentWillMount`and `componentDidMount`, but the price was not fetched properly with either.

So, I'd suggest disabling the linter rule that prevents committing the file with only the change mentioned in the first commit. And making another PR to roll back this.

PR #1440, opened 2018-05-06
Don't allow automatic night mode when dark theme is selected

### LINKED ISSUE #1005: Don't allow dark mode with automatic night mode enabled
## The Issue
If you enable automatic night mode, the theme selection should not allow the dark theme to be chosen as it overrides the setting.  Not sure about disabling the theme selection completely - right now it makes sense because there is only 1 option, but in the future when we have more, a user may want theme X and have night mode enabled.

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

btzr-io:
### Changes 
- Fix  #1005
- Minor fixes for `changelog.md`


tzarebczan:
Thanks for this @btzr-io, will reach out on Discord regarding tip :)

Issue #584, opened 2017-09-19
Automatically play free or already purchased content


btzr-io:
### Steps to reproduce
1. Open a claim video / music ( maybe images and gifs too ).
2. Click on play / open.
3. Wait for the content...

### Expected behaviour
There should be an option in settings to toggle this feature:
I open a claim and then the video starts playing no need to wait or click to play,
same for images or gifs and music.
( content already downloaded / purchased, obviously ) ^^

Also if I left before a video ends, it should trigger autoplay but:
> The video should pick up where it left off (e.g. starting at 03:55, instead of at 00:00

### Actual behaviour
Nothing, there isn't an `autoplay`  feature :stuck_out_tongue: 

## Anything Else
I think this is also needed in other issues ,
so I'm connecting: https://github.com/lbryio/lbry-app/issues/491  https://github.com/lbryio/lbry-app/issues/579

tzarebczan:
What do we want to happen if the video costs money? Prompt the user automatically for purchase? Don't do anything? 

I know there's also a recent change to allow content under a certain LBC/USD  to skip the purchase confirmation screen... Also need to decide if those should autoplay, but my guess is no.

btzr-io:
This is my initial idea:
> ( content already downloaded / purchased, obviously ) ^^

btzr-io:
> What do we want to happen if the video costs money?

I guess this sounds right to me: `Prompt the user automatically for purchase`

> I know there's also a recent change to allow content under a certain LBC/USD to skip the purchase confirmation screen

One solution could be to force the popup just in this case ( autoplay: on )

btzr-io:
@tzarebczan I need it internally so right now the safe / simple way is only for `downloaded` content

kauffj:
I don't think we need an option or settings for this. People expect content to play.

I'd propose that as a first step, content that is free or already purchased automatically play when rendered by `FilePage`.

Additionally, if we make this change, I'd like to modify the video loading component to give more of a visual indicator of the content (since you won't see the poster if you simply changed it to autoplay).

btzr-io:
> I don't think we need an option or settings for this. People expect content to play.

Ok, but disagree with this, what happen if I click in the wrong claim could be `nsfw` or have annoying music, I think the user should have control of this ^^

btzr-io:
> #### Allow autoplay disable
>  Allow the user to disable autoplay in a menu or settings  screen.

[...]

PR #1453, opened 2018-05-07
Autoplay

daovist:
This PR implements an autoplay featured for issue #584 

Autoplay is a user setting which can be checked on/off on the settings and file pages. When on, it will play videos that have downloaded at least one blob. Otherwise, it will begin the download.

This adds a new icon and some CSS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
    * New dark mode ([#1269](https://github.com/lbryio/lbry-app/pull/1269))
    * Pre-fill publish URL after clicking "Put something here" link ([#1303](https://github.com/lbryio/lbry-app/pull/1303))
    * Add Danger JS to automate code reviews ([#1289](https://github.com/lbryio/lbry-app/pull/1289))
+   * Autoplay downloaded and free media ([#584](https://github.com/lbryio/lbry-app/pull/1453))
 
 ### Changed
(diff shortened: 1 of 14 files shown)


seanyesmunt:
What about a snackbar message? That way we would display the info without affecting a playing video.

seanyesmunt:
I'm really not a fan of having it on this page at all, because it won't do anything. If a video has already started, and they un-check it, the video will continue to play. And if a video isn't playing and they check it, it won't start playing. 

For the majority of people of who keep this setting on, it will just be an extra input on the page that they never interact with (unless by accident, and they wonder why the videos aren't playing anymore)

kauffj:
@seanyesmunt if you want to remove it you are welcome to.

I will say that I think a snackbar is wrong. Snackbars are to give feedback related to actions the user took.

seanyesmunt:
Still trying to see why it isn't playing for me. I think we can keep it. I think a better spot for it would be on the right side, below the tip/subscribe buttons.

seanyesmunt:
@daovist Ah I realized why. You have a check to not autoplay if users have `obscureNSFW` as false. I think it's ok to play content if they have that setting checked, as long as the content is not marked NSFW. And in that case, there should be some help text under the checkbox saying "we won't autoplay this because you have NSFW content hidden" (not that but similar). It is really confusing in it's current state that it isn't playing because I have some other setting checked.

Another thing I would add is having some check inside `componentWillReceiveProps` to only call `handleAutoplay` if it's necessary, so it's not being called a bunch of times when data that isn't relevant to this piece of code is changing.

Issue #738, opened 2017-11-13
Poor error message for no support/tip amount input


oakey22:
Click support a claim and then click Send without inputting an amount, you get an error of 

"Something went wrong"

This should say " Please input an amount" or something along those lines.

This is on the new 0.18.0 version

![image](https://user-images.githubusercontent.com/22527575/32748376-025edf36-c8b4-11e7-8da1-a902413e67cc.png)

tzarebczan:
Thanks @oakey22! Sent you a couple LBC for filing the issue.

I've run across this in the past, never got a chance to file it.

IGassmann:
This issue will be waiting for [that issue](https://github.com/lbryio/lbry/issues/1004). The error message should be returned by the API so it can be used in the app.

liamcardenas:
@IGassmann it looks like "that issue" is done, can you get to this soon?

liamcardenas:
oh i see this is already in "next"! nice

tzarebczan:
API call now returns, is this what was expected? 
```
{
  "error": {
    "code": -32500,
    "data": [
      "  File \"site-packages\\twisted\\internet\\defer.py\", line 1299, in _inlineCallbacks",
      "    ",
      "  File \"site-packages\\lbrynet\\daemon\\Daemon.py\", line 2719, in jsonrpc_wallet_send",
      "    ",
      "lbrynet.core.Error.NullFundsError: "
    ],
    "message": "NullFundsError"
  },
  "id": 426,
  "jsonrpc": "2.0"
}
```

tzarebczan:
I think we should just disable the send box unless something is typed. @seanyesmunt can we get this done for the redesign please?

PR #1475, opened 2018-05-14
disable Send button when tip amount is zero

daovist:
I addressed the error message part of #738 in [lbryio/lbry-redux/pull/23](https://github.com/lbryio/lbry-redux/pull/23). This PR is to prevent the common cause of that.


daovist:
Good catch. I also added a condition to disable the button when balance > amount.

seanyesmunt:
Nice

PR #1476, opened 2018-05-15
Docs and remove yarnhooks

seanyesmunt:
I removed the yarnhooks for now since they are slightly annoying.

updated readme with resetting packages section

update contributing.md with lbry-redux section

Also added a nice intro after seeing this tweet
https://twitter.com/ScribblingOn/status/995715231300050944

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,9 +1,9 @@
 # Contribute to LBRY
 
-You found this page! That means you are well on your way to contributing to the LBRY app. This
-application is primarily written in JavaScript and is built on [Electron](https://electronjs.org)
-while utilizing [React](https://reactjs.org) and [Redux](https://redux.js.org) for UI and
-application state.
+
+**First:** if you're unsure or afraid of anything, just ask or submit the issue or pull request anyways. You won't be yelled at for giving your best effort. The worst that can happen is that you'll be politely asked to change something. We appreciate any sort of contributions, and don't want a wall of rules to get in the way of that.
+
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -57,10 +57,18 @@ Although all contributions should have good UX, the [UX label, when applied in c
...
+while utilizing [React](https://reactjs.org) and [Redux](https://redux.js.org) for UI and
+application state.
+
 The project comes with diverse tools for simplifying the development process and for providing
 better code quality. It's recommended to make use of them thoroughly during ongoing development. We follow the well-known [Airbnb JavaScript Style Guide](http://airbnb.io/javascript/) for defining
 our styling rules and code best practices.
 
+### lbry-redux
+
+We have started sharing some of the redux code for this project with [lbry-android](https://github.com/lbryio/lbry-android). This shared code is in a github repo called [lbry-redux](https://github.com/lbryio/lbry-redux). We plan to move additional redux code to that repo, so you may be asked to make some of your changes there. The steps to work with lbry-redux locally can be found [here](https://github.com/lbryio/lbry-redux#local-development).
```
kauffj: While I don't think we need to go full science paper and avoid all I/we statements, I do think the language frequently comes across stronger when they are (or can be) omitted and the sentences aren't too passive/awkward. Here's a rewrite of above:

This project uses [lbry-redux](https://github.com/lbryio/lbry-redux) to share Redux code with [lbry-android](https://github.com/lbryio/lbry-android) and other LBRY apps. Over time, more Redux code that is suitable to be shared will be moved into lbry-redux. If modifying Redux state, you may be asked to make some of your changes in lbry-redux rather than lbry-app. The steps to work with lbry-redux locally can be found [here](https://github.com/lbryio/lbry-redux#local-development).

seanyesmunt: I agree this rewrite sounds better.

Issue #1433, opened 2018-05-03
Use channel id instead of name when publishing


tzarebczan:
## The Issue
In order to allow channel + claim creation in the same block, we should be using the channel id instead of the channel name when publishing with channel data. See syntax: https://lbryio.github.io/lbry/cli/#publish

See comments here: https://github.com/lbryio/lbry/issues/893#issuecomment-386050685

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

kauffj:
@tzarebczan you could also try this

tzarebczan:
@seanyesmunt @kauffj I did take a brief look into it and it would require redux changes post redesign, which means re-work required pre-redesign. @seanyesmunt let me know and I can walk you through it, maybe there's an easier way, but I didn't see the channel id as an available field (I could be mistaken though).

tzarebczan:
@daovist Maybe you want to take a look at this one? I can explain to you what's happen now vs how we want it to work.

daovist:
I am working on this issue.

tzarebczan:
Fixed in https://github.com/lbryio/lbry-app/pull/1481

PR #1481, opened 2018-05-16
Publish by channel_id

daovist:
Issue #1433 

Add `channelId` to `PublishState` and use it to replace `channel` in `PublishParams`. Store both in redux while editing but always use `channel_id` to publish.

Adds `myChannels: selectMyChannelClaims(state)` to PublishForm and uses it to get a channel's id from its name when saving form data.

@tzarebczan check this out. I am able to create a new channel and publish seconds later

lbry fast publish


tzarebczan:
@daovist - nice work! https://github.com/lbryio/lbry-redux/pull/26 should fix the channel selection issue.  I'll test this out once that's merged into redux

src/renderer/component/publishForm/view.jsx
```diff
@@ -137,13 +139,16 @@ class PublishForm extends React.PureComponent<Props> {
   }
 
   handleChannelChange(channelName: string) {
-    const { name, updatePublishForm } = this.props;
+    const { name, updatePublishForm, myChannels } = this.props;
+    const form = { channel: channelName };
+    const namedChannelClaim = myChannels.filter(channel => channel.name === channelName)[0];
+    if (namedChannelClaim) {
+      form.channelId = namedChannelClaim.claim_id;
```
seanyesmunt: This should clear out the channel id if there is no `namedChannelClaim`

If I select a channel, then go back to anonymous, the same `claim_id` will be stored in redux

seanyesmunt:
Looks like this was already a bug. I thought I fixed it but I guess not.

seanyesmunt:
Looks like `channel` isn't being correctly used to create `uri` in the publish reducer.

src/renderer/component/publishForm/view.jsx
```diff
@@ -137,13 +139,14 @@ class PublishForm extends React.PureComponent<Props> {
   }
 
   handleChannelChange(channelName: string) {
-    const { name, updatePublishForm } = this.props;
+    const { name, updatePublishForm, myChannels } = this.props;
+    const form = { channel: channelName };
+    const namedChannelClaim = myChannels.filter(channel => channel.name === channelName)[0];
```
skhameneh: use `Array.prototype.find()`? (instead of filter)

daovist: Good call @skhameneh

Issue #1470, opened 2018-05-10
Right Click Anywhere On The App Interface And You Get Cut, Copy Paste(add select all option to context menu for search bar)


ceoger:
## The Issue
i get 3 options, cut, copy and paste by right clicking on any where on the app interface, the paste pastes nothing even if there is a content on the clipboard, copy copies nothing and so does cut, how ever all three options works in the search area
### Steps to Reproduce
1. Right Click anywhere on the screen and you get these options
2. Copy or cut and nothing copies or cut,
3. If you previously copied something o your clipboard, paste in the search bar works

### Expected Behaviour
Not everywhere on the app interface should be clickable, the copy, cut paste option can be in the search bar when right clicked inside the search bar.

### Actual Behaviour
What happens everywhere in the app is right clickable, everywhere... and you only get 3 fixed options, copy, cut and paste and it's void.

### Suggested Solutions
1.) if the copy cut paste will be enforced in the search bar then i'd recommend a "select all" option as well.
2.) A right click on any of the videos can give options like copy video url or any other information on the video that copy cut paste.
3.) If there is nothing in the search, the cut and copy can be greyed out leaving the paste option visible if there is something on the clipboard otherwise also greyed. 

## System Configuration

- LBRY Daemon version: 0.19.1, 0.19.3
- LBRY App version: 0.21.4
- LBRY Installation ID:
- Operating system:

## Screenshots

https://cdn.discordapp.com/attachments/377895389992321064/444128472617451541/right2.JPG
https://cdn.discordapp.com/attachments/377895389992321064/444128483938140160/rightc.JPG

PR #1486, opened 2018-05-16
Add right click context menu on all text inputs

dan1d:
Hi this PR partially fix: #1470, this pr adds the right click menu only on text inputs and copy/cut/paste works, also if you are in development mode it will show "open developers tools".

So currently works on this two items:
1.) if the copy cut paste will be enforced in the search bar then i'd recommend a "select all" option as well.
3.) If there is nothing in the search, the cut and copy can be greyed out leaving the paste option visible if there is something on the clipboard otherwise also greyed.
screenshot from 2018-05-16 18-54-43
screenshot from 2018-05-16 18-35-57

On this one: 
2.) A right click on any of the videos can give options like copy video url or any other information on the video that copy cut paste.
I need to figure out what's the best approach to fix this one, if you have an idea, let me know!


seanyesmunt:
Thanks for the PR @dan1d! We already have some code for the context menu here: https://github.com/lbryio/lbry-app/blob/master/src/main/menu/setupContextMenu.js

That would be a better place to add this.

dan1d:
Hi @seanyesmunt , I've deleted the code under `src/menu/setupContextMenu.js` and moved it to `util/contextMenu` which has a few functions exported to open the desired menu, if it is an input then it will default to this one:
screenshot from 2018-05-20 22-07-27

If it is a video image, the react component can be customized to open a different context menu:

screenshot from 2018-05-20 22-08-31
Currently when I copy the link I get an url copied like this one: `lbry://why-i-m-bullish-but-selling-alts#19d0d7e3a3c17303d4b69fcc2b78512dd58c74db`, how I should build the url to be able to share it to others users and when they click the link the lbry-app is the one that opens that link?, is there a service to use ?

Also, could you take a look at the new code on this PR please?
Also I can add a snackbar notification when the link is copied if it is required, I've only added the copy link feature just to the home page but it can be added on a lot of others components also.

Daniel

dan1d:
Hi @seanyesmunt I've updated the code on this PR to parse an URI and added a new [PR on lbry-redux](https://github.com/lbryio/lbry-redux/pull/27), read my PR comment on lbry-redux!

So for example given this URI:
lbry://why-doesn-t-capitalism-work-for#36cd5b217bcc4083a6fc35cc4de2cc1196581ef9
it will convert it to this
https://open.lbry.io/why-doesn-t-capitalism-work-for#36cd5b217bcc4083a6fc35cc4de2cc1196581ef9
and now the link can be shared by just copying the link when right click is pressed on a video card.

seanyesmunt:
@skhameneh

PR #1490, opened 2018-05-17
More release blockers

seanyesmunt:
https://github.com/lbryio/lbry-app/issues/1358
https://github.com/lbryio/lbry-app/issues/1317
https://github.com/lbryio/lbry-app/issues/1403
https://github.com/lbryio/lbry-app/issues/1272
https://github.com/lbryio/lbry-app/issues/1388 (partially) I still need to figure out the overlfow issues on the home page, but the new tooltip is being used on the search and file page.


src/renderer/component/common/icon.jsx
```diff
@@ -1,10 +1,10 @@
 // @flow
 import React from 'react';
-// import * as icons from 'constants/icons';
 import * as FeatherIcons from 'react-feather';
 import * as icons from 'constants/icons';
 
 const RED_COLOR = '#e2495e';
```
skhameneh: Would be great to have a common color palette shared that's paired with a style guide.

src/renderer/component/common/icon.jsx
```diff
@@ -19,8 +19,10 @@ class IconComponent extends React.PureComponent<Props> {
     const Icon = FeatherIcons[icon];
 
     let color;
-    if (icon === icons.HEART || icon === icons.FEATURED) {
+    if (icon === icons.TRASH || icon === icons.FEATURED) {
```
skhameneh: This may not scale too well, also icons may render different colors based on context

src/renderer/component/fileDownloadLink/view.jsx
```diff
@@ -1,19 +1,39 @@
...
+  doPause: () => void,
+};
+
+class FileDownloadLink extends React.PureComponent<Props> {
   componentWillMount() {
     this.checkAvailability(this.props.uri);
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentWillReceiveProps(nextProps: Props) {
```
skhameneh: componentWillReceiveProps is being deprecated in favor of getDerivedStateFromProps

src/renderer/component/common/icon.jsx
seanyesmunt: I agree. Currently just copying them from our scss variables file. The values should probably live somewhere outside the app.

src/renderer/component/common/icon.jsx
seanyesmunt: 👍 I'll create a `color` prop and use that instead. Then any icon can have a color.

src/renderer/component/fileDownloadLink/view.jsx
```diff
@@ -27,12 +35,7 @@ class FileDownloadLink extends React.PureComponent {
     }
   }
 
-  checkAvailability(uri) {
```
seanyesmunt: This isn't currently doing anything. The action is disabled so I figured we can just remove it until that is fixed.

seanyesmunt:
@skhameneh Made a few more small changes

Issue #1488, opened 2018-05-16
Snackbars randomly stop working / can't delete content


tzarebczan:
## The Issue
I've run into this a few times and @daovist recently did as well: snackbar notificiations cease to work. Easiest place to test if the wallet send/receive page by sending yourself credits. Refreshing the app fixes it. Not sure how to reproduce.

Another user had reported something similar on the show page - he was unable to click the delete button (could click, but not action), until a restart. 

Sean is aware and will keep an eye out. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

kauffj:
@tzarebczan @daovist dumping relevant and/or all Redux state once this happens may be helpful to debugging this.

tzarebczan:
Found the root cause - it's the rewards snackbar that's missing display type, which causes the rest of the notifications to get backed up. Thanks @seanyesmunt for showing me how to debug this!

@daovist did you run into this during your reward changes/testing? Sent you a DM on Slack

https://www.screencast.com/t/TwAaDv1f7

tzarebczan:
closed by https://github.com/lbryio/lbry-app/pull/1493

PR #1493, opened 2018-05-18
Add display type to notifications to allow snackbar to be shown.

dan1d:
Closes issue #1488, I've searched on the entire codebase to check if there is any place where `doNotify` was being called without  `displayType` property(thanks @tzarebczan, your comment helped a lot! ).
@seanyesmunt  I wonder if we should add a default `displayType`(ex: `["snackbars"]`) in case someone forgets to add it while calling the function `doNotify`, anyway currently I've not added any default and just fixed the issue reported on #1488 

This issue on #1488 :
```
Another user had reported something similar on the show page - 
he was unable to click the delete button (could click, but not action), until a restart.
```
Was also related to the notifications being queued but not dispatched because the `displayType` was not set, I could reproduce and tested the fix.

### DIFF
--- a/src/renderer/rewards.js
+++ b/src/renderer/rewards.js
@@ -56,6 +56,7 @@ rewards.claimReward = type => {
         linkText: __('Show All'),
         linkTarget: '/rewards',
         isError: false,
+        displayType: ['snackbar'],
       });
       window.app.store.dispatch(action);
 


tzarebczan:
Thanks again for the quick response and resolution @dan1d ! I'll let @seanyesmunt comment on the defaulting scenario.

seanyesmunt:
Thanks Dan. I think a default could work, but since everything is finally moved over, we should be able to check if new doNotify  contain it in future PRs. But it's still something to think about!

kauffj:
Rather than a default would it make more sense to just error out or except if the expected/required value is not provided?

seanyesmunt:
That's pobably a better solution

PR #1501, opened 2018-05-23
fix: use short form uri when preparingEdit so users know they are editing

seanyesmunt:
Fixes https://github.com/lbryio/lbry-app/issues/1499
We store the short form of the uri being edited. That was causing an issue with properly prefilling claim edits because we were passing in the long form with a claim id. In the place that we were checking if the current uri is equal to the uri being edited, they were different so it wasn't working.

```
// check to see if you are editing
// editingUri === uri

// before
lbry://@channelName#xxx.../claim-name === lbry://@channelName/claim-name

```


seanyesmunt:
@skhameneh

Issue #1329, opened 2018-04-16
App-side changes to allow reward changes without app-side changes


kauffj:
1. Display the rewards in the order returned by api.lbry.io rather than the app-coded order.
1. Look for a reward property on an individual reward called `reward_notification` that is displayed to the user when they earn a reward.
1. Have a fall-back notification when a `reward_notificaiton` isn't provided (e.g. "You earned a %s LBC reward).

tzarebczan:
The API side changes can be seen here: https://github.com/lbryio/internal-apis/pull/294

The API now returns the description and notification text:

```
"success": true,
  "error": null,
  "data": [
    {
      "id": 73950,
      "reward_type": "new_user",
      "reward_amount": 3,
      "transaction_id": "ed633e1d1000dd748948de95dc60bfaaa3f7defd49f39ad44c2ceffc769ee615",
      "created_at": "2017-07-25T04:43:21Z",
      "updated_at": "2017-12-23T06:21:43Z",
      "reward_version": null,
      "reward_title": "Your First Nickel",
      "reward_description": "A welcome bonus and thank you for your contribution to content freedom.",
      "reward_notification": "You earned 3 LBC new user reward."
    },
    {
      "id": 77407,
      "reward_type": "first_stream",
      "reward_amount": 2,
      "transaction_id": "ebe63acdf81c604ca6c1cb8de0fd0611064aa483efcb38c506df05c66954e381",
      "created_at": "2017-07-25T18:53:41Z",
      "updated_at": "2017-12-23T06:21:43Z",
      "reward_version": null,
      "reward_title": "Go for a Stream",
      "reward_description": "Just press play on something. Anything at all. Yep, even that one. Everybody's doing it.",
      "reward_notification": "You earned 2 LBC for streaming your first video."
```

tzarebczan:
@daovist this could be another one to take a look at.  Epic is here: https://github.com/lbryio/lbry-app/issues/1328

kauffj:
@daovist this is a strong candidate to knock out. Likely only a few hours work.

@tzarebczan I'm tagging this release blocker as I don't want to have to wait another month to be able to create new rewards without waiting on app releases.

PR #1509, opened 2018-05-23
Rewards reorder and snackbar notifications

daovist:
Refactor unclaimed rewards as array to keep order from api.lbry.io. Display reward notification message through snackbar with fallback for rewards with no message.

#1329

### DIFF
--- a/src/renderer/redux/actions/rewards.js
+++ b/src/renderer/redux/actions/rewards.js
@@ -1,7 +1,7 @@
 import * as ACTIONS from 'constants/action_types';
 import Lbryio from 'lbryio';
 import { doNotify, MODALS } from 'lbry-redux';
-import { selectUnclaimedRewardsByType } from 'redux/selectors/rewards';
+import { selectUnclaimedRewards } from 'redux/selectors/rewards';
 import { selectUserIsRewardApproved } from 'redux/selectors/user';
 import rewards from 'rewards';
 
@@ -30,8 +30,8 @@ export function doRewardList() {
 export function doClaimRewardType(rewardType) {
   return (dispatch, getState) => {
     const state = getState();
-    const rewardsByType = selectUnclaimedRewardsByType(state);
-    const reward = rewardsByType[rewardType];
+    const unclaimedRewards = selectUnclaimedRewards(state);
+    const reward = unclaimedRewards.find(ur => ur.reward_type === rewardType);
     const userIsRewardApproved = selectUserIsRewardApproved(state);
 
     if (!reward || reward.transaction_id) {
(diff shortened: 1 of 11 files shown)


tzarebczan:
@daovist we should clean up: https://github.com/lbryio/lbry-app/blob/master/src/renderer/rewards.js right? Since it's no longer considering those types and reward messages.

daovist:
@tzarebczan as I cleaned up `rewards.js` I realized I was duplicating the reward notification so that is fixed now too

src/renderer/redux/actions/rewards.js
```diff
@@ -61,13 +61,6 @@ export function doClaimRewardType(rewardType) {
           reward: successReward,
         },
       });
-      if (successReward.reward_type === rewards.TYPE_NEW_USER) {
```
seanyesmunt: Why are we getting rid of this?

daovist: The `new_user` reward notification is handled by `rewards.js` like the others

src/renderer/redux/reducers/rewards.js
```diff
@@ -62,24 +62,21 @@ reducers[ACTIONS.CLAIM_REWARD_STARTED] = (state, action) => {
 
 reducers[ACTIONS.CLAIM_REWARD_SUCCESS] = (state, action) => {
   const { reward } = action.data;
+  let { unclaimedRewards } = state;
 
-  const unclaimedRewardsByType = Object.assign({}, state.unclaimedRewardsByType);
-  const existingReward = unclaimedRewardsByType[reward.reward_type];
+  const index = unclaimedRewards.findIndex(ur => ur.reward_type === reward.reward_type);
+  unclaimedRewards = unclaimedRewards.slice(0, index).concat(unclaimedRewards.slice(index + 1));
```
skhameneh: It looks like you're removing the current element?
I would suggest using splice instead.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

tzarebczan:
Giving this a test with the new user reward.

src/renderer/redux/reducers/rewards.js
skhameneh: That's likely because it's modifying the original array value and React only shallow checks.
If you return `[...unclaimedRewards]` instead of `unclaimedRewards`, it'll force the update.

This is a good use case for immutable.js as well

tzarebczan:
The last refractor broke reward.type constant comparisons which breaks first reward modal and ability to claim rewards like publish / channel. Travis looking into it.

src/renderer/redux/reducers/rewards.js
daovist: Thanks. I did the same with claimed rewards and now they're showing up at the bottom like they should.

tzarebczan:
This is working as expected again. @seanyesmunt we also removed the amount from the invite button because it was coming back as 0. It may be misleading also, because sending the invite doesn't actually get you 3 credits (the person has to sign up, get approved, etc), so I think it's better this way.

PR #1510, opened 2018-05-23
Allow https to follow redirection when requesting latest release data

### LINKED ISSUE #1505: Create fallback/fix for github api lookup
After renaming lbry-app >lbry-desktop, the github api does not redirect properly and this line fails:https://github.com/lbryio/lbry-desktop/blob/master/src/main/index.js#L172 

This causes the app to crash when going to the help page that looks up version info. On Linux, it crashes on startup. 

Create a fallback to the lbry-desktop URL if lbry-app fails.

dan1d:
Closes #1505  as it allows https to follow redirections, it checks for status 301 and 302.

I've looked through [https docs](https://nodejs.org/api/https.html) and [github https](https://github.com/nodejs/node/blob/master/lib/https.js) to see if there was an easier way or option to allow redirections, but I couldn't find anything.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -176,11 +176,11 @@ ipcMain.on('version-info-requested', () => {
     },
   };
   let result = '';
-
-  const req = https.get(Object.assign(opts, url.parse(latestReleaseAPIURL)), res => {
+  const onSuccess = res => {
     res.on('data', data => {
       result += data;
     });
+
     res.on('end', () => {
       const tagName = JSON.parse(result).tag_name;
       const [, remoteVersion] = tagName.match(/^v([\d.]+(?:-?rc\d+)?)$/);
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks is great @dan1d , thanks for taking care of it so quickly! Hopefully we can get a patch release with this fix for the old app next week :) (working out some build issues at the moment).

seanyesmunt:
@tzarebczan Do you know how we can test this without renaming?

tzarebczan:
@seanyesmunt not easily, but could stage something on a local repo - I don't think it's worth it. I'll test this today by temporarily renaming lbry-app to lbry-desktop again.

tzarebczan:
Tested and looks like it's working as expected!

PR #1511, opened 2018-05-23
special links/invalid URIs

tzarebczan:
Implements https://github.com/lbryio/lbry-app/issues/1352 - tested on Windows/Mac/Linux
Fixes https://github.com/lbryio/lbry-app/issues/959

This will also enable us to use magic links on Windows, as before the `?` was not being escaped properly. 
 Will need to update: https://github.com/lbryio/lbry.io/blob/b81191c7c5312183c25d072f3068b09fcdd1719e/view/template/acquisition/verify.php#L54

Tested magic link on Windows by setting the user agent. 

Thanks to @seanyesmunt for help with the doNotify logic :)

### DIFF
--- a/src/renderer/component/router/view.jsx
+++ b/src/renderer/component/router/view.jsx
@@ -18,16 +18,21 @@ import InvitePage from 'page/invite';
 import BackupPage from 'page/backup';
 import SubscriptionsPage from 'page/subscriptions';
 
-const route = (page, routesMap) => {
+const route = (props, page, routesMap) => {
   const component = routesMap[page];
-
-  return component || DiscoverPage;
+  if (!component) {
+    props.doNotify({
+      message: __('Invalid page requested'),
+      displayType: ['snackbar'],
+    });
+  }
+  return component || routesMap.discover;
 };
 
 const Router = props => {
   const { currentPage, params } = props;
 
-  return route(currentPage, {
+  return route(props, currentPage, {
     auth: <AuthPage params={params} />,
     backup: <BackupPage params={params} />,
     channel: <ChannelPage params={params} />,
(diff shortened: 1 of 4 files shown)


src/renderer/index.js
```diff
@@ -18,6 +18,7 @@ import app from './app';
 import analytics from './analytics';
 
 const { autoUpdater } = remote.require('electron-updater');
+const specialURL = 'lbry://?';
```
seanyesmunt: I think we should call this something else. Maybe `appPageUrl` or something. Just to give some indication of what it is

tzarebczan: good call.

src/renderer/component/router/view.jsx
```diff
@@ -18,16 +18,21 @@ import InvitePage from 'page/invite';
...
 
-const route = (page, routesMap) => {
+const route = (props, page, routesMap) => {
   const component = routesMap[page];
-
-  return component || DiscoverPage;
+  if (!component) {
+        props.doNotify({
+        message: __('Invalid page requested'),
+        displayType: ['snackbar'],
```
seanyesmunt: can you try running `yarn format`? These two lines above should be indented once more.

src/renderer/index.js
```diff
@@ -18,6 +18,7 @@ import app from './app';
 import analytics from './analytics';
 
 const { autoUpdater } = remote.require('electron-updater');
+const appPageURL = 'lbry://?';
```
seanyesmunt: Generally constants like this are in all caps `APP_PAGE_URL`

tzarebczan: fixed

src/renderer/component/router/view.jsx
tzarebczan: fixed this one, there may be some others from previous commits. If you want, run prettier on this branch please.

seanyesmunt:
@tzarebczan I rebased and squashed your commits. I'll merge once it's green :slightly_smiling_face:

seanyesmunt:
Not sure why the `push` travis build is looking at the old `.travis.yml`. The PR build passed so good to go.

tzarebczan:
I didn't update the Travis config on this build, that's why.

Issue #1457, opened 2018-05-09
expected to see the search results again after clicking the back from the app browser


ceoger:
## The Issue
after searching for a content, I sometimes multiple results, maybe i am not sure which is what i want, so i try to open the content, when i find out it's the wrong one and i try going back, it shows no results for the item i searched which i previously had results for: 
.....................
My take: when we get the search results, we should be able to go back to that same results when we click on the back rather than having to type in the same search bar again...

### Steps to Reproduce
1. Search for an item
2. Get a list of them
3. Click o any item and try to going back again and the search results will be empty

### Expected Behaviour
when we get the search results, we should be able to go back to that same results when we click on the back rather than having to type in the same search bar again...

### Actual Behaviour
Instead, it shows no results for the search
### Suggested Solutions
 when we click back, the search results should display the previous search history, or the previous results from which we clicked one out of it

## System Configuration

- LBRY Daemon version: 0. 19. 1
- LBRY App version: 	0.22.0-rc.8
- LBRY Installation ID: 	3cXLZiAWXwwJbv4fjhbXVgFtaBuUiFQM7afXmFYu5ss1chD1WCpxGUp8aW68gjeNRC
- Operating system: windows 10

tzarebczan:
Great find and suggestion, thanks for filing! Tip sent on Discord.

jsigwart:
I noticed this as well and was going to file an issue for it. There is no back button for search, which is what I'd like to see. I end up going to the home page after viewing a search result, instead of going back to the search results. Also still want more than 10 search results but that's another ticket.

daovist:
I'm on it....

tzarebczan:
This is now fixed on master.

PR #1514, opened 2018-05-24
Search page (not modal)

daovist:
This is dependent on [lbry-redux/pull/31](https://github.com/lbryio/lbry-redux/pull/31)

Addresses issue #1457 

This was a small change and seems a little too simple, but the back button works as described

The only UI change was removing `noPadding` so there may be more to do there


tzarebczan:
@daovist this is what @seanyesmunt is referring to: https://www.screencast.com/t/xIzrm0y4yoI

PR #1522, opened 2018-05-26
Updated Search Page CSS and added description

tiger5226:
Added description and changed css so the search screen is more presentable to the user.


src/renderer/scss/_vars.scss
```diff
@@ -149,8 +149,8 @@ $large-breakpoint: 1760px;
   --success-msg-bg: var(--color-green-light);
 
   /* File Tile Card */
-  --file-tile--media-height: 60px;
-  --file-tile--media-width: 94px; // 60 * (1 + 9/16)
+  --file-tile--media-height: 150px;
+  --file-tile--media-width: 200px; // 60 * (1 + 9/16)
```
seanyesmunt: This should keep the same aspect ratio
`calc(150px * (1 + 9/16))`

src/renderer/component/fileTile/view.jsx
```diff
@@ -97,16 +98,19 @@ class FileTile extends React.PureComponent<Props> {
...
               <div className="card__title--small card__title--file">
                 <TruncatedText lines={2}>{title || name}</TruncatedText>
               </div>
+              <div className="card__subtext">
+                <TruncatedText lines={3}>{description}</TruncatedText>
+              </div>
               <div className="card__subtitle">
                 {showUri ? uri : channel || __('Anonymous')}
                 {isRewardContent && <Icon icon={icons.FEATURED} />}
                 {showLocal && isDownloaded && <Icon icon={icons.LOCAL} />}
```
seanyesmunt: I think the file price should stay on the bottom. The title should be the first thing they see.

seanyesmunt: IMO it should be:
```
{title}
{channel}

{description}

{price}
```

src/renderer/scss/component/_card.scss
```diff
@@ -103,6 +103,7 @@
 .card__title--file {
   font-family: 'metropolis-bold';
   font-size: 28px;
```
seanyesmunt: Since we are making the tiles bigger, I think we can increase the title size too

src/renderer/component/fileTile/view.jsx
tiger5226: I think that makes sense. This was a special request from Tom. I don't think he was especially attached to it. So no biggie.

seanyesmunt:
@tiger5226 Were you planning on finishing this up? I can also make the changes. I do really like this change so it would be great to have for the initial release

tiger5226:
ohh crap. I saw this but forgot about. I will make the change tonight. Sorry about that.

seanyesmunt:
Just merged https://github.com/lbryio/lbry-app/pull/1514

Which just moves the search modal to an actual page with forward/back support. There are some style issues which should be fixed by this PR.

tiger5226:
ok, I will take care of this.

tiger5226:
Hi Sean,

I merged master in, but there are now more issues on the search page after doing so. When you type, then click the search row, it brings you to an empty page. I saw a commit for adjusting the search mechanics. I am not good with react or javascript so its not so simple. Also I tried a rebase and there are conflicts that need to be resolved and I am not sure I am the best person to be doing that.

PR #1525, opened 2018-05-28
Add confirmation propmt before sending a transaction.

### LINKED ISSUE #947: Add confirmation prompt before sending a transaction
## The Issue
From a UX perspective, having a confirmation before sending LBC funds would be beneficial.  Also, once we have https://github.com/lbryio/lbry/issues/1045, this would be a good screen to show what the tx fee is and the total tx amount.

dan1d:
Closes #947 
It adds the feature to prompt a modal before sending a transaction.
Maybe we can add the usd value of the transaction also?

screenshot from 2018-05-28 17-01-57
screenshot from 2018-05-28 17-02-14

### DIFF
--- a/src/renderer/component/walletSend/index.js
+++ b/src/renderer/component/walletSend/index.js
@@ -1,9 +1,9 @@
 import { connect } from 'react-redux';
-import { doSendDraftTransaction, selectBalance } from 'lbry-redux';
+import { selectBalance, doNotify } from 'lbry-redux';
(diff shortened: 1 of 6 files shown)


seanyesmunt:
This is awesome. I think if we say `"Once the transaction is sent, it cannot be reversed."` it would be a little more clear.

dan1d:
@seanyesmunt Thanks, I've updated the requested text change.

PR #1530, opened 2018-05-31
Fix claim ID on when reporting a claim.

### LINKED ISSUE #1512: claim_id param missing on url when report infringing content page opens
## The Issue
While I was downloading the video I clicked on the flag icon to report infringing content and the claim_id parameter was not in the url of the page, but in a video already downloaded when clicking on the flag icon the claim_id parameter is present on the url.

### Steps to Reproduce
1. Play a video
2. While the video still downloads click the flag icon to report infringing content
3. you will notice that the page that opens does not have a claim_id parameter set, it is like this:  `https://lbry.io/dmca?claim_id=`

### Expected Behaviour
The report page that opens should have the claim_id parameter set. Infringing Content Identifier input should be autocompleted with the claim_id parameter taken from the url.

### Actual Behaviour
It shows the flag icon during the download at the right of the video and allows you to click it, and opens a Report Infringing Content page without the claim_id parameter in the url.

### Suggested Solutions
Hide the flag icon during the download of a video, or set the claim_id parameter in the URL of the Report Infringing Content page.

## System Configuration
- LBRY Daemon version: 0.19.3
- LBRY App version: 1.8.6
- LBRY Installation ID: 2jDC7VNTmY4ojVA3rSs7Kqvv6ycVTgU56HeXu4sQGRu6QbgAzeJpAow5ysYSNwwsP2
- Operating system: elementary OS 0.4.1 Loki (debian based)

## Screenshots
Missing parameter:
screenshot from 2018-05-24 00 20 25
Working url:
screenshot from 2018-05-24 00 20 48

DaniNz:
Closes issue: #1512 properly pass claim ID to the https://lbry.io/dmca?claim_id url, as @tzarebczan told me I will continue to work on https://github.com/lbryio/lbry-app/pull/1425 but for now the fix closes #1512

### DIFF
--- a/src/renderer/component/fileActions/index.js
+++ b/src/renderer/component/fileActions/index.js
@@ -12,6 +12,7 @@ const select = (state, props) => ({
   /* availability check is disabled due to poor performance, TBD if it dies forever or requires daemon fix */
   costInfo: makeSelectCostInfoForUri(props.uri)(state),
   claimIsMine: makeSelectClaimIsMine(props.uri)(state),
+  claimId: props.claimId,
 });
 
 const perform = dispatch => ({
--- a/src/renderer/component/fileActions/view.jsx
+++ b/src/renderer/component/fileActions/view.jsx
@@ -11,6 +11,7 @@ type FileInfo = {
 
 type Props = {
   uri: string,
+  claimId: string,
   openModal: ({ id: string }, { uri: string }) => void,
   claimIsMine: boolean,
   fileInfo: FileInfo,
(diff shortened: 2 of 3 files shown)


src/renderer/component/fileActions/index.js
```diff
@@ -12,6 +12,7 @@ const select = (state, props) => ({
   /* availability check is disabled due to poor performance, TBD if it dies forever or requires daemon fix */
   costInfo: makeSelectCostInfoForUri(props.uri)(state),
   claimIsMine: makeSelectClaimIsMine(props.uri)(state),
+  claimId: props.claimId,
```
seanyesmunt: Since this is already being passed in from another component, we don't need to specify it here. This is only for selecting props from state.

tzarebczan:
Woohoo @DaniNz, thanks for the contribution - glad you made it through your first PR!  Normally we have a bot that reminds devs about updating the changelog, but it's disabled at the moment. Can you please do so at your earliest convenience? See previous entries for format...it's pretty straightforward. 

If you are interested in an LBC tip for your time, please see https://lbry.io/faq/tips 

We'll also be adding you to our developer mailing list so we can keep in touch (you can unsubscribe anytime!).

PR #1532, opened 2018-05-31
reject claim_list_by_channel when no uri is passed

daovist:
@tzarebczan was seeing `claim_list_by_channel` calls on the network with no uri param. This PR prevents that behavior.


kauffj:
@daovist this is likely a less than ideal solution since it encourages/tolerates sloppy coding.

If it makes sense to add this for pragmatic reasons, I would encourage the addition of an environment check that excepts/warns when these calls happen, rather than silent failure. This way they can at least be tracked down and eliminated.

PR #1562, opened 2018-06-07
fix reward error notification

### LINKED ISSUE #1555: Reward cannot be claimed messages when visiting rewards screen
## The Issue
If you visit the rewards page after your first stream (not featured content) , it tries to claim the Watch Many Things reward and you get a "you haven't watched enough stuff yet" message. It also tries to claim the feature content reward and you get a 2nd message "you haven't watched a featured video from this time period yet". Users will. By confused as to why they are getting this message. 

We should fail silently on this check if we perform it automatically. 

### Steps to Reproduce
1. Stream soemthing on a new account that's not featured contnet
2. Visit the rewards page

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

daovist:
Fixes #1555 @tzarebczan 

This adds an `option` parameter to `doClaimRewardType` which it uses to fail silently.

`doClaimEligiblePurchaseRewards` (called by each `doDownloadFile`) uses this option when attempting to claim `rewards.TYPE_MANY_DOWNLOADS` and `rewards.TYPE_FEATURED_DOWNLOAD`

The error modal for rewards is in the `<RewardLink>` component and displayed when `errorMessage` is defined. It was getting set in redux and sitting there until a `<RewardLink>` is rendered when a user visits the rewards page. `error` is sent `undefined` in this PR because not calling the error reducer results in the two rewards pending on the rewards page.

### DIFF
--- a/src/renderer/redux/actions/rewards.js
+++ b/src/renderer/redux/actions/rewards.js
@@ -27,7 +27,7 @@ export function doRewardList() {
   };
 }
 
-export function doClaimRewardType(rewardType) {
+export function doClaimRewardType(rewardType, options) {
   return (dispatch, getState) => {
     const state = getState();
     const unclaimedRewards = selectUnclaimedRewards(state);
@@ -73,7 +73,10 @@ export function doClaimRewardType(rewardType) {
     const failure = error => {
       dispatch({
         type: ACTIONS.CLAIM_REWARD_FAILURE,
-        data: { reward, error },
+        data: {
+          reward,
+          error: !options || !options.failSilently ? error : undefined,
+        },
       });
     };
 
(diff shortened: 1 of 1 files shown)


tzarebczan:
Tested and it looks good now - no more error when going to the rewards page after streaming something.

seanyesmunt:
I guess I'm not fully following the description of this. Why do we want this to fail silently? Shouldn't it just not fail?

kauffj:
@seanyesmunt some rewards that are checked automatically on all file starts ought to silently fail (e.g. we check for "many downloads" after every start until you've started enough).

PR #1563, opened 2018-06-07
always update costInfo on FilePage componentDidMount

daovist:
#797 @tzarebczan 

Call `fetchCostInfo` whenever `<FilePage>` mounts. All `fetchCostInfo` does is update `costInfo` based on `fileInfo`; there is no api call. So running it whenever a user visits a claim is not a problem where fetching `fileInfo` from the daemon could slow things down significantly.

I don't like this solution but refactoring everything about `costInfo` would be more involved and [there may be reasons for this design which are not apparent to me](https://en.wikipedia.org/wiki/Wikipedia:Chesterton's_fence).

### DIFF
--- a/src/renderer/page/file/view.jsx
+++ b/src/renderer/page/file/view.jsx
@@ -60,9 +60,7 @@ class FilePage extends React.Component<Props> {
       fetchFileInfo(uri);
     }
 
-    if (costInfo === undefined) {
-      fetchCostInfo(uri);
-    }
+    fetchCostInfo(uri);
 
     this.checkSubscription(this.props);
   }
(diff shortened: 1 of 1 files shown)


kauffj:
Chesterson's Fence applies very much to programming :)

The reason for this code was in the original (and likely future) model, there may be an additional cost for data separate from the cost of the content itself.

This feature was turned off since data is free in 99.9+% of cases.

It's debatable whether it's worth retaining this since we expect data fees to return or to just remove it and re-add it when they do.

(@seanyesmunt can make that call)

seanyesmunt:
I think this is ok for now. It fixes the bug and is a small enough change that it will be easy to change in the future. If there are any larger changes related to this, then we may want to go a different route. I will add a comment that links to this PR/discussion above this line.

Issue #1564, opened 2018-06-08
Show hand pointer when hovering over anywhere on video thumbnail


tzarebczan:
## The Issue
We recently re-added the ability to play a video when clicking anywhere in the thumbnail, but the hand pointer still only shows when hovering over the play button. 

I may take a shot at fixing this if no one else claims it :)

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

PR #1565, opened 2018-06-08
Adds pointer on hover for video thumbnail

DaniNz:
Show hand pointer when hovering over anywhere on video thumbnail.
Close issue #1564


tzarebczan:
@DaniNz thanks for the quick fix on this one! Please see https://lbry.io/faq/tips :)

Issue #1556, opened 2018-06-06
Make it Easy For People to Find Their Log


Invariant-Change:
Add a button to auto open log or log folder or both, so that users can send debugging information easier.

![image](https://user-images.githubusercontent.com/29914179/41013960-c1ff3f52-698b-11e8-9725-00566111aa92.png)

seanyesmunt:
I really like this idea.

Invariant-Change:
```js
<div className="card__actions">
            <Button
              button="primary"
              label={__('Open Log File')}
              icon={icons.MESSAGE}
              href="{log-location}"
            />
          </div>
```

[Code Location](https://github.com/lbryio/lbry-app/blob/d27308688b05759cea551ce8250d567a5254af6f/src/renderer/page/help/view.jsx)

kauffj:
Agreed, this is a great idea @Invariant-Change!

The button should probably be somewhere else though. Either a standalone card, or in About?

Invariant-Change:
As a separate section makes it easier to find. So I guess that is the best solution.

![image](https://user-images.githubusercontent.com/29914179/41045626-1337771c-69ec-11e8-91f7-098212bb1ce8.png)

amitnndn:
I will take this up!

I had a couple of questions/clarifications - @kauffj @seanyesmunt 

* The log files location are mentioned in [this post](https://lbry.io/faq/how-to-find-lbry-log-file) and I am assuming that this is the right file. 
* I know that the electron shell plugin allows you to open external files that are on the file system, but it cannot open files with relative paths. Does anyone know how I can access properties like the user's home directory etc.?

tzarebczan:
@amitnndn take a look at how we do this on the backup page: https://github.com/lbryio/lbry-app/blob/afaad1910e84d27140abaa696af53a985349c53d/src/renderer/page/backup/view.jsx

In this case, the setting is: `data_dir`

amitnndn:
@tzarebczan Thank you! I will take a look at it.

amitnndn:
#1600 I see that there is an issue opened, that is related to the changes in this issue. Can anybody with a Windows 10 system, click on the open log folder and let me know what folder it ends up opening? Thank you!

tzarebczan:
On windows 10 and Mac, it opens to the parent folder of lbrynet, with it highlighted. Think you want to pass the logfile and have it open that folder.

amitnndn:
[...]

Issue #741, opened 2017-11-14
content address extends outside of visible area in card


seanyesmunt:
## The Issue
When going to an address that is very long, content may extend passed the edge of the card

### Steps to reproduce
1. Navigate to lbry://@seanyesmunt#2fa8ecba445648eae50cb6e98948e67a417c9017/thisisaverylongstringthatiscutoff
2. Notice content extending off the page
3. Publish new content
4. Go to the new content page before the publish is done, see it extend off the page

### Expected behaviour
I should be able to see the entire address
Long addresses should probably wrap, maybe there is a better way, does it need to be in a card?
Since the address is in the search bar, do we need the address here?

### Actual behaviour
The address is cut off

## Screenshots
<img width="400" alt="screen shot 2017-11-14 at 9 00 37 am" src="https://user-images.githubusercontent.com/16882830/32783603-514f4732-c91a-11e7-87c2-223d5d7780c0.png">

<img width="400" alt="screen shot 2017-11-14 at 9 01 56 am" src="https://user-images.githubusercontent.com/16882830/32783695-a78ce33e-c91a-11e7-92cd-7160ad60ab0e.png">

amitnndn:
Since this is a small fix, I am fixing this as a part of  #1569 PR.

tzarebczan:
Closing per #1569

PR #1569, opened 2018-06-09
Adding ability to view logs and open log directory

amitnndn:
#1556 - Adding ability to open logs and view log directory on the help page.
#741 -  Fixing content address extending outside of the visible area in the card. (I am including this fix as a part of this PR since it is a small relatively small fix - 3 lines)

### Changes
* Made changes to the code to include a new section to giving users the option to open the log file or the directory that contains the log file. 
* Added some type declarations since eslint and flow checkers were failing on the file that I edited. 
* Added a word wrap while to card's ```h1``` element so that long URLs do not extend outside the visible area.

### Preview 
#1556 
screen shot 2018-06-09 at 6 30 04 pm

#741 
screen shot 2018-06-10 at 5 15 46 pm

P.S: I have tested this only on macOS Sierra and theoretically should work on other operating systems as well. Let me know if it doesn't!

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -20,5 +20,6 @@ module.name_mapper='^lbry\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/lbry\1'
 module.name_mapper='^rewards\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/rewards\1'
 module.name_mapper='^modal\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/modal\1'
 module.name_mapper='^app\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/app\1'
+module.name_mapper='^native\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/native\1'
 
 [strict]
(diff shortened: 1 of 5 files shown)


tzarebczan:
@amitnndn thanks for the contribution! We'll get this reviewed shortly. 

Similar to last time, please see https://lbry.io/faq/tips if you'd like a tip for your contribution :)

amitnndn:
Thanks, @tzarebczan! I fixed another very simple bug in the app and included that as a part of this PR. Did not want to go through the trouble of creating another PR for such a simple fix. Let me know if that is alright!

.flowconfig
```diff
@@ -20,5 +20,6 @@ module.name_mapper='^lbry\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/lbry\1'
 module.name_mapper='^rewards\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/rewards\1'
 module.name_mapper='^modal\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/modal\1'
 module.name_mapper='^app\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/app\1'
+module.name_mapper='^native\(.*\)$' -> '<PROJECT_ROOT>/src/renderer/native\1'
```
seanyesmunt: Is there a reason this was added?

amitnndn: @seanyesmunt This was to resolve a flow error that was on [in line](https://github.com/amitnndn/lbry-app/blob/aa493f997bd99dd9ef620484bf3585d6604b4d0f/src/renderer/page/help/view.jsx#L6)

PR #1574, opened 2018-06-11
[WIP] Fix edit issues

seanyesmunt:


src/renderer/redux/selectors/publish.js
```diff
@@ -25,3 +25,49 @@ export const selectPendingPublish = uri =>
...
+  // ex: going from a channel to anonymous, the new uri won't return contentName, so we need to use claimName
+  if (!currentIsChannel && editIsChannel) {
+    return currentClaimName === editContentName;
+  } else if (currentIsChannel && !editIsChannel) {
+    return currentContentName === editClaimName;
+  } else if (!currentIsChannel && !editIsChannel) {
+    return currentClaimName === editClaimName;
+  } else {
+    return currentContentName === editContentName;
+  }
```
seanyesmunt: I don't like the above code, is there a better way to do this?

seanyesmunt: Ensures that users are still editing if they have the same claim name, but different channels (anonymous => channel)

src/renderer/redux/selectors/publish.js
```diff
@@ -25,3 +48,54 @@ export const selectPendingPublish = uri =>
...
+  // ex: going from a channel to anonymous, the new uri won't return contentName, so we need to use claimName
+  if (!currentIsChannel && editIsChannel) {
+    return currentClaimName === editContentName;
+  } else if (currentIsChannel && !editIsChannel) {
+    return currentContentName === editClaimName;
+  } else if (!currentIsChannel && !editIsChannel) {
+    return currentClaimName === editClaimName;
+  }
+  return currentContentName === editContentName;
+});
```
seanyesmunt: Not really a fan of the above code. Can this be simplified?
Ensures that users are still editing if they have the same claim name, but different channels (anonymous => channel)

kauffj: Is this a code golf challenge? Here's my entry:

```
const currentName = currentIsChannel ? currentContentName : currentClaimName;
const editName = editIsChannel ? editContentName : editClaimName;
return currentName === editName;
```

Alternatively (and possibly contentiously):
```
return (currentIsChannel ? currentContentName : currentClaimName) === (editIsChannel ? editContentName : editClaimName);
```

seanyesmunt: Much cleaner. Nice.

seanyesmunt:
@kauffj Realized I left the `WIP` tag on here. This is ready for a full review.

package.json
```diff
@@ -48,7 +48,7 @@
     "formik": "^0.10.4",
     "hast-util-sanitize": "^1.1.2",
     "keytar": "^4.2.1",
-    "lbry-redux": "lbryio/lbry-redux#7759bc6e8c482bed173d1f10aee6f6f9a439a15a",
+    "lbry-redux": "lbryio/lbry-redux#121ba56f47fff05531e27a7c99d7d417e79cd3ee",
```
kauffj: Not blocking - but what's it look like to package this properly? Does it make dev cycle more work / more annoying? Possibly worth filing.

src/renderer/component/publishForm/view.jsx
```diff
@@ -227,6 +229,7 @@ class PublishForm extends React.PureComponent<Props> {
       price,
       uri,
       channel,
+      isStillEditing
```
kauffj: It looks like the code in this section could be more succinct by just directly copying the desired properties from `this.props` directly to `publishParams` (that don't need to be modified/checked).

src/renderer/page/publish/index.js
```diff
@@ -18,34 +15,40 @@ import {
...
-  if (claimForUri) {
-    winningBidForClaimUri = claimForUri.effective_amount;
-    myClaimForUri = myClaims.find(claim => claim.name === name);
+  if (!myClaimForUri) {
+    // if the uri isn't from a users claim, find the winning bid needed for the vanity url
+    // in the future we may want to display this on users claims
+    // ex: "you own this, for 5 more lbc you will win this claim"
+    const claimsByUri = selectClaimsByUri(state);
+    claimForUri = claimsByUri[uri];
+    winningBidForClaimUri = claimForUri ? claimForUri.effective_amount : undefined;
```
kauffj: `null` seems more correct than `undefined` - we know there's not one; we're not implying it will be set later

src/renderer/redux/actions/publish.js
```diff
@@ -258,38 +248,36 @@ export const doPublish = (params: PublishParams) => (dispatch: Dispatch, getStat
...
+
+      claims.forEach(claim => {
+        if (pendingPublishMap[claim.name]) {
+          dispatch({
+            type: ACTIONS.REMOVE_PENDING_PUBLISH,
+            data: {
+              name: claim.name,
+            },
+          });
+          dispatch({
```
kauffj: Why dispatch this for every item in the loop?

src/renderer/redux/actions/publish.js
```diff
@@ -258,38 +248,36 @@ export const doPublish = (params: PublishParams) => (dispatch: Dispatch, getStat
...
-        });
-
+      const pendingPublishMap = {};
+      pendingPublishes.forEach(({ name }) => {
+        pendingPublishMap[name] = name;
+      });
+
+      claims.forEach(claim => {
+        if (pendingPublishMap[claim.name]) {
+          dispatch({
```
kauffj: Should these dispatches be grouped and bulk dispatched?

src/renderer/redux/actions/publish.js
```diff
@@ -258,38 +248,36 @@ export const doPublish = (params: PublishParams) => (dispatch: Dispatch, getStat
...
-                claims,
-              },
-            });
-
-            delete pendingPublishMap[claim.name];
-          }
-        });
-
+      const pendingPublishMap = {};
+      pendingPublishes.forEach(({ name }) => {
```
kauffj: `reduce` candidate if you want to sharpen those skills

src/renderer/redux/selectors/publish.js
```diff
@@ -8,6 +13,24 @@ export const selectPendingPublishes = createSelector(
...
 
+export const selectClaimsWithPendingPublishes = createSelector(
+  selectMyClaimsWithoutChannels,
+  selectPendingPublishes,
+  (claims, pendingPublishes) => {
+    // ensure there are no duplicates, they are being checked for in a setInterval
+    // no need to wait for it to complete though
+    // loop through myclaims
+    // if a claim has the same name as one in pendingPublish, remove it from pending
+    const claimMap = {};
```
kauffj: Another `reduce` candidate

src/renderer/redux/selectors/publish.js
```diff
@@ -25,3 +48,54 @@ export const selectPendingPublish = uri =>
...
+      // They clicked "edit" from the file page
+      // They haven't changed the channel/name after clicking edit
+      // Get the claim so they can edit without re-uploading a new file
+      myClaimForUri = claimsById[editClaimId];
+    } else {
+      // Check if they have a previous claim based on the channel/name
+      myClaimForUri = myClaims.find(claim => claim.name === currentContentName);
+    }
+
+    return myClaimForUri;
```
kauffj: Could be `return isStillEditing ? <> : <>` if that's allowed.

tzarebczan:
The UX is not changing. I gave this a run through found 2 issues:
1) got into a state where claim_list_mine kept being called in network tab
2) looks like checking of existing claims while editing broke - if I type a claim name, or select a file which populates claim name, that I already have a claim for.  Should see the "You already have a claim at lbry://...".  I think making "Publishing will update your existing claim." stick out more is a good idea too - different color or add a warning?

package.json
seanyesmunt: Not really sure. It's pretty simple right now, just build `lbry-redux` push the new build and grab the latest commit. I would think it would be more work to actually package it.

I think that could be valuable in the future, but while we are adding lots of stuff, I think this is working fine.

src/renderer/redux/actions/publish.js
seanyesmunt: We shouldn't be doing that.

PR #1582, opened 2018-06-12
add togglable QR code to wallet address component

daovist:


tzarebczan:
I'm not crazy about how the button looks on both light and dark themes. Maybe @seanyesmunt can work on styling it better. 

Also, doesn't work in dark mode - looks like some of the background is seeping through and adding some extra tiny lines to the QR. 
qr code

seanyesmunt:
I agree @tzarebczan. There was some styling to ensure the qr code had a white border so that it worked with dark mode. This should probably use the `button="link"` style.

daovist:
updated to `button="link"`

tzarebczan:
Working on dark mode now!

Issue #1579, opened 2018-06-11
Pass app version information to internal APIs


tzarebczan:
## The Issue

`lbryio.authenticate` should be modified so that after a successful authentication is completed, additional system details are sent to a new endpoint.

- The new endpoint will be `/install/new` (and thus can be called `lbryio.call("install", "new")`). 
- It should be sent the following properties:
  - `app_id` (this is already sent to `/user/new`)
  - `app_version` (from package.json)
  - `daemon_version` (FROM THE DAEMON NOT FROM PACKAGE.JSON)
  - `operating_system` (Windows, Android, etc.)
  - `platform` (Windows 10, macOS 10.12, etc.)

Only the first two values, `app_id` and `app_version` are necessary to get this merged. If there is any trickiness with the later 3 values, please PR this without them.

Note: When running in dev mode, the version doesn't show correctly on the about page. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

seanyesmunt:
Is this actually a release blocker?

PR #1583, opened 2018-06-12
Pass app version info to internal APIs

daovist:
#1579 

Edit: This passes the platform returned by `Lbry.version` which for me is `Linux-4.15.0-23-generic-x86_64-with-Ubuntu-18.04-bionic`


src/renderer/redux/actions/user.js
```diff
@@ -49,6 +66,7 @@ export function doAuthenticate() {
         });
         dispatch(doRewardList());
         dispatch(doFetchInviteStatus());
+        dispatch(doInstallNew());
```
seanyesmunt: Since we aren't storing this in the store, we don't need to wrap this call in a `dispatch`. You can just call `doInstallNew()` after.

PR #1586, opened 2018-06-13
Thumbnail edit improvements

seanyesmunt:
https://github.com/lbryio/lbry-app/issues/1584

### DIFF
--- a/src/renderer/component/publishForm/view.jsx
+++ b/src/renderer/component/publishForm/view.jsx
@@ -363,10 +363,16 @@ class PublishForm extends React.PureComponent<Props> {
           <section className="card card--section">
             <div className="card__title">{__('Thumbnail')}</div>
             <div className="card__subtitle">
-              {__(
-                'Upload your thumbnail to spee.ch, or enter the url manually. Learn more about spee.ch '
+              {uploadThumbnailStatus === STATUSES.API_DOWN ? (
+                __('Enter a url for your thumbnail.')
+              ) : (
+                <React.Fragment>
+                  {__(
+                    'Upload your thumbnail to spee.ch, or enter the url manually. Learn more about spee.ch '
+                  )}
+                  <Button button="link" label={__('here')} href="https://spee.ch/about" />.
+                </React.Fragment>
               )}
-              <Button button="link" label={__('here')} href="https://spee.ch/about" />.
             </div>
             <SelectThumbnail
               thumbnailPath={thumbnailPath}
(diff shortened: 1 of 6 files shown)


tzarebczan:
This looks good functionally according to my testing!

src/renderer/component/publishForm/view.jsx
```diff
@@ -1,6 +1,6 @@
 // @flow
 import * as React from 'react';
-import { isNameValid, buildURI, regexInvalidURI } from 'lbry-redux';
+import { isNameValid, buildURI, regexInvalidURI, STATUSES } from 'lbry-redux';
```
kauffj: what kind of statuses are these?

src/renderer/component/publishForm/view.jsx
```diff
@@ -555,7 +564,12 @@ class PublishForm extends React.PureComponent<Props> {
           </section>
 
           <div className="card__actions">
-            <Submit label={submitLabel} disabled={formDisabled || !formValid || publishing} />
+            <Submit
+              label={submitLabel}
+              disabled={
+                formDisabled || !formValid || uploadThumbnailStatus === STATUSES.IN_PROGRESS
```
kauffj: This design forces the user to wait for the upload to finish before clicking Publish and it may not be clear why they're waiting.

It would be preferable if Publish could be clicked regardless of the upload status and instead the app waited for the remote upload to finish before making the `publish` API call.

(Fixing this may be tricky, and if it's complicated probably shouldn't be considered blocking.)

src/renderer/component/publishForm/view.jsx
```diff
@@ -360,10 +363,16 @@ class PublishForm extends React.PureComponent<Props> {
           <section className="card card--section">
             <div className="card__title">{__('Thumbnail')}</div>
             <div className="card__subtitle">
-              {__(
-                'Upload your thumbnail to spee.ch, or enter the url manually. Learn more about spee.ch '
+              {uploadThumbnailStatus === STATUSES.API_DOWN ? (
```
kauffj: How feasible (and beneficial) would it be to split the entirety of this into a separate component?

seanyesmunt: Not hard. We could just move this into the `SelectThumbnail` component.

src/renderer/component/publishForm/view.jsx
seanyesmunt: spee.ch api status'. Agreed we should probably rename.

src/renderer/component/publishForm/view.jsx
seanyesmunt: Until we move add this functionality somewhere else I think it's ok right now

Issue #1283, opened 2018-04-04
Add loading screen to Downloads/Published area


tzarebczan:
## The Issue
If you navigate to the downloads/published area before the app can load file_list or claim_list_mine, you just see "Nothing has been downloaded/published". It should show a loading dialog.

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

tzarebczan:
Happening subscription page too.

tzarebczan:
Sean added a loading indicator (which could use some text such as "loading..." or "fetching..."), but when you clear app cache and go to subscriptions, it's still missing.

tzarebczan:
This LGTM!

PR #1593, opened 2018-06-13
handle fetching state on downloads and publishes pages

daovist:
This adds a `fetching` prop to each of the two pages, `selectIsFetchingFileList` for downloads and `selectIsFetchingClaimListMine` for publishes.

This one is hard to test because you need to download lots of content so that your downloads page takes a while to load. I haven't gotten there with publishes but am making this PR for @tzarebczan and @seanyesmunt to take a look at.

#1283


seanyesmunt:
We should use the `loading` prop on the `<Page />` component here instead. I know it's a little funky since it is doing some special stuff for the subscription page. I'm planning on working on that soon so we can add this after that component is cleaned up.

seanyesmunt:
@daovist Sorry this has been sitting for so long. Hoping to improve the subscriptions page so we can use this on the publishes/downloads page.

package.json
```diff
@@ -89,9 +89,6 @@
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
     "babel-preset-stage-2": "^6.18.0",
-    "danger": "^3.6.0",
```
seanyesmunt: This was causing issues with Flow for me. We aren't using danger-ci atm so we can remove it (hopefully just for now. it would be nice to get it back)

src/renderer/component/page/view.jsx
```diff
@@ -49,16 +31,29 @@ class Page extends React.PureComponent<Props, State> {
...
+    const { loading } = this.props;
+    if (!this.loaderTimeout && !prevProps.loading && loading) {
+      this.beginLoadingTimeout();
+    }
   }
 
   componentWillUnmount() {
-    this.loaderTimeout = null;
+    if (this.loaderTimeout) {
+      this.loaderTimeout = null;
```
skhameneh: This will only set the handle to null, you'll want to use `clearTimeout` in some parts of the flow.

seanyesmunt: Whoops

PR #1594, opened 2018-06-15
send lbry_id for user me call

tiger5226:
@seanyesmunt I added some information to the `user/me` api call so we can link the node_id and the user_id for rewards. Can you double check that I did this correctly? I think there may be something wrong because the app is not authenticating properly.


tiger5226:
actually wrong area. disregard.

PR #1595, opened 2018-06-15
added lbry_id to install/new api call.

tiger5226:
Also made some search fixes as they were broken from previous merge.


seanyesmunt:
Lets just keep this to the id change. Or update since there were some other css changes. If the aspect ratio is wrong we can change that but I think 150px is to big for the thumbnail height

seanyesmunt:
Removed the css changes just so I can merge this now. I will add them here https://github.com/lbryio/lbry-app/pull/1629

PR #1597, opened 2018-06-15
Design Bug in light theme

iamuditpatel:
Background Color and Text-selection-color is same so when we select text for copy or something..

For example in white background white text became invisible..

To fix this problem i change the selection text color to red so when anyone select the text in light theme they see text in red color when they select the text..

design bug

### DIFF
--- a/static/themes/light.css
+++ b/static/themes/light.css
@@ -1,1 +1,6 @@
-:root {}
+:root {
+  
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Thanks for the PR @iamuditpatel 

This has been fixed in the redesign on `master`

PR #1608, opened 2018-06-15
More css fixes

seanyesmunt:
Very minor CSS changes, mostly dealing with thumbnail placeholders in different parts of the app.

Also added `react-toggle` to style the autoplay switch. In the future, we will move all checkboxes over to this but there were some alignment issues so I didn't want to bother with it right now.
https://github.com/lbryio/lbry-app/issues/1539
screen shot 2018-06-15 at 4 11 12 pm

### DIFF
--- a/src/renderer/component/common/form-components/form-field.jsx
+++ b/src/renderer/component/common/form-components/form-field.jsx
@@ -5,6 +5,7 @@ import classnames from 'classnames';
 import MarkdownPreview from 'component/common/markdown-preview';
 import SimpleMDE from 'react-simplemde-editor';
 import 'simplemde/dist/simplemde.min.css';
+import Toggle from 'react-toggle';
 
 type Props = {
   name: string,
@@ -21,6 +22,7 @@ type Props = {
   children?: React.Node,
   stretch?: boolean,
   affixClass?: string, // class applied to prefix/postfix label
+  useToggle?: boolean,
 };
 
 export class FormField extends React.PureComponent<Props> {
@@ -37,6 +39,7 @@ export class FormField extends React.PureComponent<Props> {
       children,
       stretch,
       affixClass,
+      useToggle,
       ...inputProps
     } = this.props;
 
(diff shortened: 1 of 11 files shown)


src/renderer/component/common/form-components/form-field.jsx
```diff
@@ -68,6 +71,8 @@ export class FormField extends React.PureComponent<Props> {
         );
       } else if (type === 'textarea') {
         input = <textarea type={type} id={name} {...inputProps} />;
+      } else if (type === 'checkbox' && useToggle) {
+        input = <Toggle id={name} {...inputProps} />;
```
kauffj: There are only 11 checkboxes in the entire app. Is it worth making this all of them?

Also, does Toggle maintain the keyboard usability of native inputs? If we're replacing native form inputs with custom ones, I'd prefer they function the same as native inputs for accessibility and keyboard use. 

In this case, that'd probably mean that you can still tab to them and that space toggles the value.

src/renderer/component/common/tooltip.jsx
```diff
@@ -8,6 +8,7 @@ type Props = {
   children: ?React.Node,
   icon: ?boolean,
   direction: string,
+  noPadding?: boolean,
```
kauffj: Is there a name for this property that would describe the _type_ of tooltip this is rather than the styling of it?

src/renderer/component/common/tooltip.jsx
```diff
@@ -29,6 +30,7 @@ class ToolTip extends React.PureComponent<Props> {
           'tooltip--right': direction === 'right',
           'tooltip--bottom': direction === 'bottom',
           'tooltip--left': direction === 'left',
+          'tooltip--no-padding': noPadding
```
kauffj: Same for this class as the above comment.

src/renderer/scss/component/_tooltip.scss
```diff
@@ -5,14 +5,18 @@
...
-  font-size: 12px;
+  font-size: 14px;
   padding-left: $spacing-vertical * 1/3;
 
   .tooltip__body {
     margin-top: 5px;
   }
 }
 
+.tooltip.tooltip--no-padding {
```
kauffj: These `.tooltip` rules can be nested since it's SCSS

src/renderer/component/common/form-components/form-field.jsx
seanyesmunt: It was just one checkbox on the settings page that was messed up due to the postfix overlflowing onto a second line. Didn't spend much time on it but yes, all of the checkboxes will use this style (which supports nice usability)

PR #1632, opened 2018-06-18
Revert "fixed isPlayable to only play for video or audio."

seanyesmunt:
Reverts lbryio/lbry-app#1623

@tiger5226 reverting this because it's causing issues with images not being playable. I think the fix for that would be pretty easy but I'm not sure off the top of my head.


btzr-io:
@seanyesmunt  I can fix this in: https://github.com/lbryio/lbry-app/pull/1576

tiger5226:
Thanks @btzr-io I just talked to tom. I think the name isPlayable threw me for a loop. I thought that would only be video or audio. We should probably rename that to something like canOpen or isLaunchable. Sorry about that.

PR #1639, opened 2018-06-19
Added config setting to say how many search results you want to see

tiger5226:


kauffj:
@tiger5226 can this be placed somewhere on the search page itself, rather than the settings page?

tiger5226:
Yeah, I could do that. However, I did this with the intention of being able to give people an opportunity to see more search results for the new release. I am admittedly slow when it comes to react/redux/js. This took me almost 5 hours to accomplish(excluding the lbry-redux changes). I chose this place because it was the easiest to extend while being able to deliver the end goal of more search results. Also it was important not to force the result size onto others, so there is a default configuration. I have a more in depth thing I am working on for pagination on the search page. That should make the page result size clickable like 5,10,25,50,100 there too. 

Putting it on the search page, feels like a much bigger lift, because all of the settings information is not available there, like storing, retrieving, etc. Also the HTML and CSS structure is not there and has to be reworked a bit. Maybe this is much simpler that I am thinking. Others with more experience would know better. 

Where did you anticipate this input being?

screen shot 2018-06-19 at 7 50 07 pm

Also the user won't have access to it without searching first.

seanyesmunt:
Yeah I think I like this being on the search page more. The code shouldn't be any different, just in a different place.

This should probably be on the top right, similar to the filter select on transactions/published/downloads pages.


I will increase the search results size just so we can get that in for tomorrow.

PR #1641, opened 2018-06-19
Fix Linux upgrade path

tzarebczan:
And show manual install message


tzarebczan:
Something similar to this, but cleaned up the text, $ and `: https://www.screencast.com/t/vMxNOHhh8J

PR #1645, opened 2018-06-19
throttle fetchFeaturedUris

daovist:
#1048 

Throttles the `fetchFeaturedUris` action call in the `componentWillMount` lifecycle hook of the `<DiscoverPage>` component such that it is only called at most once every 30 minutes.

### DIFF
--- a/src/renderer/page/discover/view.jsx
+++ b/src/renderer/page/discover/view.jsx
@@ -9,9 +9,17 @@ type Props = {
   featuredUris: {},
 };
 
+let featuredUrisThrottled = false;
+
 class DiscoverPage extends React.PureComponent<Props> {
   componentWillMount() {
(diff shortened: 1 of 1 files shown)


kauffj:
What was the motivation behind this? 

(I'm wondering if this is masking a deeper issue.)

tzarebczan:
So we can support channel definitions (already merged) on internal-apis and have the app automatically update the homepage when new content arrives. Maybe there's a better way to do it rather than refreshing the homepage?

kauffj:
@tzarebczan I might be misunderstanding, but I believe this reduces the frequency with which the homepage is fetched.

daovist:
I think I misunderstood the issue. It wants to load new content every couple hours in case they stay on the homepage?

PR #1646, opened 2018-06-19
set 1hr interval on fetchFeaturedUris

daovist:
#1048 

Pattern copied from [/component/shapeShift/internal/active-shift.jsx](https://github.com/lbryio/lbry-app/blob/master/src/renderer/component/shapeShift/internal/active-shift.jsx#L26)


kauffj:
I would like to briefly discuss this. @seanyesmunt should be included to, so I'll plan to join app standup tomorrow rather than discuss in the office.

Issue #1636, opened 2018-06-18
Adjust Resend verification email button on welcome prompt


tzarebczan:
## The Issue
This looks funky on the welcome prompt because of the not now button, but looks fine with just 2 buttons on rewards entry screen for email.

![buttonssss](https://user-images.githubusercontent.com/8120721/41557680-63cfb148-730c-11e8-8502-99e43ea44f67.jpg)

### Steps to Reproduce
1. Start new user flow

### Expected Behaviour
Tell us what should happen.

### Actual Behaviour
Tell us what happens instead.

### Suggested Solutions
Tell us your suggested solutions if you have any.

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

Issue #1598, opened 2018-06-15
Styling Fixes


Invariant-Change:
**Open File** button is not `inline` with the other buttons creating unnecessary **whitespace** and looks broken.

![image](https://user-images.githubusercontent.com/29914179/41453172-05423cbe-70b8-11e8-8a6a-ad9afe613a78.png)

![image](https://user-images.githubusercontent.com/29914179/41453523-6c755b4a-70b9-11e8-8d38-75ee6658ba91.png)

![image](https://user-images.githubusercontent.com/29914179/41453532-712bc016-70b9-11e8-82d9-3ceee3a4eb6b.png)

A few minor changes, including fixing the auto-play tick-box and you end up with:

![image](https://user-images.githubusercontent.com/29914179/41454342-645b2b6c-70bc-11e8-84d4-b96d7649f38d.png)

Invariant-Change:
If I get time this weekend, I'll give it a go.

seanyesmunt:
@Invariant-Change I'm a fan of this. A couple of thoughts:

- We should keep the border-radius so we have a consistent button style.
- I think the subscription button should still stay on the right side, it is related to the channel, not the file, so I don't think it should be grouped with the file actions
- You can hold off the autoplay styling. It will be improved here https://github.com/lbryio/lbry-app/pull/1599

seanyesmunt:
One of the main reasons for keeping the "Download/Open" button on it's own line was because it loses the button as it's being downloaded. It may look fine inline, but we might need to alter it's styling while the file is being downloaded.

<img width="251" alt="screen shot 2018-06-15 at 3 15 53 am" src="https://user-images.githubusercontent.com/16882830/41455373-efa71068-704a-11e8-93bc-843457984be1.png">

Invariant-Change:
* **Q) We should keep the border-radius so we have a consistent button style.**
* A) Yes, consistency paramount. I was playing around with theme ideas. The app could do with some more. I've got some great ideas in this department, which I am working on in my secret basement in groom lake, hehe!

* **Q) I think the subscription button should still stay on the right side, it is related to the channel, not the file, so I don't think it should be grouped with the file actions**
* A) Yes. However, it currently looks like it's there because the code is broken. It needs some love so that it's meant to be there.

* **Q) You can hold off the autoplay styling. It will be improved here #1599**
* A) Sure. IMO, a nice toggle button would work well.

[...]

PR #1650, opened 2018-06-20
UX changes

seanyesmunt:
### Changes
- Simplified search autocomplete. Tried to make it easier to notifiy users if they are going to a claim/channel/search page
- #1598 
- #1636 
- Removed auto search, we may add it back but it did not need to be on the `FileList` component, that was when we had the search modal.
- Other minor fixes accross the app

### Images
screen shot 2018-06-19 at 11 38 14 pm

screen shot 2018-06-20 at 3 17 34 pm

screen shot 2018-06-20 at 3 17 45 pm

### DIFF
--- a/src/renderer/component/common/icon.jsx
+++ b/src/renderer/component/common/icon.jsx
@@ -60,7 +60,7 @@ class IconComponent extends React.PureComponent<Props> {
     }
     const inner = <Icon size={size} className="icon" color={color} />;
 
-    return tooltip ? (
+    return tooltip && tooltipText ? (
       <Tooltip icon body={tooltipText} direction={tooltip}>
         {inner}
       </Tooltip>
--- a/src/renderer/component/common/icon.jsx
+++ b/src/renderer/component/common/icon.jsx
@@ -60,7 +60,7 @@ class IconComponent extends React.PureComponent<Props> {
     }
     const inner = <Icon size={size} className="icon" color={color} />;
 
-    return tooltip && tooltipText ? (
+    return tooltipText ? (
       <Tooltip icon body={tooltipText} direction={tooltip}>
         {inner}
       </Tooltip>
--- a/src/renderer/component/fileActions/view.jsx
+++ b/src/renderer/component/fileActions/view.jsx
@@ -1,9 +1,9 @@
 // @flow
-import React from 'react';
+import * as React from 'react';
 import Button from 'component/button';
 import { MODALS } from 'lbry-redux';
-import classnames from 'classnames';
 import * as icons from 'constants/icons';
+import Tooltip from 'component/common/tooltip';
 
 type FileInfo = {
   claim_id: string,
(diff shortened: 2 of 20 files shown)


seanyesmunt:
@tzarebczan a quick test on this would be great

src/renderer/component/common/icon.jsx
```diff
@@ -60,7 +60,7 @@ class IconComponent extends React.PureComponent<Props> {
     }
     const inner = <Icon size={size} className="icon" color={color} />;
 
-    return tooltip ? (
+    return tooltip && tooltipText ? (
```
kauffj: first check no longer adds anything, this can just check for `tooltipText`

src/renderer/component/fileActions/view.jsx
```diff
@@ -15,34 +15,35 @@ type Props = {
...
+          </Tooltip>
         )}
         {!claimIsMine && (
-          <Button
-            button="alt"
-            icon={icons.REPORT}
-            href={`https://lbry.io/dmca?claim_id=${claimId}`}
-            label={__('Report content')}
-          />
+          <Tooltip onComponent body={__('Report content')}>
```
kauffj: Not a blocker, but it'd probably be superior if tooltip could be a property of a button, rather than a component that wraps a button.

src/renderer/scss/component/_tooltip.scss
```diff
@@ -13,43 +13,57 @@
...
 }
 
-.tooltip--top .tooltip__body {
-  bottom: 100%;
-  left: 50%;
-  margin-left: -100px;
+.tooltip--top {
+  .tooltip__body {
+    bottom: 100%;
+    left: 50%;
```
kauffj: The margins on `<main>` appear to make this safe for realistic use cases, but what would happen if a tooltip is on a button with a width that  would extend past the canvas size?

kauffj:
@seanyesmunt there's also a growing number of instances where components are being passed flags that effectively require them to know about what's outside of them rather than operate independently. E.g. `onModal`, `onComponent`. We ought to avoid this pattern whenever possible.

kauffj:
And here's my UX/UI feedback:

- Autoplay has got to change it's place in the layout. It probably shouldn't be it's own row.
- Spacing is weird. Here are the relative spaces I measured:
  - 36px between "published on..." and the channel indicator
  - 32px between the card action /button row and the channel indicator
  - 24px between autoplay and action row
  - 24px between "About" and autoplay
- Some of the above may stem from the usage of `form-row` in an inappropriate manner.
- Would still like to consider other icons than a paper plane for send
- I dislike that the first action in the row is "Unsubscribe" once I've subscribed, but I suppose that might be unavoidable.
- Why is download faded?
- View on web did not work for this URL (lbry://joe-haldeman-from-the-vietnam-war-to#33ad70ec22e757c828d934f571880b42f6cedb77), the first and only one I tried it on. Not sure where to file this. I did not debug this enough to figure it out.

src/renderer/component/fileActions/view.jsx
seanyesmunt: Ooh that's a good idea.

src/renderer/scss/component/_tooltip.scss
seanyesmunt: Tooltips can extend past the margin with no issues

PR #1661, opened 2018-06-21
Additional UX fixes

seanyesmunt:
#### Changes
Some changes based on comments [here](https://github.com/lbryio/lbry-app/pull/1650#issuecomment-398911888)
- Brought back the sticky header. Based on @Invariant-Change's design
- Send tip icon is now a present
- Change the download icon color, purple looked on light mode.
  - Green looks better, but doesn't stand out as much as black. Maybe we should just use black on light mode?
- Slightly improve spacing on file page (you are right, there are some kinks to be worked out with form row/form elements that are causing inconsistent spacing)
  - Will revist, but better than what we currently have

#### Images
screen shot 2018-06-25 at 2 21 11 am
screen shot 2018-06-25 at 2 24 16 am
screen shot 2018-06-25 at 2 26 30 am
screen shot 2018-06-25 at 2 48 53 am

screen shot 2018-06-25 at 2 08 15 am
screen shot 2018-06-25 at 2 08 46 am

### DIFF
--- a/src/renderer/scss/component/_header.scss
+++ b/src/renderer/scss/component/_header.scss
@@ -1,17 +1,37 @@
 .header {
+  position: fixed;
+  height: var(--header-height);
+  width: 100%;
   display: flex;
   z-index: 1;
   justify-content: space-between;
-  padding: $spacing-width $spacing-width 0 $spacing-width;
+  align-items: center;
+  padding: 0 $spacing-width;
   background-color: var(--color-bg);
+  box-shadow: var(--box-shadow-header);
+}
+
+.header__navigation {
+  display: flex;
+  justify-content: space-between;
+}
+
+.header__history {
+  display: flex;
+  padding: 0 $spacing-width * 1/2;
+
+  @media only screen and (min-width: $medium-breakpoint) {
+    .btn {
+      padding: 0 $spacing-width * 1/6;
+    }
+  }
 }
 
 .header__actions-right {
   margin-left: auto;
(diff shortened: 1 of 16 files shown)


src/renderer/scss/component/_header.scss
```diff
@@ -1,17 +1,37 @@
 .header {
+  position: fixed;
```
kauffj: Now that the header is fixed, should the address bar stretch to take the remaining space available?

src/renderer/scss/component/_header.scss
```diff
@@ -1,17 +1,37 @@
...
+.header__navigation {
+  display: flex;
+  justify-content: space-between;
+}
+
+.header__history {
+  display: flex;
+  padding: 0 $spacing-width * 1/2;
+
+  @media only screen and (min-width: $medium-breakpoint) {
```
kauffj: Is there a reason we set rules to only screen?

src/renderer/scss/component/_header.scss
seanyesmunt: yes

src/renderer/scss/component/_header.scss
seanyesmunt: I guess not since screen will be the only one used.

Invariant-Change: `Media Types` are **optional** and if **not** used it defaults to `All`. Unless targeting a **printer** or **screen reader**, there is no need for them.

Issue #1502, opened 2018-05-23
Navigation button resulting incorrect page designation


ceoger:
## The Issue
 back navigation button resulting incorrect page designation
### Steps to Reproduce
1. go to brightsunfilms channel
2.  click page 11
3. click back button
4. It shows page 1  but highlights page 11.

### Expected Behaviour
Back should take you to previous page and highlight page number

### Actual Behaviour
It shows page 1  but highlights page 11. 
same for other pages, be on page 11, click page 10, click back and it takes you back to page 11 with pag 10 still highlighted

### Suggested Solutions
Back should take you to previous page and highlight page number

## System Configuration

- LBRY Daemon version: 0.19.1
- LBRY App version: 	0.22.0-rc.8
- LBRY Installation ID: 3cXLZiAWXwwJbv4fjhbXVgFtaBuUiFQM7afXmFYu5ss1chD1WCpxGUp8aW68gjeNRC
- Operating system: Windows (Windows-10-10.0.16299)

## Anything Else

https://cdn.discordapp.com/attachments/377895389992321064/447849109462843402/1.JPG
https://cdn.discordapp.com/attachments/377895389992321064/447849121521336320/2.JPG

tzarebczan:
Can confirm this is a bug pre-redesign as well. 

@ceoger please upload those images directly into the ticket, otherwise the links may get taken down by Discord. 

Thanks again for filing, sent you some LBC on Discord!

ceoger:
@tzarebczan there is no channel pages in the predesign... i double checked

tiger5226:
@tzarebczan this was fixed recently with another PR. I can't find it though. I tested and it appear to be working fine ( the back button ) in the redesign. Can you confirm?

tzarebczan:
@tiger5226 just tried it on 0.20rc10 and it still occurs.

tiger5226:
Ohh sorry wrong issue. The search page back button wasn't working and is now. This issue is different, my apologies.

btzr-io:
Fixed and merged in master.

PR #1665, opened 2018-06-21
Fix incorrect page designation

btzr-io:
### Changes 
 Navigation button resulting incorrect page designation #1502 

> We can know that if the forcePage number still the same, the page will re-render, but won't change. So when you do your search function, you should reset the forcePage number.
https://github.com/AdeleD/react-paginate/issues/198#issuecomment-381465298


seanyesmunt:
Also added the cursor not being correct in some cases. Nice work 👍

Issue #1175, opened 2018-03-24
Shouldn't allow to open multiple export and choose file dialogs


namujan78:
## The Issue
User can open many "Choose file" windows to select files to publish
Same for "Choose directory" in Settings page
Same for "EXPORT" in History page (transaction history) and Overview page

### Steps to reproduce
1. Click on "PUBLISH" button
2. Click on "CHOOSE FILE" button ==> "Choose file" window will display
3. Click on "CHOOSE FILE" button again ==> another "Choose file" window will display
4. Repeat step 3 many time ==> many "Choose file" windows will display

### Expected behaviour
User should open only one "Choose file" windows, and it will be in top selection, User cannot focus on the Publish page after selected file success or close the "Choose file" windows

### Actual behaviour
User can open many "Choose file" windows by clicking on "Choose file" button.
Same for "Choose directory" in Settings page
Same for "EXPORT" in History page (transaction history)

## System Configuration
- LBRY Daemon version: 0.19.1
- LBRY App version: 0.21.2
- LBRY Installation ID: N/A
- Operating system: Windows (Windows-10-10.0.16299)
- Wallet (lbryum): 3.2.0

## Anything Else

Same for "Choose directory" in Settings page
Same for "EXPORT" in History page (transaction history) and Overview page

## Screenshots

N/A

tzarebczan:
@namujan78 thanks for the detailed write-up! I've sent you a tip on Discord :)

btzr-io:
Fixed and merged in master.

namujan78:
@tzarebczan This issue still happens when you quickly click multiple times on CHOOSE FILE after the PUBLISH screen appeared.
CC @btzr-io 
![image](https://user-images.githubusercontent.com/37721875/42723471-89758d56-8788-11e8-97e6-14158d74bba0.png)

btzr-io:
I can confirm this is working on `linux`, not sure about `macOS`,
Maybe this is a `windows-bug` ?

tzarebczan:
He's clicking it really quickly... I'll try on windows. This is an edge case though, someone won't click like that :)

btzr-io:
Clicking really quick freezes the window:
 https://github.com/electron/electron/issues/11323

namujan78:
It’s not freeze, but the app become slow responsive...

namujan78:
One more thing, it happen only for first time you go to Publish app.
Next time it doesn’t happen. Maybe the app load complete all components successfull.

[...]

PR #1673, opened 2018-06-21
Fix multiple dialog windows bug

btzr-io:
### Changes 
- Fix shouldn't allow to open multiple export and choose file dialogs #1175


tzarebczan:
nice one @btzr-io !

PR #1679, opened 2018-06-24
Spell Checker CSS Style Changes

Invariant-Change:
These changes follow **typical** styling rules for **spelling errors** in **most** applications which means people are **familiar** with the **markup** which avoids **confusion**. The changes also mean that the incorrect spelling style is more **visible** especially in the **Dark theme** where it was almost impossible to see.

image


Invariant-Change:
On a related but separate note. The spell checking functionality is currently **disabled** and for some unknown reason, it **incorrectly** highlights some **correctly** spelt words such as "spelt" ironically. 😂

PR #1680, opened 2018-06-24
Description Correction For Community Top Bids Tooltip.

Invariant-Change:
Corrections on the `tooltip` detailing that you can bid on **one to ten**, not **one to five**.

image


Invariant-Change:
I also shortened the text. Too much text is usually ignored and can look cluttered.

PR #1685, opened 2018-06-25
use spee.ch publish config endpoint for thumbnail status

seanyesmunt:


src/renderer/redux/actions/publish.js
```diff
@@ -43,17 +43,22 @@ export const doResetThumbnailStatus = () => (dispatch: Dispatch): PromiseAction
     },
   });
 
-  return fetch('https://spee.ch/api/channel/availability/@testing')
-    .then(() =>
-      dispatch({
+  return fetch('https://spee.ch/api/config/site/publishing')
+    .then(res => res.json())
+    .then(status => {
+      if (!status.disabled) {
```
billbitt: I think this should be `if (status.disabled) {` because you want to throw the error in the case where disabled is `true`.  i.e. if it _is_ disabled, _then_ api is down

seanyesmunt: whoops. Used that when I was testing

PR #1692, opened 2018-06-26
Tourniquets Text Over 20VW to Fix Table Length Issue

Invariant-Change:
Fixes table length issue https://github.com/lbryio/lbry-app/issues/1690

### BEFORE

image

### AFTER

image

### DIFF
--- a/src/renderer/scss/component/_button.scss
+++ b/src/renderer/scss/component/_button.scss
@@ -66,6 +66,11 @@ button:disabled {
   min-width: 0;
   box-shadow: none;
   text-align: left;
+/*Tourniquets text over 400px*/
+  max-width: 400px;
+  white-space: nowrap;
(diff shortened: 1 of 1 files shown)


Invariant-Change:
Using Viewport Width (vw) makes it responsive, thus it works at all sizes without out the need for conditional media queries.

seanyesmunt:
This is great.

Invariant-Change:
Thanks @seanyesmunt I was looking for it to **auto tourniquet** based on the current `%width`, but it would only work with a set `VW` (viewport width).

I will spend more time on this at a later date and play around with things like converting the current % values to VW to see if that is a better solution that will be globally acceptable for any sized area, which I think was @kauffj's concern.

For now, these are a good fix with the current layout.

PR #1694, opened 2018-06-26
Tourniquets Text to Fix Table Length Issue in Help

Invariant-Change:
Same **CSS** issue and fix as https://github.com/lbryio/lbry-app/pull/1692 which fixes tables issues when the app is at minimum width.

These little details help keep the app looking professional, thus helping brand reputation/authority.

### BEFORE

image

### AFTER

image

### DIFF
--- a/src/renderer/scss/component/_table.scss
+++ b/src/renderer/scss/component/_table.scss
@@ -13,11 +13,6 @@ table.table,
   tr td:last-of-type,
   tr th:last-of-type {
     padding-right: $spacing-vertical * 2/3;
-/*Tourniquets text over 20VW*/
-    max-width: 20vw;
-    white-space: nowrap;
-    overflow: hidden;
-    text-overflow: ellipsis;
   }
 
   thead {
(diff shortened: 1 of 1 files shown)


src/renderer/scss/component/_table.scss
```diff
@@ -13,6 +13,11 @@ table.table,
   tr td:last-of-type,
   tr th:last-of-type {
     padding-right: $spacing-vertical * 2/3;
+/*Tourniquets text over 20VW*/
```
kauffj: This probably can't become a property of all final table cells.

Invariant-Change:
Was better placed in the HELP section of the CSS. This way, it's only affecting the HELP table and not global assets.

```css
table.table--help {
  td:nth-of-type(1) {
    color: var(--text-color);
    font-family: 'metropolis-semibold';
    min-width: 130px;
  }
    td:nth-of-type(2) {
/*Tourniquets text over 20VW*/
    max-width: 20vw;
    overflow: hidden;
    white-space: nowrap;
    text-overflow: ellipsis;
  }
}
```

PR #1697, opened 2018-06-26
minor fix url or Uppercase

ykris45:


Invariant-Change:
@ykris45 You missed the other Url from the issues I posted in Discord. Here is the fix for that one. https://github.com/lbryio/lbry-app/pull/1704

image

PR #1716, opened 2018-06-28
Publish fixes

seanyesmunt:
#### Fixes
Disabled the file input when `publishing === true`
Fix an issue with form error styling at the bottom of the publish form
https://github.com/lbryio/lbry-app/issues/1714
https://github.com/lbryio/lbry-app/issues/1647

Moved button styling to be specific to the transactions table because it was affecting buttons it shouldn't.

### DIFF
--- a/src/renderer/component/publishForm/view.jsx
+++ b/src/renderer/component/publishForm/view.jsx
@@ -251,23 +251,15 @@ class PublishForm extends React.PureComponent<Props> {
   renderFormErrors() {
     const { name, nameError, title, bid, bidError, tosAccepted } = this.props;
 
-    if (nameError || bidError) {
-      // There will be inline errors if either of these exist
-      // These are just extra help at the bottom of the screen
-      // There could be multiple bid errors, so just duplicate it at the bottom
-      return (
-        <div className="card__subtitle form-field__error">
-          {nameError && <div>{__('The URL you created is not valid.')}</div>}
-          {bidError && <div>{bidError}</div>}
-        </div>
-      );
-    }
-
+    // These are extra help
+    // If there is an error it will be presented as an inline error as well
     return (
-      <div className="card__content card__subtitle card__subtitle--block form-field__error">
+      <div className="card__content card__subtitle form-field__error">
         {!title && <div>{__('A title is required')}</div>}
         {!name && <div>{__('A URL is required')}</div>}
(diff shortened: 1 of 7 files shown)


tzarebczan:
Tested and working as expected for me. When I edit the name, I get the warning about selecting a file. Publish works after going down that route. Also able to change the name back and edit as expected.

src/renderer/component/publishForm/view.jsx
```diff
@@ -237,32 +237,54 @@ class PublishForm extends React.PureComponent<Props> {
...
+    // These are extra help
+    // If there is an error it will be presented as an inline error as well
     return (
-      <div className="card__content card__subtitle card__subtitle--block form-field__error">
-        {!title && <div>{__('A title is required')}</div>}
-        {!name && <div>{__('A URL is required')}</div>}
-        {!bid && <div>{__('A bid amount is required')}</div>}
-        {!tosAccepted && <div>{__('You must agree to the terms of service')}</div>}
-      </div>
+      !isFormValid && (
```
kauffj: This is fine, but it's worth considering in the future how at least some of these could be funneled directly to the fields rather than all shown in a global area.

src/renderer/scss/component/_table.scss
```diff
@@ -99,6 +99,15 @@ table.table--transactions {
   }
   td:nth-of-type(3) {
     width: 22.5%;
+
+    .btn--link {
```
kauffj: I like this tourniquitting CSS but I dislike how it is being applied. Is tourniqueting text really just a property of button links when they're inside the 3rd `td` element of a `<table.table--help>`? This seems unlikely.

If we're struggling to find a clear global rule for these, I think it makes more sense to apply turniquitting as either a table or cell level class property.

kauffj: Also, this breaks BEM

src/renderer/component/publishForm/view.jsx
seanyesmunt: They are shown in the field, this is just extra help at the bottom of the page so it isn't confusing why the `publish` button is greyed out.

kauffj: :+1:

kauffj: (On second thought, this still feels somewhat wrong. Maybe the superior pattern is one that provides a link to scroll to the fields with errors? But obviously non-blocking)

PR #1725, opened 2018-07-02
take previous bid into account for total available to deposit

seanyesmunt:
https://github.com/lbryio/lbry-app/issues/1712

### DIFF
--- a/src/renderer/component/publishForm/view.jsx
+++ b/src/renderer/component/publishForm/view.jsx
@@ -152,14 +152,21 @@ class PublishForm extends React.PureComponent<Props> {
   }
 
   handleBidChange(bid: number) {
-    const { balance, updatePublishForm } = this.props;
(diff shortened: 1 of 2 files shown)


src/renderer/component/publishForm/view.jsx
```diff
@@ -152,14 +152,21 @@ class PublishForm extends React.PureComponent<Props> {
...
   handleBidChange(bid: number) {
-    const { balance, updatePublishForm } = this.props;
+    const { balance, updatePublishForm, myClaimForUri } = this.props;
+
+    let previousBidAmount = 0;
+    if (myClaimForUri) {
+      previousBidAmount = myClaimForUri.amount;
+    }
+
+    const totalAvailableBidAmount = previousBidAmount + balance;
```
kauffj: This logic would conceivably be better placed in a selector, but otherwise LGTM

PR #1726, opened 2018-07-02
clear lbry-redux cache on preinstall

seanyesmunt:
https://github.com/lbryio/lbry-app/issues/1634

Now clears the `lbry-redux` cache from yarn on `yarn` installs


kauffj:
:clap: :clap: :clap:

PR #1739, opened 2018-07-08
Handle LBC unavailable error from ShapeShift without breaking the app

seanyesmunt:
If LBC is unavailable on shapeshift (like it is right now) the app breaks on the Get Credits page. We need to explicitly check that it is available before worrying about other coins availability.

If it's unavailable, stop and show an error message.

### DIFF
--- a/src/renderer/constants/shape_shift.js
+++ b/src/renderer/constants/shape_shift.js
@@ -1,3 +1,5 @@
 export const NO_DEPOSITS = 'no_deposits';
 export const RECEIVED = 'received';
 export const COMPLETE = 'complete';
+export const AVAILABLE = 'available';
+export const UNAVAILABLE = 'unavailable';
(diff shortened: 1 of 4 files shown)


src/renderer/constants/shape_shift.js
```diff
@@ -1,3 +1,5 @@
 export const NO_DEPOSITS = 'no_deposits';
 export const RECEIVED = 'received';
 export const COMPLETE = 'complete';
+export const AVAILABLE = 'available';
+export const UNAVAILABLE = 'unavailable';
```
kauffj: :+1: good stuff

src/renderer/redux/actions/shape_shift.js
```diff
@@ -65,9 +66,16 @@ export const shapeShiftInit = () => (dispatch: Dispatch): ThunkAction => {
   return shapeShift
     .coinsAsync()
     .then(coinData => {
+      if (coinData.LBC.status === SHAPESHIFT_STATUSES.UNAVAILABLE) {
+        dispatch({
```
kauffj: `return dispatch` ?

PR #1748, opened 2018-07-10
Actually hide NSFW content

seanyesmunt:
Dependent upon https://github.com/lbryio/lbry-redux/pull/53

Currently causes issues with the community top bids section. That will be fixed in a future PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "formik": "^0.10.4",
     "hast-util-sanitize": "^1.1.2",
     "keytar": "^4.2.1",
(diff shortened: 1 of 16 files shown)


tzarebczan:
This LGTM on the channel and search pages. Gave @seanyesmunt a heads up about subs page - if you are subbed to a channel that has NSFW (and have NSFW off), there's no indication on the sub page that they are hidden.

tzarebczan:
Actually, found an edge case when my own content is marked as NSFW and going to the channel page - says result is hidden, but it's not. We should not hide it if it's our own content, right? So just need to suppress the message. 
https://www.screencast.com/t/vmUvZLTkuPv

Issue #1255, opened 2018-04-02
Channel page claim caching / other optimizations


tzarebczan:
## The Issue
When you go into a channel page likelbry://@lumbridgecity, the page starts fetching content which takes a bit of time. If you enter a claim or go to another page, and then go back, it re-fetches content again making this a slow process. We should try to add some type of caching to reduce the subsequent load times. Have to take into account new/updated claims as well. 

Update: The current process also runs claim_list_by_channel on the first page which returns all the claim data. It then proceeds to run a resolve on each of the claims (if they have not been visited in the current session). The resolve should not be needed as the prior call has all claim data returned. If we do need to run a resolve, we should batch it. 

### Expected Behaviour
Faster subsequent load times

### Actual Behaviour
Reload every page as if it were the first time seeing it

### Suggested Solutions
Employ caching strategy

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

tzarebczan:
Closed by https://github.com/lbryio/lbry-redux/pull/60

PR #1750, opened 2018-07-10
Channel page caching

daovist:
Issue #1255 

Moves `<BusyIndicator>` to title bar.

Refactors `contentList` to show "No content found" only after fetch is complete; `null` while fetching.

channel page fetching

There is still some chance that a user will click the wrong tile but the update is very fast and pretty smooth. The best way I see to prevent this is to disable navigation for ~1sec following the update. This could be done by adding a throttle property to the state along with corresponding actions, reducers, and a thunk in doNavigate, or migrating from thunks to middleware, both of which seem well beyond the scope of this issue.

### DIFF
--- a/src/renderer/page/channel/view.jsx
+++ b/src/renderer/page/channel/view.jsx
@@ -48,7 +48,7 @@ class ChannelPage extends React.PureComponent<Props> {
     this.props.navigate('/show', newParams);
   }
 
-  paginate(e, totalPages: number) {
+  paginate(e: SyntheticKeyboardEvent<*>, totalPages: number) {
     // Change page if enter was pressed, and the given page is between
     // the first and the last.
     const pageFromInput = Number(e.target.value);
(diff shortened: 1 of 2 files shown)


src/renderer/page/channel/view.jsx
```diff
@@ -67,22 +67,23 @@ class ChannelPage extends React.PureComponent<Props> {
...
-    }
+
+    const contentList =
+      claimsInChannel && claimsInChannel.length ? (
+        <FileList sortByHeight hideFilter fileInfos={claimsInChannel} />
+      ) : !fetching ? (
+        <span className="empty">{__('No content found.')}</span>
+      ) : (
+        null
+      );
```
seanyesmunt: maybe 
```
...
) : !fetching && <span className="empty">{__('No content found.')}</span>;
```

To avoid the nested ternarys

src/renderer/page/channel/view.jsx
```diff
@@ -67,22 +67,23 @@ class ChannelPage extends React.PureComponent<Props> {
...
 
     return (
       <Page notContained>
         <section className="card__channel-info card__channel-info--large">
-          <h1>{name}</h1>
+          <h1>
+            {name}
+            {fetching ? <BusyIndicator /> : null}
```
seanyesmunt: `{!fetching && <BusyIndicator />}`

seanyesmunt:
This is great. So much smoother.

daovist:
I also cleaned up types in `<BusyIndicator>`.

PR #1752, opened 2018-07-11
Changed the error message given when the daemon didn't update

42degrees:
Changed the error message given when the daemon and LBRY app are on different versions of the API to make it more clear for the user what their next steps are.

Regarding ticket: https://github.com/lbryio/lbry-app/issues/1749

ticket complete.


tzarebczan:
Woohoo, your first PR, congrats!

Issue #1572, opened 2018-06-10
Fetch new rewards automatically


tzarebczan:
## The Issue
Currently, the app only fetches new rewards on startup. Need to fix/replace this code: https://github.com/lbryio/lbry-app/blob/master/src/renderer/page/rewards/view.jsx#L28 so that it fetches when the rewards page is accessed or at some interval. 

I believe @daovist briefly looked into this. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

PR #1754, opened 2018-07-11
fetch rewards on page load

daovist:
#1572 

Removes the `componentDidReceiveProps` hook from the `<RewardsPage>` component, which was firing repeatedly.

Passes in claimed rewards through the container to differentiate between "none available" and "you claimed them all".

### DIFF
--- a/src/renderer/page/rewards/view.jsx
+++ b/src/renderer/page/rewards/view.jsx
@@ -28,11 +28,7 @@ type Props = {
 
 class RewardsPage extends React.PureComponent<Props> {
   componentDidMount() {
-    const { fetching, rewards, fetchRewards } = this.props;
-
-    if (!fetching && (!rewards || !rewards.length)) {
-      fetchRewards();
-    }
+    this.props.fetchRewards();
   }
 
   renderPageHeader() {
--- a/src/renderer/page/rewards/view.jsx
+++ b/src/renderer/page/rewards/view.jsx
@@ -117,7 +117,7 @@ class RewardsPage extends React.PureComponent<Props> {
     } else if (!rewards || rewards.length <= 0) {
       return (
         <div className="card__content">
-          {claimed && Object.keys(claimed).length
+          {claimed && claimed.length
             ? __(
                 "You have claimed all available rewards! We're regularly adding more so be sure to check back later."
               )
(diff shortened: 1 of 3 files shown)


src/renderer/page/rewards/view.jsx
```diff
@@ -30,21 +32,13 @@ class RewardsPage extends React.PureComponent<Props> {
     It can safely be disabled since we fetch all rewards after authentication, but should be re-enabled once fixed.
```
seanyesmunt: This comment can be removed

src/renderer/page/rewards/view.jsx
```diff
@@ -9,9 +9,11 @@ import classnames from 'classnames';
 
 type Props = {
   doAuth: () => void,
+  fetchRewards: () => void,
   navigate: string => void,
   fetching: boolean,
   rewards: Array<{ reward_type: boolean }>,
+  claimed: {},
```
seanyesmunt: Could you add a new type in `types/reward`? Then we can improve the above two lines with actual types.

src/renderer/page/rewards/view.jsx
```diff
@@ -6,12 +6,15 @@ import RewardTile from 'component/rewardTile';
...
+import type { Reward } from 'types/reward';
 
 type Props = {
   doAuth: () => void,
+  fetchRewards: () => void,
   navigate: string => void,
   fetching: boolean,
-  rewards: Array<{ reward_type: boolean }>,
+  rewards: Array<Reward>,
+  claimed: Reward,
```
seanyesmunt: I don't think `claimed` would be a reward right? ` { 'reward name': Reward } `, not super sure how to specify that in Flow

tzarebczan:
@daovist won't be able to test this in our live environment until next week's LBRY cast. Other option is to run a local internal-api server. I know @seanyesmunt has one setup...I haven't got around to trying it on Linux yet (failed on Windows in the past).

tzarebczan:
Thanks for the fix @daovist, now it fires. 
There is a slight loading delay, which isn't a big deal. Are the rewards cachable similar to how we handle some of the other pages?

daovist:
@tzarebczan yes, I want to rewrite the rewards page (the whole component is a mess rn) to look and function like the new channel page in terms of caching and loading but I'm thinking we should iron out the details there before using it as a model for other pages. This may include re-fetching new data while a user stays on a page. @seanyesmunt

seanyesmunt:
I think this is good for now, but I agree it would be nice to add styling similar to how the channel pages work now.

Issue #1362, opened 2018-04-19
Validate thumbnails and show tile preview during publish 


tzarebczan:
## The Issue
Currently, there is no validation done on the thumbnail URL, so a user may enter an incorrect URL which results in a blank tile. We should validate the thumbnail input to be a jpg/png/gif (others?) and have a button that shows a preview of what the tile will look like in the LBRY app. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

PR #1755, opened 2018-07-11
Validate thumbnails

daovist:
#1362 

Inform the user whether their manually entered thumbnail URL is valid and display a preview of it.

I mostly forgot how to do CSS after Bootstrap came out and most of my early website gigs were pixel-to-HTML so my design skills are severely lacking. This will probably need some input/touchup from others.

manual thumbnail preview


kauffj:
Great idea! (Agreed UI should likely change a bit)

seanyesmunt:
This is awesome. Will think about how we can display this in the app.

seanyesmunt:
Current design:
screen shot 2018-07-18 at 11 56 20 am

screen shot 2018-07-18 at 11 55 51 am

screen shot 2018-07-18 at 11 53 45 am

seanyesmunt:
I think the only thing left would be to create a proper "No thumbnail chosen" (or whatever we want to call it) image to use instead of the quick one I added.

Issue #1327, opened 2018-04-16
Downloads list should be unique by outpoint/key, not claimid


tzarebczan:
## The Issue

You get the below console warning if two downloads have the same claim id (which is normal if a claim is updated with a new file). Currently, both claims will show the same content as it is pulled up by the claimid, but it should really use the outpoint or key from file_list. This may be somewhat related to https://github.com/lbryio/lbry-app/issues/1203

```
 C:\Users\SoloDolo\Desktop\new3\lbry-app\node_modules\fbjs\lib\warning.js:33 Warning: Encountered two children with the same key, `lbry://one#008401a48cecef2a13cf5376b696169bb36b18c6`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in div (created by FileList)
    in section (created by FileList)
    in FileList (created by Connect(FileList))
    in Connect(FileList) (created by FileListDownloaded)
    in main (created by Page)
    in Page (created by Connect(Page))
    in Connect(Page) (created by FileListDownloaded)
    in FileListDownloaded (created by Connect(FileListDownloaded))
    in Connect(FileListDownloaded) (created by Router)
    in Router (created by Connect(Router))
    in Connect(Router) (created by App)
    in div (created by App)
    in main (created by App)
    in div (created by App)
    in App (created by Connect(App))
    in Connect(App)
    in div
    in Provider
```
### Steps to Reproduce
1. Download a claim
2. Update claim with new data
3. Download the claim again

### Expected Behaviour
Should be able to access both files in the LBRY downloads

### Actual Behaviour
Can only access by claim id, which leads to a single file.

### Suggested Solutions
Use the outpoint/key 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

kauffj:
The long-term fix here is not straightforward, though switching to outpoint as a unique key might be an acceptable short term change .

If claim id stays the same, it is supposed to be the publisher saying this is the same piece of content. If the claim id stays the same but the outpoint changes, the user should not have to re-pay a key fee and instead should be prompted to update to the latest version of the content.

Part of the problem is probably that all edits keep the claim id the same, when instead we should ask the creator if the edit is actually a new publish against the same name. And in that case, change the claim id.

tzarebczan:
[...]

PR #1759, opened 2018-07-11
Add or claimId clause to avoid getting a console log.

dan1d:
Work around to remove the warning log is add an or clause as the outpoint is undefined.
Why outpoint is undefined ? I didn't understand that part.
I've saw a comment referencing issue: #1327 

Anyway, feel free to ask me any changes!

### DIFF
--- a/src/renderer/component/fileList/view.jsx
+++ b/src/renderer/component/fileList/view.jsx
@@ -168,7 +168,8 @@ class FileList extends React.PureComponent<Props, State> {
       const uri = buildURI(uriParams);
 
       // See https://github.com/lbryio/lbry-app/issues/1327 for discussion around using outpoint as the key
-      content.push(<FileCard key={outpoint} uri={uri} checkPending={checkPending} />);
+      // Outpoint is undefined on this list that's why the `|| claimId` fix the key console log error.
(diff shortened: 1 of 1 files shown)


tzarebczan:
@dan1d per Sean's comments [here](https://github.com/lbryio/lbry-desktop/issues/1327#issuecomment-401892560) he already made this change but kept the ticket open for other discussion around editing + claimIds.

dan1d:
@tzarebczan that branch seems like already merged but the code on master still throw the log warning. 
I've saw a comment on the code referencing that issue. Maybe @seanyesmunt can clarify it? My code is just one line change.

tzarebczan:
@dan1d ah, gotcha. I don't get a console error here, so maybe it's some other edge case where this is happening. All download should have an outpoint, but I don't think this fix hurts anything.

seanyesmunt:
I'm not sure why outpoint is undefined. The real fix is probably ensuring that always exists, but maybe that is a daemon issue. I'll spend a little bit of time looking into this before we merge this.

seanyesmunt:
@dan1d We are going to create an issue for the daemon. The outpoint should be included in your list of publishes.

seanyesmunt:
I'm going to close this PR for now and open an issue.

PR #1760, opened 2018-07-12
hide community row if nsfw content is hidden

seanyesmunt:
I think we should keep the "row" styling somewhat there. Then users know this would be a scrollable row just like the other rows, if they change their settings.

screen shot 2018-07-11 at 9 54 28 pm

### DIFF
--- a/src/renderer/component/categoryList/view.jsx
+++ b/src/renderer/component/categoryList/view.jsx
@@ -1,11 +1,11 @@
 // @flow
-import React from 'react';
+import * as React from 'react';
 import { normalizeURI } from 'lbry-redux';
 import ToolTip from 'component/common/tooltip';
 import FileCard from 'component/fileCard';
 import Button from 'component/button';
 import * as icons from 'constants/icons';
-import Claim from 'types/claim';
+import type { Claim } from 'types/claim';
 
 type Props = {
   category: string,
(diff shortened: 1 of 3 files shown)


src/renderer/scss/component/_card.scss
```diff
@@ -21,7 +21,6 @@
 }
 
 .card--small {
-  width: var(--card-small-width);
```
seanyesmunt: this wasn't doing anything.

src/renderer/scss/component/_card.scss
```diff
@@ -283,6 +282,12 @@
   &:last-of-type {
     padding-bottom: $spacing-vertical * 2/3;
   }
+
+  // For the message if community top bids are disabled
+  .card__content.help {
```
seanyesmunt: This is needed because of the special CSS that happens inside the CategoryList component, that allows content to scroll "off" the side of the edge. That behavior doesn't exist anywhere else in the app.

src/renderer/component/categoryList/view.jsx
```diff
@@ -1,11 +1,11 @@
 // @flow
-import React from 'react';
+import * as React from 'react';
```
skhameneh: why this change?

seanyesmunt: It's needed after adding `React.Fragment`. Flow acts funny without it.

tzarebczan:
This looks good to me and confirmed with Rob that will move trending to top (or 2nd from top) so that Community top bids will be at the bottom. This means we won't have to worry about the weird spacing.

Issue #1612, opened 2018-06-16
Suppress download error messages when video is playing / subscribed downloads


tzarebczan:
## The Issue
If a user is watching a video, we should suppress any download failure related error messages. 
This would happen if a user started a download, then went to watch another video, and the original download fails. 

Another scenario to suppress download failure notifications would be automated downloads started when subscriptions are added/new claim comes in on a subscription. 

### Steps to Reproduce
1. Start a bad video download 
2. Go to a new claim, start watching
3. Bad video download will result in an error modal

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

PR #1768, opened 2018-07-13
suppress load video error

daovist:
#1612 

Depends on [/lbryio/lbry-redux/issues/57](https://github.com/lbryio/lbry-redux/pull/57)

Refactor video loading error into a single function `handleLoadVideoError` and use for timeout and catch.

Suppress error notification when the user is playing another video. We could also suppress the error when no video is playing?

This fix introduced an issue with autoplay where it keeps trying again after it fails and stacks error messages. I added (in '`lbry-redux`) an `error` property to `fileInfo`. When a video fails to load the `uri` is added to the `error` object with value true. If this value is true for a given uri, that media will not autoplay. This value is deleted when the video starts to load/play again.


kauffj:
This issue/solution might highlight that what the app really needs is an error/notification queue (e.g. something like the one on GitHub). 

(This is separate from the PR, which would still be an approvement over the status quo.)

Issue #608, opened 2017-09-22
Transaction List - categorize and fix amount for abandon transactions


tzarebczan:
## The Issue
Currently, a claim or channel abandoning transaction shows as a "Spend" and only displays the fee (twice).  It should be possible to determine when a transaction is an abandon and show the reclaimed LBC (positive) as the Amount.  The claim name should also be shown (even though it's abandoned - it will go to a blank claim) 

Opened https://github.com/lbryio/lbry/issues/946 on the lbry side - this is where the fix/enhancement needs to occur. After we have the amount/data correctly, categorize on the app side. 

### Expected behaviour
Abandon tx should be categorized as such and show amount of deposit returned

### Actual behaviour
Abandon tx treated as a spend and shows fee as amount - can't tell how many credits came back to the wallet. 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

kauffj:
@hackrush01 good one for you?

tzarebczan:
![abandon](https://user-images.githubusercontent.com/8120721/31454386-4953468e-ae83-11e7-9b47-3914f49b4183.JPG)

tzarebczan:
Issue moved to [lbryio/lbry #946](https://github.com/lbryio/lbry/issues/946) via [**ZenHub**](https://www.zenhub.com/)

kauffj:
Blocked by https://github.com/lbryio/lbryum/pull/159

kauffj:
App will need to use new `balance_delta` field provided by LBRY API.

tzarebczan:
@kauffj what terminology should we use to categorize these in the tx list/filter? Abandon? Revoke? Delete?  

This applies to tips/supports/claims/channels that are abandoned.

kauffj:
Not delete, since this is mildly erroneous (nothing is ever deleted from the blockchain).

I think abandon (abandoned?) is good.

kauffj:
Revoke is also good though.

lyoshenka:
This is waiting on daemon 0.19

tzarebczan:
Reopening
@daovist  - Please add this to the filter dropdown on the transaction history page and ensure that the export feature works with it as well.

PR #1769, opened 2018-07-13
add abandon to transaction list filter

daovist:
#608


seanyesmunt:
Maybe these should all be constants?

seanyesmunt:
We can just add it to this PR, it's not a big rush.

PR #1771, opened 2018-07-16
Better Error Message

Invariant-Change:
The current error message sounds like LBRY is broken, when in fact it's a ShapeShift issue.
This message is friendlier and more accurate, which helps to keep LBRY's brand reputation stronger :)


Invariant-Change:
image

PR #1778, opened 2018-07-17
updated the url for successful thumbnail publishes

billbitt:
This is an update for the app to use a static embed url for successful thumbnails published via spee.ch.
This is dependent upon merging the following PR on spee.ch: https://github.com/lbryio/spee.ch/pull/529


seanyesmunt:
Whoops. We can merge this after that PR is merged.

PR #1798, opened 2018-07-21
Prevent downloading the daemon if it's not necessary

seanyesmunt:
We shouldn't download the daemon if it's already downloaded.


seanyesmunt:
closing in favor of #1803

Issue #1808, opened 2018-07-24
Differentiate between blank and incorrect URL on Publish thumbnail preview


tzarebczan:
## The Issue
Currently, the same "broken" image is shown when there is no URL or an incorrect URL is entered. We should differentiate between the two scenarios to make it clearer for the user. 

![thumb preview](https://user-images.githubusercontent.com/8120721/43147556-4e1f441c-8f31-11e8-9e6f-d97acb41d6cf.jpg)

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

dan1d:
Hi, I would need an icon image to be inserted on this issue. Code is ready to be reviewed but I'm missing a second icon.
Fail url:
![screenshot from 2018-07-27 17-53-10](https://user-images.githubusercontent.com/3293616/43346123-6f106d80-91c6-11e8-95a4-f16f6a07fa90.png)
Blank url
![screenshot from 2018-07-27 17-53-21](https://user-images.githubusercontent.com/3293616/43346127-71e09b20-91c6-11e8-9dc0-62cd7a188ac3.png)

tzarebczan:
Implemented on master via 1829, will be in next release.

PR #1829, opened 2018-07-27
Differentiate between failed thumbnail url and emtpy url images.

dan1d:
#1808 
Hi, I would need an icon image to be inserted on this issue. Code is ready to be reviewed but I'm missing a second icon, I'm very bad at design, so if anyone can chime in, thanks a lot!

On Failed url:
screenshot from 2018-07-27 17-53-10
On Emtpy url:
screenshot from 2018-07-27 17-53-21


tzarebczan:
@NetOperatorWibby, think you can give us a hand?

NetOperatorWibby:
@tzarebczan How do I add images to an existing PR?

tzarebczan:
@dan1d where is the image pulling from? I don't see it in the PR.

@NetOperatorWibby since this is a community PR, the branch is on his forked repo. I'd assume you can edit there, or just add the thumbnail to this thread.

tzarebczan:
@dan1d we'd probably want to define the warning thumbnail the same way, right? 

@NetOperatorWibby might just be easier to include a png here (it would be the same size/aspect ratio as the one we provided for the no thumb scenario: https://github.com/lbryio/lbry-desktop/pull/1755/files

NetOperatorWibby:
failed-thumbnail

This should work.

dan1d:
Looks good!
screenshot from 2018-07-30 15-42-57

kauffj:
Above could probably be done in pure CSS. More importantly, it should probably be the error color.

NetOperatorWibby:
+1 for error color but I went with image since the other message is also an image. Is there a unified icon set? It would make sense to take the CSS approach then.

NetOperatorWibby:
error-thumbnail

Used the error color found [here](https://github.com/lbryio/lbry-desktop/blob/master/src/renderer/scss/_vars.scss#L44).

dan1d:
Image updated to red, but should we add them with CSS then ?

PR #1832, opened 2018-07-28
Update electron-builder.json

kodxana:
Fixed error when building keytar
gyp: Call to 'pkg-config --libs-only-l libsecret-1' returned exit status 127 while in binding.gyp. while trying to load binding.gyp
gyp ERR! configure error
gyp ERR! stack Error: `gyp` failed with exit code: 1

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,7 @@ The LBRY app is a graphical browser for the decentralized content marketplace pr
 ## Install
 
 We provide installers for Raspberry Pi 3
-| Linux                                        
-| -------------------------------------------- |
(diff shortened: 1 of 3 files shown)


tzarebczan:
@kodxana when would this fail for you? Is this to help with building for the pi, or something else you ran into locally? What Node version?

tzarebczan:
@kodxana did you mean to update that readme locally? See my previous question too. Was that meant to be locally as well? If so, please close this PR.

kodxana:
Ups I didnt wanted to do that pull

Issue #1788, opened 2018-07-18
Feature Suggestion: Option to enable/disable desktop notification


tzarebczan:
## The Issue
Currently, we use desktop notifications when a file has completed downloading. These notifications may extend to other areas in the future. 

A setting to enable/disable these notifications should be available for users who don't want to to see them.

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

dan1d:
Hi I've added the setting there, should I create a new section?
![screenshot from 2018-07-30 21-10-53](https://user-images.githubusercontent.com/3293616/43429990-170e0434-943d-11e8-9cea-01cdac97c374.png)

tzarebczan:
Merged in master with https://github.com/lbryio/lbry-desktop/pull/1834, will be in next release.

PR #1834, opened 2018-07-31
Add setting to skip desktop nofifications

dan1d:
Closes issue #1788 
It persist the setting on localstorage, defaults to true(notifications enabled), I've created a selector to check if the notification should be run.
Example:
`if(selectDesktopNotificationsEnabled(state)) thenNotify()`

The setting is present here, let me know If I should change the text/section of the checkbox.
screenshot from 2018-07-30 21-10-53


kauffj:
1) This should not be clustered until the Download Directory. I'd suggest making a new section with a header of "Notifications". The label for the checkbox can be "Notify me when a download completes" or just "On Download Completion", whichever looks better.

2) @seanyesmunt is the checkbox used on the file show page intended to be the new checkbox used everywhere? If so, it should be used here (and really, replace all checkboxes). Mixed styles for the same element are a big design inconsistency.

dan1d:
@kauffj on point #1, I've made the code to enable/disable all kind of desktop notifications, not just when it finishes downloading, should I add code to cover only this case:  "Notify me when a download completes" ?
Daniel

dan1d:
Updated checkbox to toggle style on settings page.
screenshot from 2018-07-31 19-40-44

seanyesmunt:
Yes the plan is to add that to all checkboxes. @dan1d I would check to make sure the description is still lined up properly on smaller screens when the description text has to wrap to the next line.

tzarebczan:
The notifications include more than just download complete - we also do notifications for subscriptions when new content is posted.  And we'll probably add more (think there's an issue for adding a published complete notification). 

IMO, it should be called something along the lines of OS/System Notifications so as not to be confused with the in-app notification system (when it's added).

seanyesmunt:
Yeah I think for the first step we will just add a blanket mute for all notifications. I like @tzarebczan's idea of calling them `OS Notifications`

dan1d:
I've changed the variable names to `osNotificationsEnabled` and the selector to `selectOsNotificationEnabled`.

The check sign has a weird behavior on smaller screen. 
screenshot from 2018-08-01 15-29-05

kauffj:
@dan1d looks great! Can you label it "Show Desktop Notifications"?

seanyesmunt:
Nice. You may need to make some change to `_toggle.scss`. Not sure what would be causing that.

dan1d:
Added `margin-top: auto; margin-bottom:  auto;` to `.react-toggle` class and it adjust automatically to the center of the parent box. Which fix the issue. Let me know if I should change that part.
screenshot from 2018-08-01 16-47-43

seanyesmunt:
I think that fix is fine (or a similar one), but in my opinion the toggle should be at the top. In-line with the first line of the description.

dan1d:
Done!
screenshot from 2018-08-01 19-38-32

seanyesmunt:
@dan1d Just tested this and it looks great! One thing I forgot to say is that we should probably just get rid of the `useToggle` prop and always use this toggle. That was only when we only used the toggle in one place. Once that is done and a changelog entry is added this can be merged. 

Nice work! 🙂

dan1d:
@seanyesmunt Thanks for the review. Removed the useToggle prop and added the changelog entry.

Issue #1843, opened 2018-08-01
Middle Click Link Error


Invariant-Change:
When someone **middle-clicks** on **links** within the app, it opens them up in an app window with **incorrect** formatting and also **doesn't** support the Shapeshift site.

But if clicked as normal with the **left mouse button**, they open in the default browser as **normal**.

![image](https://user-images.githubusercontent.com/29914179/43533044-3547ca78-95f7-11e8-80d4-d88fc1812657.png)

btzr-io:
https://github.com/electron/electron/issues/2770

tzarebczan:
Fixed in https://github.com/lbryio/lbry-desktop/issues/1844

PR #1844, opened 2018-08-01
Fix middle-click open link event

btzr-io:
### Fixes
- Middle Click Link Error #1843

  > When someone middle-clicks on links within the app, it opens them up in an app window with incorrect formatting and also doesn't support the Shapeshift site.


seanyesmunt:
Sweet. Can you add a changelog entry to this?

btzr-io:
ok, done :+1:

Issue #2, opened 2017-02-16
No URI Handling


jobevers:
Currently the app doesn't do anything on `lbry://` uris.

jobevers:
See

- http://electron.atom.io/docs/api/app/#appsetasdefaultprotocolclientprotocol-path-args-macos-windows
- https://github.com/electron-userland/electron-builder/wiki/Options#Protocol

alexliebowitz:
@lyoshenka: Are you planning to work on this or just supervising it? I can own this, looks pretty small (at least in theory)

lyoshenka:
Go for it

alexliebowitz:
Fixed for Mac by #116 and Windows by #121. Linux still in progress (#122).

alexliebowitz:
We decided to merge Linux (although there's an unrelated issue that means it doesn't work perfectly yet). So I'm marking this resolved.

PR #1845, opened 2018-08-02
get recommended videos from search based on title

seanyesmunt:
screen shot 2018-08-07 at 1 13 57 am


tzarebczan:
3 things I noticed:
1) the URL is missing from the search bar when you go into it the first time. It comes back the next time you re-enter. 
2) if you go into lbry://three - you get a suggestion for lbry://three#claimid (that matches the same claim id). Probably a vanity url issue .
3) I noticed the play button black > green (takes about a second or two) - can we speed this up? Why does it happen at all?

seanyesmunt:
@tzarebczan 
#1 will be fixed with https://github.com/lbryio/lbry-redux/pull/65
Still thinking about how I can fix #2
#3 seems to be an issue on previous versions. Nothing from this PR should mess with that (it is definitely a valid issue though)

seanyesmunt:
@skhameneh This is ready for review. There are still two issues but I would like to get this merged now so we can get some more eyes on it since I changed a lot of styles across the app.

Known issues:
- if you go into lbry://three - you get a suggestion for lbry://three#claimid (that matches the same claim id). Probably a vanity url issue .
- https://github.com/lbryio/lighthouse/issues/101

Issue #1009, opened 2018-02-09
Can't see history of content/channels I've viewed


seanyesmunt:
## The Issue
Can't see history of content/channels I've viewed

### Steps to reproduce
1. Navigate to a channel/content from the homepage but don't watch it
2. Wait for it to be taken off the home page
3. Try to get back to it

### Expected behaviour
I should be able to see my navigation history, similar to a web browser. This would consist of content pages only (I think?) and not app pages like wallet, settings, etc.

### Actual behaviour
I have to remember channels/videos if I want to go back to them. I think this leads to a lot of content that is lost as soon as it's taken off the homepage

daovist:
@seanyesmunt I'm having some trouble making this work in `navigation.stack`

when a user reloads the app, should they be taken to the last page they visited? Are we pushing `/discover` to the end of the `navigation.stack` array?

because if `index` is 0 or 1 and the user clicks a link it will remove all history beyond `index`--a browser with back and forward navigation wouldn't work otherwise

or I'm just approaching this from the wrong angle again...

seanyesmunt:
If the user reloads the app should stay on the same screen. @daovist Is it not working by just adding `'navigation'` here? 
https://github.com/lbryio/lbry-app/blob/master/src/renderer/store.js#L101

daovist:
Oh, okay. I didn't think we would want it to work like that. I would prefer it start on the home page with new content but am not picky.

What you've described is now the case in the master branch and it loads the last page, so I'll just make the History page to display.

FYI the new master isn't working quite right on my machine (Ubuntu 16.04.3). There is no scrolling except for non-video content e.g. UC Berkeley index. When trying to access a page I searched for by name I get reference error: normalizeURI not found. Should I make be making submitting issues or PRs to fix these?

I'm dealing with merge issues all over the place and trying not to revert to making all new PRs but may end up going that route.

tzarebczan:
Recently Viewed would be a good title for this area. Should we track this in the sqlite database through the daemon? Decision would be based on cross device syncing solution (is there one yet?)

PR #1846, opened 2018-08-02
User history

daovist:
See #1009 

Requires [/lbryio/lbry-redux/pull/64](https://github.com/lbryio/lbry-redux/pull/64)

- [x] structures media playback position data by `state.content.[claimId][outpoint]` rather than `state.media[outpoint]`--this will useful down the road when we handle file updates
- [x] remove duplicated data representing a paused file 
- [x] use changes to `state.content.playingUri` to pause video on file opening rather than `state.media.paused`
- [ ] fix UI to not show layover after pausing from opening a file?
- [x] record user history browsing `/show` pages
- [x] add `<UserHistoryPage>` to display a paginated history list with "last viewed" info
- [x] add resume icons to `<FileTile>`s
- [x] add delete icon to `<FileTile>`s shown on `<UserHistoryPage>`
- [ ] style page and tiles


seanyesmunt:
screen shot 2018-08-02 at 1 07 47 pm

I don't think we need the thumbnail in this view since they would have already seen it when actually navigating to it.

seanyesmunt:
@daovist @skhameneh This is ready for review. I would appreciate some testing, most of the data logic from travis' work stayed, I just changed the ui a bit.

daovist:
I am busy so I only looked at some of the code but the UI looks good. `react-pagination` is most excellent.

I might add a little padding to the top of the darker area but that's not a big deal. Alternating colors (ever so slightly) per item might be nice too, or a hover color change.

seanyesmunt:
@skhameneh 

The app currently doesn't work with the latest `lbry-redux` because I removed the `selectNavLinks` function. Since we have a community PR with `lbry-redux` changes I'm going to merge this, so they can get the app to build properly. I will open another PR if needed to address any comments you have.

PR #1853, opened 2018-08-03
log time to play

seanyesmunt:
#### Notes
This ended up being a little more complex after realizing we need to reset analytics data when users navigate directly to another file (where the File page won't be re-mounted)

There are three scenarios:
- User has already downloaded a file: do nothing
- User started downloading but navigated away before it started playing: log `file_view` with no `time_to_start`
- User stayed on the page until the file started playing: log `file_view` with `time_to_start`

### DIFF
--- a/src/renderer/component/fileViewer/view.jsx
+++ b/src/renderer/component/fileViewer/view.jsx
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react';
 import classnames from 'classnames';
+import analytics from 'analytics';
 import type { Claim } from 'types/claim';
 import LoadingScreen from 'component/common/loading-screen';
 import Player from './internal/player';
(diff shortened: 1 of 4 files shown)


seanyesmunt:
Currently outpoint is still not included in the claim information (https://github.com/lbryio/lbry/issues/1306). Previously, we were calling `Lbry.get` which returns the `outpoint` and we used that for this api call.

We can't read it from `fileInfo` because that will not exist if we try to fire the analytics event in response to a user navigating away before the file is downloaded.

I think the simplest approach would be to wait for https://github.com/lbryio/lbry/issues/1306 and grab `outpoint` along with the rest of the data from the claim information.

@skhameneh @kauffj Thoughts?

kauffj:
Three comments:

1) I think this overlaps some with @daovist work on media history. Please discuss on stand-up if it makes sense to combine this work or for media history to contain this value.

2) View tracking may currently supports parameters other than outpoints.

3) Please don't hesitate to point this issue out to a daemon team member and/or comment on that ticket itself. I suspect this is a fairly small one to fix.

daovist:
A claim's outpoint is just: `${claim.txid}:${claim.nout}`. I don't understand why #1306 is an issue.

src/renderer/redux/actions/content.js
```diff
@@ -244,6 +236,29 @@ export function doSetPlayingUri(uri) {
   };
 }
 
+function handleLoadVideoError(uri, errorType = '') {
```
seanyesmunt: Lint was telling me to move this up

src/renderer/component/fileViewer/view.jsx
```diff
@@ -107,9 +147,42 @@ class FileViewer extends React.PureComponent<Props> {
...
+  logTimeToStart() {
+    const { claim } = this.props;
+
+    if (this.startTime) {
+      this.playTime = Date.now();
+      this.fireAnalyticsEvent(claim, this.startTime, this.playTime);
+    }
+  }
+
+  fireAnalyticsEvent = (claim, startTime, playTime) => {
```
skhameneh: Why would we log the same parameters with every analytics event?
There doesn’t appear to be any event name associated as well, this could make things difficult to differentiate events.

seanyesmunt: Not sure what you mean. We log the same parameters, but only once per file. The actual api call is here https://github.com/lbryio/lbry-desktop/blob/master/src/renderer/analytics.js#L47

seanyesmunt: Whoops. You are right @tzarebczan 

I need to make sure it doesn't fire if the video hasn't started downloading. Will be easy to add.

seanyesmunt:
@tzarebczan Just updated this to only fire if a user leaves the page _after_ it has started downloading. Before it was firing if they clicked play and left even if it wasn't downloading.

seanyesmunt:
@skhameneh Did you have any concerns before merging?

kauffj:
@seanyesmunt 

> Before it was firing if they clicked play and left even if it wasn't downloading.

My concern is that we want this information!

seanyesmunt:
@kauffj Tom had a comment
`
my only concern is that it's now even easier to get a "view" - the file doesn't even have to start playing. Maybe we need to only take into account files with time to play for rewards purposes?
`

If we change the behavior by adding a view event, we may need to change rewards logic.

skhameneh:
@seanyesmunt this is why I had mentioned tagging different events explicitly, isn’t the immediate approach to move forward and change how we track analytics to be more explicit?
We need an approach where we capture a lot of raw data and do transforms to SQL for more specific tables that correlate to data captured.

skhameneh:
@seanyesmunt the rewards logic would ideally be a separate issue

tzarebczan:
We can worry about the rewards logic on the internal-api side. I think this is ready to be merged in right?

PR #1858, opened 2018-08-06
update electron to 2.0.6

tzarebczan:
Also includes fix from https://github.com/electron/electron/issues/13008#issuecomment-400261941

Tested on Windows/Mac/Linux, don't see any issues. We'll revisit electron 3 when they are out of beta.


seanyesmunt:
I am seeing 
```
Electron Security Warning (Node.js Integration with Remote Content)
```
in the devtools. Do we need to make any changes to how we create the browser window?

tzarebczan:
@seanyesmunt should be fixed now!

PR #1869, opened 2018-08-12
Use codemirror minified version

btzr-io:


skhameneh:
We already have minification upon distribution; this loses debugging information for the app's runtime in dev mode.  I would suggest keeping the original packages for this.

I'll let @seanyesmunt make the call

skhameneh:
I checked the package changes and they're not maintained by the codemirror team; the versions are behind.

seanyesmunt:
I think we can keep the original package.

Issue #1861, opened 2018-08-07
Improve subscription experience in the App


alyssaoc:
This epic is a collection of issues to improve and consolidate the subscription experience:

-Load subscription data on app startup
-Subscriptions should autodownload on startup
-Autodownload settings for subscriptions
-Subscriptions quirks
-Automatically claim subscription reward after subbing to channel

Issue #1852, opened 2018-08-02
Load subscription data on app startup


tzarebczan:
## The Issue
Currently, we don't check the API or load subscription data until the user subscribes to something or visits the subscriptions page. This should be loaded on startup instead.

Side effect: Clearing cache or logging into a new PC with old account will result in a subscription 101 notification if you subscribe to something before visiting the subscription page. 

Thanks @ceogear for finding this one! 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

kauffj:
@daovist probably easy to roll into current sub work?

daovist:
Yes, this is going in my subscriptions epic.

tzarebczan:
Completed and confirmed in https://github.com/lbryio/lbry-desktop/pull/1872

Issue #1835, opened 2018-07-31
Automatically claim subscription reward after subbing to channel


tzarebczan:
## The Issue
Similar to how we claim (and fail silently) other rewards like featured download/many download after download actions, we should attempt to claim the new subscription reward after a subscribe action occurs. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

tzarebczan:
@daovist do you want to take this one since you are looking at the subscription code? Should be a quick one.

tzarebczan:
Completed in https://github.com/lbryio/lbry-desktop/pull/1872

Issue #1775, opened 2018-07-16
Subscriptions should autodownload on startup


tzarebczan:
## The Issue
We tell users that "This app will automatically download new free content from channels you are subscribed to." but this only occurs if you had the LBRY app opened while the content was posted. We should be auto-downloading subscribed free content (lastest single upload) on app startup. 

## System Configuration

- LBRY Daemon version: 0.20.3
- LBRY App version: 0.22.2
- LBRY Installation ID: Private
- Operating system: Windows

tzarebczan:
@daovist you are working on this in your autodownload branch, right?

daovist:
yes

PR #1872, opened 2018-08-13
Subscriptions overhaul

daovist:
Epic #1861 

- [x] show new/incoming content immediately on subscriptions page #1734
- [x] prevent window notification when viewing subscriptions page; show snackbar instead
- [x] add auto-download client setting; use as necessary #1114
- [x] increase `SUBSCRIPTION_DOWNLOAD_LIMIT` from 1 to 3 and apply limit to total downloads rather than per-channel* 
- [x] automatically (attempt to) claim subscription reward after subbing to channel (fail silently) #1835
- [x] load subscription data on app startup #1852
- [x] auto-download on startup #1775
- [ ] Do #1852 and #1775 correctly. My fix uses setTimeout which is a hack. We should be using `redux-persist`'s [`<PersistGate>`](https://github.com/rt2zz/redux-persist/blob/master/docs/PersistGate.md) component/feature so actions are dispatched once redux has been rehydrated. This requires upgrading `redux-persist` from version `4.x` to `5.x`. There is a [migration guide](https://github.com/rt2zz/redux-persist/blob/master/docs/MigrationGuide-v5.md).
- [x] The last commit prevents downloading anything immediately following a user subscribing to a channel. This wasn't in any of the issues but it was very weird from a UX perspective to auto-download multiple videos upon subscribing. It still sets latest so it will work like other subscription from then on.

*As it was, the app would download the limit (1) for each channel which could be too much if a user is subscribed to many channels. One issue with this approach is that it will dispatch downloads in the order it iterates through subscriptions. To fix that would require calling all the endpoints, flattening the data, and choosing which ones to download. I found this to be a reasonable compromise for when a user opens the app and may have lots of new content available. Any new content that is prevented from downloading this way still creates a `NOTIFY_ONLY` notification, adding to the badge count next to Subscriptions. We may want to allow a user to determine the value of `SUBSCRIPTION_DOWNLOAD_LIMIT`.


daovist:
One weird thing about this @seanyesmunt 

I'm using the existing reducer `FETCH_CHANNEL_CLAIMS_COMPLETED` to update the claim info so it will show on the subscriptions page; the lack of doing so was causing the main quirk in all this #1734 . I am not using `FETCH_CHANNEL_CLAIMS_STARTED` because that prevented the subscriptions page from even loading. The code I started with dispatched `CHECK_SUBSCRIPTION_STARTED` and `CHECK_SUBSCRIPTION_COMPLETED` which had no reducers and thus did nothing.

src/renderer/page/settings/view.jsx
```diff
@@ -265,6 +272,13 @@ class SettingsPage extends React.PureComponent<Props, State> {
                 checked={autoplay}
                 postfix={__('Autoplay media files')}
               />
+              <FormField
+                type="checkbox"
+                name="auto_download"
+                onChange={this.onAutoDownloadChange}
+                checked={autoDownload}
+                postfix={__('Automatically download new content')}
```
seanyesmunt: This should reference that it is for subscriptions. 
`Automatically download new content from your subscriptions`? That might be a little long

src/renderer/redux/actions/content.js
```diff
@@ -139,7 +140,14 @@ export function doUpdateLoadStatus(uri, outpoint) {
             0
           );
 
-          if (selectosNotificationsEnabled(getState())) {
+          if (state.navigation.currentPath === '/subscriptions') {
+            dispatch(
+              doNotify({
+                message: `'${fileInfo.metadata.title}' has been downloaded`,
+                displayType: ['snackbar'],
```
seanyesmunt: I think the idea behind snackbars is that they are reserved for user actions, which wouldn't make sense for this use case. Maybe we should create our own notification component that mimics desktop notifications but has some additional style. 

Not needed for this PR but I think we should stick with the OS notifications for now.

src/renderer/redux/reducers/settings.js
```diff
@@ -26,6 +26,7 @@ const defaultState = {
     automaticDarkModeEnabled: getLocalStorageSetting(SETTINGS.AUTOMATIC_DARK_MODE_ENABLED, false),
     autoplay: getLocalStorageSetting(SETTINGS.AUTOPLAY, false),
     resultCount: Number(getLocalStorageSetting(SETTINGS.RESULT_COUNT, 50)),
+    autoDownload: getLocalStorageSetting(SETTINGS.AUTO_DOWNLOAD, false),
```
seanyesmunt: I think this should default to `true`. Users can turn it off if they want.

seanyesmunt:
Found one issue (still testing).
> I subscribed to a channel 
> Started downloading the first video
> Saw a badge of `(9)` in the side bar
> Finished downloading
> Badge chagned to `(10)`

seanyesmunt:
Possibly we should look at the date you subscribed to a channel, and if it is newer than that, _then_ show the notification/download? Not sure.

daovist:
The badge 9/10 thing is because of how that number is calculated. It's the count of all download that are *not* `DOWNLOADING`, so just `NOTIFY_ONLY` and `DOWNLOADED`. I think counting downloading files as notifications seems like a good fix.

seanyesmunt:
I think that number is ok, but it should only show the badge if it's actually new content, not just the first page of content on an old channel that I subscribed to.

In this case:
> I subscribe to a channel
> Close the app
> Channel uploads 10 files
> Open the app
> See notification with `(10)`

tzarebczan:
@seanyesmunt and I also discussed exposing the autodownload setting on the sub page too. 

Is autodownload on startup supposed to work with the hack? When I go into a claim I'm subbed to (and not downloaded), I see it starting to download automatically - is this intentional? 

We should incorporate the download limit as a setting if we are going to impose it. I think it would also work better as a per channel instead of across all subs setting.

kauffj:
Encountered this error running this branch:

```
subscriptions.js?421d:133 Uncaught ReferenceError: savedSubscription is not defined
    at eval (subscriptions.js?421d:133)
    at /home/kauffj/code/lbry-app/node_modules/redux-thunk/lib/index.js:11
    at dispatch (/home/kauffj/code/lbry-app/node_modules/redux/lib/applyMiddleware.js:45)
    at eval (subscriptions.js?421d:273)
    at Array.forEach (<anonymous>)
    at eval (subscriptions.js?421d:272)
    at /home/kauffj/code/lbry-app/node_modules/redux-thunk/lib/index.js:11
    at dispatch (/home/kauffj/code/lbry-app/node_modules/redux/lib/applyMiddleware.js:45)
    at eval (subscriptions.js?421d:282)
```

PR #1881, opened 2018-08-15
Handle new daemon startup for encryption

seanyesmunt:
for daemon v0.21.0


seanyesmunt:
Merging now to fix master.

PR #1882, opened 2018-08-15
Update view.jsx

ykris45:


seanyesmunt:
Hey @ykris45. We decided to rename this button as `Share`. We are going to keep that name for now. 🙂

PR #1895, opened 2018-08-21
update node_id

tzarebczan:
per https://github.com/lbryio/internal-apis/commit/3f7bd0d969e18742fa326fa9d724e4d33e1122da


seanyesmunt:
https://github.com/lbryio/lbry-desktop/blob/master/src/renderer/component/splash/view.jsx#L70

Do we need to check that `dht` exists here?

tzarebczan:
Yea, it's probably safer to do so, but I ran a status very early on in the startup process and it was there. I could see DHT being disabled and not being there. Will add the check.

PR #1908, opened 2018-08-27
add message for 0 returned results in related content

seanyesmunt:
Show message if no search results are returned for a specific title for related content.

screen shot 2018-08-26 at 11 21 57 pm


tzarebczan:
This looks good when running it along with the redux change. Also, the bug with this URI is fixed: lbry://@Madiator2011-Gaming/LBRYmeVR3

PR #1909, opened 2018-08-27
Handle channels properly on top of search page

neb-b:


src/renderer/component/channelTile/view.jsx
```diff
@@ -36,27 +38,54 @@ class ChannelTile extends React.PureComponent<Props> {
   }
 
   render() {
-    const { claim, navigate, isResolvingUri, totalItems, uri } = this.props;
-    let channelName, channelId;
+    const { claim, navigate, isResolvingUri, totalItems, uri, size } = this.props;
```
skhameneh: Should probably split this into multiple lines

seanyesmunt: prettier will split it if it exceeds x characters

PR #1917, opened 2018-08-28
added learn more faq to wallet security section

ykris45:
unknown

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "formik": "^0.10.4",
     "hast-util-sanitize": "^1.1.2",
     "keytar": "^4.2.1",
-    "lbry-redux": "lbryio/lbry-redux#feffbe966fc568cef595ab66b314766261e855df",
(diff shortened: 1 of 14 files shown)


src/renderer/page/settings/view.jsx
```diff
@@ -348,12 +348,17 @@ class SettingsPage extends React.PureComponent<Props, State> {
                 name="encrypt_wallet"
                 onChange={e => this.onChangeEncryptWallet(e)}
                 checked={walletEncrypted}
-                postfix={__('Encrypt my wallet with a custom password.')}
-                helper={__(
```
seanyesmunt: We should still use this `helper` prop. You can just add the button at the end of it.

ykris45:
@seanyesmunt done

tzarebczan:
@ykris45 how are you committing these changes, through the web interface? They should be submitted through git, so that formatting tools like prettier run on them. You can also run prettier locally in programs like Atom.

PR #1928, opened 2018-09-01
add app icon (256x256)

jessopb:
On launching either the .deb production package or the dev package, the boring default app icon appears in OS window.
I chose to add the 256x256 icon, and it works in linux.
This seems like low hanging fruit, perhaps it was removed from before? My fix didn't work when I tested on mac: mac displays the default electron icon instead of the icon provided. 
I will test on windows later tonight.

### DIFF
--- a/src/main/createWindow.js
+++ b/src/main/createWindow.js
@@ -24,11 +24,9 @@ export default appState => {
     // Create the window using the state information.
     x: windowState.x,
     y: windowState.y,
-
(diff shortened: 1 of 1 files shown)


jessopb:
The 256 doesn't look great in windows, and yet the docs say to specify a 256 for windows. Is this an icon format/antialiasing thing?

jessopb:
I should have used the .ico file. The 256 png does not scale well in my windows test.

PR #1931, opened 2018-09-02
Re: Issue 1810: creates facebook share and moves social to fileActions

jessopb:
Re: Issue 1810
The facebook link facilitates posting the associated spee.ch link.
I like the idea of grouping channel-specific actions away from file-specific action.
Alternately, socialActions could be a component parallel to fileActions.
However, the best solution may be to do social sharing in a modal, as the real estate under the viewer is going to get crowded.
Feedback would be appreciated.
social

Re: linting and flow: Fixed all errors except an unrelated unused subscription prop in page/file/view.jsx

### DIFF
--- a/src/renderer/component/fileActions/view.jsx
+++ b/src/renderer/component/fileActions/view.jsx
@@ -4,6 +4,8 @@ import Button from 'component/button';
 import { MODALS } from 'lbry-redux';
 import * as icons from 'constants/icons';
 import Tooltip from 'component/common/tooltip';
+import ViewOnWebButton from 'component/viewOnWebButton';
+import ShareOnFacebook from 'component/shareOnFacebook';
 
 type FileInfo = {
   claim_id: string,
@@ -12,6 +14,8 @@ type FileInfo = {
 type Props = {
   uri: string,
   claimId: string,
+  claimName: string,
+  speechSharable: boolean,
   openModal: ({ id: string }, { uri: string }) => void,
   claimIsMine: boolean,
   fileInfo: FileInfo,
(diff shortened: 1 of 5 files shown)


jessopb:
Prop "speechSharable" was put together in page/file/view because that's where it was initially needed. It's probably no longer in a good place. I'm happy to disentangle it if this development path looks like a good one.

seanyesmunt:
This is awesome @jessopb! Thanks for the contribution

I think you are right that it's getting a little cramped with all the buttons under there. Maybe we should just move to a modal now, similar to how YouTube does it. 

We could have one button with the share icon `Share2`, which would trigger a modal. The top part could just be the basic spee.ch link, then below it, we could have buttons for Facebook and Twitter. How does that sound?

Also If you'd like to be tipped for your work, please check out http://lbry.io/faq/tips 🙂

jessopb:
The link to be copied and pasted at the top and buttons at the bottom was exactly what I was thinking. Thanks for the feedback. I've been researching just that approach.

seanyesmunt:
Great! I look forward to seeing how you choose to do it! Just ping me in this PR and I'll take a look when you are done.

jessopb:
Progress update: Currently, socialShare component displays the claim props necessary to build a spee.ch web link as a test inside modalSocialShare and dismisses with a "Done" button, following the walletSendTip modal example. I plan to have the share buttons, link formview, and linting sorted out before friday. 

I am Unsure whether I should include an update to lbry/lbry-redux MODAL types constants, or leave it alone.

jessopb:
Is there a way to bundle the PRs for lbry-redux and lbry-desktop or do I just do them separately? The ModalShare seems to work well.
modalshare

seanyesmunt:
There will need to be two separate PR's

seanyesmunt:
@jessopb I merged your lbry-redux PR then pushed up and updated build. You can change the commit id in the package.json for `lbry-redux` to use `e85d16fe03c2ff3ff613946fc3e5a4bd1ed559e0`

seanyesmunt:
Closing this in favor of #1943

PR #1938, opened 2018-09-07
Creates a modal SocialShare component

jessopb:
This was closely modeled after the modalSendTip and walletSendTip components.
Daovist pointed out it should share both anonymous and channel based content - it seemed to do both when that content could be located.
It continues to only be available if the content meets shareable criteria.
It uses share urls from http://www.sharelinkgenerator.com/ to share to Facebook, Twitter, and Google Plus.
There are some improvements that could be made:
- feathericons has no actual google-plus icon, I think
- it could share a tinyurl-ified version
- the formatting and spacing could be better
- the mouseover tooltips could be more responsive: they seem to give up sometimes - ideas?
- the link could come preselected, rather than wait for a click in the text field
- the modal could give feedback that the link copy was successful: 'link copied to clipboard'
- it could close the modal when clicking the background or pressing esc

Finally, the ViewOnWeb component is apparently no longer in use.

I look forward to more guidance in these areas.

### DIFF
--- a/src/renderer/component/socialShare/index.js
+++ b/src/renderer/component/socialShare/index.js
@@ -1,0 +1,9 @@
+import { connect } from 'react-redux';
+import { makeSelectClaimForUri } from 'lbry-redux';
+import SocialShare from './view';
+
+const select = (state, props) => ({
+  claim: makeSelectClaimForUri(props.uri)(state),
+});
+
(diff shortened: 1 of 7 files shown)


jessopb:
Do not pull: - there seems to be a problem with wallet address copy throwing errors at my new modalSocialShare.

PR #1940, opened 2018-09-07
Better messaging for takeover amounts on publish page

seanyesmunt:
More clarity for takeover amounts on claims

Show amount needed to win a claim when editing. Ex: "You have a claim at xxx, for 5 more LBC you will control it"

### Changes
I created a new component `NameHelpText` for variable help text under the name input. I moved `BidHelpText` under the bid input (which is probably where it should have been in the first place)

#### Before
screen shot 2018-09-24 at 8 24 53 pm

#### After
screen shot 2018-09-24 at 8 23 39 pm

### DIFF
--- a/src/renderer/component/publishForm/internal/bid-help-text.jsx
+++ b/src/renderer/component/publishForm/internal/bid-help-text.jsx
@@ -20,7 +20,9 @@ class BidHelpText extends React.PureComponent<Props> {
       } else {
         bidHelpText = `${__('If you bid more than')} ${amountNeededForTakeover} LBC, ${__(
           'when someone navigates to'
-        )} ${uri} ${__('it will load your published content')}.`;
+        )} ${uri} ${__('it will load your published content')}. ${__(
+          'However, you can get a longer version of this URL for any bid'
+        )}.`;
       }
     }
 
--- a/src/renderer/component/publishForm/internal/bid-help-text.jsx
+++ b/src/renderer/component/publishForm/internal/bid-help-text.jsx
@@ -19,7 +19,7 @@ class BidHelpText extends React.PureComponent<Props> {
         bidHelpText = __('Any amount will give you the winning bid.');
       } else {
         bidHelpText = `${__('If you bid more than')} ${amountNeededForTakeover} LBC, ${__(
-          'when somone navigates to'
+          'when someone navigates to'
         )} ${uri} ${__('it will load your published content')}.`;
       }
     }
--- a/src/renderer/component/publishForm/internal/bid-help-text.jsx
+++ b/src/renderer/component/publishForm/internal/bid-help-text.jsx
@@ -1,72 +1,34 @@
 // @flow
 import * as React from 'react';
-import Button from 'component/button';
-import { buildURI } from 'lbry-redux';
-import type { Claim } from 'types/claim';
 
 type Props = {
   uri: ?string,
   isResolvingUri: boolean,
-  winningBidForClaimUri: ?number,
-  myClaimForUri: ?Claim,
-  isStillEditing: boolean,
-  onEditMyClaim: (any, string) => void,
+  amountNeededForTakeover: ?number,
 };
 
 class BidHelpText extends React.PureComponent<Props> {
   render() {
-    const {
-      uri,
-      isResolvingUri,
-      winningBidForClaimUri,
-      myClaimForUri,
-      onEditMyClaim,
-      isStillEditing,
-    } = this.props;
+    const { uri, isResolvingUri, amountNeededForTakeover } = this.props;
+    let bidHelpText;
 
-    if (!uri) {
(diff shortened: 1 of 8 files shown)


src/renderer/component/publishForm/internal/bid-help-text.jsx
```diff
@@ -58,16 +59,35 @@ class BidHelpText extends React.PureComponent<Props> {
...
     }
-
-    return winningBidForClaimUri ? (
+    
+    if (!winningBidForClaimUri && !winndingBidForClaimWithChannel) {
+      return __('Any amount will give you the winning bid.');
+    }
+    
+    const generateTakeOverMessage = (amount, uri) => {
+      return `${__('A deposit greater than')} ${amount} ${__('is needed to win')} ${uri}`
```
tzarebczan: Is "win" the right word here? Maybe control or mention vanity? We use it in a few places...

src/renderer/component/publishForm/internal/bid-help-text.jsx
```diff
@@ -58,16 +59,35 @@ class BidHelpText extends React.PureComponent<Props> {
...
+    }
+    
+    let content;
+    if (winndingBidForClaimWithChannel) {
+      content = `Winning bid is ${winndingBidForClaimWithChannel}`
+    }
+    
+    // <div>
+    //   {__('However, you can still get')}{' '}
+    //   {(winndingBidForClaimWithChannel && winningBidForClaimUri) ? __('these URLs for any amount')}
```
tzarebczan: I'm assuming you'll add this back in :)

src/renderer/component/publishForm/internal/bid-help-text.jsx
seanyesmunt: I like control. I think we can definitely improve this copy.

src/renderer/component/publishForm/internal/bid-help-text.jsx
seanyesmunt: Yes. This PR still needs a lot of cleanup.

seanyesmunt:
@tzarebczan I think this is working correctly. Before cleaning up/renaming stuff could you give this a test?

I'm open to adding new messages/re-wording it but I mainly want to see if it's working.

How it should work:
```
uri on publish page:
lbry://@sean/sean

messages:
A deposit greater than 1.234 is needed to control lbry://sean
Any amount will let you control lbry://@sean/sean
```

If there is no channel in the uri:
```
uri on publish  page:
lbry://sean

messages:
A deposit greater than 1.234 is needed to control lbry://sean
```

If there are no matches for the short uri or uri with channel:
```
uri on publish page:
lbry://@somerandomchannel/somerandomclaim

messages:
Any amount will let you control lbry://somerandomclaim
Any amount will let you control lbry://@somerandomchannel/somerandomclaim
```

tzarebczan:
the messaging about both urls in the last case might be confusing, I think it's enough only to tell them `Any amount will let you control lbry://somerandomclaim` - what do you think? 

`Any amount will let you control lbry://@sean/sean`  what if someone takes over @sean? 

I still haven't run through it, will give that a shot today.

seanyesmunt:
I think it would be ok to only show the short url

tzarebczan:
@seanyesmunt we also discussed showing vanity top bids when editing claims as part of this PR.

src/renderer/component/publishForm/internal/name-help-text.jsx
```diff
@@ -0,0 +1,49 @@
...
+        'You are currently editing this claim. If you change the URL, you will need to reselect a file.'
+      );
+    } else if (uri && myClaimForUri) {
+      const editUri = buildURI({
+        contentName: myClaimForUri.name,
+        claimId: myClaimForUri.claim_id,
+      });
+
+      nameHelpText = (
+        <React.Fragment>
```
skhameneh: Can also use the short syntax (empty tags):
```
<>
  {/* markup here */}
</>
```

seanyesmunt: Do we need to do something with webpack for that to work? Doesn't seem to build properly when I do that.

tzarebczan:
`If you bid more than 110082.30000001 LBC, when somone navigates to lbry://one it will load your published content.`
How can we make it clearer that they can still publish to this URL and have their content at the exact URL? We can't show them the claim id because it's not known until the transaction is created. 

(just noticed the misspelling on `someone`)

seanyesmunt:
@tzarebczan "However, you can get a longer version of this URL for any price"?

tzarebczan:
"However, you can get a longer version of this URL for any bid"?

tzarebczan:
We should also add similar bid help to the channel creation process (this is opened as a separate issue I believe).

PR #1943, opened 2018-09-07
feature: ModalSocialShare

jessopb:
Resubmitted after verifying the bug I was worried about was due to redux lacking my MODAL constant.

This was closely modeled after the modalSendTip and walletSendTip components.
It should shares either anonymous and channel-based spee.ch links
Sharing is available only if the content meets shareable criteria.
It uses share urls from http://www.sharelinkgenerator.com/ to share to Facebook, Twitter, and Google Plus.

There are some improvements that could be made:
- feathericons has no actual google-plus icon, I think
- it could share a tinyurl version
- the formatting and spacing could be better
- the mouseover tooltips could be more responsive: they seem to give up sometimes - ideas?
- the link could come preselected, rather than wait for a click in the text field
- the modal could give feedback that the link copy was successful: 'link copied to clipboard'
- it could close the modal when clicking the background or pressing esc
Also, the ViewOnWebButton component is apparently no longer in use.

I look forward to more guidance in these areas.

### DIFF
--- a/src/renderer/component/socialShare/view.jsx
+++ b/src/renderer/component/socialShare/view.jsx
@@ -1,0 +1,77 @@
+// @flow
+import React from 'react';
+import type { Claim } from 'types/claim';
+import Button from 'component/button';
+import * as icons from 'constants/icons';
+import Tooltip from 'component/common/tooltip';
+import Address from 'component/address';
+
+type Props = {
+  claim: Claim,
+  onDone: () => void,
+};
+
+class SocialShare extends React.PureComponent<Props> {
+  constructor(props: Props) {
+    super(props);
+
+    this.input = undefined;
+  }
+
+  input: ?HTMLInputElement;
+
+  render() {
+    const { claim_id: claimId, name: claimName, channel_name: channelName } = this.props.claim;
+    const { onDone } = this.props;
+    const speechPrefix = 'http://spee.ch/';
+    const speechURL = channelName
+      ? `${speechPrefix}${channelName}/${claimName}`
+      : `${speechPrefix}${claimName}#${claimId}`;
+
+    return (
+      <div>
+        <div className="card__title">
+          <h2>{__('Share This Content')}</h2>
+
+          <div className="card__content">
+            <Address address={speechURL} noSnackbar />
+          </div>
+          <div className="card__actions card__actions--center">
+            <Tooltip onComponent body={__('Facebook')}>
+              <Button
+                iconColor="blue"
+                icon={icons.FACEBOOK}
+                button="alt"
+                label={__('')}
+                href={`https://facebook.com/sharer/sharer.php?u=${speechURL}`}
+              />
+            </Tooltip>
+            <Tooltip onComponent body={__('Twitter')}>
+              <Button
+                iconColor="blue"
+                icon={icons.TWITTER}
+                button="alt"
+                label={__('')}
+                href={`https://twitter.com/home?status=${speechURL}`}
+              />
+            </Tooltip>
+            <Tooltip onComponent body={__('View on Spee.ch')}>
+              <Button
+                icon={icons.GLOBE}
+                iconColor="blue"
+                button="alt"
+                label={__('')}
+                href={`${speechURL}`}
+              />
+            </Tooltip>
+          </div>
+          <div className="card__actions">
+            <Button button="link" label={__('Done')} onClick={onDone} />
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+export default SocialShare;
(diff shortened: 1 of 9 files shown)


src/renderer/component/socialShare/view.jsx
```diff
@@ -0,0 +1,100 @@
...
+              />
+              <Button
+                noPadding
+                button="secondary"
+                icon={icons.CLIPBOARD}
+                onClick={() => {
+                  clipboard.writeText(speechURL);
+                }}
+              />
+            </FormRow>
```
seanyesmunt: Since the above code is pretty much identical to the `Address` component at `component/address/view.jsx` we should probably break this into it's own component.

Maybe `CopyableInput`?

I'm only talking about everything inside of the `FormRow`. It would just need two props (I think), `value` and `copySuccessMessage` (for the snackbar after you copy it)

src/renderer/component/socialShare/view.jsx
```diff
@@ -0,0 +1,100 @@
...
+                href={`https://twitter.com/home?status=${speechURL}`}
+              />
+            </Tooltip>
+            <Tooltip onComponent body={__('Post on Google Plus')}>
+              <Button
+                icon={icons.GOOGLE_PLUS}
+                button="alt"
+                label={__('')}
+                href={`https://plus.google.com/share?url=${speechURL}`}
+              />
```
seanyesmunt: I don't think we need google plus, especially since we don't have an icon for it.

src/renderer/component/socialShare/view.jsx
```diff
@@ -0,0 +1,100 @@
...
+                noPadding
+                button="secondary"
+                icon={icons.CLIPBOARD}
+                onClick={() => {
+                  clipboard.writeText(speechURL);
+                }}
+              />
+            </FormRow>
+          </div>
+          <div className="card__content">
```
seanyesmunt: This should be `card__actions`. It should give the buttons proper spacing.

src/renderer/component/socialShare/view.jsx
jessopb: I originally tried to use <Address /> and it crashed at that point. Using <Address /> now has the odd behavior that it doesn't show the snackbar until after the modal is closed (modal over modal?). I will need to really look into how doNotify on top of doNotify works. Perhaps it's putting the next notification in a queue that doesn't get activated until my modal is closed?

If so, taking doNotify functionality out of Address and making it an optional passed prop might be best?

seanyesmunt: Ah you are right. I forgot about that. I just created an issue for it.
https://github.com/lbryio/lbry-desktop/issues/1945

I'm fine with passing in a prop for now to disable this (it should default to notify) for now. Just include a link to this issue #1945 in a comment above it.

We should probably rename the `Address` component if we are going to use it here.

jessopb:
Ok, I think it's ready to be looked at again.

I'm hesitant to mess too much with `<Address>`, but I added a temporary noSnackbar bool to it (3 ways until the linter stopped complaining). With my change, The wallet copy button throws the snackbar, but the modal copy button does not. Renaming `<Address>` I think is a change with its own scope? 

Messing with Z-index and overflow (and size) in scss for .tooltip_body didn't do anything to help the tooltip clipping. For now I'd like to center it and figure out how to make tooltips play with modals later.

I made the remaining changes of removing google_plus, making the modal close on esc, and changing the style className for better spacing on the buttons.

modalsharecropped

I hate how many steps it is to change a commit message; 
commit message for 434c is supposed to use "bool" not "binary"

seanyesmunt:
I'm ok with the centering of buttons for now. I do they they should be on the bottom though. I'll give this another test/better look tomorrow.

seanyesmunt:
This looks good. I think once the share buttons are moved below the address component this can be merged.

jessopb:
ssa
ssb

Which do you like better?

seanyesmunt:
@kauffj I like the top one more, it is more consistent with other modals in the app. Also I don't think we need the `:` in the title

jessopb:
This seems to match the sendTip modal best: Green buttons are actions, link buttons are exits.
socialsharee

jessopb:
The latest commit looks like this, adding an available blue stroke color to the Icon class.
modalsharelinks

seanyesmunt:
@jessopb This looks great. I will give it a better test later today.

tzarebczan:
Hey @jessopb, thanks again for another successful PR! We appreciate all the contributions, check out https://lbry.io/faq/tips :)

I've added you to our developer mailing list also :)

PR #1944, opened 2018-09-07
modalSendTip closes on esc/click outside

jessopb:
clicking off of send tip now closes the modal.


tzarebczan:
@jessopb thanks for the recent contributions! Please see https://lbry.io/faq/tips :)

PR #1947, opened 2018-09-08
fix: checks tip input re:#1889

jessopb:
Issue #1889 mentions the tip input allowing any number of decimal places, far beyond one satoshi.
This should do it. I'm looking for input on the Error message, though.

I also noticed that the field would fail to complain about 5.5.5.5 as an input. That's a bit better now, though not perfect. 5.5. slips by for some reason, but 5.5.5 throws an error.

### DIFF
--- a/src/renderer/component/walletSendTip/view.jsx
+++ b/src/renderer/component/walletSendTip/view.jsx
@@ -50,6 +50,9 @@ class WalletSendTip extends React.PureComponent<Props, State> {
     const { balance } = this.props;
     const tipAmount = parseFloat(event.target.value);
     let newTipError;
+    if (!String(tipAmount).match(/^(\d*([.]\d{0,8})?)$/)) {
+      newTipError = __('Tip must be a valid number with no more than 8 decimal places');
+    }
     if (tipAmount === balance) {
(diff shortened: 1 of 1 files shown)


src/renderer/component/walletSendTip/view.jsx
```diff
@@ -50,6 +50,9 @@ class WalletSendTip extends React.PureComponent<Props, State> {
     const { balance } = this.props;
     const tipAmount = parseFloat(event.target.value);
     let newTipError;
+    if (!String(tipAmount).match(/^(\d*([.]\d{0,8})?)$/)) {
```
seanyesmunt: I think it would be simpler to just do
```
if (tipAmount < 0.00000001) {

}
```

jessopb:
I appreciate your patience. I agree - it's pretty ugly to cast to string just to use regex. 
I was trying to kill all possible birds, but I may be bikeshedding and overthinking it.
Is there any concern about other invalid inputs, or are those handled on the backend? My console.log tests show other invalid tipAmounts like 2.123456789 (9 decimal places) and 'undefined' (from 5.5.5) being passed as the tipAmount state.

PR #1964, opened 2018-09-13
refactors and enhances tip modal

jessopb:
Tip modal now disables send button on all invalid actions and shows feedback.
tip_
tipequalsbalance
tipgreaterthanbalance
tipmanydots
tipmaxdecimal
tippointfive
tiptoomanydecimals
tiptwenty

### DIFF
--- a/src/renderer/component/walletSendTip/view.jsx
+++ b/src/renderer/component/walletSendTip/view.jsx
@@ -18,7 +18,7 @@ type Props = {
 
 type State = {
   tipAmount: number,
-  newTipError: string,
+  tipError: string,
 };
 
 class WalletSendTip extends React.PureComponent<Props, State> {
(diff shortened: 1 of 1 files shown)


jessopb:
Oh, I remember noticing that before. Modification on line 85.

kauffj:
Not strictly necessary, but splitting the "tip must be a valid number with no more than 8 decimals places" into more messages would improve UX:

- "Tip must be a number" (shown if anything other than a number is entered - can also drop valid, it conveys no additional information)
- "Tip must be positive" (I've now deduced what was meant by the word valid)
- "Tip must not be more than 8 decimal places" (shown if a 9+ decimal number is entered)

jessopb:
I hear you. I'm glad to get specific input on the actual messages.
 I'll have something shortly.

jessopb:
I'm noticing this functionality duplicated in 3 places: Publish>deposit, page>file>...>walletSendTip, and Wallet>walletSend. All three are basically the same FormField that informs a button to be disabled. The form field could stand to be wider as well, I think. Thoughts? Make it component?

seanyesmunt:
I think making the input/label into it's own component sounds like a great idea!

tzarebczan:
@jessopb any updates here? Will the last comments also occur in this PR or a separate one?

jessopb:
The latest commit does precisely what @kauffj suggests.

Issue #1969, opened 2018-09-16
Allow typing of encryption password without clicking entry box


tzarebczan:
## The Issue
When starting up the app with an encrypted wallet, you are prompted for a password. In order to type the password, you must first click the entry box and then you can type.

The app should automatically select the entry box so the password can be typed without having to click anything. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

kauffj:
Implementation suggestion: make this a property of either an input or form, not specific to encryption component.

PR #1977, opened 2018-09-22
AutoFocus form-fields

btzr-io:
### Changes
- Add `autoFocus` prop  of `form-field` component. 
- Auto focus modals form.
- Allow typing of encryption password without clicking entry box #1969


seanyesmunt:
🙂

Issue #1978, opened 2018-09-22
Make tooltip smarter


seanyesmunt:
## The Issue
Currently the tooltip can be cut off if it is near the edge of the screen

<img width="400" alt="screen shot 2018-09-22 at 1 46 20 am" src="https://user-images.githubusercontent.com/16882830/45913865-7fa06000-be09-11e8-825f-e919640b39f1.png">

### Steps to Reproduce
1. Hide NSFW content
2. Scroll to bottom of explore
3. Hover over "whats this" tooltip

### Expected Behaviour
All of the tooltip should be visible

### Suggested Solutions
The quick fix for this one is to always show the tooltip ontop with `direction="top"`
The more long term fix is to improve the tooltip component. If any of the tooltip is cutoff, it should move to the opposite side

PR #1979, opened 2018-09-23
Smart tooltip

btzr-io:
### Changes
- Make tooltip smarter #1978

### DIFF
--- a/src/renderer/component/common/tooltip.jsx
+++ b/src/renderer/component/common/tooltip.jsx
@@ -68,6 +68,7 @@ class ToolTip extends React.PureComponent<Props> {
     const { direction } = this.state;
     const visibility = this.getVisibility();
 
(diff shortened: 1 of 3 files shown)


src/renderer/component/common/tooltip.jsx
```diff
@@ -16,15 +16,75 @@ class ToolTip extends React.PureComponent<Props> {
...
+    if (inverted) {
+      this.setState({ direction: inverted });
+    }
+  };
+
+  handleVisibility = () => {
+    const { direction } = this.state;
+    const visibility = this.getVisibility();
+
+    if (!visibility[direction]) {
```
seanyesmunt: Could you add a short comment above this explaining why we might need to invert? The code is very readable but it wouldn't hurt to add some additional explanation.

PR #1980, opened 2018-09-24
Add proper html title to card/tile components for long titles

seanyesmunt:
A lot of titles were cut off with no way to see the entire thing without navigating to the claim. Now it will show if you hover over it.


kauffj:
@seanyesmunt if this is every instance of `TruncatedText`, it looks like you could have a single prop called `text` rather than using both `title` and `props.children`.

seanyesmunt:
@kauffj That's a good point. I'll change it.

PR #1981, opened 2018-09-24
hotfix: share is broken; use currently working spee.ch share url

jessopb:
Share on spee.ch urls using channelName need clarification and testing. Revert back to working url without channelName.

Must confirm with spee.ch group how to properly format urls using channelName.
If someone can get a url using channelName to work consistently, I can put that format in.

### DIFF
--- a/src/renderer/component/socialShare/view.jsx
+++ b/src/renderer/component/socialShare/view.jsx
@@ -21,12 +21,20 @@ class SocialShare extends React.PureComponent<Props> {
   input: ?HTMLInputElement;
 
   render() {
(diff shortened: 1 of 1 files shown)


jessopb:
There we go. Thanks for the guidance.

tzarebczan:
Sweet, nice fix. I was going to comment about that after noticing that myself over the weekend. 

I'd also like to see the open.lbry.io URL as a share option - especially since not all claims can be viewed on spee.ch currently. I'll reopen the original issue as I think that was mentioned there.

PR #1985, opened 2018-09-25
Update _card.scss

ykris45:
Invites section doesn't expand past a certain height
https://media.discordapp.net/attachments/363087331475062785/493920826505035776/unknown.png?width=693&height=428
Displays correctly when position is set from relative to initial in the CSS.

### DIFF
--- a/src/renderer/scss/component/_card.scss
+++ b/src/renderer/scss/component/_card.scss
@@ -4,7 +4,6 @@
   border-radius: var(--card-radius);
   user-select: text;
   display: flex;
-  position: relative;
   flex-direction: column;
   max-width: var(--card-max-width);
 }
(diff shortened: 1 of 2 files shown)


src/renderer/scss/component/_card.scss
```diff
@@ -4,7 +4,7 @@
   border-radius: var(--card-radius);
   user-select: text;
   display: flex;
-  position: relative;
+  position: initial;
```
btzr-io: maybe if you just remove this line should also work?

seanyesmunt: That would be ideal if it works.

ykris45: @seanyesmunt yes it works

ykris45: image

seanyesmunt:
@ykris45 We use the `.card` class in a lot of places. Have you tested that this doesnt affect other parts of the app? I will give it a test in a little bit.

ykris45:
@seanyesmunt i doesnt have a lot of referalls. and i will test it on locally.

PR #1986, opened 2018-09-26
Use video.js for audio/video

seanyesmunt:


seanyesmunt:
The new video player is  added for web so I'm closing this. We still need to strip out the old player for electron once the sdk provides streaming support.

PR #1998, opened 2018-10-03
Support for lbrynet daemon v0.30.0

eukreign:
@skhameneh This is ready for a review. @akinwale I would appreciate your eyes too to see if I missed anything else needed for v30. 

Editing claims aren't working until https://github.com/lbryio/lbry/issues/1555 is merged. I _think_ it will work as is. I would like to merge it soon so we can get an RC out and begin testing.

I was able to remove a lot of logic for tracking pending publishes since we can use the new `confirmations` field. It makes things a lot simpler.

Requires https://github.com/lbryio/lbry-redux/pull/79 and https://github.com/lbryio/lbryinc/pull/11 but you can run this branch as is (i'm just pointing at the commits from my PR)


akinwale:
@eukreign You can make use of
`parseFloat(amount).toFixed(precision || 1).replace(/\.?0+$/, '');`
This is with the assumption that the amount passed is a valid numeric value.

seanyesmunt:
I'll try finishing this up

src/renderer/component/selectChannel/view.jsx
```diff
@@ -159,7 +159,6 @@ class ChannelSection extends React.PureComponent<Props, State> {
           <FormField
             key="channel"
             type="select"
-            tabIndex="1"
```
skhameneh: What is this change fixing?

seanyesmunt: It wasn't doing anything so I removed it.

PR #1999, opened 2018-10-04
Style improvements and user history bugs

seanyesmunt:
#### Made some general style changes
- Changed text color for easier reading
  - Black text is no longer pure black, changed it to #&#xfeff;444
  - White text is no longer pure white, changed it to #efefef
- Defaulted `font-family` to `metropolis-medium`
  - This was being used in more places than `metropolis-semibold` so it makes sense to default to this

#### User history changes
- Down with tables!
- Created new css class family `item-list` for vertical lists of items
- Fixed the issue where the layout was wonky on large screens (due to table weirdness)
- Fixed link issue when navigating to file page

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -40,6 +40,8 @@
     "react/require-default-props": 0,
     "react/jsx-closing-tag-location": 0,
     "jsx-a11y/no-noninteractive-element-to-interactive-role": 0,
-    "class-methods-use-this": 0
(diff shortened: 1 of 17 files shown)


tzarebczan:
Can we improve the view when there is no history?  "You don't have anything saved in history yet, go check out some content on LBRY!" ? 
image

seanyesmunt:
Yes good catch

PR #2003, opened 2018-10-04
add {cmd,ctrl} + l for wunderbar focus

seanyesmunt:
[] - tested on windows
[] - tested on linux
[x] - tested on mac

Also got rid of prop types and unused files


src/renderer/component/wunderbar/internal/autocomplete.jsx
```diff
@@ -45,129 +44,129 @@ function getScrollOffset() {
...
-     * Used to override the internal logic which displays/hides the dropdown
-     * menu. This is useful if you want to force a certain state based on your
-     * UX/business logic. Use it together with `onMenuVisibilityChange` for
-     * fine-grained control over the dropdown menu dynamics.
-     */
-    open: PropTypes.bool,
-    debug: PropTypes.bool,
-  };
+  // static propTypes = {
+  //   /**
```
skhameneh: Why is this commented out?

src/renderer/component/wunderbar/view.jsx
```diff
@@ -16,15 +19,23 @@ type Props = {
...
     (this: any).handleChange = this.handleChange.bind(this);
-    this.input = undefined;
+  }
+
+  componentDidMount() {
+    window.addEventListener('keydown', this.handleKeyDown.bind(this));
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('keydown', this.handleKeyDown);
```
skhameneh: Does this correctly remove the event?  Using `bind` above creates a handle that isn't stored.

src/renderer/component/wunderbar/internal/autocomplete.jsx
seanyesmunt: I wanted to get rid of `prop-types` since we aren't using it. This is from a third party library but we needed to make a change to the component so I just copied it in.

I still wanted to keep them for documentation.

src/renderer/component/wunderbar/view.jsx
seanyesmunt: 👍 I'll move the `bind` into the constructor and use `this.handleKeyDown` for both.

kauffj:
Love the addition keyboard shortcuts.

This isn't a PR blocker and can be considered more for the next round/addition, but if keyboard shortcuts are intended to be global, a design that allows them to be cleanly separated and grouped is likely preferable to one that puts them throughout individual components.

Issue #2000, opened 2018-10-04
Transactions filter menu collides with transaction table


seanyesmunt:
## The Issue
Transactions filter menu collides with transaction table on small screens. This must have happened after https://github.com/lbryio/lbry-desktop/pull/1985 was merged.

### Steps to Reproduce
1. View transactions page
2. Shrink screen until the filter menu overlaps with the table

<img width="400" alt="screen shot 2018-10-03 at 10 41 18 pm" src="https://user-images.githubusercontent.com/16882830/46450124-ae99c880-c75d-11e8-83fd-403f4261249d.png">

### Suggested Solutions
It currently uses `position: absolute` to achieve the positioning. Ideally the filter menu should just appear first in the dom so we don't need to hoist it up.

It should probably be in line with the title, just moved over to the right side.

estevanjantsk:
nice! I'll try that one 👍

seanyesmunt:
Awesome! Let me know if you have any more questions 🙂

ykris45:
@seanyesmunt i tested it with position relative and the Transactions filter menu uncolides

ykris45:
close close..hehehe

seanyesmunt:
@estevanjantsk We found a fix for this issue but I still think the solution we discussed would be preferable. If you are still working on it feel free to submit a PR!

seanyesmunt:
I'm going to close this for now

PR #2005, opened 2018-10-04
Fix Transaction menu colides [#2000]

ykris45:
i tested on my local development 
image
if the position unchecked on my inspect element i will looks like
image
as @seanyesmunt said on issues #2000 ..


seanyesmunt:
@ykris45 I belive this messes with the invite table. Isn't that why you removed it originally?

ykris45:
@seanyesmunt on invite table it use same element with transaction table, i added position: relative; to tis element.

seanyesmunt:
Based on #1985 

This will break the invite table. I think a better fix is described in #2000

ykris45:
@seanyesmunt 
image i think it wouldnt break the invite table

ykris45:
https://css-tricks.com/absolute-relative-fixed-positioining-how-do-they-differ/

ykris45:
image
@seanyesmunt it will show based on my screenshot

ykris45:
@seanyesmunt thanks brow

Issue #1859, opened 2018-08-06
Make "new content" in subscriptions stand out more


seanyesmunt:
## The Issue
Make "new content" in subscriptions stand out more
add publish date to the tile

### Steps to Reproduce
1. Subscribe to a channel
2. Wait for new content to be posted from that channel
3. Navigate to the subscriptions page

### Expected Behaviour
The new content should be more prominent

### Actual Behaviour
They are at the top of the list, but there is no way to tell what is new

### Suggested Solutions
Some sort of top level header "This is new"

tzarebczan:
Notes from Slack:
summary of subscription discussion:
- show the age of the content on the tile card (`x time units old`) similar to youtube
- make sure there is a clear indicator of what's "new" - there is a number in parentheses next to subscriptions, I should easily be able to tell what this corresponds to after clicking
- also make sure the badge number has consistent/understandable behavior
- consider allowing sorting by both date and channel
- consider extending the visual indicator for "you're subbed to this and haven't seen it" to be used throughout the entire app, not just the subscription area
- show your subscription status in more places (most likely, show the heart on all file tiles next to the channel name if you are subbed, regardless of where they are rendered)
- if we're emailing people about subscriptions, consider how and where we tell people this, as well as how they can disable this (most likely by linking them to the new email management page)

One more: support the competitionist use case, not just the "find interesting stuff" use-case. Subscriptions should facilitate me not missing a publish from my favorite creator if I want to use it that way.

tzarebczan:
Possibly consider https://github.com/lbryio/lbry-desktop/issues/1605 to go along with these changes.

PR #2011, opened 2018-10-05
Simplify FileCard & FileTile styles

neb-b:
Working on this before #1859 since we will need to make some ui changes around the file card fo displaying stuff like the publish date.

Trying to simplify `FileCard` and `FileTile` components so that there is only one top-level class ex: `.file-tile--small` and everything else adjusts accordingly. Removing all of the sub-level size specific classes ex: `card__title--small card__title--large`, now there will just be one `card__title` and it will adjust based on the top level sizing ex: `card--small`

Some testing needs to be done with prices being displayed across the app


kauffj:
Beautiful!

PR #2022, opened 2018-10-10
Add content placeholders and add time info to FileCard/FileTile

seanyesmunt:
### Changes
- Added placeholder styles for all file cards and tiles (related content, search, channels)
- Added the published date (based on block height) to every file card and tile
- Added back the upgrades to the share modal (I accidentally removed it)

### Pics
screen shot 2018-10-10 at 1 25 20 am
screen shot 2018-10-10 at 1 18 33 pm

screen shot 2018-10-10 at 1 25 34 am

### DIFF
--- a/src/renderer/component/fileCard/view.jsx
+++ b/src/renderer/component/fileCard/view.jsx
@@ -10,6 +10,7 @@ import * as icons from 'constants/icons';
 import classnames from 'classnames';
 import FilePrice from 'component/filePrice';
 import { openCopyLinkMenu } from 'util/contextMenu';
+import DateTime from 'component/dateTime';
 
 type Props = {
   uri: string,
@@ -24,15 +25,11 @@ type Props = {
   /* eslint-disable react/no-unused-prop-types */
   resolveUri: string => void,
   isResolvingUri: boolean,
-  showPrice: boolean,
   /* eslint-enable react/no-unused-prop-types */
+  subscribed: boolean,
 };
 
 class FileCard extends React.PureComponent<Props> {
-  static defaultProps = {
-    showPrice: true,
-  };
-
   componentWillMount() {
     this.resolve(this.props);
   }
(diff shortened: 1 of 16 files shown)


src/renderer/component/fileCard/view.jsx
```diff
@@ -59,9 +56,20 @@ class FileCard extends React.PureComponent<Props> {
...
+      subscribed,
     } = this.props;
 
+    if (!claim) {
+      return (
+        <div className="card card--small">
+          <div className="card--placeholder card__media" />
+          <div className="card--placeholder placeholder__title" />
+          <div className="card--placeholder placeholder__channel" />
+          <div className="card--placeholder placeholder__date" />
```
seanyesmunt: Should this be its own component? I'm not sure.

I think it could be helpful to see the classes that are used below to keep the layout consistent

skhameneh: Doesn't contain a lot, do you expect to reuse this within other components?

skhameneh: I see similar below, but it's not the same.

src/renderer/component/fileTile/view.jsx
```diff
@@ -108,43 +134,38 @@ class FileTile extends React.PureComponent<Props> {
...
-              )}
+              {__('This location is unused.')}{' '}
+              <Button
+                button="link"
+                label={__('Put something here!')}
+                onClick={e => {
+                  // avoid navigating to /show from clicking on the section
+                  e.stopPropagation();
+                  // strip prefix from the Uri and use that as navigation parameter
+                  const nameFromUri = uri.replace(/lbry:\/\//g, '').replace(/-/g, ' ');
```
skhameneh: Why is this logic here?  Seems it should be shared logic

src/renderer/scss/component/_placeholder.scss
```diff
@@ -0,0 +1,38 @@
...
+  background: var(--color-placeholder);
+}
+
+// Individual items we need a placeholder for
+// FileCard
+.placeholder__title {
+  margin-top: $spacing-vertical * 1/3;
+  height: 3em;
+}
+.placeholder__channel {
```
skhameneh: line break

src/renderer/component/fileTile/view.jsx
seanyesmunt: 👍 We should probably use `const { name } = parseUri(uri)`

seanyesmunt:
@skhameneh Can't respond to your comment inline, but I don't expect to use those placeholders in other components. They are very specific to FileCard/FileTile.

PR #2026, opened 2018-10-11
fix 1972 - invalid characters in URL

tzarebczan:
Adds a snackbar if the lbry:// URL is invalid
I tried to do the doNotify similar to most of the other pages, but failed. If this is not the right way to go about it, I'd love to learn because we do it 4 different ways haha.

### DIFF
--- a/src/renderer/component/wunderbar/view.jsx
+++ b/src/renderer/component/wunderbar/view.jsx
@@ -1,7 +1,7 @@
 // @flow
 import React from 'react';
 import classnames from 'classnames';
-import { normalizeURI, SEARCH_TYPES } from 'lbry-redux';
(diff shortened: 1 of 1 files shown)


seanyesmunt:
[Right here](https://github.com/lbryio/lbry-desktop/blob/master/src/renderer/component/wunderbar/index.js#L40) you will want to do something like [this](https://github.com/lbryio/lbry-desktop/blob/22207d2636851bd4b60d45b1af0b3f329067ed9b/src/renderer/component/externalLink/index.js#L9).

Then we have access to `doNotify` (or whatever you call it, but it shouldn't be `doOpenModal` in `this.props`, i.e. `this.props.doShowSnackBar()`

PR #2030, opened 2018-10-11
fix thumbnail height/width

seanyesmunt:
Messed it up when I was refactoring the filecard

The thumbnail component is different from `.card__media` where it is a static height/width

### DIFF
--- a/src/renderer/scss/_gui.scss
+++ b/src/renderer/scss/_gui.scss
@@ -393,4 +393,7 @@ p {
 .thumbnail-preview {
   height: var(--thumbnail-preview-height);
   width: var(--thumbnail-preview-width);
+  background-size: cover;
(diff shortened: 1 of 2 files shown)


src/renderer/scss/_gui.scss
```diff
@@ -393,4 +393,7 @@ p {
 .thumbnail-preview {
   height: var(--thumbnail-preview-height);
   width: var(--thumbnail-preview-width);
+  background-size: cover;
+  background-repeat: no-repeat;
+  background-position: 50% 50%;
```
skhameneh: Is this line needed?

seanyesmunt: Yeah. It makes sure the image is always centered.

PR #2035, opened 2018-10-13
Proposed: Centralized redux action

chrisza4:
While I work with flow type, I found that in the codebase there are multiple places that define Redux Dispatching type. 

So I took the liberty to centralised redux dispatch type definition into one single place.

If I break any codebase rules or established pattern, please let me know and you can close this pull request.


seanyesmunt:
This is great!

tzarebczan:
@chrisza4 thanks again for the contributions! Send us time estimates for these and we'll get some [LBC over to you as appreciation for your time](https://lbry.io/faq/appreciation)

Issue #2024, opened 2018-10-10
Flow type errors


seanyesmunt:
We are currently trying to use Flow in every file. We have been slowly making progress but there are still a lot of errors.

If you open a PR with some Flow error fixes (or anything else) we will send you some LBC as appreciation!

Some of the biggest offenders right now are in `pages/file/view.jsx` and anything with subscriptions. 

To see all of the errors you can run `./node_modules/.bin/flow  --color always`

izanf:
I will work on this issue

seanyesmunt:
@izanf and @feralCode there are a ton of issues that can be fixed. I would open a PR as you fix them (or leave another comment here) saying which component/file you are fixing so no work is duplicated.

chrisza4:
While I am working on this issue, I also need to fix flow type error at `publishForm/view.jsx`. I am working on that file.

dannycalleri:
Hi @seanyesmunt , I'm opening a PR to fix flow errors in a few `view.jsx` files: 
`renderer/component/channelTile/view.jsx`
`renderer/component/button/view.jsx`
`...`

PR #2038, opened 2018-10-14
Fix flow errors of fileRender component

btzr-io:
Small Hacktober fest contribution: Flow type errors #2024


seanyesmunt:
Nice 🙂

PR #2039, opened 2018-10-15
Fix flow type errors of all viewers

btzr-io:
Hacktober fest contribution: Flow type errors #2024

### DIFF
--- a/src/renderer/component/viewers/threeViewer/index.jsx
+++ b/src/renderer/component/viewers/threeViewer/index.jsx
@@ -34,10 +34,11 @@ type State = {
 };
 
 class ThreeViewer extends React.PureComponent<Props, State> {
-  static testWebgl = new Promise((resolve, reject) => {
-    if (detectWebGL()) resolve();
-    else reject();
-  });
+  static testWebgl = (): Promise<void> =>
+    new Promise((resolve, reject) => {
+      if (detectWebGL()) resolve();
+      else reject();
+    });
 
   static createOrbitControls(camera: any, canvas: any) {
     const controls = new THREE.OrbitControls(camera, canvas);
@@ -92,8 +93,6 @@ class ThreeViewer extends React.PureComponent<Props, State> {
   constructor(props: Props) {
     super(props);
     const { theme } = this.props;
-    this.viewer = React.createRef();
-    this.guiContainer = React.createRef();
     // Object defualt color
     this.materialColor = '#44b098';
     // Viewer themes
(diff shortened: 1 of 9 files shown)


btzr-io:
@seanyesmunt I still don't know how handle `react.createRef` errors, any ideas?
https://flow.org/en/docs/react/refs/

src/renderer/component/viewers/threeViewer/index.jsx
```diff
@@ -187,7 +193,37 @@ class ThreeViewer extends React.PureComponent<Props, State> {
     }
   }
 
-  transformGroup(group) {
+  // Define component types
+  theme: viewerTheme;
+  themes: { dark: viewerTheme, light: viewerTheme };
+  materialColor: string;
+  // Too complex to add a custom type
+  // Posible solution: https://www.npmjs.com/package/@types/three
```
btzr-io: Should we use this instead of `any`?
 https://www.npmjs.com/package/@types/three

seanyesmunt: Nice. Definitely

btzr-io: Looks like they are `TypeScript` type definitions :disappointed:

btzr-io: I can't find anything for flow :stuck_out_tongue:

seanyesmunt:
@btzr-io I'm not sure. You can just stick with assigning it in the callback

```
// constructor
this.textArea = null

// render
<textarea ref={input => this.textArea = input} />

```

src/renderer/component/viewers/threeViewer/index.jsx
btzr-io: I can't use / import it:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/three/index.d.ts

seanyesmunt: I gotcha. Oh well.

btzr-io:
@chrisza4 I keep getting this errors:
```
Cannot call React.createRef because property createRef is missing in object type [1].

     src/renderer/component/viewers/codeViewer.jsx
      29| class CodeViewer extends React.PureComponent<Props> {
      30|   constructor(props: Props) {
      31|     super(props);
      32|     this.textarea = React.createRef();
      33|     this.codeMirror = null;
      34|   }
      35|
```
```
Cannot create textarea element because a callable signature is missing in object type [1] but exists in function
type [2] in property ref.

     src/renderer/component/viewers/codeViewer.jsx
 [1]  57|   textarea: { current: null | HTMLTextAreaElement };
      58|   codeMirror: any;
      59|
      60|   render() {
      61|     const { value } = this.props;
      62|     return (
      63|       <div className="code-viewer" onContextMenu={stopContextMenu}>
      64|         <textarea ref={this.textarea} disabled value={value} />
      65|       </div>
      66|     );
      67|   }

```

Issue #1974, opened 2018-09-19
Obstructive notifications (snackbar)


btzr-io:
## The Issue
Snackbar notifications obstruct some elements on the header component.
~Also would be nice to have an option to remove it without having to wait.~

### Steps to Reproduce
1. Trigger a snackbar notification (click on the copy wallet address button)

### Expected Behaviour
Non-obstructive notifications.

## Screenshots

![snackbar](https://user-images.githubusercontent.com/14793624/45723967-45b52c80-bb71-11e8-95bc-d4aa709c9c73.png)

btzr-io:
@seanyesmunt some good suggestions:
https://material.io/design/components/snackbars.html#placement

sanketsingh24:
If this issue is free, I would like to work on this, but would need some more info on where to place the snackbar, I think adding a close button would be necessary.

kauffj:
Material specifically does not have close buttons. Is there something unique about our app that requires them vs. Google which does not? Or do we think Google is wrong in not having them?

Unless we're answering in the affirmative to one of those questions, it is probably better to start with different/smarter placement and consider tweaking timing or usage if people feel a need to close.

sanketsingh24:
Usually, snackbar is placed on bottom-right or bottom-left, the reason being the fact that people usually try to keep the relevent info nearer to top of the screen, and hence the bottom side of website is not used. Also, size and timing can be tweaked.

seanyesmunt:
I think if we move the snackbar to the bottom left, a close button would not be necessary. The only reason an argument could be made to have one now is because it covers the search bar.

btzr-io:
Yeah, we don't need the close button if isn't obstructing important stuff :+1:

sanketsingh24:
Bottom-left is cool but might hide some menu buttons, which is my only concern.

kauffj:
Just noticed Gmail snackbars do have close buttons.

sanketsingh24:
Well, they have the same issue I had concerns with, and the snackbar disappear upon clicking any action button.

btzr-io:
@sanketsingh24  Are you still working on this?

PR #2040, opened 2018-10-15
Change snackbar notification postion to bottom-left

btzr-io:
### Fixes
- Obstructive notifications (snackbar) #1974

### DIFF
--- a/src/renderer/scss/component/_snack-bar.scss
+++ b/src/renderer/scss/component/_snack-bar.scss
@@ -12,7 +12,7 @@
   opacity: 0.95;
   padding: 14px 20px 10px 20px;
   position: fixed;
-  right: $spacing-vertical;
+  left: $spacing-vertical;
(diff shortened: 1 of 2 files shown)


src/renderer/scss/component/_snack-bar.scss
```diff
@@ -5,16 +5,15 @@
...
   margin-left: auto;
   margin-right: auto;
   max-width: var(--snack-bar-width);
   min-width: 300px;
   opacity: 0.95;
   padding: 14px 20px 10px 20px;
   position: fixed;
-  right: 0;
-  top: $spacing-vertical;
+  right: $spacing-vertical;
```
btzr-io: Should it be `bottom-left` or `bottom-right` ?
I think in the right looks good :thinking:

seanyesmunt: I think I like the left side more

Issue #1933, opened 2018-09-04
[Feature Request] Progress overview of ongoing downloads


m160:
## The Issue
When downloading a lot of content at the same time there's no convenient way to check the progress of each download. You'd have to make a mental note of the content you are currently downloading and then go to each individual claim to see the progress.

### Suggested Solutions
A downloads button similar to what you can find Firefox and other browsers would make it a lot easier to check the status of your ongoing downloads.

![](https://2.bp.blogspot.com/-pjn0wvYUwbk/T5FWuH_MLUI/AAAAAAAAIjA/FC-5PRB3gOg/s1600/firefox-nightly-download-button.png)

seanyesmunt:
This is a great idea.

I think there can be two approaches:
- Adding the downloaded percentage to the `FileCard` component in "Your Downloads" (and possibly subscriptions too)
- Adding this new drop-down component for quick access to your x latest downloads

PR #2041, opened 2018-10-15
Add download percentage to file card

chrisza4:
Based on this [issue](https://github.com/lbryio/lbry-desktop/issues/1933):

Add percentage download to file card in case of downloading file
screen shot 2561-10-15 at 17 18 51

I wonder if we should have new icon for these downloading files? Using LOCAL icon for downloading content may mislead user and a little bit confused in term of user interface.

Also:
- Since we have places that need file progress calulcation, I created function `calculateDownloadProgress` and let all logic refer to there.
- Add flowtype for `common/tooltip.jsx`
- Add flowtype for `FileInfo`. We have a lot of components using `FileInfo` type

### DIFF
--- a/src/renderer/component/cardMedia/view.jsx
+++ b/src/renderer/component/cardMedia/view.jsx
@@ -4,17 +4,19 @@ import classnames from 'classnames';
 
 type Props = {
   thumbnail: ?string, // externally sourced image
+  opaque?: boolean,
 };
 
 class CardMedia extends React.PureComponent<Props> {
   render() {
-    const { thumbnail } = this.props;
+    const { thumbnail, opaque } = this.props;
 
     return (
       <div
         style={thumbnail ? { backgroundImage: `url('${thumbnail}')` } : {}}
         className={classnames('card__media', {
           'card__media--no-img': !thumbnail,
+          'card__media--opaque': opaque,
         })}
       >
         {!thumbnail && <span className="card__media-text">LBRY</span>}
--- a/src/renderer/component/common/tooltip.jsx
+++ b/src/renderer/component/common/tooltip.jsx
@@ -20,9 +20,8 @@ class ToolTip extends React.PureComponent<Props, State> {
     direction: 'bottom',
   };
 
-  constructor(props) {
+  constructor(props: Props) {
     super(props);
-    this.tooltip = React.createRef();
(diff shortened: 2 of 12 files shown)


seanyesmunt:
@chrisza4 Thanks for another PR. This looks great and seems to work well. I think we should something next to the percentage that shows what it is. Maybe the download icon? I also don't think we need the `(``)` that wraps it.

You can use `.card--space-between` so that the percentage and icon is always on the right side, just above the other icons. I think that would look good.

seanyesmunt:
Actually I think this should just be at the very bottom of the card. Then we don't have to deal with very long titles pushing the download progress to the next row.

alyssaoc:

chrisza4:
@seanyesmunt I have try different layout and here are results:

After the icons:
screen shot 2561-10-21 at 12 57 47

I think this look weird.

Before the icons:
screen shot 2561-10-21 at 13 05 39

This look okay but once there are many icons it force the time indicator to have 2 rows.

Personally I think this look best:
screen shot 2561-10-21 at 13 10 17

I will commit this layout to this branch first. Let's see your opinion on it.

chrisza4:
Also let me push another idea.

In My LBRY page I think it is better to add some opacity into downloading media since it cannot be viewed yet. The result look like this:

screen shot 2561-10-21 at 13 35 06

What do you think?

seanyesmunt:
Hey @chrisza4. I think we can move it to the very very bottom (in it's own row). I agree it looks kind of weird when there are a lot of icons and text has to wrap.

Also I don't think we need the opacity, you can actually view items before they are finished downloading.

kauffj:
What if we eliminate the folder while it is downloading, or combine them somehow? Currently, the folder convey no information to the user while the percentage is there. Only the folder absent a percentage is conveying information (which feels somewhat absurd when spelled out that way)

chrisza4:
@seanyesmunt Yes. I will finish this up. Last few weeks just get very busy.

seanyesmunt:
@chrisza4 I'm going to close this as I do some repo cleanup (there have been quite a few commits since this was opened so there are a few conflicts). Please re-open if/when you fix the merge conflicts and finish this up! I think it would be a nice addition. 🙂

Issue #1860, opened 2018-08-06
file_list call continues indefinitely if a file is removed while downloading


tzarebczan:
## The Issue
If a user is in the middle of a download and deletes the file (trying to stop the download), the file_list call that we use to check progress continues to be called indefinitely. The result is null ( `"result": []`)

This will bog down their app util it's refreshed/restarted. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

PR #2042, opened 2018-10-15
Fix issue file_list call continues indefinitely if a file is removed while downloading

chrisza4:
Based on this [issue](https://github.com/lbryio/lbry-desktop/issues/1860)

Since when ACTIONS.FILE_DELETE dispatch in `lbry-redux`, we actually remove the outpoint from store. So in order to update status, it makes sense to check if the outpoint still exists first before we dispatch update.

Also, centralized the update interval function.

### DIFF
--- a/src/renderer/redux/actions/content.js
+++ b/src/renderer/redux/actions/content.js
@@ -3,10 +3,7 @@ import * as NOTIFICATION_TYPES from 'constants/notification_types';
 import { ipcRenderer } from 'electron';
 import { doAlertError } from 'redux/actions/app';
 import { doNavigate } from 'redux/actions/navigation';
-import {
-  setSubscriptionLatest,
-  setSubscriptionNotification,
-} from 'redux/actions/subscriptions';
+import { setSubscriptionLatest, setSubscriptionNotification } from 'redux/actions/subscriptions';
 import { selectNotifications } from 'redux/selectors/subscriptions';
 import { selectBadgeNumber } from 'redux/selectors/app';
 import {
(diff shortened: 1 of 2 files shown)


seanyesmunt:
Another one! Could you add a comment above `setNextStatusUpdate` saying why we need to check if outpoint exists? I think that would be helpful in the future. Then just add a changelog update and it's ready to merge.

alyssaoc:

chrisza4:
Changelog updated, and also comment added.

src/renderer/redux/actions/content.js
```diff
@@ -31,17 +29,24 @@ import analytics from 'analytics';
 
 const DOWNLOAD_POLL_INTERVAL = 250;
 
-export function doUpdateLoadStatus(uri, outpoint) {
+export function doUpdateLoadStatus(uri: string, outpoint: string) {
   return (dispatch, getState) => {
+    const setNextStatusUpdate = () =>
+      setTimeout(() => {
+        // We need to check if outpoint still exists first because user are able to delete file (outpoint) while downloading.
+        // If fiel is already deleted, no point to still try update load status
```
seanyesmunt:
```suggestion
        // If a file is already deleted, no point to still try update load status
```

chrisza4:
Thanks for fixing the typo

tzarebczan:
@chrisza4 , please don't forget to reach out to us via email so we can get some LBC over to you and send you the form for the Hacktoberfest t-shirt.

Issue #1851, opened 2018-08-02
Pre-cache next page when viewing channels


tzarebczan:
## The Issue
The experience when switching between pages in a channel has improved where pages load fairly quickly, but I think we can go another step further to make it even more seamless. If a user goes to a channel, we should call the next page (or maybe two/three pages forward?) so when they switch, it will already have all the next data. Same would happen on page 2 (get page 3). 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

dan1d:
@tzarebczan  Working on this one!

dan1d:
Thanks @seanyesmunt , PR sent!

tzarebczan:
Not required anymore, we'll make sure the API can return this very quickly!

PR #2044, opened 2018-10-15
Add next page cache on channels.

dan1d:
This PR add a prefetch and caches the response when the user visits the channel page.
It should close issue #1851

### DIFF
--- a/src/renderer/page/channel/index.js
+++ b/src/renderer/page/channel/index.js
@@ -23,7 +23,7 @@ const select = (state, props) => ({
 });
 
 const perform = dispatch => ({
-  fetchClaims: (uri, page) => dispatch(doFetchClaimsByChannel(uri, page)),
+  fetchClaims: (uri, page) => dispatch(doFetchClaimsByChannel(uri, page, true)),
   fetchClaimCount: uri => dispatch(doFetchClaimCountByChannel(uri)),
   navigate: (path, params) => dispatch(doNavigate(path, params)),
 });
--- a/src/renderer/redux/actions/content.js
+++ b/src/renderer/redux/actions/content.js
@@ -26,6 +26,7 @@ import {
   creditsToString,
 } from 'lbry-redux';
 import { makeSelectClientSetting, selectosNotificationsEnabled } from 'redux/selectors/settings';
+import { makeSelectTotalPagesForChannel } from 'redux/selectors/content';
 import setBadge from 'util/setBadge';
 import setProgressBar from 'util/setProgressBar';
 import analytics from 'analytics';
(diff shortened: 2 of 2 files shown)


alyssaoc:

dan1d:
Hi @seanyesmunt, I think we need a new state because we don't want to re-fetch the same page again, going back and forth right ?, so somehow we need to keep track of the page numbers that were already been fetched correct?, currently I've added a call to fetch the page after fetching the current page.  Also I've added a function to fetch the next available page to fetch, for example if you click on the page 1 the first time it will fetch 2 and 3, if you click on the page 2, then It will try to fetch page 4 and 5, that's why the state is needed.

Let me know what do you think!

seanyesmunt:
If we go to page 1 (and fetch page 2/3 in the background), it's ok to fetch them again when we get to that page. What if the content was updated? The main goal with this is to have content appear on the screen faster when going to other pages.

dan1d:
cool @seanyesmunt , I will work on that this night.

dan1d:
Hi @seanyesmunt , I've just updated the code! Thanks for the insight I was overthinking a lot on this one. 👍

seanyesmunt:
@dan1d This is looking better but I still think it can be simplified.

What if `doFetchClaimsByChannel` took a third argument `cacheNextPage`.

If that argument is passed in, you can check if there is an additional page and then call `doFetchClaimsByChannel` from inside the function with the next page.

```
    // Somewhere else
    doFetchClaimsByChannel(uri, 1, true)

    function doFetchClaimsByChannel(uri, page, cacheNextPage) {
      ...
      if (cacheNextPage && page + 1 < totalPages) {
        doFetchClaimsByChannel(uri, page)
      }
    }

```

seanyesmunt:
Doing that would allow opting into this behavior too

tzarebczan:
@dan1d see above, let us know when you think and when you can get this fixed up :)

dan1d:
Today will be ready, I'm going to implement the recursive changes that you
asked!

dan1d:
@seanyesmunt  Code updated! Thanks for the insight on the recursive function, I'm working with Ruby/Ruby on Rails currently so I'm a bit rusty with React/Redux/JS.

seanyesmunt:
@dan1d This is looking good. When testing I see that there are 3 `claim_list_by_channel` calls. I see one for `page: "1"`, one for `page: "1"`, and a third one with no params (I'm not sure where that is coming.)

Could you check it out? I'm not sure if it's your new code or not but we shouldn't be making that third call.

seanyesmunt:
Also I rebased and pulled in the latest changes from master, so you should pull from your fork. You might notice few minor changes from recent PR's

seanyesmunt:
I'm going to close this while I do some repo cleanup. If you plan to finish this up please re-open! I think it would be a great feature in the app.

PR #2045, opened 2018-10-16
Replacing existing colors with npm module, and refactored Sass

NetOperatorWibby:
_This PR is unfinished, still have to go through the app and tweak._


src/renderer/component/wunderbar/view.jsx
```diff
@@ -100,17 +100,15 @@ class WunderBar extends React.PureComponent<Props> {
     if (suggestion) {
       if (suggestion.type === 'search') {
         onSearch(query, resultCount);
+      } else if (isURIValid(query)) {
```
NetOperatorWibby: The linter changed this file, not me.

src/renderer/redux/selectors/settings.js
```diff
@@ -30,10 +30,15 @@ export const selectThemePath = createSelector(
...
   (theme, automaticDarkModeEnabled, isNight) => {
     const dynamicTheme = automaticDarkModeEnabled && isNight ? 'dark' : theme;
-    return `${staticResourcesPath}/themes/${dynamicTheme || 'light'}.css`;
+    // return `${staticResourcesPath}/themes/${dynamicTheme || 'light'}.css`;
+    return dynamicTheme || 'light';
   }
 );
 
+// TODO: I have no idea what I am doing
+export const whatIsTheTheme = () =>
```
NetOperatorWibby: Pretty sure this needs fixing.

seanyesmunt:
This is looking great. I really like the new font. I found a few issues and also have a few more generic comments.

- The input borders are gone. It looks like it can't find the variable
- The modal opacity makes it hard to read text actual content on the modal
screen shot 2018-10-19 at 12 59 10 pm
- It's hard to read the help text next to the search suggestion
screen shot 2018-10-19 at 1 02 39 pm
- The exact url match on the search page needs to be updated for the new dark mode
screen shot 2018-10-19 at 1 13 19 pm
- Input/label input color needs to be updated for dark mode
screen shot 2018-10-19 at 1 14 40 pm
- The table row headings are hard to see on dark mode, also the grey on grey text is hard to read on even numbered rows
screen shot 2018-10-19 at 1 16 00 pm

seanyesmunt:
I'm not super sure about the new green color. I like the actual color but it's a lot harder to read buttons with white text. The contrast ratio is significantly lower than the previous combination which was already pretty low.
screen shot 2018-10-19 at 1 24 20 pm
screen shot 2018-10-19 at 1 20 16 pm

src/renderer/redux/selectors/settings.js
seanyesmunt: We can remove this now

NetOperatorWibby:
@seanyesmunt Pushed up some changes. The `$lbry-teal-5` color is being used for all buttons now, due to its excellent contrast. It doesn't pass WCAG AAA though, only WCAG AA.

seanyesmunt:
@NetOperatorWibby I don't think we have to pass all, `teal-5` looks great. Way easier to read compared to `teal-3`

seanyesmunt:
The search page is looking better on dark mode. There are still some issues with placeholders.
screen shot 2018-10-19 at 4 52 57 pm

The input background color is also a little tough to read on dark mode (both search and copyable inputs)
And the markdown input on the publish page
screen shot 2018-10-19 at 4 53 38 pm

NetOperatorWibby:
Alright, fixed inputs. Found a rather interesting bug/feature. When you highlight text and then switch themes, the selected text is white and will not fix itself unless you reload the app.

NetOperatorWibby:
This is another good site for contrast testing: https://colorable.jxnblk.com/ffffff/2f9176

seanyesmunt:
@NetOperatorWibby After looking at some of the code more closely I'm starting to think we should keep the `vars` file. That would avoid the need for using nested rules in the dark theme scss file. 

Currently in some cases, if any layout changes, the dark mode would break.

NetOperatorWibby:
@seanyesmunt We lose the flexibility of using `rgba` functions and end up having more colors than necessary. I don't think layout changes are a strong argument. Dark mode would have to be checked out if changes are made regardless.

NetOperatorWibby:
@seanyesmunt Do I need to do anything else to get this merged?

seanyesmunt:
I think the only thing left I've noticed is the text in the search bar is really hard to read on light mode. After that is fixed I think we can merge this.

NetOperatorWibby:
Cool beans, I'll go through everything finely.

PR #2048, opened 2018-10-18
persists transaction filter

jessopb:
DEPENDS ON: https://github.com/lbryio/lbry-redux/pull/84

Filter choices in transaction list persist.


src/renderer/component/transactionList/view.jsx
```diff
@@ -111,31 +104,14 @@ class TransactionList extends React.PureComponent<Props, State> {
...
-                {__(`${this.capitalize(TRANSACTIONS.SUPPORT)}s`)}
-              </option>
-              <option value={TRANSACTIONS.UPDATE}>
-                {__(`${this.capitalize(TRANSACTIONS.UPDATE)}s`)}
-              </option>
-              <option value={TRANSACTIONS.ABANDON}>
-                {__(`${this.capitalize(TRANSACTIONS.ABANDON)}s`)}
-              </option>
+              {transactionTypes
+                .concat('all')
```
seanyesmunt: After looking at this now, I think we should just add a new `TRANSACTION` value for `"all"` instead of referencing it in a couple places. Sorry for merging the lbry-redux PR before realizing this.

src/renderer/component/transactionList/view.jsx
```diff
@@ -50,15 +44,11 @@ class TransactionList extends React.PureComponent<Props, State> {
...
-      filter: event.target.value,
-    });
+    this.props.setTransactionFilter(event.target.value);
   }
 
   filterTransaction(transaction: Transaction) {
-    const { filter } = this.state;
-
-    return filter === 'all' || filter === transaction.type;
+    return this.props.filterSetting === 'all' || this.props.filterSetting === transaction.type;
```
seanyesmunt: This should reference `TRANSACTIONS.ALL`

src/renderer/component/transactionList/view.jsx
```diff
@@ -98,7 +91,7 @@ class TransactionList extends React.PureComponent<Props, State> {
           <div className="card__actions-top-corner">
             <FormField
               type="select"
-              value={filter}
+              value={filterSetting || 'all'}
```
seanyesmunt: This one should also reference TRANSACTIONS.ALL

Issue #1967, opened 2018-09-15
Sort order should be retained when returning to page


42degrees:
My publishes library has a lot of duplicates in it.  These are real duplicates.  So, I sorted my list by name and started going through it to remove the duplicates.  But every time I hit Back to return to the main list, it changes the sort order back!  Very annoying, especially if I'm scrolled down the page as the sort control scrolls off.

Also, if a change has happened (like I deleted an item) then when I return to by publishes page it's going to update, but it can take 20 seconds.  If I change the sort order before it updates, then when it updates it changes my sort order.  This can be very annoying especially if it refreshes just as I'm going to click on something since I end up clicking on the wrong item.

Sort of on a related note, when it refreshes like this, it also resets the page, so in those 20 seconds I may have changed the sort order and scrolled down the page.  When the reset happens it not only resets the sort order, it doesn't remember my scroll point.

### Steps to reproduce:
1.  Go to My LBRY | publishes.
2. Choose any sort except for "Newest First"
3. Click on any of your published items.
4. Hit the back button
5. Note that the sort always reverts to "Newest First"

### Expected Behaviour
It should remember that I changed the sort order.

### Actual Behaviour
It forgets the sort order.

### Suggested Solutions
I would vote to remember the sort order of both Downloads and Publishes, even if I'm clicking between them or away to Explore or Subscriptions (not just for back).  The user changed the sort order for a reason, the page should honor that.  Also, when resetting the page it should remember the sort order as well as the current scroll position on the page.

## System Configuration
- LBRY App version: 0.25.0
- LBRY Installation ID: 8jTfoiRnN7Gux5LUhcAb5Twq3cU4XVhnoF4Y6nr6mVVjP6piiRwXBxt2stQc6xAKfE
- Operating system: Windows 10

tzarebczan:
Thanks for filing this @42degrees ! Similarly, we should persist these filters in other areas of the app, i.e. https://github.com/lbryio/lbry-desktop/issues/1154

jessopb:
I have spent some time researching this and https://github.com/lbryio/lbry-desktop/issues/1154
It seems like there is a choice to create a new state object in local app redux state, or to just use the existing navigation object that records, for example, scrollY. 

Supposing the best choice is to persist during the session in the navigation local redux object. This will involve:
1) an action in src/renderer/redux/reducers/navigation.js
 - named doRecordFiltered type ACTIONS.LIST_FILTERED in ,
[...]

PR #2050, opened 2018-10-22
persist fileList sorting

jessopb:
Re: https://github.com/lbryio/lbry-desktop/issues/1967


seanyesmunt:
@jessopb This is great. I'll give the lbry-redux PR a review in a little bit and do some testing.

seanyesmunt:
@jessopb Seems to work well. Can you add a filter [here](https://github.com/lbryio/lbry-desktop/blob/master/src/renderer/store.js#L104-L111) so these new values are saved when you refresh/close the app?

seanyesmunt:
@jessopb I was blocked on another PR so I finished this one up.

PR #2053, opened 2018-10-23
WIP: Responsiveness and refactoring

NetOperatorWibby:
_In progress_


src/renderer/component/button/view.jsx
```diff
@@ -58,22 +58,22 @@ class Button extends React.PureComponent<Props> {
     const combinedClassName = classnames(
       'btn',
       {
-        'btn--no-padding': noPadding,
+        'no-padding': noPadding,
```
seanyesmunt: I think we should stick with BEM naming. Or at least, I don't see a reason to get rid of it. I think it's been pretty nice now that the app is _mostly_ using it.

NetOperatorWibby: These classes are just modifying the state of the parent element. I ran into several instances where an element would have two or more additional classes overriding each other.

Nesting these additional classes within the element itself allowed me to use `:not`s to stop overriding from happening.

Using strict BEM in this instance would've resulted in a bunch of `:not`s so I got rid of them.

seanyesmunt: Why do you need a lot of `:not`'s? We didn't need any before

NetOperatorWibby: Because styles should not be overriding. What's the point of having styles if you're not going to see them?

seanyesmunt: If there are styles that are never seen, I agree we should remove those. I don't see why we need to move away from BEM to fix this though. Maybe we just need to improve how we assign the classes and simplify (and possibly add new modifiers) the class styles.

NetOperatorWibby: Agreed. I'll add them back when I go through the buttons again, most likely when I update the "boring" pages like settings.

NetOperatorWibby:
So squashing my commits into one and pushing closed this PR.

NetOperatorWibby:
DO NOT DELETE BRANCH

I'll fix this from my desktop in the AM.

PR #2057, opened 2018-10-25
Improve type checking for card verify

amelzer:
Improved type checking for CardVerify component.

### DIFF
--- a/src/renderer/component/cardVerify/view.jsx
+++ b/src/renderer/component/cardVerify/view.jsx
@@ -7,6 +7,20 @@ let scriptLoading = false;
 let scriptLoaded = false;
 let scriptDidError = false;
 
+declare class CardVerify {
+  static stripeHandler: {
+    open: Function,
+    close: Function,
+  };
+}
+
+declare class StripeCheckout {
+  static configure({}): {
+    open: Function,
+    close: Function,
+  };
+}
+
 type Props = {
   disabled: boolean,
   label: ?string,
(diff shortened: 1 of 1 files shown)


src/renderer/component/cardVerify/view.jsx
```diff
@@ -7,6 +7,20 @@ let scriptLoading = false;
...
+  };
+}
+
+declare class StripeCheckout {
+  static configure({}): {
+    open: Function,
+    close: Function,
+  };
+}
+
```
seanyesmunt: We don't need these. They aren't doing anything as is. Also I belive the `declare class` syntax is wrong.

amelzer: Thank you for your feedback.

The classes are types for the global `CardVerify` object that you are accessing after adding the script. Not declaring classes for them results in Flow errors

`Cannot resolve name CardVerify.`
`Cannot resolve name StripeCheckout.`

As this is a global variable, it would make sense in my opinion to also declare a global class like suggested here: https://flow.org/en/docs/libdefs/creation/#toc-declaring-a-global-class

The goal of my changes was to eliminate the lint & flow errors and therefore I added class definitions for the accessed global objects.

seanyesmunt: Ah you are right! I missed the part where we access `CardVerify` when I first looked this over. Thanks for that link too 🙂 I'll definitely read over that.

seanyesmunt:
Thanks again for the PR! If you haven't yet, make sure to check out the [contributing guide](https://github.com/lbryio/lbry-desktop/blob/master/CONTRIBUTING.md).

Can we [send you some LBC as appreciation](https://lbry.io/faq/appreciation)?

amelzer:
@seanyesmunt Not needed, thank you :)

seanyesmunt:
@amelzer I reverted your PR. I guess I didn't look close enough but upon testing Tom noticed it was breaking the app. I can now see it was because of your variable name changes. 

`CardVerify.stripeHandler` is supposed to reference the component, not the type. We could change that to `this.stripeHandler`.

amelzer:
@seanyesmunt Thanks for letting me know. I will create a new PR

PR #2058, opened 2018-10-26
adds sortBy default prop to subscriptions and channel pages

jessopb:
Channel page crashes due to missing sortBy. FileList now requires it as a prop.
requires https://github.com/lbryio/lbry-redux/pull/90

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "formik": "^0.10.4",
     "hast-util-sanitize": "^1.1.2",
     "keytar": "^4.2.1",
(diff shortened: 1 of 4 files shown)


seanyesmunt:
I'm starting to think we should just use dateNew as a default prop and remove the subscriptions selector/state. That and channel pages will never (I think) use anything other than dateNew

seanyesmunt:
@jessopb Can you point the lbry-redux version in the package json to `ecd1ab156fb755c8bff4d890679399d66f41ad17`? It's the latest commit after merging your PR.

Then this can be merged.

PR #2063, opened 2018-10-27
Update CHANGELOG.md

twishajain:
Corrected repetition of fixed heading under the same date


seanyesmunt:
@twishajain Thanks for the PR 🙂 

Not sure how it ended up in that state

tzarebczan:
@twishajain , Can we send you [some LBC as appreciation](https://lbry.io/faq/appreciation)?

PR #2064, opened 2018-10-28
Reduce unused var warnings. 

amelzer:
Hey, as a little housekeeping quest, I removed unused vars from components where it was easily doable without triggering more linting errors while committing.
I also removed redux connections where they were not needed.


seanyesmunt:
Woohoo!

Issue #2065, opened 2018-10-28
Linting errors fixable with `--fix`


amelzer:
Currently, the project contains a ton of lint errors that get can and will get autofixed by running `yarn lint` . This adds several modified files to a commit / PR and therefore not helpful at the current stage. 

### Steps to Reproduce
1. Create new branch
2. Run `yarn flow`
3. Run `git status`

### Expected Behaviour
No changes made, because master is clean of autifixable errors

### Actual Behaviour
Tons of modified files

### Suggested Solutions
My suggestion would be fixing each of those files in seperate PRs. One big PR would probably create a lot of merge conflicts for branches currently worked on. Modifying one of those files will prompt you to fix any other errors before commit. Might be a good time to do so :)

- [ ]  main/index.js
- [ ]   renderer/component/address/index.js
- [ ]   renderer/component/button/index.js
- [ ]  renderer/component/cardMedia/index.js
- [ ]  renderer/component/channelTile/index.js
- [ ]  renderer/component/common/credit-amount.jsx
- [ ]  renderer/component/common/file-exporter.jsx
- [ ]  renderer/component/dateTime/index.js
- [ ]  renderer/component/externalLink/index.js
- [ ]  renderer/component/fileActions/index.js
- [ ]  renderer/component/fileDetails/index.js
- [ ]  renderer/component/fileDetails/view.jsx
- [ ]  renderer/component/filePrice/index.js
- [ ]  renderer/component/formFieldPrice/index.js
- [ ]  renderer/component/nsfwOverlay/index.js
- [ ] renderer/component/publishForm/index.js
- [ ]  renderer/component/router/index.js
- [ ]  renderer/component/selectChannel/index.js
- [ ]  renderer/component/shapeShift/index.js
- [ ]  renderer/component/snackBar/index.js
- [ ]  renderer/component/spinner/index.js
- [ ]  renderer/component/theme/index.js
- [ ]  renderer/component/transactionListRecent/index.js
- [ ]  renderer/component/uriIndicator/index.js
- [ ]  renderer/component/userVerify/view.jsx
- [ ]  renderer/component/walletAddress/index.js
- [ ]  renderer/component/walletBalance/index.js
- [ ]  renderer/component/walletSend/index.js
- [ ]  renderer/component/walletSendTip/index.js
- [ ]  renderer/modal/modalAffirmPurchase/index.js
- [ ]  renderer/modal/modalAuthFailure/index.js
[...]

PR #2066, opened 2018-10-28
Linting/address

amelzer:
This PR lints & type checks the address component and its consumers as part of #2065


seanyesmunt:
Another one! 🙂 

Thanks for the PR!

Issue #2073, opened 2018-10-30
Recent issues on master after design/sdk updates


tzarebczan:
## The Issue
- [x]  Rewards page text is jumbled
- [x] Channels names are different shades and cutoff sometimes on Publish page  (See below)
- [x] transaction overview history is backwards when new transactions are created (pending show at the bottom of the list, and so do any txs created int he current sessionI). A refresh fixes this. 
- [x] all modals/toasts stop working after abandoning a claim via the Delete button 
- [x] heart (subbed) icon mising on explore/channel pages
- [x] subscription button doesn't change to "unsubscribe" after clicking it from the channel page (thanks @Invariant-Change )
- [x] spell check syntax in description on publish page doesn't show properly (@Invariant-Change said he'd fix)
- [x]  `CardVerify is not defined`/app crash after email verification and leaving the verification page
- [x] Amount for channel creation needs to be converted to string decimal
- [x] Text on Publish button in dark mode looks too "dark" 
- [x] pending publish does not show on Publishes page until re-navigating to it (might be related to checking of the daemon response?)
- [x] when choosing a filter on the transaction history page, it also gets applied to the overview. We either need to stop that, or add the filter to the overview page (cc @jessopb ).
- [x] publish button in dark mode blends into background when inactive 
- [x] During edits on Publish screen, a "none" license type shows up as `Copyrighted...` with None as the notice. 
- [ ] claim_list_mine claims are re-resolved when going to publishes page
- [x] trashcan/unlock icon does not disappear after abandoning on transaction list page (until refresh/going switching pages). This used to disappear a few seconds after the abandon was called. 

![image 1](https://user-images.githubusercontent.com/8120721/47724594-5c00de80-dc2d-11e8-8202-e6dac12b1b0c.png)

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

seanyesmunt:
Fixing these

seanyesmunt:
[...]

PR #2082, opened 2018-10-31
Rc fixes

seanyesmunt:
More fixes for the RC
https://github.com/lbryio/lbry-desktop/issues/2073

- Added subscription icons to the home page
- Fixed channel subscribe failing on the channel page
- Fixed channel creation failing by changing the bid amount to a string
- Fixed dark mode markdown editor

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -42,6 +42,7 @@
     "jsx-a11y/no-noninteractive-element-to-interactive-role": 0,
     "class-methods-use-this": 0,
     "jsx-a11y/interactive-supports-focus": 0,
-    "jsx-a11y/click-events-have-key-events": 0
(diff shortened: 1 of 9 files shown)


.eslintrc.json
```diff
@@ -42,6 +42,7 @@
     "jsx-a11y/no-noninteractive-element-to-interactive-role": 0,
     "class-methods-use-this": 0,
     "jsx-a11y/interactive-supports-focus": 0,
-    "jsx-a11y/click-events-have-key-events": 0
+    "jsx-a11y/click-events-have-key-events": 0,
+    "consistent-return": 0
```
seanyesmunt: This has always caused issues, mainly in react components where you might not want to return anything:
ex: 
```
someMethod() {
  if (!props.value) return;
  
  doStuff()
}
```

PR #2087, opened 2018-11-01
Remove ToS checkbox

seanyesmunt:
We now default add a label saying "by continuing you agree..." so you don't have to check the box every time you publish.

screen shot 2018-11-01 at 5 48 43 pm


kauffj:
@seanyesmunt fine with me. It may need a comma: "By continuing, ..."

PR #2091, opened 2018-11-05
Fix: correctly search for daemon process on startup

seanyesmunt:
When we added the v30 daemon changes, we changed our process search to `lbrynet`, which would return results if you had any `lbrynet` files open (including `lbrynet.log`)

This fixes that issue.


seanyesmunt:
Tested with @nikooo777 and @tzarebczan

PR #2092, opened 2018-11-06
Rc fixes

seanyesmunt:


seanyesmunt:
@tzarebczan The decimal fix is in https://github.com/lbryio/lbry-redux/pull/100

seanyesmunt:
Moved to time estimates based on block heights in https://github.com/lbryio/lbry-redux/pull/101/commits

PR #2095, opened 2018-11-07
Comments support button

seanyesmunt:
### Changes
- Adds a button that will register support for adding comments in the app
- Adds `<Expandable>` component

Still deciding on the copy for button/message


src/renderer/constants/action_types.js
```diff
@@ -36,6 +37,7 @@ export const SKIP_UPGRADE = 'SKIP_UPGRADE';
 export const START_UPGRADE = 'START_UPGRADE';
 export const AUTO_UPDATE_DECLINED = 'AUTO_UPDATE_DECLINED';
 export const AUTO_UPDATE_DOWNLOADED = 'AUTO_UPDATE_DOWNLOADED';
+export const CLEAR_UPGRADE_TIMER = 'CLEAR_UPGRADE_TIMER';
```
seanyesmunt: This was added to the lbry-redux action types but it should have never been there. I will do some cleanup in the future to make sure we aren't keeping types there that only belong in the app

src/renderer/component/expandable/index.js
```diff
@@ -0,0 +1,7 @@
+import { connect } from 'react-redux';
+import Expandable from './view';
+
+export default connect(
```
skhameneh: 😐

src/renderer/component/expandable/view.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+type State = {
+  expanded: boolean,
+};
+
+export default class Expandable extends PureComponent<Props, State> {
+  constructor() {
+    super();
+
+    this.state = {
```
skhameneh: Can also do:
```
class MyComponent extends PureComponent {
  state = {};
}
```

and use arrow fn's or bind syntax https://babeljs.io/blog/2015/05/14/function-bind

src/renderer/component/fileDetails/view.jsx
```diff
@@ -1,77 +1,128 @@
...
+    }
+
+    const { description, language, license } = metadata;
 
-  return (
-    <React.Fragment>
-      {description && (
-        <React.Fragment>
-          <div className="card__subtext-title">About</div>
+    const mediaType = contentType || 'unknown';
```
skhameneh: Should `'unknown'` be a shared constant?

seanyesmunt: Probably. Although I'm not sure where else it's used.

PR #2102, opened 2018-11-12
Changes and improvements

### LINKED ISSUE #1976: App frontpage redesign
This issue pertains to what you see in the app as soon as it loads. The "homepage", if you will.

NetOperatorWibby:
## PR Checklist
- [x] This PR introduces breaking changes and I have provided a detailed explanation below.

## PR Type
- [x] Breaking changes (bugfix or feature that introduces breaking changes)

## Fixes

I borked https://github.com/lbryio/lbry-desktop/pull/2053 while attempting to have a sane rebase. This PR refactors the CSS and makes semantic changes in the rendered HTML.

Closes #1976.

Nothing the average user would notice, aside from styling.

## What is the new behavior?

New hotness. Sass is (more) organized and easier to manage. It still isn't perfect but it's close.


src/renderer/scss/component/_button.scss
```diff
@@ -1,224 +1,161 @@
...
+  &--link {
+    transition: color 0.2s;
+
+    &:not(:hover) {
+      color: $lbry-teal-5;
 
-    &:disabled {
-      opacity: 0.3;
+      html[data-theme='dark'] & {
+        color: $lbry-teal-3;
```
seanyesmunt: After looking over this code more, I think we should move back to variables. I think it will be hard to keep consistency with this new pattern. (And add new theme options if we ever decided to do so).

src/renderer/scss/component/_card.scss
```diff
@@ -1,367 +1,140 @@
...
-.card__list--rewards {
-  column-count: 2;
-  column-gap: $spacing-vertical * 1/3;
-
-  .card {
-    display: inline-block;
-    margin: 0 0 $spacing-vertical * 1/3;
-    width: 100%;
-  }
+.card-media__internal-links {
```
seanyesmunt: There's also `card-media__` which makes it even more confusing.

src/renderer/scss/component/_card.scss
```diff
@@ -1,367 +1,140 @@
...
+  &--section {
+    padding: 2rem;
 
-.card--wallet-balance {
-  background: url('../../../static/img/stripe-background.png') no-repeat;
-  background-size: cover;
-  color: $lbry-white;
-  justify-content: space-between;
+    .card__content:not(:last-child) {
+      margin-bottom: 1.5rem;
```
seanyesmunt: Why the change the `margin-bottom` instead of `margin-top`? IME always using `margin-top`/`margin-left` makes it a lot simpler to add new pieces of ui, because you don't rely on the markup before your new component to provide the spacing. 

It also requires removing the margin on `:last-child` everywhere.

src/renderer/scss/component/_credit.scss
```diff
@@ -0,0 +1,49 @@
...
+      // color: $lbry-gray-5;
+    }
+
+    &.credit-amount--file-page {
+      background-color: $lbry-yellow-2;
+      // color: $lbry-black;
+    }
+  }
+
+  &--file-page {
```
seanyesmunt: This should just use something like `$--large`.

seanyesmunt:
screen shot 2018-11-12 at 5 35 10 pm

Moving to `position: absolute` is messing with this (and probably affects a few other places)

seanyesmunt:
screen shot 2018-11-12 at 5 36 58 pm

This uses `card--space-between` which was basically just `display: flex; justify-content: space-between;`

src/renderer/component/common/credit-amount.jsx
```diff
@@ -77,14 +77,24 @@ class CreditAmount extends React.PureComponent<Props> {
...
-          'credit-amount--large': large,
+        // className={classnames('credit-amount', {
+        //   'credit-amount--large': large,
+        //   // TODO: remove inheritStyle prop
+        //   // It just complicates things
+        //   'credit-amount--inherit': inheritStyle,
+        //   badge,
+        //   'badge--cost': badge && !isFree,
+        //   'badge--free': badge && isFree,
+        // })}
```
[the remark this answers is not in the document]
NetOperatorWibby: For reference while I was fixing my branch.

src/renderer/component/dateTime/view.jsx
```diff
@@ -23,28 +23,27 @@ class DateTime extends React.PureComponent<Props> {
...
 
   componentWillMount() {
-    // this.refreshDate(this.props);
+    this.refreshDate(this.props);
   }
 
-  componentWillReceiveProps() {
-    // this.refreshDate(props);
+  componentWillReceiveProps(props) {
+    this.refreshDate(props);
```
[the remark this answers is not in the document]
NetOperatorWibby: This was an oversight on my part.

seanyesmunt:
screen shot 2018-11-12 at 5 51 27 pm

The exact search tile match is off.

src/renderer/page/file/view.jsx
```diff
@@ -162,90 +162,85 @@ class FilePage extends React.Component<Props> {
...
-                    button="alt"
-                    icon={icons.GIFT}
-                    label={__('Send a tip')}
-                    onClick={() => openModal({ id: MODALS.SEND_TIP }, { uri })}
-                  />
-                )}
+          ))}
+
+        <div className="media__content media__content--file-page">
+          <h1 className="media__title media__title--file-page">{title}</h1>
```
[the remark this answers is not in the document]
NetOperatorWibby: After seeing the changes in master, it became apparent that `--file-page` wasn't the best name.

I separated `media` and `card` b/c it was quite confusing. I'll have to give them another pass.

seanyesmunt:
screen shot 2018-11-12 at 5 52 03 pm
screen shot 2018-11-12 at 5 52 43 pm


Why the switch to cards for search? I think tiles make more sense if we want to display the description. It offers a lot more room.

src/renderer/page/fileListDownloaded/view.jsx
```diff
@@ -3,34 +3,41 @@ import React from 'react';
...
 class FileListDownloaded extends React.PureComponent<Props> {
   render() {
-    const { fetching, fileInfos, navigate, sortBy } = this.props;
+    const { fetching, fileInfos, navigate } = this.props;
     const hasDownloads = fileInfos && Object.values(fileInfos).length > 0;
 
     return (
       <Page notContained loading={fetching}>
         {hasDownloads ? (
-          <FileList fileInfos={fileInfos} sortBy={sortBy} page={PAGES.DOWNLOADED} />
```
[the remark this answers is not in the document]
NetOperatorWibby: Another oversight.

seanyesmunt:
screen shot 2018-11-12 at 5 38 04 pm

Cards with no thumbnail don't have the same height as other cards.

Also the placeholders don't match up with the actual text.

seanyesmunt:
screen shot 2018-11-12 at 5 38 37 pm

The spacing under the title looks way to big. And the spacing between the two inputs should be increased.

src/renderer/scss/component/_card.scss
NetOperatorWibby: Copying from Bootstrap's docs:

> Avoid margin-top. Vertical margins can collapse, yielding unexpected results. More importantly though, **a single direction of margin is a simpler mental model**.

It's consistent with how the rest of the app is styled. Also, the removal of the margin on `:last-child` is only inside this block. I haven't seen anything to indicate visual issues.

src/renderer/component/app/view.jsx
```diff
@@ -89,9 +89,17 @@ class App extends React.PureComponent<Props> {
...
+    const scrollAmount = document.querySelector('#content').scrollTop;
     if (this.mainContent) {
       recordScroll(this.mainContent.scrollTop);
     }
+
+    // 2rem (padding) is 24px b/c body text is 12px
+    if (scrollAmount >= 24) {
+      document.querySelector('main.page').classList.add('scrolled');
+    } else {
+      document.querySelector('main.page').classList.remove('scrolled');
```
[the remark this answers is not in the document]
NetOperatorWibby: When scrolling on the homepage past the padding, the wunderbar doesn't have a border/divider that separates it from the content.

[Comment out this line](https://github.com/lbryio/lbry-desktop/blob/newnew/src/renderer/scss/component/_page.scss#L21) and scroll on a page in Settings to see what I mean.

NetOperatorWibby:
I think we should have an default image for uploads without thumbnails. That'll make things easier for that issue.

src/renderer/scss/component/_card.scss
seanyesmunt: 👍 I'm fine moving to `margin-bottom`

NetOperatorWibby:
Having a wholly different layout for search means it's yet another thing to maintain. It seemed easier to just keep it consistent. The description field has more content than expected though.

src/renderer/component/app/view.jsx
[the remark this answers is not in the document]
seanyesmunt: Why not make that the default behavior? It seems like it's just adding a class to give it the same look as before scrolling (with the border bottom). Why can't we always keep the z-index on the header?

seanyesmunt:
> Having a wholly different layout for search means it's yet another thing to maintain. It seemed easier to just keep it consistent. The description field has more content than expected though.

It's not an entirely different layout. It's just a list of `FileTile`'s, which is the same thing as the related content. Just scaled up

NetOperatorWibby:
Okay, I'll look at the current look of that page in the app.

src/renderer/component/app/view.jsx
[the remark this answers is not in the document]
NetOperatorWibby: It doesn't look good.

src/renderer/component/fileDetails/view.jsx
```diff
@@ -39,37 +39,42 @@ const FileDetails = (props: Props) => {
     <React.Fragment>
       {description && (
         <React.Fragment>
-          <div className="card__subtext-title">About</div>
-          <div className="card__subtext">
+          <div className="media__info__title">About</div>
```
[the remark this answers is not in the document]
NetOperatorWibby: Doesn't make sense here. This is inside a `media__info` block.

src/renderer/component/app/view.jsx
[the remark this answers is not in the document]
seanyesmunt: In what case doesn't it look good? If we move the z-index down to this line https://github.com/lbryio/lbry-desktop/blob/newnew/src/renderer/scss/component/_page.scss#L25, we never get to a state where the border-bottom disappears

screen shot 2018-11-15 at 11 23 25 am

seanyesmunt:
I can't seem to reply in the original thread but for https://github.com/lbryio/lbry-desktop/pull/2102#discussion-diff-232837723L43

The `element__element1__element2` naming is not following BEM naming
https://en.bem.info/methodology/quick-start/#nesting-1

src/renderer/component/app/view.jsx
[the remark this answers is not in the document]
NetOperatorWibby: When you're on the first view of the app, the border doesn't look good.

NetOperatorWibby:
I realized that a while after I made the comment but wasn't able to find it.

src/renderer/component/inviteNew/view.jsx
```diff
@@ -47,8 +47,11 @@ class FormInviteNew extends React.PureComponent {
             }}
           />
         </FormRow>
-        <div className="card__actions">
-          <Submit label="Invite" disabled={isPending} />
+
+        <div className="card__content">
```
seanyesmunt: I don't think I'm a fan of this. Before it was super simple to add a list (or single) button. It was one `card__actions` class that had a `padding-top`. I think this could work, but it just takes more effort. A good example is on the wallet page. I see three buttons, with three different top/bottom spacings

screen shot 2018-11-15 at 5 05 02 pm

src/renderer/component/recommendedContent/view.jsx
```diff
@@ -50,7 +50,7 @@ export default class RecommendedContent extends React.PureComponent<Props> {
     const { recommendedContent, isSearching } = this.props;
 
     return (
-      <section className="card__list--recommended">
+      <section className="media-group--list--recommended">
```
seanyesmunt: There should only be one modifier per class.

src/renderer/scss/component/_search.scss
```diff
@@ -1,106 +1,105 @@
...
-    left: 10px;
+    &:not(:last-of-type) {
+      margin-bottom: 1.5rem;
+    }
 
-    position: absolute;
-    z-index: 1;
+    &:not(.media--large) {
+      .media__thumb {
+        width: 20rem; height: 10rem;
```
seanyesmunt: This isn't the correct aspect ratio.

src/renderer/scss/component/_search.scss
```diff
@@ -1,106 +1,105 @@
-.wunderbar {
-  cursor: text;
-  display: flex;
-  flex: 1;
-  min-width: 175px;
-  position: relative;
-  z-index: 1;
+.search {
+  .media {
```
seanyesmunt: The media styles shouldn't change just because they are on the search page. They should be contained to the `_media.scss` file.

src/renderer/scss/component/_placeholder.scss
```diff
@@ -26,17 +26,3 @@
 .placeholder__title--tile {
   height: 3em;
 }
-
-@keyframes pulse {
```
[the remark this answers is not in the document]
NetOperatorWibby: It made sense to put all the animation in one file.

src/renderer/component/channelTile/view.jsx
```diff
@@ -51,20 +51,20 @@ class ChannelTile extends React.PureComponent<Props> {
       <section
         onClick={onClick}
         role="button"
-        className={classnames('file-tile card--link', {
-          'file-tile--small': size === 'small',
-          'file-tile--large': size === 'large',
+        className={classnames('media card--link', {
+          'media--small': size === 'small',
+          'media--large': size === 'large',
```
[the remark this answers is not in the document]
NetOperatorWibby: Agreed, it's a bit of a mess. `media-tile` sounds like a good name. So we'll have `card`, `media`, and `media-tile`.

src/renderer/scss/component/_placeholder.scss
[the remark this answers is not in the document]
seanyesmunt: Ah I see. I missed that.

src/renderer/component/channelTile/view.jsx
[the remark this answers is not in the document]
seanyesmunt: Maybe `media-card`? Or if the base class is `media` the tile style would be a modifier so `media--tile`

src/renderer/component/fileDetails/view.jsx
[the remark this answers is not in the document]
seanyesmunt: The docs say to avoid double element naming https://en.bem.info/methodology/quick-start/#nesting-1

src/renderer/scss/component/_search.scss
NetOperatorWibby: The media on the search page are not displayed the same as everywhere else. My previous design kept the design uniform but that was unwanted.

src/renderer/component/common/lbc-symbol.jsx
```diff
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
 
-const LbcSymbol = () => <span>LBC</span>;
+const LbcSymbol = () => <span>℄</span>;
```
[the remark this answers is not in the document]
kauffj: Yes, do not switch this yet.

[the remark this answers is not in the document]
NetOperatorWibby: screen shot 2018-11-19 at 1 32 41 pm

Should I also remove it from here? I added it to spur conversation and hopefully get a definitive answer. I'll revert the change.

src/renderer/component/inviteNew/view.jsx
NetOperatorWibby: I just inspected the code and the buttons themselves don't have any spacing. The container they are in also don't have spacing unless they are inside another class. I could remove `.form-row--padded` from the address in **Receive Credits** (which has top/bottom spacing) to reduce the vertical spacing.

Other than that, I don't see anything wrong with taking the top-down approach to making layouts (pushing down rather than up).

seanyesmunt: My comment was more aimed at the inconsistency of the button spacings. It seems like this still isn't fixed. The `Claim Rewards` button has a large amount of space above, and small amount of space below. The `Get New Address` button has a large amount of space on top and bottom. And the `Full History` button has a small amount of space on top and bottom.

seanyesmunt: > The container they are in also don't have spacing unless they are inside another class.

This seems like it could cause issues. Why not just give the base level class the spacing instead of only adding it if it's inside of another class?

src/renderer/component/header/view.jsx
```diff
@@ -38,68 +38,67 @@ const Header = (props: Props) => {
...
-          description={__('Your wallet')}
+          onClick={() => navigate('/wallet')}
         />
 
         <Button
-          uppercase
-          button="primary"
-          className="btn--header-publish"
-          onClick={() => navigate('/publish')}
-          icon={icons.UPLOAD}
```
seanyesmunt: With the new icons that were added we now have two ways of adding icons to buttons. We should consolidate this into one.

seanyesmunt:
Forgot another thing, the subscription button still moves down a bit when the `heart` icon is added/removed.

src/renderer/component/header/view.jsx
NetOperatorWibby: What is being consolidated?

src/renderer/component/fileCard/view.jsx
```diff
@@ -64,17 +63,16 @@ class FileCard extends React.PureComponent<Props> {
       pending,
       isSubscribed,
       isNew,
-      showSubscribedLogo,
```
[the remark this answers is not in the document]
NetOperatorWibby: I removed this b/c the linter was complaining. Thought I got everything.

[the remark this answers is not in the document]
NetOperatorWibby: Should `static defaultProps` still be there? Just empty, or removed entirely?

src/renderer/component/header/view.jsx
seanyesmunt: We should either always use a class and add the icons through css with `background-image` like you do [here](https://github.com/lbryio/lbry-desktop/pull/2102/files#diff-88b5fbd05c1ee805da8402786b5cc714R58), or stick with using props and explicitly render an icon like [we are currently doing](https://github.com/lbryio/lbry-desktop/blob/master/src/renderer/component/common/icon.jsx#L3) where we use `react-feather`

src/renderer/component/fileCard/view.jsx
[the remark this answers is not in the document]
seanyesmunt: If there aren't any default props, it can be removed entirely

src/renderer/scss/component/_search.scss
```diff
@@ -1,106 +1,47 @@
...
-
-  .search-result__row {
-    padding: $spacing-vertical 0;
-
-    &:first-of-type {
-      padding-top: 0;
-    }
-  }
+.search__results-section {
+  margin-bottom: 4rem;
```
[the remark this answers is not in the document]
NetOperatorWibby: I don't think everything having the same margin makes sense.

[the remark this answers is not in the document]
seanyesmunt: They don't need to have the same margins, but some factor of the base level marign. If we just had small, medium, and large margins, where small is 1/3 the base margin, medium is 1/2 and large is the base margin (these are just random numbers), it would greatly help with consistency.

If everything is on it's own, every component could have different margins if future changes aren't tracked carefully after this is merged.

src/renderer/scss/component/_credit.scss
```diff
@@ -0,0 +1,39 @@
...
+  white-space: nowrap;
+}
+
+.credit-amount--cost {
+  &.credit-amount--file-page {
+    background-color: $lbry-yellow-2;
+  }
+}
+
+.credit-amount--file-page {
```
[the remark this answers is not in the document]
NetOperatorWibby: This isn't being used anymore.

src/renderer/scss/component/_channel.scss
```diff
@@ -0,0 +1,25 @@
+.channel-info {
+  align-items: center;
+  display: flex;
+  justify-content: space-between;
```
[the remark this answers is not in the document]
NetOperatorWibby: The Publish link in the URL bar is also to the right. Should that be removed too?

src/renderer/component/inviteNew/view.jsx
NetOperatorWibby: These buttons are also in different contexts.

- `Claim Rewards` is below a header. Everything below a header has the same spacing.
- `Get New Address` is below a copyable input. That spacing ensures you don't click on something unintentionally.
- `Full History` is below a table and _pertains to that table_. Why would it have large spacing above it? The spacing below it is consistent with every single card.

If you want all the buttons to have "consistent spacing", they should all be in the same contexts. But they're not. Because it doesn't make sense.

src/renderer/component/header/view.jsx
NetOperatorWibby: What happens when we want to use an icon that doesn't exist with `react-feather`?

NetOperatorWibby:
`'media--search-result': isSearchResult` is not working for `channelTile`. Not sure what needs to be done there.

NetOperatorWibby:
screen shot 2018-11-26 at 1 41 28 pm

What is messed up about this style?

seanyesmunt:
>'media--search-result': isSearchResult is not working for channelTile. Not sure what needs to be done there.

You have to pass in that value when you use it `<ChannelTile ... isSearchResult />`

>What is messed up about this style?

General inconsistency compared to the rest of the app. There is no margin above the textarea/submit button.

src/renderer/component/channelTile/view.jsx
```diff
@@ -75,7 +75,7 @@ class ChannelTile extends React.PureComponent<Props> {
             </React.Fragment>
           )}
           {subscriptionUri && (
-            <div className="card__actions">
+            <div className="media__actions">
```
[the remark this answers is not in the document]
NetOperatorWibby: The only modifier class that's the same is `--between`.

src/renderer/component/channelTile/view.jsx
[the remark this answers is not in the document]
seanyesmunt: This still exists

src/renderer/component/expandable/view.jsx
```diff
@@ -48,6 +48,7 @@ export default class Expandable extends PureComponent<Props, State> {
         </div>
         <Button
           button="link"
+          className="expandable__button"
```
seanyesmunt: Does a regular `link` not work? If not this should be a button class `btn--expandable`, but it should just use a regular link button

src/renderer/component/fileCard/view.jsx
```diff
@@ -101,34 +95,29 @@ class FileCard extends React.PureComponent<Props> {
         tabIndex="0"
         role="button"
         onClick={!pending ? () => navigate('/show', { uri }) : () => {}}
-        className={classnames('card card--small', {
-          'card--link': !pending,
-          'card--pending': pending,
+        className={classnames('media', {
```
seanyesmunt: As re-opened above, this should be something like `media-card`

src/renderer/component/header/view.jsx
seanyesmunt: In the past, we have just chosen a different icon. This comment was more about still using the `Icon` component. I think it's ok if we use different icons there (it should be avoided, but fine if absolutely necessary), but we definitely shouldn't be adding icons in two different ways.

Everything should go through the `Icon` component and we shouldn't be adding icons by using a class (`.play`). I have some ideas but maybe we should discuss over slack. We could check to see if we have a custom icon, if one isn't there, then fallback to react-feather.

src/renderer/component/fileViewer/internal/play-button.jsx
```diff
@@ -15,10 +15,19 @@ class VideoPlayButton extends React.PureComponent<Props> {
...
     const label = doesPlayback ? __('Play') : __('View');
 
-    return <Button button="primary" disabled={disabled} label={label} icon={icon} onClick={play} />;
+    return (
+      <Button
+        disabled={disabled}
+        label={label}
+        className={classnames('btn--icon', {
+          'btn--play': doesPlayback,
+          'btn--view': !doesPlayback,
```
seanyesmunt: This should still use the `icon` prop.

`icon={ICONS.PLAY}`

src/renderer/component/header/view.jsx
```diff
@@ -38,68 +38,67 @@ const Header = (props: Props) => {
...
-          label={isUpgradeAvailable ? '' : __('Publish')}
+          className="header__navigation-item header__navigation-item--back"
+          description={__('Navigate back')}
+          disabled={isBackDisabled}
+          onClick={back}
+        />
+
+        <Button
+          className="header__navigation-item header__navigation-item--forward"
```
seanyesmunt: These should use the `icon` prop

src/renderer/component/inviteNew/view.jsx
seanyesmunt: I understand not having a large space on the `Full History` button. I think that is fine.

For `Get New Address` and `Claim Rewards`, they are both directly above help text, but have two different spacings. Looks like one is `p.help` and the other is `div.help`. Might just be a typo.

seanyesmunt:
screen shot 2018-11-29 at 1 54 03 am

The button under the textarea should have the same spacing as it would under a table (like transaction history) right? I think those are the same context.

Also the spacing seems really weird when there are cards with subtitles and without. The margin-bottom on the `card-header` class is causing inconsistent spacing. The space between the textarea and the subtitle on the top card is ~36 pixes, and the spacing between the top link and the title on the bottom card is ~42 pixels.

src/renderer/component/header/view.jsx
NetOperatorWibby: The icons I'm using here are not in the react-feathers repo.

src/renderer/component/fileViewer/internal/play-button.jsx
NetOperatorWibby: The icon I'm using here is not in the react-feathers repo.

Issue #1825, opened 2018-07-26
Explore presenting channel suggestions on subscriptions page / modal


tzarebczan:
## The Issue
Similar to how we currently fetch featured content, we will be implementing an endpoint  (https://github.com/lbryio/internal-apis/issues/262 - internal issue) which lists suggested channels for users to subscribe/view content from. I've also suggested a community-driven category be added to the endpoint which would list channels with the highest claim bids. The endpoint for featured subs will have a description

Since we don't have channel metadata, we'd need a clever way to engage the user/show the suggestions. 

For a user without any subscriptions, this could be shown when they enter the subscriptions page.  For those with subs already, could show it as a compact view near the top or have an option to view the suggestions.

We could also incorporate this as part of the first run experience to drive exploration/engagement.

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1. Completed design and implementation
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

tzarebczan:
@alyssaoc I'll let you determine if we should be created an Epic for this as a separate issue, or possibly turn this one into an Epic

tzarebczan:
API sample output:
https://api.lbry.io/subscription/suggest
```
{
  "success": true,
  "error": null,
  "data": {
    "featured": {
      "Crypto Candor": "@CryptoCandor#9152f3b054f692076a6882d1b58a30e8781cc8e6",
      "Dota 2 Divine": "@Dota2Divine#2d2c1a3dc98ca868d1f2bfe76b8a1bb820857c5a",
      "UC Berkeley": "@UCBerkeley#de0fcd76d525b1db36f24523e75c28b542e92fa2",
      "Mothers Basement": "@MothersBasement#65dee913b4f26144d573118b2e87704e03d01b8f",
      "bluedrake42": "@bluedrake42#32de523ba228dd3f3159eb5a6cc07b6fd51f4dc0",
      "Tech Fox": "@TechFox#2305db36455aa0d18571015b9e9bd0950262aa0f",
[...]

PR #2109, opened 2018-11-16
Show recommended subscriptions

seanyesmunt:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Breaking changes (bugfix or feature that introduces breaking changes)
- [x] Code style update (formatting)
- [ ] Documentation changes

## Fixes

Issue Number: #1825

We don't recommend any channels for subscriptions

## What is the new behavior?
Channels are recommended on your first run, if you clear out all of you unread subscriptions, and if you unsubscribe from all of the channels you originally subscribed too.

## Additional changes
- Cleaned up the `util/` folder. Now all files use the same naming convenstions `file-name.js`
- Split the subscriptions page into sub components, it was starting to get very big.


src/renderer/component/subscribeButton/view.jsx
```diff
@@ -17,23 +17,31 @@ type Props = {
   doChannelSubscribe: ({ channelName: string, uri: string }) => void,
   doChannelUnsubscribe: SubscribtionArgs => void,
   doNotify: ({ id: string }) => void,
+  firstRunCompleted: boolean,
 };
 
 export default (props: Props) => {
   const {
-    channelName,
```
seanyesmunt: Before we were passing the uri and the channel name in, but we can just get the channel name from the uri

src/renderer/component/subscribeButton/view.jsx
```diff
@@ -42,14 +49,19 @@ export default (props: Props) => {
       onClick={e => {
         e.stopPropagation();
```
skhameneh: I would suggest moving this logic out of the inline markup

src/renderer/store.js
```diff
@@ -116,14 +115,7 @@ const persistOptions = {
...
   // read the data
-  transforms: [
-    subscriptionsFilter,
-    walletFilter,
-    contentFilter,
-    fileInfoFilter,
-    appFilter,
-    compressor,
-  ],
+  transforms: [walletFilter, contentFilter, fileInfoFilter, appFilter, compressor],
```
skhameneh: Why the change in formatting?
*Mostly curious

seanyesmunt: I removed the `subscriptionsFilter` so prettier decided it could all be on one line

PR #2144, opened 2018-12-12
Newnew

### LINKED ISSUE #1976: App frontpage redesign

seanyesmunt:
Created a new PR for #2102 because that had so many comments.

## PR Checklist
- [x] This PR introduces breaking changes and I have provided a detailed explanation below.

## PR Type
- [x] Breaking changes (bugfix or feature that introduces breaking changes)

## Fixes

I borked https://github.com/lbryio/lbry-desktop/pull/2053 while attempting to have a sane rebase. This PR refactors the CSS and makes semantic changes in the rendered HTML.

Closes #1976.

Nothing the average user would notice, aside from styling.

## What is the new behavior?

New hotness. Sass is (more) organized and easier to manage. It still isn't perfect but it's close.


src/renderer/scss/component/_button.scss
```diff
@@ -1,232 +1,160 @@
...
-      box-shadow: none;
-    }
+  &:hover {
+    background-color: $lbry-green-3;
   }
 
-  &.btn--external-link {
-    color: $lbry-blue-1;
+  &.btn--play {
+    background-image: url("data:image/svg+xml,%3Csvg viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E %3Cg stroke='white' stroke-width='2' fill='white' fill-rule='evenodd' stroke-linejoin='round'%3E %3Cpolygon points='5 21 5 3 21 12'/%3E %3C/g%3E %3C/svg%3E");
```
seanyesmunt: These should stil use the `Icon` component

src/renderer/scss/component/_button.scss
```diff
@@ -1,232 +1,160 @@
...
-    background-color: inherit;
-    border-radius: 0;
-    box-shadow: none;
-    color: inherit;
-    font-size: inherit;
-    font-weight: inherit;
-    min-width: 0;
-  }
+.btn--primary {
+  align-self: center; // fixes buttons next to tall elements inside one with `display: flex`
```
seanyesmunt: I think is causing alignment issues. See the transactions table. The `Export` button isn't aligned with the filter dropdown because it's centered vertically. Maybe we could remove this and give buttons a `height`?

seanyesmunt: Eh, it's not a big deal. Your solution is probably better.

src/renderer/scss/component/_form-field.scss
```diff
@@ -84,30 +39,39 @@
...
 .form-field__help,
 .form-field__label,
 .form-field__error {
-  font-size: 12px;
   font-weight: 500;
 }
 
 .form-field__error {
-  color: $lbry-red-3 !important;
+  color: $lbry-red-5;
```
seanyesmunt: This doesn't seem to work for labels on inputs. See the `Send & Recieve` page.

src/renderer/scss/component/_header.scss
```diff
@@ -4,39 +4,144 @@
...
-  margin-left: auto;
 
-  .btn {
-    margin-left: $spacing-width * 1/3;
+  &:first-of-type {
+    html[data-theme='dark'] & {
+      border-color: $lbry-gray-5;
+    }
+
+    @media (min-width: 601px) {
```
seanyesmunt: Should these be variables? If it's specific to header stuff maybe just add a comment because it's confusing from a glance.

src/renderer/scss/component/_header.scss
```diff
@@ -4,39 +4,144 @@
...
+  &:disabled {
+    opacity: 0.3;
+  }
 }
 
-.header__navigation {
-  display: flex;
-  justify-content: space-between;
+.header__navigation-item--back {
+  background-image: url("data:image/svg+xml,%3Csvg viewBox='0 0 24 24' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink'%3E %3Cg stroke='black' stroke-width='2' fill='none' fill-rule='evenodd' stroke-linecap='round'%3E %3Cpath d='M4, 12 L21, 12'/%3E %3Cpolyline stroke-linejoin='round' transform='translate(7.000000, 12.000000) scale(-1, 1) translate(-7.000000, -12.000000)' points='3 4 11 12 3 20'/%3E %3C/g%3E %3C/svg%3E");
```
seanyesmunt: These should all use the `Icon` component

src/renderer/scss/init/_gui.scss
```diff
@@ -0,0 +1,204 @@
+// Generic html styles used accross the App
```
seanyesmunt: I'm not able to highlight anything in the app. We shouldn't be preventing this.

src/renderer/page/search/view.jsx
```diff
@@ -48,52 +48,54 @@ class SearchPage extends React.PureComponent<Props> {
...
                   body={__('This is the resolution of a LBRY URL and not controlled by LBRY Inc.')}
                 >
                   <Icon icon={ICONS.HELP} />
                 </ToolTip>
-              </div>
+              </h1>
               {isChannel ? (
                 <ChannelTile size="large" uri={uri} />
               ) : (
                 <FileTile size="large" displayHiddenMessage uri={uri} />
```
seanyesmunt: The placeholder seems broken for this top result on the search page. For channels it looks fine.

src/renderer/component/categoryList/view.jsx
```diff
@@ -250,38 +251,30 @@ class CategoryList extends PureComponent<Props, State> {
...
             <Button button="link" navigate="/settings" label={__('here')} />.
-          </div>
+          </p>
         ) : (
-          <div
-            className={classnames('card-row__scrollhouse', {
-              'card-row__scrollhouse--sub-component': isSubComponent,
-            })}
+          <ul
+            className="media-scrollhouse"
```
seanyesmunt: There used to be a min-height on this for cases like the empty cards on line 307. Currently suggested subscriptions don't render any cards until something is loaded because there is no height. Maybe the placeholders just need a height?

src/renderer/component/channelTile/view.jsx
```diff
@@ -75,7 +77,7 @@ class ChannelTile extends React.PureComponent<Props> {
             </React.Fragment>
           )}
           {subscriptionUri && (
-            <div className="card__actions">
+            <div className="media__actions">
               <SubscribeButton uri={subscriptionUri} />
```
seanyesmunt: This is misaligned on the search page
screen shot 2018-12-12 at 3 22 17 pm

src/renderer/page/file/view.jsx
```diff
@@ -163,90 +159,85 @@ class FilePage extends React.Component<Props> {
...
-            ) : (
-              <div
-                className={classnames('content__empty', {
-                  'content__empty--nsfw': shouldObscureThumbnail,
-                })}
-              >
-                <div className="card__media-text">
-                  {__("Sorry, looks like we can't preview this file.")}
-                </div>
+        {showFile && <FileViewer className="content__embedded" uri={uri} mediaType={mediaType} />}
```
seanyesmunt: If we are removing the wrapper on this, we should probably add some sort of max-height, then add a background so the thumbnail, content doesn't get huge. Most thumbails aren't great on large screens, and this will only be exacerbated on larger screens. 
 
screen shot 2018-12-12 at 3 28 55 pm

screen shot 2018-12-12 at 3 30 24 pm

src/renderer/component/categoryList/view.jsx
```diff
@@ -216,26 +214,29 @@ class CategoryList extends PureComponent<Props, State> {
...
       obscureNsfw,
-      isSubComponent,
     } = this.props;
     const { canScrollNext, canScrollPrevious } = this.state;
     const isCommunityTopBids = category.match(/^community/i);
     const showScrollButtons = isCommunityTopBids ? !obscureNsfw : true;
 
-    // isSubComponent is a hack, this component should be able to handle this with proper overflow styling
-
     return (
```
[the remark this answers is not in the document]
NetOperatorWibby: This is probably due to the additional icon div that appears when you have something downloaded. If you scroll, do you see icons?

src/renderer/component/sideBar/view.jsx
```diff
@@ -23,64 +23,59 @@ const SideBar = (props: Props) => {
...
-                    : label
-                }
-                icon={icon}
-              />
-            </li>
-          ))}
-        </ul>
-        <hr />
+    <nav className="navigation">
+      <div className="navigation__links">
```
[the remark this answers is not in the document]
NetOperatorWibby: No, there's already a `ul` surrounding the account links. I don't think Explore and Subscriptions need to be `li`s.

src/renderer/scss/init/_gui.scss
NetOperatorWibby: Do we need the ability to highlight everything? Titles, markdown content, wallet addresses and so on make sense but not things like transaction fees.

src/renderer/component/categoryList/view.jsx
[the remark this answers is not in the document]
seanyesmunt: Yeah it is. Should this always keep a space for that div? If you start downloading something, then go to the home page before it starts, it jumps when the icon is added.

src/renderer/scss/init/_gui.scss
seanyesmunt: Probably not everything, but I don't think we should prevent highlighting if users want to. It's a lot more work to have to opt into something if we want it to be highlightable.

src/renderer/component/sideBar/view.jsx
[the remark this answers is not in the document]
seanyesmunt: It is a list of items though, even if it's only two. I don't think we lose anything from adding it, but we gain accessibility improvements, which we will need to focus on more for web/tv uis to better handle keyboard/remote navigation.

src/renderer/component/categoryList/view.jsx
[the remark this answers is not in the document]
NetOperatorWibby: I'll add a height to it so the jumping doesn't happen.

src/renderer/scss/component/_button.scss
NetOperatorWibby: I thought this was going to be resolved after this PR was merged?

seanyesmunt: If we are adding new icons I would prefer they are implemented using the `Icon` component, just so it doesn't slip through the cracks.

NetOperatorWibby: We had a whole conversation about this. It'd require creating a new repo to replicate the react-feathers module. Should this be done now, or later?

seanyesmunt: It wouldn't require a new repo, that might be a better long-term solution. This could be implemented by something like this:

```
// in component/common/icon

const customIcons = {
      icons.PLAY: <svg>...</svg>,
      icons.EYE: <svg>...</svg>,
      ...
    }
const Icon = customIcons[this.props.icon] || FeatherIcons[this.props.icon];
```

Then we have all of these new custom icons in one place.

NetOperatorWibby: I wish this was brought up when we had that conversation.

seanyesmunt: My fault. Sorry about that, I thought we were on the same page.

src/renderer/component/walletBalance/view.jsx
```diff
@@ -10,10 +10,11 @@ const WalletBalance = (props: Props) => {
   const { balance } = props;
   return (
     <section className="card card--section card--wallet-balance">
-      <div>
-        <div className="card__title">{__('Balance')}</div>
-        <span className="card__subtitle">{__('You currently have')}</span>
-      </div>
+      <header className="card__header">
```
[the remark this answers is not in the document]
NetOperatorWibby: I thought we weren't worrying about making things responsive.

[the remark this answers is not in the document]
NetOperatorWibby: Why are we showing all these decimal points? We don't in the header bar. We don't in the hover text either.

```sass
display: block;
max-width: 400px;
overflow: hidden;
padding-right: 7rem;
text-overflow: ellipsis;
white-space: nowrap;
```

This does not look good on the `<span/>` surrounding the amount. You have ellipsis and extra space to add "LBC" in an `::after` pseudo-element.

Another option is to have the numbers fade (to replace the ellipsis).

Let me know which direction you want me to go in.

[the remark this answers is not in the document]
seanyesmunt: We aren't worrying about mobile design, but the minimum app width still needs to look good.

We should show the full amount here, it's the only place you can see it. I think it's fine to show an estimate in the header because it goes to this page when you click on it, and shows the full amount in the hover text. It's a bug that we show an estimate when hovering over the large balance (I guess we don't even really need that though). 

What about just dropping the `font-size` down to `4rem`?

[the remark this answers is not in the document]
NetOperatorWibby: I'll make it `3.5rem` so the Rewards box has more breathing room at `950px`.

src/renderer/page/search/view.jsx
NetOperatorWibby: I must have a great connection because I don't see these hitches. I'm adding min-heights to things now.

seanyesmunt: Try clearing cache then try again. It might be already fetched. Or just search for another channel.

src/renderer/scss/component/_header.scss
seanyesmunt: If 600px is a global width that will trigger ui changes, it should be a variable. But if only the header is changing at 600px then a comment explaining that will help.

NetOperatorWibby: Ah gotcha. I'll add comments to all media queries then.

Issue #2151, opened 2018-12-19
Redesign Fixes


seanyesmunt:
### Outstanding issues
- <> navigation issues sometimes (https://spee.ch/7/redesign-scroll)

### Fixed
- ~call reward list after getting confirmation from lbry.io - this ensures existing accounts show correct rewards.~
- ~Publish page - Price > value and drop down alignment (same with other value/dropdown pairs, i.e. settings page)~
- ~Publish page - choose thumbnail / enter url spacing~
- ~last item in row is not sized properly~
- ~File page viewer doesn't fit most thumbnails and related content requires more scrolling. I liked the related content to be on the right hand side.~
- ~Requesting stream text hard to see in some cases (previous app greyed out the thumb)~
- ~subscribe/tip/share text alignment~
- ~reward ribbon alignment in title~
- ~file action icons alignment~
- ~Play/view icon is greyed out until mouse over - intentional ? Think it looks better colored always.~
  - Going to keep this
- ~autoplay button removed from file show page?~
    - Will be re-added at a future date
- ~Media properties hard to see on search/recommended~
- ~can't click hyperlink if "more" is not clicked:~ 
  - ~lbry://test1121212121#a8f41412179debf5883cced81656357a9eb559c3~
- ~Should spacing between different elements in the same column? See Rewards area of Reward overview page.~
- ~positive value on tx history are black text in dark mode~
- ~pending publishes not greyed out~
- ~All subscriptions / Latest only should be buttons similar to Refresh? ~
    - They act as "links" so the green text makes more sense
- ~Dark mode channel page - hover over on page number whites out value~
- ~Publish success message could use some styling~
- ~Homepage extra spacing at bottom of featured section~
- ~Rewards are all in a single column now - do we want this?~
- ~Clear the cache icon is not clear~
   - possibly will increase button size so this is more clear, but keeping it for now.
- ~Subscription - header needs more space~
- ~glossy tile loading gradient looks bad (looks like the app locks up before cards are rendered)~
-  ~Meta text in light mode hard to see~
- ~Button alignment on auth screen~
- ~Invite button misaligned~
- ~History table wrong color in dark mode~
- ~Thumbnail stretches past the container on large screens on the file page.~
  - Looks ok, but I think it could be handled better
- ~File action buttons on File Page not aligned on the smallest screen size~
- ~Comments button is very close to the "Comments" header~
- ~Outline buttons are hard to see in dark mode when hovering~
[...]

PR #2164, opened 2019-01-04
Redesign fixes

seanyesmunt:
Fixes from the new redesign changes

Issue Number: #2151


src/renderer/component/categoryList/view.jsx
```diff
@@ -1,12 +1,13 @@
 // @flow
```
seanyesmunt: I didn't do much here, mostly switched to `createRef` and made it so it played nice with the new scrollable wrapper.

src/renderer/redux/actions/publish.js
```diff
@@ -119,16 +119,17 @@ export const doUploadThumbnail = (filePath: string, nsfw: boolean) => (
...
+      json =>
+        json.success
+          ? dispatch({
+              type: ACTIONS.UPDATE_PUBLISH_FORM,
+              data: {
+                uploadThumbnailStatus: THUMBNAIL_STATUSES.COMPLETE,
+                thumbnail: `${json.data.url}${fileExt}`,
+              },
+            })
+          : uploadError(json.message)
```
seanyesmunt: Not sure why we keep seeing these format changes when I didn't touch this stuff :(

skhameneh: It's probably that atom plugin... I warned you of linter hell! 😆

seanyesmunt: Hm good point. I'll explore

PR #2169, opened 2019-01-08
feat: auto email confirmation

seanyesmunt:


src/renderer/component/userEmailVerify/view.jsx
```diff
@@ -1,104 +1,76 @@
...
+    this.emailVerifyCheckInterval = null;
     (this: any).handleResendVerificationEmail = this.handleResendVerificationEmail.bind(this);
   }
 
-  handleCodeChanged(event: SyntheticInputEvent<*>) {
-    this.setState({
-      code: String(event.target.value).trim(),
-    });
+  componentDidMount() {
+    this.emailVerifyCheckInterval = setInterval(() => this.checkIfVerified(), 5000);
```
skhameneh: Five seconds?  Seems a bit long, we should look into doing this with websockets instead of polling.

seanyesmunt: Yeah I figured that would be less hammering on the server. I agree.

Issue #2154, opened 2018-12-27
Unread count reset when navigating other channel pages


tzarebczan:
## The Issue
Thanks to a community member (Brendon) for spotting this one. 
If you are subscribed to a channel and then navigate to any other page (besides 1), the unread count gets reset to 10 and the `new` tag comes back. It takes a few seconds for this to occur after navigating to the other page.

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

seanyesmunt:
Fixed in #2176

PR #2176, opened 2019-01-10
Fixes/Changes for 0.27.0

seanyesmunt:
### Minor fixes
- #2154 
- Lower priority for channel name/date so they aren't the same color as the title
- Lots of spacing consistency improvements (modals will need some work moving forward)

### Changes
- Add "Invite" link to the left nav
- Add gerbil to the "All caught up" subscriptions view
- Moved related back to right side, tried to use grid to move it below/right with media queries. It's super close but I kept running into a lot of issues.

### DIFF
--- a/src/renderer/scss/component/_content.scss
+++ b/src/renderer/scss/component/_content.scss
@@ -20,7 +20,7 @@
   background-repeat: no-repeat;
   background-size: cover;
   margin: auto;
-  // max-height: var(--file-max-height);
+
   &:not(.card__media--nsfw) {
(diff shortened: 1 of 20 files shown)


src/renderer/scss/component/_content.scss
```diff
@@ -19,9 +19,8 @@
   background-position: 50% 50%;
   background-repeat: no-repeat;
   background-size: cover;
-  max-width: calc(var(--file-max-width) * (16 / 9));
   margin: auto;
-
+  // max-height: var(--file-max-height);
```
skhameneh: remove

PR #2178, opened 2019-01-11
Add support for sandbox applications

skhameneh:
Can test with `lbry://FriendsOrFood#40b303c6714775a9863c8c1bda97bf396359c0c6`

This should be close to final working functionality for this app (minimum functionality, wallet APIs have not been started).

I have investigated WebView and BrowserView, concluding that BrowserView lacks in functionality and may introduce additional security concerns.  With the current implementation of WebView, there is no IPC access (I consider this a good thing, since module exploits should be rendered null), as a result I am working on an acceptable IPC channel.  I am currently consider WebSocket support for IPC communication as this will be the most platform agnostic communication that also offers adequate API security.

This current implementation has no finalized API channel for packages to request funds from the user.  This implementation for the in-app container is final (100% of a static view area is usable for all packages, just like an iframe would provide on other platforms).


seanyesmunt:
screen shot 2019-01-11 at 11 53 38 am

Seeing this when I try and run the game

package.json
```diff
@@ -46,18 +46,21 @@
...
+    "keytar": "^4.3.0",
+    "lbry-format": "https://github.com/lbryio/lbry-format.git",
     "lbry-redux": "lbryio/lbry-redux#820b6eeadf9e58c1e5027fd4d92808ba817b410e",
     "lbryinc": "lbryio/lbryinc#68c8ce6b7608dabc9180fff9b4841d10e1c62284",
     "localforage": "^1.7.1",
     "mammoth": "^1.4.6",
     "mime": "^2.3.1",
     "mixpanel-browser": "^2.17.1",
     "moment": "^2.22.0",
+    "node-fetch": "^2.3.0",
```
seanyesmunt: We aleady bring in `axios` for fetching in the build process. I'm in favor of using `node-fetch` for the main process, it might help to dump axios while were adding this.
https://github.com/lbryio/lbry-desktop/blob/master/build/downloadDaemon.js#L46

src/main/startSandbox.js
```diff
@@ -0,0 +1,28 @@
...
+import unpackByOutpoint from './unpackByOutpoint';
+
+// Polyfills and `lbry-redux`
+global.fetch = require('node-fetch');
+
+global.window = global;
+// eslint-disable-next-line import/no-commonjs,global-require
+const { Lbry } = require('lbry-redux');
+
+delete global.window;
```
seanyesmunt: Why delete this?

src/main/startSandbox.js
```diff
@@ -0,0 +1,28 @@
+import express from 'express';
+import unpackByOutpoint from './unpackByOutpoint';
+
+// Polyfills and `lbry-redux`
+global.fetch = require('node-fetch');
+
+global.window = global;
+// eslint-disable-next-line import/no-commonjs,global-require
+const { Lbry } = require('lbry-redux');
```
seanyesmunt: Does this need to be here? Or could we use `import` and move it to the top, then remove the eslint disable line.

src/main/startSandbox.js
```diff
@@ -0,0 +1,28 @@
...
+  const sandbox = express();
+  const port = 5278;
+
+  sandbox.get('/set/:outpoint', async (req, res) => {
+    const { outpoint } = req.params;
+    const resolvedPath = await unpackByOutpoint(Lbry, outpoint);
+
+    sandbox.use(`/sandbox/${outpoint}/`, express.static(resolvedPath));
+
+    res.send(`/sandbox/${outpoint}/`);
```
seanyesmunt: Is this needed?

src/renderer/component/fileRender/view.jsx
```diff
@@ -59,6 +120,20 @@ class FileRender extends React.PureComponent<Props> {
     return viewer || unsupported;
   }
 
+  escapeListener(e) {
+    if (e.keyCode==27){
```
seanyesmunt: I'm not able to enter full screen. Is this added?

src/renderer/component/fileViewer/internal/player.jsx
```diff
@@ -193,19 +194,36 @@ class MediaPlayer extends React.PureComponent {
...
 
   renderFile() {
     // This is what render-media does with unplayable files
-    const { fileName, downloadPath, contentType, mediaType } = this.props;
+    const { claim, fileName, downloadPath, contentType, mediaType } = this.props;
+
+    if (MediaPlayer.SANDBOX_TYPES.indexOf(contentType) > -1) {
+      const outpoint = `${claim.txid}:${claim.nout}`;
+
+      return fetch(`http://localhost:5278/set/${outpoint}`)
```
seanyesmunt: This endpoint should be a constant

src/renderer/component/fileViewer/internal/player.jsx
```diff
@@ -259,8 +277,9 @@ class MediaPlayer extends React.PureComponent {
 
     // Files
     const isLoadingFile = !fileSource && isFileType;
+    const isLbryPackage = /application\/x(-ext)?-lbry$/.test(contentType);
```
seanyesmunt: Can this just be replaced with the `this.supportedType()` function?

src/renderer/component/fileViewer/internal/player.jsx
```diff
@@ -274,6 +293,8 @@ class MediaPlayer extends React.PureComponent {
       // Show unsupported error message
     } else if (isUnsupported || isUnplayable) {
       loader.loadingStatus = isUnsupported ? unsupportedMessage : unplayableMessage;
+    } else if (isLbryPackage && !isLoadingFile) {
+      loader.loadingStatus = false;
```
seanyesmunt: Nevermind, I see it's used down here. It just seems kinda weird having very similar functionality.  Maybe `isLbryPackage` can be moved to its own function?

src/main/startSandbox.js
skhameneh: Should never try to emulate window inside Node, can cause side effects with other applications.
This is a patch to force this to work with `lbry-redux`, which I don't need, but `Lbry` is bundled with it, which I do need so I can make a daemon call for the file list.

See line 7.  This was grouped (spacing) with line 7.  The linter forced my grouping to break.

src/main/startSandbox.js
skhameneh: Try it with `import`, doesn't work.  See my other comment.

src/renderer/component/fileRender/view.jsx
skhameneh: Not yet, a lot of work needs to be done to make it work with Electron sandboxing.
The request from the sandbox goes through, but there are issues with dynamically sizing content that has already request a WebGL context.

src/renderer/component/fileViewer/internal/player.jsx
skhameneh: See next comment

src/renderer/component/fileViewer/internal/player.jsx
skhameneh: If you have logic written out for this, I'm all ears.
See the original line 263, I'm doing this to avoid breaking existing logic.

src/main/startSandbox.js
skhameneh: Yes, see `src/renderer/component/fileViewer/internal/player.jsx`, it uses the response to know where to go when the sandbox is ready.

package.json
skhameneh: I am in favor of this, do we want to do this now? (upcoming release, at the edge of release)
That entire function could use a rewrite to `async`/`await`.

src/renderer/component/fileRender/view.jsx
skhameneh: This is added so if fullscreen happens, the user can always escape.  Nothing is stopping existing apps from requesting it. *It can't be cancelled but it can be exited, Electron team hardcoded the events.

package.json
seanyesmunt: I think we might as well do it now. the `async/await` stuff might add a lot of work though. It should be super simple to swap axios for fetch in `downloadDaemon.js`

seanyesmunt:
This looks good (pending removal of axios). Lets wait until the release to merge this.

src/renderer/component/fileViewer/internal/player.jsx
```diff
@@ -11,7 +11,10 @@ import LoadingScreen from 'component/common/loading-screen';
 
 class MediaPlayer extends React.PureComponent {
   static MP3_CONTENT_TYPES = ['audio/mpeg3', 'audio/mpeg'];
+  static SANDBOX_TYPES = ['application/x-lbry', 'application/x-ext-lbry'];
```
kauffj: This is the most/only objectionable part for me. I would still strongly like to seek a way of triggering the behavior that does not require switching on media type. This is not me saying a custom media type is wrong; it is me saying that _only_ being able to trigger to the sandbox behavior via media type is wrong.

Please at least file/determine what metadata and sdk changes need to happen to facilitate this behavior without requiring this. @eukreign @lyoshenka

eukreign: There is a W3C proposal for something that seems very similar to `.lbry`, could we use that proposal instead of having our own?

Proposal: https://github.com/WICG/webpackage

The extension suggested in that proposal is `.wpk`:

https://github.com/WICG/webpackage/blob/master/draft-yasskin-dispatch-web-packaging.md#internet-media-type-registration

skhameneh: A lot that makes sense, but it's a proposal 19 years in the making with zero adoption and no working tools.

I will reach out to the current maintainer.

PR #2185, opened 2019-01-14
Simplify upgrade text

kauffj:
Proposed more concise upgrade prompt. Do we need to say upfront a restart is required or is the implication enough?


skhameneh:
@seanyesmunt this is merged

PR #2200, opened 2019-01-16
[WIP] update: babel@7, electron@4, webpack@4

seanyesmunt:
Doesn't seem to build properly on windows/linux but I've been messing around some of the new stuff.

Also changing to production mode in webpack for production. I think the performance increases are worth the degraded debugging.


seanyesmunt:
This was kinda just me messing around/exploring babel 7, and webpack 4. Closing for now, but some of these changes will be brought in after #2214 is merged.

PR #2207, opened 2019-01-17
Fix content aspect ratio breaking WebGL

skhameneh:
Removes forced aspect ratio from the top level, this is breaking the webview


skhameneh:
Bad rebase, ignore.

PR #2210, opened 2019-01-18
Less intrusive first run experience

seanyesmunt:
### Changes
- Removing first run modals for inline elements.
    - `ModalEmailCollection` => `FirstRunEmailCollection`
- Add yrbl to first run
- Hard code settings in reducer
- Default to dark mode

Issue #1683, opened 2018-06-25
Set and document icon naming standard


kauffj:
Should icon constants be named after the icon display or the icon purpose?

For example, should the constant for the subscribe icon, currently a heart, be `HEART` or `SUBSCRIBE`?

Typically (e.g. FontAwesome), icons are named after the display. But I'd actually argue that for a single application, the icons should be named after the purpose (i.e. `SUBSCRIBE` would be a better name than `HEART`).

Traditionally icons are named after the display, but I think once you are in a specific application, naming for purpose is superior since it helps encourage you to use the icon in one and only one way.

Either way, I'd like a choice made here and a short comment added to the top of the file indicating which you should do.

seanyesmunt:
I'm beginning to think it should be based on the purpose. Mainly to prevent the same icon being used for two different actions.

Ex: If we had a like button, maybe someone would just see the `HEART` icon and use that. If it was called `SUBSCRIBE` instead the chance of that happening is much lower.

seanyesmunt:
Done in https://github.com/lbryio/lbry-desktop/pull/2221

PR #2221, opened 2019-01-22
First run changes for invite page

seanyesmunt:
### Changes
- Standardize icon names for https://github.com/lbryio/lbry-desktop/issues/1683
- Brought publishes/downloads/history up to the top level of the side bar
- Add constants for all page names
- Add tooltip to direct users to the invite page
- Replace empty invites table with gerbil
screen shot 2019-01-26 at 7 28 23 pm
screen shot 2019-01-26 at 7 32 16 pm

### Notes
- Dependent on #2210 
- A lot of the work on the sidebar selector is to _hopefully_ make it super easy to add new guided tour items to links in the future if we want.
- Still determining exact copy.


src/renderer/redux/reducers/settings.js
```diff
@@ -19,6 +19,7 @@ const defaultState = {
     showUnavailable: getLocalStorageSetting(SETTINGS.SHOW_UNAVAILABLE, true),
     welcome_acknowledged: getLocalStorageSetting(SETTINGS.NEW_USER_ACKNOWLEDGED, false),
     email_collection_acknowledged: getLocalStorageSetting(SETTINGS.EMAIL_COLLECTION_ACKNOWLEDGED),
+    invite_acknowledged: getLocalStorageSetting(SETTINGS.INVITE_ACKNOWLEDGED, false),
```
seanyesmunt: All of these first run "settings" probably shouldn't live here. Not sure how tricky it would be to migrate them just to regular values saved by redux-persist. Then we can get rid of `getLocalStorageSetting`.

Works fine for now though, but something to keep in mind.

kauffj: kill it with :fire:

src/renderer/redux/selectors/app.js
```diff
@@ -127,124 +141,92 @@ export const selectNavLinks = createSelector(
...
+      active: currentPage === page,
+      guide: getGuideIfNecessary(page),
+    });
 
     const walletSubLinks = [
       {
-        label: 'Overview',
-        path: '/wallet',
-        active: currentPage === 'wallet',
+        ...buildLink('Overview', PAGES.WALLET),
```
skhameneh: Could make an array of these pairs and do a map.

seanyesmunt: Ah doh. Good idea.

seanyesmunt: Eh I think I'll leave it. Some of the links are slightly different, and then I would need to change all of those into objects, so there will still be some repetition.

PR #2225, opened 2019-01-24
Lbryweb

seanyesmunt:
I think this can be merged. The desktop app seems to be working fine.

#### Working on web
- It loads

#### Not working on web
- Resolve calls. The lbryweb daemon needs to be updated for 0.32.3 
  - @sayplastic I'm not sure where this should be done. I am still using https://github.com/sayplastic/docker-lbryweb-js to run this
- Images
  - Not sure where they need to be moved to inside `dist/web/`
- I think starting up with an encrypted account, I commented it out on `splash/view.jsx`


skhameneh:
Thanks @seanyesmunt I'm working on fixing some of the issues introduced

seanyesmunt:
@skhameneh I'll let you give this another look, but I think we can merge this then fix individual issues.

PR #2227, opened 2019-01-24
Lazy resolve homepage claims to avoid unnecessary page updates

seanyesmunt:
Dependent on https://github.com/lbryio/lbry-redux/pull/112

#### Minor
- Removed unused packages
- Removed some logic in the tooltip component that was slowing down rendering (trying to spend more time looking at rendering performance)
#### The big one
- Lazy resolve homepage claims

### Why
Currently, when you refresh the app/launch for the first time, there is a noticable ui block the moment all of the cards are rendered/images are fetched. This can be especially troublesome for the first run experience. 

This has been slightly improved with the new first run flow, but users can still run into a case where they see a non-disabled button, but are unable to click it until the homepage cards are completely rendered.

#### Old behavior (i try to click the button but end up double clicking and accidentally submitting the form because the ui was locked up)
giphy

#### New behavior
giphy 1


src/renderer/component/common/tooltip.jsx
```diff
@@ -29,63 +29,6 @@ class ToolTip extends React.PureComponent<Props, State> {
     };
   }
 
-  componentDidMount() {
```
seanyesmunt: this was slowing down the render. it wasn't really needed, more of a safe guard. It now requires manual testing to make sure a tooltip doesn't extend off the screen. I think this may be achievable with css, but haven't spent much time on it.

PR #2247, opened 2019-01-30
fix: yrbl style on subscriptions page

seanyesmunt:
The first run view on subscriptions wasn't using the new `<Yrbl />` component.


Invariant-Change:
At least I don't have to bother with the update log now. lol

PR #2250, opened 2019-01-30
feat: add mouse back/forward navigation

tzarebczan:
Tested on Windows, but it should work on Mac too (Sean to verify). For Linux, we'd need Electron 3+ (https://github.com/electron/electron/pull/15441)  or https://github.com/jostrander/mouse-forward-back


seanyesmunt:
I think we can remove the mac swipe since it doesn't seem to be doing anything.

tzarebczan:
Done...I had some trouble with npm packages last night.

PR #2263, opened 2019-02-04
fix: show 'no preview available' for files correctly

seanyesmunt:
The games code removed a line that correctly showed a message saying "No preview available" for certain files.


seanyesmunt:
Manually tested with @tzarebczan 

@skhameneh Let me know if there are issues with this logic, but it seems to work.

Issue #2231, opened 2019-01-26
0.31 daemon changes needed app side


tzarebczan:
## The Issue
- [x] change download_directory > download_dir
- [x] settings_set format is new for setting/disabling key fee, download directory and possibly others we allow:
```{"jsonrpc":"2.0","method":"settings_set","params":{"key": "max_key_fee", "value": null}}```
```{"jsonrpc":"2.0","method":"settings_set","params":{"key": "max_key_fee", "value": "1 LBC"}}```
- [x] no need to pass sd hash when updating claim without a file (will fail)
- [x] if a file is deleted from hard disk, `file_path` will now be blank. This crashes the app currently. It's how the daemon keeps track of what's removed. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

tzarebczan:
These have been made on https://github.com/lbryio/lbry-desktop/commits/031-sdk-fixes. Needs a couple more fixes around unavailable files.

PR #2277, opened 2019-02-11
Support 0.31 SDK, vrooom

tzarebczan:
Fixes https://github.com/lbryio/lbry-desktop/issues/2231

### DIFF
--- a/src/renderer/page/file/view.jsx
+++ b/src/renderer/page/file/view.jsx
@@ -58,8 +58,7 @@ class FilePage extends React.Component<Props> {
   ];
 
   componentDidMount() {
-    // eslint-disable-next-line no-unused-vars
-    const { uri, fileInfo, fetchFileInfo, fetchCostInfo, setViewed, isSubscribed } = this.props;
+    const { uri, fetchFileInfo, fetchCostInfo, setViewed, isSubscribed } = this.props;
 
     if (isSubscribed) {
       this.removeFromSubscriptionNotifications();
--- a/src/renderer/page/file/view.jsx
+++ b/src/renderer/page/file/view.jsx
@@ -58,15 +58,14 @@ class FilePage extends React.Component<Props> {
   ];
 
   componentDidMount() {
+    // eslint-disable-next-line no-unused-vars
     const { uri, fileInfo, fetchFileInfo, fetchCostInfo, setViewed, isSubscribed } = this.props;
 
     if (isSubscribed) {
       this.removeFromSubscriptionNotifications();
     }
-
-    if (fileInfo === undefined) {
-      fetchFileInfo(uri);
-    }
+    // always refresh file info when entering file page
+    fetchFileInfo(uri);
 
     // See https://github.com/lbryio/lbry-desktop/pull/1563 for discussion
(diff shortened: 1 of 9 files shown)


src/renderer/page/file/view.jsx
```diff
@@ -58,15 +58,14 @@ class FilePage extends React.Component<Props> {
   ];
 
   componentDidMount() {
+    // eslint-disable-next-line no-unused-vars
```
seanyesmunt: Can we just remove `fileInfo` so we don't need this?

src/renderer/page/settings/view.jsx
```diff
@@ -80,15 +84,18 @@ class SettingsPage extends React.PureComponent<Props, State> {
...
   onKeyFeeChange(newValue: Price) {
+    this.props.setClientSetting(SETTINGS.LOCAL_MAX_KEY_FEE, newValue);
     this.setDaemonSetting('max_key_fee', newValue);
   }
 
   onKeyFeeDisableChange(isDisabled: boolean) {
-    this.setDaemonSetting('disable_max_key_fee', isDisabled);
+    this.props.setClientSetting(SETTINGS.DISABLE_MAX_KEY_FEE, isDisabled);
+    // null is default value passed to clear key fee
+    if (isDisabled) this.setDaemonSetting('max_key_fee');
```
seanyesmunt: We shouldn't keep this as a local setting. It should stay in the daemon only. Then we don't have the same setting being tracked in two places.

seanyesmunt:
To add defaults, we can add something here:
https://github.com/lbryio/lbry-desktop/blob/master/src/renderer/redux/reducers/settings.js#L47

```
daemonSettings: {
  [DAEMON_SETTINGS.MAX_KEY_FEE]: ...,
```

It would be nice if they are constants, but not a huge deal right now. We can open an issue for that with help wanted maybe.

src/renderer/page/settings/view.jsx
tzarebczan: But it will either be null or a value - then we need to check for the case of a value. I can change this...

seanyesmunt: Yeah.

```
this.setDaemonSetting('max_key_fee', isDisabled ? null : true);
```

I think that will work.

tzarebczan: Just removed the top part...otherwise, true is not a valid value. When you choose the other option, it sets a LBC/USD amount.  Changes have been pushed.

src/renderer/page/settings/view.jsx
```diff
@@ -80,15 +83,16 @@ class SettingsPage extends React.PureComponent<Props, State> {
   }
 
   onDownloadDirChange(newDirectory: string) {
-    this.setDaemonSetting('download_directory', newDirectory);
+    this.setDaemonSetting('download_dir', newDirectory);
   }
 
   onKeyFeeChange(newValue: Price) {
+    this.props.setClientSetting(SETTINGS.LOCAL_MAX_KEY_FEE, newValue);
```
seanyesmunt: This shouldn't need a client setting either.

PR #2282, opened 2019-02-18
feat: search filters

seanyesmunt:
### Changes
- Adds filters for search
- Removes `resultCount` from settings reducer, not it lives in search

screen shot 2019-02-18 at 12 21 55 pm

### DIFF
--- a/src/renderer/component/searchOptions/view.jsx
+++ b/src/renderer/component/searchOptions/view.jsx
@@ -1,0 +1,123 @@
+// @flow
+import * as ICONS from 'constants/icons';
+import React, { useState } from 'react';
+import { SEARCH_OPTIONS } from 'lbry-redux';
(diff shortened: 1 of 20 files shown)


src/renderer/component/searchOptions/view.jsx
```diff
@@ -0,0 +1,123 @@
...
+                  onChange={() => setSearchOption(SEARCH_OPTIONS.CLAIM_TYPE, option)}
+                />
+              ))}
+            </fieldset>
+
+            <fieldset>
+              <legend className="search__legend--2">{__('File Types')}</legend>
+              {[
+                {
+                  option: SEARCH_OPTIONS.MEDIA_VIDEO,
```
skhameneh: Could generate this from an object

PR #2310, opened 2019-03-07
Web stability improvements and global error handler

seanyesmunt:


.eslintrc.json
```diff
@@ -53,6 +60,8 @@
     "react/prefer-stateless-function": 0,
     "react/sort-comp": 0,
     "jsx-a11y/media-has-caption": 0,
-    "no-underscore-dangle": 0
+    "no-underscore-dangle": 0,
+    "import/extensions": 0,
+    "react/default-props-match-prop-types": 0
```
seanyesmunt: There are a bunch of issues with flow not doing normal things with defaultProps. It will work correctly, but this lint rule doesn't interpret the code correctly.
https://github.com/facebook/flow/issues/6882

seanyesmunt:
@skhameneh I think this is ready. Still need to fix up some of the build setup to work better with lint. Let me know if you have a better way to do the prod stuff. I was trying to avoid having separate dev/prod configs.

package.json
```diff
@@ -23,11 +23,11 @@
     "compile:electron": "webpack --progress --config webpack.electron.config.js",
     "compile:web": "webpack --config webpack.web.config.js",
     "compile": "yarn compile:electron && yarn compile:web",
-    "dev:electron": "webpack-dev-server --hot --progress --config webpack.electron.config.js",
-    "dev:web": "webpack-dev-server --hot --progress --config webpack.web.config.js",
+    "dev:electron": "cross-env webpack-dev-server --hot --progress --config webpack.electron.config.js",
```
skhameneh: What is `cross-env` providing without flags?
I'd suggest setting `NODE_ENV=development` or removing it.

src/ui/component/errorBoundary/index.js
```diff
@@ -0,0 +1,4 @@
+import ErrorBoundary from './view';
+
+// TODO: bring in device/user(?) info in the future
+export default ErrorBoundary;
```
skhameneh: What is the intended purpose of this component?  The name gives me the impression this may be something we want to discuss.

package.json
seanyesmunt: Whoops. I ended up removing that, but never removed `cross-env`

src/ui/component/errorBoundary/index.js
seanyesmunt: The purpose is a global error handler. If we don't catch anything below, catch this at the top and report it. My idea was that we could include user info if we wanted to reach out/more info on what could have caused the issue, but if we go to internal-apis we should have everything there.

nikooo777:
@seanyesmunt this is bad. you should not include the slack hook, anyone can use it. I can create an API endpoint if needed, but this should go and the token should be regenerated/voided.

PR #2331, opened 2019-03-13
better web streaming

seanyesmunt:
Still some things to fix up but streaming actually works now for the web


src/ui/component/fileViewer/view.jsx
```diff
@@ -152,6 +152,7 @@ class FileViewer extends React.PureComponent<Props> {
   playContent() {
     const { play, uri, fileInfo, isDownloading, isLoading } = this.props;
 
+    // @if TARGET='app'
```
seanyesmunt: This will need to change, but not worried about it for now.

PR #2346, opened 2019-03-18
add unsupported messages

seanyesmunt:
A few messages for unsupported pages and file types


src/ui/component/publishForm/view.jsx
```diff
@@ -346,297 +347,308 @@ class PublishForm extends React.PureComponent<Props> {
     const shortUri = buildURI({ contentName: name });
 
     return (
-      <Form onSubmit={this.handlePublish}>
```
seanyesmunt: I just added the `React.Fragment` wrapper and the unsupported message for web here.

PR #2360, opened 2019-03-19
Blockers

seanyesmunt:
- Fix wallet balance background image
- Daemon download path so windows users don't see a firewall prompt
- check if the games server is already running so we can run two apps at the same time
- Fix channel publish counts
- Fix channel navigation from search


.eslintrc.json
```diff
@@ -35,6 +35,7 @@
     "standard/no-callback-literal": 0,
     "comma-dangle": ["error", "always-multiline"],
     "space-before-function-paren": ["error", "never"],
+    "jsx-quotes": ["error", "prefer-double"],
```
seanyesmunt: @skhameneh Do you know why this wouldn't be used? It's still using single quotes.

seanyesmunt: Nevermind. Not sure what I was doing earlier but if works.

PR #2387, opened 2019-03-27
Initial lazy-load and code-splitting

skhameneh:
This could probably be merged, merge if you'd like to be aggressive on testing.

Working on more additions.


src/ui/component/viewers/audioVideoViewer.jsx
```diff
@@ -38,7 +42,14 @@ class AudioVideoViewer extends React.PureComponent<Props> {
...
     };
 
-    this.player = videojs(this.videoNode, videoJsOptions, () => {});
+    import(
+        /* webpackChunkName: "videojs" */
+        /* webpackMode: "lazy" */
+        /* webpackPreload: true */
+        'video.js'
+    ).then((videojs) => {
+      me.player = videojs(me.videoNode, videoJsOptions, () => {});
```
seanyesmunt: Can we just use `this.player`?

webpack.base.config.js
```diff
@@ -12,9 +14,22 @@ const UI_ROOT = path.resolve(__dirname, 'src/ui/');
 const STATIC_ROOT = path.resolve(__dirname, 'static/');
 const DIST_ROOT = path.resolve(__dirname, 'dist/');
 
+console.log(ifProduction('production', 'development'))
```
seanyesmunt: Don't need

src/ui/component/viewers/audioVideoViewer.jsx
seanyesmunt: I see it's added in a few places, do we need it?

webpack.base.config.js
skhameneh: It's useful to see what's actively building

src/ui/component/viewers/audioVideoViewer.jsx
skhameneh: Since there's arrow fn's, yes

seanyesmunt: Arrow fn's don't scope `this`. If it was a regular function you would need it.

seanyesmunt: We have something similar here https://github.com/lbryio/lbry-desktop/blob/master/src/ui/component/fileViewer/internal/player.jsx#L286

seanyesmunt: Seems to work ok with `this`

PR #2395, opened 2019-03-28
New routing setup

seanyesmunt:
First off, I'd like to apologize. It isn't _as_ bad as it looks. There are a few large changes, then it's mostly making sure all of our current links work.

### Changes
- The largest change is in `src/ui/components/router/view.jsx` which sets up our new routing
- killed the `navigation` reducer
  - The browser is now the source of truth for routing
- `dispatch(doNavigate(...))` becomes `reachRouter.navigate(...)`

### Notes
There are still a few things broken that I am tracking. I think this is to a point where we can get it in now because the PR is already so big.


src/ui/component/common/icon-custom.jsx
```diff
@@ -114,4 +114,5 @@ export const customIcons = {
   [ICONS.UNSUBSCRIBE]: buildIcon(
     <path d="M 12,5.67 10.94,4.61 C 5.7533356,-0.57666427 -2.0266644,7.2033357 3.16,12.39 l 1.06,1.06 7.78,7.78 7.78,-7.78 1.06,-1.06 c 2.149101,-2.148092 2.149101,-5.6319078 0,-7.78 -2.148092,-2.1491008 -5.631908,-2.1491008 -7.78,0 L 9.4481298,8.2303201 15.320603,9.2419066 11.772427,13.723825" />
   ),
+  [ICONS.LBRY]: <span>LBRY</span>, // Fix this
```
seanyesmunt: Will add

src/ui/component/sideBar/view.jsx
```diff
@@ -1,4 +1,6 @@
 // @flow
```
seanyesmunt: This is the other big file. No longer bringing in the nav links from redux. Will add the tutorial back in a future PR.

PR #2405, opened 2019-04-02
Outstanding routing cleanup

seanyesmunt:
- Now jumps to top on navigate (will fix this not working on forward/back, but it's better for now)
- File card hover effect
- Removed insufficient credits modal for inline elements on the publish/file page. Now it will always be shown if users can't take any actions without getting more LBC.
  - Will come back to the copy after hearing from the growth team


src/ui/component/app/view.jsx
```diff
@@ -70,14 +70,14 @@ class App extends React.PureComponent<Props> {
     return (
       <div id="window" onContextMenu={e => openContextMenu(e)}>
         <Header />
-        <main className="page">
+        <section className="page">
```
seanyesmunt: The `<Page>` component contained `<main>` too. It makes more sense to keep there.

PR #2407, opened 2019-04-03
0.36.0 SDK changes

seanyesmunt:
### Notes
- The actual 0.36.0 changes weren't that bad, a ton of this is finally fixing flow errors in the files that had some. 
- The nsfw/tags still need a lot of work, currently if `nsfw` is checked on publish, it applies the `mature` tag.
- The goal of this was to just get a stable build with 0.36.0


src/ui/component/viewers/audioViewer.jsx
```diff
@@ -1,5 +1,3 @@
-// @flow
```
seanyesmunt: I removed flow from this file because there were ~30 type errors. If we have flow in a file, it should cover everything so we don't need to constantly use `--no-verify`

src/ui/component/common/form-components/form-field-price.jsx
```diff
@@ -23,7 +27,7 @@ export class FormFieldPrice extends React.PureComponent<Props> {
 
   handleAmountChange(event: SyntheticInputEvent<*>) {
     const { price, onChange } = this.props;
-    const amount = event.target.value ? parseFloat(event.target.value) : '';
+    const amount = event.target.value ? parseFloat(event.target.value) : undefined;
```
kauffj: `null`, not `undefined`, I think

src/ui/page/file/view.jsx
```diff
@@ -22,15 +20,15 @@ import getMediaType from 'util/get-media-type';
 import RecommendedContent from 'component/recommendedContent';
 
 type Props = {
```
kauffj: It looks like this component both receives the entire `Claim` and properties of the claim independently. Should it always be one or the other?

src/ui/redux/actions/content.js
```diff
@@ -319,7 +319,7 @@ export function doFetchClaimsByChannel(
                 uri: buildURI(
                   {
                     contentName: latest.channel_name,
-                    claimId: latest.value.publisherSignature.certificateId,
+                    claimId: latest.claim_id,
```
kauffj: Nice :D

flow-typed/publish.js
```diff
@@ -0,0 +1,52 @@
+// @flow
+
+declare type UpdatePublishFormData = {
```
kauffj: Naming is a little inconsistent between these two types.

Also, you may be able to use Intersection Types to avoid repeating the shared values across both (but I've never used this feature).

seanyesmunt: Yeah... the publish types were one of the first added. Now that I have a better idea of flow and have developed some patterns in other parts of the app this can be cleaned up a lot. Planning to clean it up when I add the tags to the publish page

src/ui/page/file/view.jsx
seanyesmunt: I think we should start moving to bring in individual properties that components use: see `makeSelectTitleForUri`, instead of large objects if the component only uses a couple of values.

It's faster for react to update because then it isn't doing tons of un-needed checks to see if an object changed at all.

src/ui/component/common/form-components/form-field-price.jsx
seanyesmunt: Inputs can't use `null` or react throws errors. Something to do with how javascript handles `null`
https://github.com/facebook/react/issues/11417

PR #2408, opened 2019-04-04
Fix everything I broke (I hope) - Use react-router

seanyesmunt:
- Will fix auth, scroll in a different PR


src/ui/analytics.js
```diff
@@ -1,7 +1,6 @@
 // @flow
 import { Lbryio } from 'lbryinc';
 import ReactGA from 'react-ga';
-import { globalHistory } from '@reach/router';
```
seanyesmunt: Will add back analytics stuff in another PR. They were busted anyway. Just want to get this in.

src/ui/component/router/view.jsx
```diff
@@ -21,49 +21,32 @@ import UserHistoryPage from 'page/userHistory';
 import SendCreditsPage from 'page/sendCredits';
 import NavigationHistory from 'page/navigationHistory';
 
-const ScrollHandler = props => {
```
seanyesmunt: Will add this back later, it will be a little different with react-router

PR #2435, opened 2019-04-18
Add setting to allow content Lbry Inc has censored

anonRQYQ5yb1jnL2y7XF7W9:
Lbry falsely advertises as censorship free. Users should be able to decide what content they can view.

### DIFF
--- a/src/ui/index.jsx
+++ b/src/ui/index.jsx
@@ -219,7 +219,10 @@ const init = () => {
   // @endif
 
   app.store.dispatch(doDownloadLanguages());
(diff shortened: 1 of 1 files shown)


kauffj:
@anonRQYQ5yb1jnL2y7XF7W9 I think everyone working at LBRY would prefer a world in which this setting were not legally necessary.

Our current legal advice is that our application ought to to have this "feature". We are looking at whether there are certain conditions under which we can remove it.

One user in the past has forked the app to remove it, but it appears out of date: https://github.com/PublicLBRY/lbry-desktop

PR #2439, opened 2019-04-19
displays snackbar instead of greenscreen on bad url

jessopb:
## Changes

Enables a showSnackBar that was previously erroring.

### DIFF
--- a/src/ui/component/wunderbar/view.jsx
+++ b/src/ui/component/wunderbar/view.jsx
@@ -21,7 +21,7 @@ type Props = {
   doFocus: () => void,
   doBlur: () => void,
   focused: boolean,
-  doShowSnackBar: ({}) => void,
+  doShowSnackBar: string => void,
 };
 
 type State = {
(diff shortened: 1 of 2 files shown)


src/ui/component/wunderbar/view.jsx
```diff
@@ -96,9 +96,9 @@ class WunderBar extends React.PureComponent<Props, State> {
 
     const query = value.trim();
     const showSnackError = () => {
-      doShowSnackBar({
-        message: __('Invalid LBRY URL entered. Only A-Z, a-z, 0-9, and "-" allowed.'),
-      });
+      doShowSnackBar(
+        <span>{'Invalid LBRY URL entered. Only A-Z, a-z, 0-9, and "-" allowed.'}</span>
```
seanyesmunt: This doesn't need to have a `<span>` wrapped around it. 

Also now that I'm thinking about it. We can use `isError: true`, to get the error snack bar.

Issue #2412, opened 2019-04-08
[CSS] things aren't lined up/ style inconsistencies


seanyesmunt:
- Subscriptions latest only page
  - Not horizontally lined up
<img width="136" alt="Screen Shot 2019-04-08 at 11 41 50 AM" src="https://user-images.githubusercontent.com/16882830/55737541-5769fe80-59f3-11e9-9dde-74e0d933058c.png">

- Subscriptions overview page
  - Also not lined up
  - Not bold like the other view
<img width="265" alt="Screen Shot 2019-04-08 at 11 42 39 AM" src="https://user-images.githubusercontent.com/16882830/55737593-74063680-59f3-11e9-9863-df30e1f07429.png">

- Subscriptions download toggle
  - Not lined up
<img width="265" alt="Screen Shot 2019-04-08 at 11 41 46 AM" src="https://user-images.githubusercontent.com/16882830/55737544-5a64ef00-59f3-11e9-8090-0ffe39ada60b.png">

seanyesmunt:
Working on this

seanyesmunt:
resolved in #2450

PR #2450, opened 2019-04-28
Make alignment great again

seanyesmunt:
For https://github.com/lbryio/lbry-desktop/issues/2412

~Dependent on https://github.com/lbryio/components/pull/8 and https://github.com/lbryio/components/pull/9~

#### Changes
- Tried to simplify the header/button styles
  - Removed excessive html markup
  - Tried to get rid of as many `top: {x}px` type styles as I could
- Brought back the round buttons
  - Square buttons are hard to make look polished
  - A lot of extra styles are needed to really differentiate buttons from inputs/static markup


src/ui/component/button/view.jsx
```diff
@@ -20,7 +20,6 @@ type Props = {
   description: ?string,
   type: string,
   button: ?string, // primary, secondary, alt, link
-  noPadding: ?boolean, // to remove padding and allow circular buttons
```
kauffj: Nice. We should kill properties that are style only like this whenever possible.

src/ui/scss/component/_header.scss
```diff
@@ -2,108 +2,81 @@
   position: fixed;
   width: 100%;
   height: var(--header-height);
-  align-items: center;
-  background-color: $lbry-white;
   display: flex;
   justify-content: space-between;
-  z-index: 2;
+  z-index: 2; // Main content uses z-index: 1, this ensures it always scrolls under the header
```
kauffj: Not necessary to adjust in this PR, but it can be good to make your z-index values SASS variables for better clarity.

src/ui/scss/component/_icon.scss
```diff
@@ -0,0 +1,10 @@
+// Not all icons are created equally... at least the react-feather ones aren't
+// Minor adjustments to ensure icons line up vertically
+
+.icon--Heart {
+  top: -1px;
```
kauffj: How will this scale when icons are rendered at different sizes?

seanyesmunt: That's a good point. This won't scale well. I think it will work if we use `em`. Currently our icons are all locked in at the same size, regardless of font size. Will keep that in mind going forward.

PR #2464, opened 2019-05-07
Issue2431 disable send tip

jessopb:
Issue reported: Send button is not disabled on first view when amount is empty.
Now: Send button is disabled/grayed out while box is empty.


package.json
```diff
@@ -185,7 +185,7 @@
     "yarn": "^1.3"
```
seanyesmunt: update this file to point to the new `lbry-redux` commit
`5fb6efae24205dfdfa133ca4f26bccce033570e1`

src/ui/component/dateTime/view.jsx
```diff
@@ -1,9 +1,10 @@
 // @flow
 import React from 'react';
 import moment from 'moment';
+import i18n from '../../i18n';
```
seanyesmunt: You should be able to just do `import i18n from 'i18n'`

package.json
seanyesmunt: you can update `lbryinc` too since I just merged that PR.

`43d382d9b74d396a581a74d87e4c53105e04f845`

jessopb:
Addressed in #2465 2465

PR #2466, opened 2019-05-07
General fixes

seanyesmunt:
### Changes
- Style consistency with inputs/cards to match new button styles
- More noticeable hover effect on cards/tiles

### Fixes
- Channels with only one page are now being fetched correctly

#### What it looks like
Screen Shot 2019-05-07 at 5 51 06 PM

Screen Shot 2019-05-07 at 6 23 37 PM

### DIFF
--- a/src/ui/component/transactionList/view.jsx
+++ b/src/ui/component/transactionList/view.jsx
@@ -2,7 +2,6 @@
 import * as icons from 'constants/icons';
 import * as MODALS from 'constants/modal_types';
 import * as React from 'react';
-import { List } from 'react-virtualized';
 import { FormField, Form } from 'component/common/form';
 import Button from 'component/button';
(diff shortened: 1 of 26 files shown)


src/ui/component/transactionList/view.jsx
```diff
@@ -103,7 +102,7 @@ class TransactionList extends React.PureComponent<Props> {
             </div>
           )}
         </header>
-        {!transactionList.length && <p>{emptyMessage || __('No transactions to list.')}</p>}
+        {!transactionList.length && <p className="card__subtitle">{emptyMessage || __('No transactions to list.')}</p>}
```
kauffj: If there's not a style for empty lists/cards/elements, can we add one? Probably more for @NetOperatorWibby than @seanyesmunt

PR #2471, opened 2019-05-09
moar release fixes

tzarebczan:
Fixes claim editing while keeping any existing tags + setting/clearing fee.
Fixes not being able to abandon claims if file was deleted

Beware of lint errors..some old and new. Everything works and we should clean them up later once tags are properly integrated.


src/ui/component/publishForm/view.jsx
```diff
@@ -217,10 +217,12 @@ class PublishForm extends React.PureComponent<Props> {
       otherLicenseDescription,
       name: this.props.name || undefined,
       contentIsFree: this.props.contentIsFree,
-      price: this.props.price,
+      feeAmount: this.props.fee.amount,
```
seanyesmunt: This should just be `fee`

Because if `fee` is undefined, it will error when trying to read `amount`

src/ui/redux/actions/file.js
```diff
@@ -33,15 +33,10 @@ export function doDeleteFile(outpoint, deleteFromComputer, abandonClaim) {
     // If the file is for a claim we published then also abandon the claim
     const myClaimsOutpoints = selectMyClaimsOutpoints(state);
     if (abandonClaim && myClaimsOutpoints.indexOf(outpoint) !== -1) {
-      const byOutpoint = selectFileInfosByOutpoint(state);
-      const fileInfo = byOutpoint[outpoint];
+      const txid = outpoint.slice(0, -2);
+      const nout = Number(outpoint.slice(-1));
```
seanyesmunt: This might be a little more readable:
```
const [txid, nout] = outpoint.split(':')
```

src/ui/redux/actions/publish.js
```diff
@@ -214,36 +222,56 @@ export const doPublish = (params: PublishParams) => (dispatch: Dispatch, getStat
...
     name,
     bid: creditsToString(bid),
     title,
     license,
-    license_url: licenseUrl,
     languages: [language],
     description,
-    thumbnail_url: thumbnail,
-    tags: [],
+    tags: claim && claim.value.tags,
```
seanyesmunt:
```
tags: claim && claim.value.tags || []
```

so we don't have to check that `tags` exitsts inside `publishPayload`

src/ui/redux/actions/publish.js
```diff
@@ -214,36 +222,56 @@ export const doPublish = (params: PublishParams) => (dispatch: Dispatch, getStat
...
+    publishPayload.thumbnail_url = thumbnail;
+  }
+
+  if (claim && claim.value.release_time) {
+    publishPayload.release_time = claim && Number(claim.value.release_time);
+  }
+
   if (nsfw) {
-    publishPayload.tags.push('mature');
+    if (publishPayload.tags && !publishPayload.tags.includes('mature')) {
```
seanyesmunt: We shouldn't need to check that `publishPayload.tags` exists here

PR #2473, opened 2019-05-10
RC fixes

seanyesmunt:
#### Changes
- Use port 9090 for webpack so internal apis can use it's default port
- Add back support for unlocking supports
  - Dependent on https://github.com/lbryio/lbry-redux/pull/144
- Fix phone collection modal closing on success
- Fix navigation away from error page in electron
  - also maintains navigation state, so you can still go back after
- Style consitency cleanup


src/platforms/electron/devServer.js
```diff
@@ -1,18 +1,13 @@
...
 
-// TODO: Spawn separate threads so realtime status logging can be used
-// without overwriting information/execptions logged by the compilers
-const logRealtime = str => {
-  let lineCount = (str.match(/\n/) || []).length + 1;
-  console.log('\u001B[' + lineCount + 'F\u001B[G\u001B[2K' + str);
-};
+// Ideally this would come from the same source that sets this port in webpack.web.config.js
+// but we can't inject a variable into this file because it is not being touched by webpack
+const WEBPACK_PORT = 9090;
```
kauffj: It looks like 9090 appears in several places in the code. Since we have to DRY break for pragmatic reasons here, the best way to handle this is:

- Note in a comment by the primary definition (presumably webpack) where all secondary usages occur
- Reference that devs should look at the primary definition in all secondary usages

src/platforms/electron/devServer.js
```diff
@@ -46,8 +41,8 @@ app.use(require('webpack-hot-middleware')(renderCompiler));
 app.use(renderInstance);
 app.use(express.static('dist/electron/static'));
 
-app.listen(8080, () => {
-  console.log(chalk.yellow.bold('Renderer listening on port 8080 (still compiling)'));
+app.listen(WEBPACK_PORT, () => {
+  console.log(chalk.yellow.bold('Renderer listening on port 9090 (still compiling)'));
```
kauffj: inject WEBPACK_PORT into string

src/ui/component/channelThumbnail/view.jsx
```diff
@@ -1,15 +1,44 @@
...
+  const { thumbnail, uri } = props;
+
+  function getColorClass() {
+    // Generate a random color class based on the first letter of the channel name
+    const { channelName } = parseURI(uri);
+    const initializer = channelName.charCodeAt(0) - 65; // will be between 0 and 57
+
+    let thumbnailColorClass = `channel-thumbnail__default--`;
+
+    switch (true) {
```
kauffj: I believe that `thumbnailColorClass += (initializer % 15)` or something close to this is equivalent to the entire `switch`

src/ui/component/transactionList/view.jsx
```diff
@@ -42,10 +43,8 @@ class TransactionList extends React.PureComponent<Props> {
   }
 
   isRevokeable(txid: string, nout: number) {
-    const { myClaims } = this.props;
-    // a claim/support/update is revokable if it
-    // is in my claim list(claim_list_mine)
-    return myClaims.has(`${txid}:${nout}`);
+    const { mySupports, myClaims } = this.props;
+    return !!mySupports[txid] || myClaims.has(`${txid}:${nout}`);
```
kauffj: I believe you can have multiple supports in a single transaction. Should supports also be indexed by outpoint rather than txid?

src/ui/scss/component/_channel.scss
```diff
@@ -1,49 +1,70 @@
 $cover-z-index: 0;
 $metadata-z-index: 1;
 
-.channel__cover {
+.channel-cover {
```
kauffj: thank you

src/ui/component/transactionList/view.jsx
seanyesmunt: Good point. Yeah, we probably should use it. Is this a blocker? RC testing has been using it. I can update after the release

src/ui/component/channelThumbnail/view.jsx
seanyesmunt: Yep. `initializer % 4`

src/ui/component/sideBar/view.jsx
```diff
@@ -74,35 +74,30 @@ class SideBar extends React.PureComponent<Props> {
         <ul className="navigation__links">
           {[
             {
-              ...buildLink(PAGES.ACCOUNT, 'Overview', ICONS.ACCOUNT),
+              ...buildLink(PAGES.ACCOUNT, __('Overview'), ICONS.ACCOUNT),
```
kauffj: ty for this

src/ui/component/socialShare/view.jsx
```diff
@@ -70,7 +70,7 @@ class SocialShare extends React.PureComponent<Props> {
       <React.Fragment>
         {speechShareable && (
           <div className="card__content">
-            <label className="card__subtitle">{__('Web link')}</label>
+            <label className="help">{__('Web link')}</label>
```
kauffj: Heh, I think you changed this class because I complained, but I'd argue these aren't "help" items either.

`card__label`, `card__action-label`? or something like this? it's fine if the classes are styled exactly the same, but using `help` when it is not "help" will render future style evolution harder

PR #2495, opened 2019-05-20
I18n fixes

kauffj:
This makes i18n workable.


kauffj:
@seanyesmunt fyi this will result in `index.js` accumulating translations while doing routing work/usage. Let me know if you take issue with that.

PR #2528, opened 2019-06-03
app uses auth token header

jessopb:


seanyesmunt:
Will merge after we release 0.33.0

PR #2539, opened 2019-06-07
crude i18n support

kauffj:
This adds a basic language selector, stores the language setting, refreshes the UI on change, and adds the first polish language file.

Much of this should be improved for multiple language support and so there is a better ongoing workflow, but this should help @Madiator2011 continue his work (@jsigwart).


src/ui/component/app/view.jsx
```diff
@@ -64,12 +66,12 @@ class App extends React.PureComponent<Props> {
...
   render() {
-    const { enhancedLayout } = this.props;
+    const { enhancedLayout, language } = this.props;
 
     return (
       <div id="window" onContextMenu={e => openContextMenu(e)}>
-        <Header />
-        <SideBar />
+        <Header language={language} />
+        <SideBar language={language} />
```
kauffj: @seanyesmunt I did this to trigger re-renders when the language changes. Is this the proper way to do this [given the `__` function is not a React component]?

kodxana: At this moment menu not work :)

kauffj: @kodxana it should now

src/ui/component/app/view.jsx
```diff
@@ -14,14 +14,16 @@ const TWO_POINT_FIVE_MINUTES = 1000 * 60 * 2.5;
...
   theme: string,
   updateBlockHeight: () => void,
   toggleEnhancedLayout: () => void,
   enhancedLayout: boolean,
 };
 
 class App extends React.PureComponent<Props> {
-  componentWillMount() {
+  // eslint-disable-next-line
+  UNSAFE_componentWillMount() {
```
seanyesmunt: This can be changed to `componentDidMount`

src/ui/component/app/view.jsx
seanyesmunt: Discussed on the standup, but I would move these to `{header, sideBar}/index.js`

PR #2545, opened 2019-06-12
fix: allow auth token to persist if the browser is closed

### LINKED ISSUE #2544: Users are not captured properly on lbry.tv
I have been going to  beta.lbry.tv and lbry.tv multiple times a week for a number of weeks now. Almost every time I visit, it thinks that I am a new user. 

This has the side effect of misinforming our analytics of the number of users to the site. It is also incorrectly flooding internal-apis with users that are not real users. I flagged it a few weeks back but I see its not fixed yet so I wanted to create an issue so we don't overlook it. 

I would expect we can easily leverage cookies,local storage and ip addresses to get this more accurate.

seanyesmunt:
Fixes #2544 

Not setting an expire time expires the auth token when the session ends.

### DIFF
--- a/src/ui/index.jsx
+++ b/src/ui/index.jsx
@@ -29,7 +29,7 @@ import { formatLbryUriForWeb } from 'util/uri';
 import 'scss/all.scss';
 
 const APPPAGEURL = 'lbry://?';
-
+const COOKIE_EXPIRE_TIME = 60 * 60 * 24 * 365; // 1 year
 // @if TARGET='app'
 const { autoUpdater } = remote.require('electron-updater');
 autoUpdater.logger = remote.require('electron-log');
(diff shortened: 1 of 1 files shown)


src/ui/index.jsx
```diff
@@ -101,6 +104,7 @@ Lbryio.setOverride(
         // @endif
         // @if TARGET='web'
         const { auth_token: authToken } = cookie.parse(document.cookie);
+        console.log('retrieved auth', authToken);
```
kauffj: Impressive inclusion for such a short PR :grinning: 

Any reason linter did not catch?

seanyesmunt: 🤦‍♂ 

I turned that lint rule off because it was a pain with all the console logs in our electron files. I'll add it back.

Issue #2410, opened 2019-04-05
OG image support for lbry.tv links


tzarebczan:
## The Issue
OG Images should pull from channel/claim thumbnails when shared. If no thumb is provided, use a generic one. 

Tell us your suggested solutions if you have any.

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

PR #2546, opened 2019-06-12
Dynamic meta tags for content

NetOperatorWibby:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #2410

Shared `lbry.tv` links are plain.

## What is the new behavior?

Shared `lbry.tv` links are enriched with metadata to make them more appealing to scrollers.

### DIFF
--- a/src/platforms/web/server.js
+++ b/src/platforms/web/server.js
@@ -1,12 +1,76 @@
+const { readFileSync } = require('fs');
 const express = require('express');
+const fetch = require('node-fetch');
 const path = require('path');
 const app = express();
 const port = 1337;
+const headRegex = /(<head>).*(<\/head>)/g;
 
 app.use(express.static(__dirname));
 
-app.get('*', function(req, res) {
-  res.sendFile(path.join(__dirname, '/index.html'));
+app.get('*', async (req, res) => {
+  let finalHtml = '';
+  let identifier = req.url.split('/').pop();
+  let isAppPage = req.url.includes('/$/');
+  let isChannel = req.url.includes('/@');
+  let queryUrl = '';
+
+  switch (true) {
+    case isAppPage:
+      return res.sendFile(path.join(__dirname, '/index.html'));
+
+    case isChannel:
+      queryUrl = `https://chainquery.lbry.io/api/sql?query=SELECT channel.name as channel, claim.name, claim.description, claim.language, claim.thumbnail_url, claim.title FROM claim left join claim channel on claim.publisher_id = channel.claim_id where claim.claim_id="${identifier}"`;
+      break;
+
+    default:
+      // isContentPage
+      queryUrl = `https://chainquery.lbry.io/api/sql?query=SELECT * FROM claim where claim_id="${identifier}"`;
+      break;
+  }
+
+  [finalHtml] = await Promise.all([
+    fetch(queryUrl)
+      .then(response => response.json())
+      .then(claim => {
+        if (!claim.success || !claim.data[0]) {
+          return path.join(__dirname, '/index.html');
+        }
+
+        claim = claim.data[0];
+
+        // const claimAuthor = claim.channel_name ? claim.channel_name : 'Anonymous'; // Currently unable to get creator name
+        const claimDescription =
+          claim.description && claim.description.length > 0 ? claim.description : `Watch ${claim.title} on LBRY.tv`;
+        const claimLanguage = claim.language || 'en_US';
+        const claimThumbnail = claim.thumbnail_url || '/og.png';
+        const claimTitle = `${claim.title} on LBRY.tv` || 'LBRY.tv'; // `${claim.title} from ${claimAuthor} on LBRY.tv`;
+        const claimUrl =
+          `https://beta.lbry.tv/${claim.name}${claim.claim_id ? `/${claim.claim_id}` : ''}` || 'https://beta.lbry.tv';
+        // This comment was inlined and preventing the rest of the script to load
+        const indexHtml = readFileSync(path.join(__dirname, '/index.html'), 'utf8')
+          .replace(/\s\s/gm, '')
+          .replace(/>\s</gm, '><')
+          .replace('// Use relative path if we are in electron', '');
+        let finalTags = '';
+
+        finalTags += '<meta charset="utf8"/>';
+        finalTags += `<meta name="description" content="${claimDescription}"/>`;
+        finalTags += `<meta name="keywords" content="${claim.tags ? claim.tags.toString() : ''}"/>`;
+        finalTags += `<meta name="twitter:image" content="${claimThumbnail}"/>`;
+        finalTags += `<meta property="og:description" content="${claimDescription}"/>`;
+        finalTags += `<meta property="og:image" content="${claimThumbnail}"/>`;
+        finalTags += `<meta property="og:locale" content="${claimLanguage}"/>`;
+        finalTags += `<meta property="og:site_name" content="LBRY.tv"/>`;
+        finalTags += `<meta property="og:type" content="website"/>`;
+        finalTags += `<meta property="og:url" content="${claimUrl}"/>`;
+        finalTags += `<title>${claimTitle}</title>`;
+
+        return indexHtml.replace(indexHtml.match(headRegex)[0], finalTags);
+      }),
+  ]);
+
+  res.send(finalHtml);
 });
 
 app.listen(port, () => console.log(`UI server listening at http://localhost:${port}`));
--- a/static/index.html
+++ b/static/index.html
@@ -8,7 +8,7 @@
     <!-- @if TARGET='web' -->
     <title>lbry.tv</title>
     <meta property="og:url" content="https://beta.lbry.tv" />
-    <meta property="og:title" content="LBRY On The Web" />
+    <meta property="og:title" content="LBRY.tv" />
     <meta property="og:description" content="All your favorite LBRY content in your browser." />
     <meta property="og:image" content="/og.png" />
     <!-- @endif -->


NetOperatorWibby:
Included in this PR is commented code for tag inclusion, whenever that drops. Also, when navigating away from content to a regular page, the `title` attribute doesn't update (yet). It _should_ default to `LBRY.tv` or something similar.

src/ui/page/file/view.jsx
```diff
@@ -147,6 +148,45 @@ class FilePage extends React.Component<Props> {
       showToast,
     } = this.props;
 
+    const claimAuthor = `${claim.value.author} (${claim.channel_name})`;
```
tzarebczan: We don't really expose author even though it's available - it should use channel if there is one (could be anonymous too)

seanyesmunt:
Does this work? I figured we would need to change code in `platforms/web/server.js` since we aren't server rending our app.

NetOperatorWibby:
Idk how to test on the web but it updates in the app.

seanyesmunt:
To test the basic og image I used (ngrok)[https://ngrok.com/]. I would run the web version with `yarn run:web` (not `yarn dev:web`, that didn't seem to work right since it was a webpack-dev-server).

Then point ngrok at `localhost:1337` and it gives you a regular url that points to your computer. Then sharing that url will show the og tags.

NetOperatorWibby:
It doesn't appear to work.

You can just use your computer's name and send an iMessage to yourself to test. (`http://computer.local:1337/path/to/content`)

NetOperatorWibby:
I don't see how making edits to `platforms/web/server.js` will work. Navigating the app doesn't trigger updates there. **Hard** refreshes trigger updates but obviously that doesn't happen naturally.

Finally, useful information isn't found here aside from the path.

NetOperatorWibby:
Is it possible to have the video URL part of the claim data? There's no way to have a video preview if I don't have the direct mp4 link.

tzarebczan:
Let's stick to thumbnail previews for now, that would have to be tackled another day.

We don't know what streaming solution LBRY.tv will go with yet, but it should be possible to make some url later on that will point to the video itself.

tzarebczan:
Per our discussion today, might be better to grab this data via a chainquery fetch. We'll have to do this based on if we know the claim id or not.  This should also work for channels, CQ uses the same field for thumbnail url (source url for banner if we wanted that). 

Here's an example by claim id: 
https://chainquery.lbry.io/api/sql?query=SELECT%20*%20FROM%20claim%20where%20claim_id=%2294a844e1f9e7dab814378de404e35ec69b57435b%22

just claim name: 
https://chainquery.lbry.io/api/sql?query=SELECT%20*%20FROM%20claim%20where%20name=%22@theGavin2000%22&bid_state=%22controlling%22

src/platforms/web/server.js
```diff
@@ -1,12 +1,61 @@
...
+      if (!claim.success || !claim.data[0]) {
+        res.sendFile(path.join(__dirname, '/index.html'));
+      }
+
+      claim = claim.data[0];
+
+      // const claimAuthor = claim.channel_name ? claim.channel_name : 'Anonymous';
+      const claimDescription = claim.description && claim.description.length > 0 ? claim.description : 'Visit LBRY.tv for more content!';
+      const claimLanguage = claim.language || 'en_US';
+      const claimThumnail = claim.thumbnail_url;
```
tzarebczan: thumbnail misspelled :)

src/platforms/web/server.js
```diff
@@ -1,12 +1,61 @@
...
 
-app.get('*', function(req, res) {
-  res.sendFile(path.join(__dirname, '/index.html'));
+app.get('*', async(req, res) => {
+  if (req.url.includes('/$/')) {
+    res.sendFile(path.join(__dirname, '/index.html'));
+  }
+
+  const claimId = req.url.split('/').pop();
+  const queryUrl = `https://chainquery.lbry.io/api/sql?query=SELECT * FROM claim where claim_id="${claimId}"`;
```
tzarebczan: Query with a channel: SELECT channel.name as channel, claim.name, claim.description, claim.language, claim.thumbnail_url, claim.title FROM claim left join claim channel on claim.publisher_id = channel.claim_id where claim.claim_id="93a61971c473ade02300361dc077eb564b558a8d"

Will we always have the claim id? does that come from the claim that's resolved? What if someone shares beta.lbry.tv/jiggytom Sounds like we'll need to use claim.name = xxx and bid_state = controlling per my original suggestion in that case.

tzarebczan: I saw we grab claim id from the URL, so we'll need to adjust the query as such.

src/platforms/web/server.js
```diff
@@ -1,12 +1,74 @@
...
+      finalTags += `<meta property="og:image" content="${claimThumbnail}"/>`;
+      finalTags += `<meta property="og:locale" content="${claimLanguage}"/>`;
+      finalTags += `<meta property="og:site_name" content="LBRY.tv"/>`;
+      finalTags += `<meta property="og:type" content="website"/>`;
+      finalTags += `<meta property="og:url" content="${claimUrl}"/>`;
+      finalTags += `<title>${claimTitle}</title>`;
+
+      try {
+        writeFileSync(path.join(__dirname, '/index.share.html'), indexHtml.replace(indexHtml.match(headRegex)[0], finalTags), 'utf8');
+        return path.join(__dirname, '/index.share.html');
```
kauffj: Is this code writing the HTML to disk only to read it back? If I'm understanding what this does correctly, it seems like it could be skipped entirely.

(Additionally, this approach could cause a race condition among two requests that came in simultaneously)

src/platforms/web/server.js
```diff
@@ -1,12 +1,74 @@
...
+      const claimDescription = claim.description && claim.description.length > 0 ? claim.description : 'Visit LBRY.tv for more content!';
+      const claimLanguage = claim.language || 'en_US';
+      const claimThumbnail = claim.thumbnail_url || '/og.png';
+      const claimTitle = `${claim.title} on LBRY.tv` || 'LBRY.tv'; // `${claim.title} from ${claimAuthor} on LBRY.tv`;
+      const claimUrl = `https://beta.lbry.tv/${claim.name}${claim.claim_id ? (`/${claim.claim_id}`) : ''}` || 'https://beta.lbry.tv';
+      // This comment was inlined and preventing the rest of the script to load
+      const indexHtml = readFileSync(path.join(__dirname, '/index.html'), 'utf8').replace(/\s\s/gm, '').replace(/>\s</gm, '><').replace('// Use relative path if we are in electron', '');
+      let finalTags = '';
+
+      finalTags += '<meta charset="utf8"/>';
```
kauffj: Can these values be injected into the HTML file rather than built in JS?

src/platforms/web/server.js
```diff
@@ -1,12 +1,74 @@
...
+    .then(response => response.json())
+    .then(claim => {
+      if (!claim.success || !claim.data[0]) {
+        return path.join(__dirname, '/index.html');
+      }
+
+      claim = claim.data[0];
+
+      // const claimAuthor = claim.channel_name ? claim.channel_name : 'Anonymous';
+      const claimDescription = claim.description && claim.description.length > 0 ? claim.description : 'Visit LBRY.tv for more content!';
```
kauffj: How about a fallback of "Watch %content_title% on lbry.tv" or "Watch %content_title% from %creator% on lbry.tv"?

src/platforms/web/server.js
```diff
@@ -1,12 +1,74 @@
...
+  let isChannel = req.url.includes('/@');
+  let queryUrl = '';
+
+  switch (true) {
+    case isAppPage:
+      res.sendFile(path.join(__dirname, '/index.html'));
+      break;
+
+    case isChannel:
+      queryUrl = `https://chainquery.lbry.io/api/sql?query=SELECT channel.name as channel, claim.name, claim.description, claim.language, claim.thumbnail_url, claim.title FROM claim left join claim channel on claim.publisher_id = channel.claim_id where claim.claim_id="${identifier}"`;
```
kauffj: @sayplastic (or @nikooo777?) is there a better way for @NetOperatorWibby to be getting data from chainquery for lbry.tv than this?

src/platforms/web/server.js
```diff
@@ -1,12 +1,74 @@
...
 
 app.use(express.static(__dirname));
 
-app.get('*', function(req, res) {
-  res.sendFile(path.join(__dirname, '/index.html'));
+app.get('*', async(req, res) => {
+  let finalHtml = '';
+  let identifier = req.url.split('/').pop();
+  let isAppPage = req.url.includes('/$/');
+  let isChannel = req.url.includes('/@');
```
kauffj: This test probably needs to be stronger, as I think `@` can technically appear in the middle of a claim name

tzarebczan: @ is a reversed character.

src/platforms/web/server.js
nikooo777: if this is serverside we can use a mysql connector, otherwise no, unless we proxy the calls.
Also this risks getting throttled by chainquery, I think there is a limit to how many queries can be sent to the web API

NetOperatorWibby: How do I improve this?

nikooo777: as first version we agreed that this is fine. however please update from .io to .com

tzarebczan:
@NetOperatorWibby / @seanyesmunt can this be merged?

seanyesmunt:
@kauffj wanted to test or make changes to this once lbry.tv was running the new app

tzarebczan:
Cool, I can try pushing this up on beta / dev later today.

PR #2553, opened 2019-06-17
Moar discovery

seanyesmunt:
Building off my last PR

#### Changes
- Added follow/unfollow on the tags page
- Added menu buttons in the header to navigate to different pages
- Reworked the "Account" page to be the lbry inc hub


src/ui/component/fileList/view.jsx
```diff
@@ -11,19 +12,31 @@ const SORT_OLD = 'old';
...
+  meta?: Node,
   // If using the default header, this is a unique ID needed to persist the state of the filter setting
   persistedStorageKey?: string,
 };
 
 export default function FileList(props: Props) {
-  const { uris, header, headerAltControls, injectedItem, loading, persistedStorageKey, noHeader, slim, empty } = props;
+  const {
+    uris,
+    header,
```
kauffj: `header` and `noHeader` could be a single param, unless it's considered a bad practice to say a type can be a `Node` or  `boolean` (true = default, false = none, node = inject)

src/ui/component/fileListItem/view.jsx
```diff
@@ -110,7 +110,7 @@ function FileListItem(props: Props) {
           {!slim && (
             <div>
               {isChannel && <SubscribeButton uri={uri.startsWith('lbry://') ? uri : `lbry://${uri}`} />}
-              <FileProperties uri={uri} />
+              {!isChannel && <FileProperties uri={uri} />}
```
kauffj: channels aren't files

if this component is now rendering streams and channels, I'd rename it `ClaimListItem` or similar

src/ui/component/fileListItem/view.jsx
```diff
@@ -122,7 +122,7 @@ function FileListItem(props: Props) {
             <div>{isChannel ? `${claimsInChannel} ${__('publishes')}` : <DateTime timeAgo uri={uri} />}</div>
           </div>
 
-          {!slim && <FileTags uri={uri} />}
+          <FileTags uri={uri} slim={slim} />
```
kauffj: Same for this, if this is being used on channels and files, call it `ClaimTags`

src/ui/component/fileTags/view.jsx
```diff
@@ -1,19 +1,25 @@
 // @flow
 import * as React from 'react';
+import classnames from 'classnames';
 import Button from 'component/button';
 
-const MAX_TAGS = 4;
+const SLIM_TAGS = 2;
+const NORMAL_TAGS = 4;
+const LARGE_TAGS = 10;
```
kauffj: good pattern!

src/ui/component/fileTags/view.jsx
```diff
@@ -1,19 +1,25 @@
...
 
 type Props = {
   tags: Array<string>,
   followedTags: Array<Tag>,
+  large: boolean,
+  slim: boolean,
```
kauffj: slightly worse pattern, because passing true for both large and slim should not be possible and would have unexpected behavior

src/ui/component/rewardSummary/view.jsx
```diff
@@ -2,59 +2,45 @@
...
-            <Button button="link" label={__('Learn more')} href="https://lbry.com/faq/rewards" />.
-          </p>
+          <h2 className="card__title">{__('Rewards')}</h2>
         </header>
 
+        <p className="card__subtitle">
+          {fetching && __('You have...')}
+          {!fetching && hasRewards ? (
+            <React.Fragment>
+              {__('You have')}
```
kauffj: While I know we've used this approach elsewhere, it's a bad practice and doesn't really work for i18n.

Is there any ability to inject nodes inside of an i18n string? e.g. `__('You have %amount% LBC', { amount: <span>5</span> })`

If not, let's use a comment tag so we can track where these exist. Something like `@i18nfixme`.

src/ui/component/rewardTotal/view.jsx
```diff
@@ -0,0 +1,25 @@
...
+function RewardTotal(props: Props) {
+  const { rewards } = props;
+  const rewardTotal = rewards.reduce((acc, val) => acc + val.reward_amount, 0);
+  const total = useTween(rewardTotal * 25);
+  const integer = Math.round(total * rewardTotal);
+
+  return (
+    <section className="card  card--section card--reward-total" style={{ backgroundImage: `url(${TotalBackground})` }}>
+      <span className="card__title">
+        {integer} LBC {__('Earned From Rewards')}
```
kauffj: https://github.com/lbryio/lbry/issues/2245

src/ui/component/userEmail/view.jsx
```diff
@@ -0,0 +1,60 @@
...
+import UserEmailNew from 'component/userEmailNew';
+import UserEmailVerify from 'component/userEmailVerify';
+
+type Props = {
+  cancelButton: React.Node,
+  email: string,
+  resendVerificationEmail: string => void,
+  checkEmailVerified: () => void,
+  user: {
+    has_verified_email: boolean,
```
kauffj: Should user be a first class type and/or should this value simply be selected directly off of the user?

src/ui/component/header/view.jsx
```diff
@@ -53,54 +70,55 @@ const Header = (props: Props) => {
...
+                {__('Wallet')}
+              </MenuItem>
+              <MenuItem className="menu__link" onSelect={() => history.push(`/$/publish`)}>
+                <Icon aria-hidden icon={ICONS.UPLOAD} />
+                {__('Publish')}
+              </MenuItem>
+            </MenuList>
+          </Menu>
+          <Menu>
+            <MenuButton className="header__navigation-item menu__title">
```
kauffj: I happened to have my app in Polish while running this branch and noticed this menu looks a little weird when the text of the items is longer than they are in English.

kauffj: I also kind of dislike this from a UX perspective. I'd rather just put light/dark in settings, give Help first-class status, and lose the dropdown entirely if these are the only 3 items in it.

src/ui/page/auth/view.jsx
```diff
@@ -40,42 +36,9 @@ class AuthPage extends React.PureComponent<Props> {
...
-              <Button button="link" navigate="/" label={__('Return home.')} />
-            </p>
-          </section>
-        ) : (
-          innerContent
-        )}
-      </Page>
-    );
+    const { user, email } = this.props;
+    return <Page>{user && email && !user.is_identity_verified ? <UserVerify /> : <UserEmail />}</Page>;
```
kauffj: looks like a nice refactor!

src/ui/page/file/view.jsx
```diff
@@ -305,10 +281,39 @@ class FilePage extends React.Component<Props> {
...
+                });
+              }}
+            />
+            <div className="file-properties">
+              {isRewardContent && (
+                <Icon
+                  size={20}
+                  iconColor="red"
+                  icon={icons.FEATURED}
+                  // Figure out how to get the tooltip to overlap the navbar on the file page and I will love you
```
kauffj: I'm pretty sure I can explain/fix this, but easier to do so synchronously. Please bring up on next standup we're both on.

src/ui/page/invite/view.jsx
```diff
@@ -27,7 +26,7 @@ class InvitePage extends React.PureComponent<Props> {
     const { isPending, isFailed } = this.props;
 
     return (
-      <Page>
+      <div className="">
```
kauffj: empty class

src/ui/page/subscriptions/view.jsx
```diff
@@ -1,53 +1,78 @@
...
+
+    history.push(url);
+  }
 
   useEffect(() => {
     doFetchMySubscriptions();
     doFetchRecommendedSubscriptions();
   }, [doFetchMySubscriptions, doFetchRecommendedSubscriptions]);
 
+  const idString = subscribedChannels.map(channel => channel.uri.split('#')[1]).join(',');
```
kauffj: is the channel uri guaranteed to have a `#`? they will never be URIs without a claim id?

src/ui/scss/component/_file-list.scss
```diff
@@ -14,6 +15,20 @@
   fieldset-section {
     margin-bottom: 0;
   }
+
+  // Normal link buttons are too dark on the black file list background
+  .button--link {
```
kauffj: minor tsk tsk for BEM violation

src/ui/scss/component/_file-properties.scss
```diff
@@ -3,11 +3,21 @@
   position: relative;
   align-items: center;
 
-  & > *:not(:first-child) {
-    margin-left: var(--spacing-small);
+  & > *:not(:last-child) {
+    margin-right: var(--spacing-small);
```
kauffj: minor boos

src/ui/util/use-tween.js
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react';
```
kauffj: I approve of copy and pasting this wherever you copy and pasted it from

src/ui/page/subscriptions/view.jsx
seanyesmunt: The channel id is required (for now?) in internal-apis for subscriptions. The app always uses the full uri for the subscribeButton component

src/ui/util/use-tween.js
seanyesmunt: 📈

src/ui/component/userEmail/view.jsx
seanyesmunt: Yes to both of these

src/ui/component/fileListItem/view.jsx
seanyesmunt: Nice good call. Will rename all of these to `claimX`

src/ui/component/fileList/view.jsx
seanyesmunt: No I think that makes more sense. Will update.

src/ui/component/fileTags/view.jsx
seanyesmunt: Changed to `type={"small", "large"}`

src/ui/component/header/view.jsx
seanyesmunt: I'm fine with moving help out of the menu, but I would really like to keep the dark mode toggle. YouTube, Twitter, and Reddit all have a dark mode toggle in their drop-down menu.

src/ui/component/rewardSummary/view.jsx
seanyesmunt: Not with `y18n`. I would imagine this exists somewhere though

PR #2576, opened 2019-06-26
Discovery fixes

seanyesmunt:
### Changes
- Add tags to publish page
  - Got tired of working with the `PublishForm` component so I broke it into a bunch of small components
- Moved to reach/tooltip for better tooltips (and added one for channels)
  - tooltip behavior is hard
- General fixes

### Notes
- Depends on https://github.com/lbryio/components/pull/11


src/ui/component/claimList/view.jsx
```diff
@@ -36,10 +36,10 @@ export default function ClaimList(props: Props) {
   return (
     <section className={classnames('file-list')}>
       {header !== false && (
-        <div className={classnames('file-list__header', { 'file-list__header--small': type === 'small' })}>
+        <div className={classnames('claim-list__header', { 'claim-list__header--small': type === 'small' })}>
```
kauffj: good refactor

src/ui/component/publishFile/view.jsx
```diff
@@ -0,0 +1,54 @@
...
+      <header className="card__header">
+        <h2 className="card__title card__title--flex-between">{isStillEditing ? __('Edit') : __('Publish')}</h2>
+        {isStillEditing && <p className="card__subtitle">{__('You are currently editing a claim.')}</p>}
+      </header>
+
+      <div className="card__content">
+        <FileSelector currentPath={filePath} onFileChosen={handleFileChange} />
+        {!!isStillEditing && name && (
+          <p className="help">
+            {__("If you don't choose a file, the file from your existing claim")}
```
kauffj: When you notice split i18n strings like this, can you add an annotation of some kind? (I propose `//@i18nfixme`)

src/ui/component/publishFormErrors/view.jsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <div className="card__content error-text">
+      {!title && <div>{__('A title is required')}</div>}
+      {!name && <div>{__('A URL is required')}</div>}
+      {!bid && <div>{__('A deposit amount is required')}</div>}
+      {uploadThumbnailStatus === THUMBNAIL_STATUSES.IN_PROGRESS && (
+        <div>{__('Please wait for thumbnail to finish uploading')}</div>
+      )}
+      {!!editingURI && !isStillEditing && !filePath && (
+        <div>{__('You need to reselect a file after changing the LBRY URL')}</div>
```
kauffj: Can this be re-written to say "Please"?

(also, why is this necessary?)

src/ui/component/publishName/bid-help-text.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+  let bidHelpText;
+
+  if (uri) {
+    if (isResolvingUri) {
+      bidHelpText = __('Checking the winning claim amount...');
+    } else if (!amountNeededForTakeover) {
+      bidHelpText = __('Any amount will give you the winning bid.');
+    } else {
+      bidHelpText = `${__('If you bid more than')} ${amountNeededForTakeover} LBC, ${__(
```
kauffj: `@i18nfixme`

(also, these split strings are basically not worth even calling the function on, but should be tagged either way)

src/ui/component/publishName/view.jsx
```diff
@@ -0,0 +1,124 @@
...
+      prepareEdit(myClaimForUri, uri);
+    }
+  }
+
+  useEffect(() => {
+    let nameError;
+    if (!name) {
+      nameError = __('A name is required');
+    } else if (!isNameValid(name, false)) {
+      nameError = __('LBRY names cannot contain that symbol ($, #, @)');
```
kauffj: Some symbols are missing here. It's probably worth adding a `RESERVED_CHARACTER` const to `lbryUri.js`.

src/ui/component/selectThumbnail/view.jsx
```diff
@@ -160,6 +161,18 @@ class SelectThumbnail extends React.PureComponent<Props, State> {
         )}
 
         {status === THUMBNAIL_STATUSES.IN_PROGRESS && <p>{__('Uploading thumbnail')}...</p>}
+        <p className="help">
+          {(status === undefined && __('You should reselect your file to choose a thumbnail')) ||
+            (status === THUMBNAIL_STATUSES.API_DOWN ? (
+              __('Enter a URL for your thumbnail.')
+            ) : (
+              <React.Fragment>
+                {__('Upload your thumbnail (.png/.jpg/.jpeg/.gif) to')}{' '}
```
kauffj: I do not think the file types need to be mentioned here. Not a strong opinion, though.

src/ui/component/publishFormErrors/view.jsx
seanyesmunt: It shouldn't be, but we aren't currently tracking the previous claim, after you change the name, because that could be another claim.

Issue #561, opened 2017-09-13
Support publishing channel descriptions, cover photos and other metadata


tzarebczan:
## The Issue
The current channel page looks a bit empty with the "Channel pages are empty for all publishers currently, but will be coming in a future update." message.  Can we add this issue to the Publishing Epic to at least add some descriptions to channels? 

In the future, other channel data could include: 
Channel Image
Stats (creation date, total subscribers, total videos, total views, etc)
Tip Button (channel side vs content side)
Contact information
Linked channels?

```

        Create a new channel by generating a channel private key and establishing an '@' prefixed claim.

        Usage:
            channel_create (<name> | --name=<name>) (<bid> | --bid=<bid>)
                           [--allow_duplicate_name=<allow_duplicate_name>]
                           [--title=<title>] [--description=<description>] [--email=<email>] [--featured=<featured>...]
                           [--tags=<tags>...] [--languages=<languages>...] [--locations=<locations>...]
                           [--website_url=<website_url>] [--thumbnail_url=<thumbnail_url>] [--cover_url=<cover_url>]
                           [--account_id=<account_id>] [--claim_address=<claim_address>] [--preview]

        Options:
            --name=<name>                  : (str) name of the channel prefixed with '@'
        --allow_duplicate_name=<allow_duplicate_name> : (bool) create new channel even if one already exists with
                                              given name. default: false.
            --bid=<bid>                    : (decimal) amount to back the claim
            --title=<title>                : (str) title of the publication
            --description=<description>    : (str) description of the publication
            --featured=<featured>          : (list) claim_ids of featured content in channel
            --tags=<tags>                  : (list) content tags
            --languages=<languages>        : (list) languages used by the channel,
                                                    using RFC 5646 format, eg:
                                                    for English `--languages=en`
                                                    for Spanish (Spain) `--languages=es-ES`
                                                    for Spanish (Mexican) `--languages=es-MX`
                                                    for Chinese (Simplified) `--languages=zh-Hans`
                                                    for Chinese (Traditional) `--languages=zh-Hant`
            --locations=<locations>        : (list) locations of the channel, consisting of 2 letter
                                                    `country` code and a `state`, `city` and a postal
                                                    `code` along with a `latitude` and `longitude`.
                                                    for JSON RPC: pass a dictionary with aforementioned
                                                        attributes as keys, eg:
                                                        ...
                                                        "locations": [{'country': 'US', 'state': 'NH'}]
                                                        ...
[...]

PR #2584, opened 2019-06-28
Channel edit4

jessopb:
Feature: 
Channel page updates
Re: #561 

Should be broken out of the channel page - probably requires useContext wrapper.
Not happy with upload function living in component.
Still a bug with updating and status messaging on submit.
Mostly unstyled.


src/ui/component/selectAsset/view.jsx
```diff
@@ -0,0 +1,150 @@
...
+                  }}
+                >
+                  Clear
+                </Button>
+              </div>
+            )}
+          </>
+        )}
+        {assetSource === SOURCE_URL && (
+          <>
```
seanyesmunt: Not needed

src/ui/page/channel/view.jsx
```diff
@@ -1,36 +1,87 @@
...
+  website: string,
+  email: string,
+  balance: number,
+  updateChannel: any => void,
+  tags: Array<string>,
+  locations: Array<string>,
+  languages: Array<string>,
 };
 
 function ChannelPage(props: Props) {
```
seanyesmunt: For now lets just create a `channelEdit` component and put all of the edit stuff in there.

src/ui/page/channel/view.jsx
```diff
@@ -48,44 +99,179 @@ function ChannelPage(props: Props) {
...
+          </Tabs>
+        )}
+        {editing && (
+          <div className={'card--section'}>
+            <section>
+              <div className={'card__title--flex-between'}>
+                <header className="card__header">
+                  <h2 className="card__title">{__('Edit')}</h2>
+                </header>
+                <div>
```
seanyesmunt: `<div className="card__actions">`

src/ui/component/selectAsset/view.jsx
```diff
@@ -0,0 +1,150 @@
...
+    } else {
+      return null;
+    }
+
+    const makeid = () => {
+      let text = '';
+      const possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+      for (let i = 0; i < 24; i += 1) text += possible.charAt(Math.floor(Math.random() * 62));
+      return text;
+    };
```
seanyesmunt: We already use the `uuid` module, lets just use that here.

src/ui/page/channel/view.jsx
```diff
@@ -48,44 +99,179 @@ function ChannelPage(props: Props) {
...
+    if (bid <= 0.0 || isNaN(bid)) {
+      setBidError(__('Deposit cannot be 0'));
+    } else if (totalAvailableBidAmount === bid) {
+      setBidError(__('Please decrease your deposit to account for transaction fees'));
+    } else if (totalAvailableBidAmount < bid) {
+      setBidError(__('Deposit cannot be higher than your balance'));
+    } else if (bid <= MINIMUM_PUBLISH_BID) {
+      setBidError(__('Your deposit must be higher'));
+    }
+  };
```
seanyesmunt: We use this same logic in the publish form. Lets pull it into it's own function.

It can accept `bid, balance, amount` and return an error if it's bad. If there is no error, update the form, if there is an error, setError

```
const error = checkBidStuff(amount, balance, bid)
if (error) { 
  setBidError(error)
} else {
  setParams()
}

src/ui/component/selectAsset/view.jsx
jessopb: flow doesn't like import {v4 as uuidv4} from uuid;

src/ui/page/channel/view.jsx
jessopb: DIdn't do this yet. 
I'm thinking checkBidStuff(amount, balance, bid, errorFn, updateFn) would be the signature? Or do I misunderstand you?

jessopb:
This has been updated.

PR #2593, opened 2019-07-01
Add tags to publish page, infinite scroll, navigation improvements 

seanyesmunt:
Probably should have opened this sooner, but this is general fixes/infinite scroll v0.1, tags on publish page.


src/ui/component/claimList/view.jsx
```diff
@@ -33,33 +48,63 @@ export default function ClaimList(props: Props) {
...
 
+  const urisLength = uris && uris.length;
+  useEffect(() => {
+    function handleScroll(e) {
+      if (onScrollBottom) {
+        const x = document.querySelector(`.${MAIN_WRAPPER_CLASS}`);
+
+        if (x && window.scrollY + window.innerHeight >= x.offsetHeight) {
+          // fix this
+          if (!loading && urisLength > 19) {
```
seanyesmunt: @kauffj does this seem like a reasonable way to do this or is there something simpler?

It is to check if a full page was loaded, (if 15 items are loaded) don't fetch more, because page 2 wont' exist. If that seems fine I can create a prop called `claimListSearchCount` or something and add a constant that we use for the `claim_search` params

kauffj: It could potentially make sense to move this out of here and instead have the scroll state be something that is generally consumable by any component (i.e. track the scroll state app-wide). 

I'm not up to date on the best techniques for this though.

PR #2599, opened 2019-07-02
~Final~ Discovery fixes/cleanup

seanyesmunt:


src/ui/component/claimPreview/view.jsx
```diff
@@ -59,7 +59,7 @@ function ClaimPreview(props: Props) {
   let shouldHide = abandoned || (!claimIsMine && obscureNsfw && nsfw);
 
   // This will be replaced once blocking is done at the wallet server level
-  if (claim && !shouldHide) {
+  if (claim && !shouldHide && blackListedOutpoints) {
```
kauffj: Below for loop can be functional. I admit that I may just enjoy rewriting these more than it is actually useful:

```
shouldHide = blackListedOutpoints.reduce((hide, outpoint) => 
  return hide || (outpoint.txid === claim.txid && outpoint.nout === claim.nout);
)
```

src/ui/util/use-hover.js
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react';
```
kauffj: No need to do this now, but I suspect these should end up with their own folder or another organizational method rather than being dumped in `util`.

src/ui/component/claimPreview/view.jsx
seanyesmunt: I was going to use this, but then I realized we should add a `break` statement to the for loop so we don't loop over the list when we don't need to. Now it will stop if it finds a matching item.

PR #2607, opened 2019-07-06
[merge after 34 release] Lbry redux publish WIP

jessopb:
Old behavior: 
app uses lbry-desktop redux store for publish.
New behavior:
all apps can use lbry-redux store for publish.

Still requires:
fix notifications
extensive rebase
delete lbry-desktop publish redux (after rebase)


src/ui/component/publishForm/view.jsx
```diff
@@ -87,6 +92,45 @@ function PublishForm(props: Props) {
     submitLabel = !publishing ? __('Publish') : __('Publishing...');
   }
 
+  const success = successResponse => {
```
seanyesmunt: Instead of bringing this into the component, we can still use a `doPublish` function that lives inside of the app. All it will do is call the publish function from lbry-redux, and pass in this success callback.

src/ui/component/publishForm/view.jsx
```diff
@@ -160,7 +204,7 @@ function PublishForm(props: Props) {
           <div className="card__actions">
             <Button
               button="primary"
-              onClick={publish}
+              onClick={() => publish(success, fail)}
```
seanyesmunt: this should stay as `publish`

src/ui/component/publishForm/index.js
```diff
@@ -34,7 +33,7 @@ const perform = dispatch => ({
   updatePublishForm: value => dispatch(doUpdatePublishForm(value)),
   clearPublish: () => dispatch(doClearPublish()),
   resolveUri: uri => dispatch(doResolveUri(uri)),
-  publish: params => dispatch(doPublish(params)),
+  publish: (success, fail) => dispatch(doPublish(success, fail)),
```
seanyesmunt: We don't need to pass in the callbacks here, lets do it in redux (in the desktop codebase).

src/ui/component/publishName/index.js
```diff
@@ -5,17 +5,18 @@ import {
...
+} from 'lbry-redux';
+
 import PublishPage from './view';
 
 const select = state => ({
   name: makeSelectPublishFormValue('name')(state),
   channel: makeSelectPublishFormValue('channel')(state),
   bid: makeSelectPublishFormValue('bid')(state),
   uri: makeSelectPublishFormValue('uri')(state),
-  bid: makeSelectPublishFormValue('bid')(state),
```
seanyesmunt: I don't think this should be removed?

jessopb: it's a duplicate

tzarebczan:
Reminder: remove unused reducer code

PR #2641, opened 2019-07-18
adds tags to channelUpdate

jessopb:
depends on https://github.com/lbryio/lbry-redux/pull/168

### DIFF
--- a/src/ui/component/channelEdit/view.jsx
+++ b/src/ui/component/channelEdit/view.jsx
@@ -4,12 +4,11 @@ import { parseURI } from 'lbry-redux';
 import { Form, FormField } from 'component/common/form';
 import Button from 'component/button';
 
-import SelectAsset from '../selectAsset/view';
-import TagSelect from '../tagsSelect/view';
+import SelectAsset from 'component/selectAsset';
+import TagSelect from 'component/tagsSelect';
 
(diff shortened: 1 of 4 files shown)


src/ui/component/channelEdit/view.jsx
```diff
@@ -177,6 +182,25 @@ function ChannelForm(props: Props) {
             disabled={false}
             onChange={text => setParams({ ...params, description: text })}
           />
+          <TagSelect
+            title={false}
+            suggestMature
+            help={__('The better your tags are, the easier it will be for people to discover your content.')}
```
seanyesmunt: `... to discover your channel`

src/ui/component/channelEdit/view.jsx
```diff
@@ -5,6 +5,7 @@ import { Form, FormField } from 'component/common/form';
 import Button from 'component/button';
 
 import SelectAsset from '../selectAsset/view';
+import TagSelect from '../tagsSelect/view';
```
seanyesmunt: These should be `component/selectAsset`/`component/tagsSelect`

@jessopb I think this might be why you were running into those flow issues, because you weren't importing the `index.js` file so no props will be injected.

PR #2649, opened 2019-07-22
improve style consistency and discovery cleanup

seanyesmunt:
#### Changes
- Improved font-size consistency
- Removed `card__content` and `card__header` styles
- Tried to improve consitency with spacing/titles/font anywhere I could
  - I didn't rename/move many classNames in this PR, mostly just made it so components used the same classNames, but I'm thinking the entire `card.scss` will be deleted eventually
- Ended up fixing up a lot of RC bugs in this PR too 😞. Main changes:
  - New startup process with `redux-persist` v5
    - This makes it so the app won't start until the persisted data is loaded (ex: tags you're following). Before the app was calling `claim_search` twice with the default tags, then the persisted tags
  - Tom fixed an issue with supports


seanyesmunt:
@kauffj There are probably a few questionable lines in this (re/ discovery components) but it's been tested pretty good. I'm going to merge this just so it isn't holding up other PR's causing issues on master.

I will definitely need to open new PR's to continue to fix discovery bugs so the questionable code will bubble up soon (infinite scroll is still a little wonky in a few cases)

PR #2657, opened 2019-07-24
Fix language setting

tzarebczan:
Seems to work locally when building.

### DIFF
--- a/src/ui/i18n/index.js
+++ b/src/ui/i18n/index.js
@@ -3,7 +3,7 @@ import y18n from 'y18n';
 import path from 'path';
 
 const i18n = y18n({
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Tested and working. Nice fix!

kauffj:
@seanyesmunt do you understand why `__static` didn't work here? Isn't `__static` designed to handle this? @tzarebczan said this broke it in dev mode.

kauffj:
"Read the code, Jeremy" - @seanyesmunt

seanyesmunt:
I am curious to what code you read to understand this. I'm still not sure why it wasn't working with `__static`

PR #2667, opened 2019-07-24
comments and expandable changes

jessopb:
comments are only expandable if larger than 120 rect.height
expandable checks the useRect ref height before inserting itself

Originally I implemented the useRect in the Comment component but it seems to work contained in Expandable.

For now, I commented out Expandable border-bottom style - I don't think expandable should be in charge of that sort of thing.

### DIFF
--- a/src/ui/component/expandable/view.jsx
+++ b/src/ui/component/expandable/view.jsx
@@ -4,9 +4,7 @@ import classnames from 'classnames';
 import Button from 'component/button';
 import { useRect } from '@reach/rect';
 
-// Note:
-// When we use this in other parts of the app, we will probably need to
-// add props for collapsed height
+const COLLAPSED_HEIGHT = 120;
 
 type Props = {
   children: React$Node | Array<React$Node>,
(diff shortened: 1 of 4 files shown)


src/ui/scss/component/_expandable.scss
```diff
@@ -1,5 +1,5 @@
 .expandable {
-  border-bottom: 1px solid $lbry-gray-1;
+  //border-bottom: 1px solid $lbry-gray-1;
```
jessopb: This will be removed before merge if we agree it doesn't belong.

src/ui/component/expandable/view.jsx
```diff
@@ -11,48 +12,38 @@ type Props = {
...
-        <div
-          className={classnames({
-            'expandable--open': expanded,
-            'expandable--closed': !expanded,
-          })}
-        >
-          {children}
+  return (
+    <div ref={ref}>
+      {rect && rect.height > 120 ? (
```
jessopb: This 120 should be a constant, I think. I'm not sure how I'd use that constant for the styles too, though. Base size is 12px and expandable min height is 10rem (=120).

seanyesmunt: I just moved it up to the top. I think as long as it's reasonable it's fine. In the browser (and possibly desktop sometime) users can scale the font-size so anything exact falls apart. 120 seems fine.

Issue #1394, opened 2018-04-26
Add detailed view to Downloads/Publishes with additional card/batch options


tzarebczan:
## The Issue
Discussed this with Sean briefly and it was also brought up by a reddit user. 

Problem: Currently, to take any actions (edit/delete/tip) you need to go into each claim to perform them. 

The enhancement would be to have a more detailed, list like, view for both Downloads/Publishes that would condense the claim information so they can be more easily managed. This would also include direct actions to delete/tip/edit/report/open externally as appropriate. 

Batch actions (by selecting multiple files/claims, select all option) can be added to allow deleting multiple items with a few clicks. 

We would also display detailed claim information like the URL, claim bid, total supports, total tips, if it's the vanity URL, etc. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

tzarebczan:
With the new preview, we have room to add additional quick options like download/delete/tip. Users don't want to go into each claim separately in order to delete something (heard this complaint a few times recently). 

We should also consider adding a multi-select feature like the Android app did, but we can break that out into a new issue.

walidmujahid:
Specifically on the action of "deletion" for viewers, it would be nice to have some kind of button on each video on the Library page. It could be an trash bin icon would be explanatory when viewed on the Library page, but I personally prefer buttons that have actual words like a red button that says "remove".

I am not much of a UX designer, so the decision is up the design team...or whoever makes those decisions here.

PR #2676, opened 2019-07-26
Adds icon to unsubscribe from side bar

### LINKED ISSUE #1396: Add "unsubscribe" button in the subscriptions overview
## The Issue
Currently, if a user wants to unsubscribe, they have to click on a creator name, and then unsubscribe. To make it easier, there could be a unsubscribe button placed in the subscription overview. It would make subscription management process easier.

abueide:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes
#1396
I know this is closed, but I think its a great stop gap solution until #1394 is finished. It allows users to quickly prune their subscriptions. It should also allow you to unsubscribe from deleted channels (which I think @eggplantbren mentioned was an issue). It might even be better than #1394, because it keeps things simple for the end user, and I'm not sure how useful editing and tipping without entering the claim would be compared to the complexity cost to the user.

No unsubscribe button next to the list of subscribers.
## What is the new behavior?
Unsubscribe button next to the list of subscribers. It looks like this 
image

### DIFF
--- a/src/ui/component/sideBar/view.jsx
+++ b/src/ui/component/sideBar/view.jsx
@@ -1,4 +1,5 @@
 // @flow
+import SubscribeButton from 'component/subscribeButton';
 import * as PAGES from 'constants/pages';
 import * as ICONS from 'constants/icons';
 import React from 'react';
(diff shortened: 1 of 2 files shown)


tzarebczan:
hey @abueide , thanks for the PR! We'll get a review on this, but I'm not sure if we need quick unsubscribe actions from the side bar - how often would that actually get used? It also adds lots of repeating elements to the UI. I'll let other team members review.

seanyesmunt:
Thanks for the PR @abueide!

I think this is an interesting idea, but I think it will lead to more issues than help. The main one being when someone accidentally clicks on the heart icon. In every other button, clicking on the icon leads is the same as clicking on the button label.

I'm going to close this.

PR #2698, opened 2019-07-30
Channel filter/block

jessopb:
Users can block channels and visit their blocked channels through settings.
Todo: 
 - apply blocked channels in claim search using not_channel_ids
 - block more than just channels? Claim filter/block as well?


jessopb:
closed for #2706 which is rebased and improved

PR #2706, opened 2019-08-02
Channel filter mm

jessopb:
Users can block channels and visit their blocked channels through settings.
Images on blocked channels are obscured
channel ids are filtered in discovery claim search

### DIFF
--- a/src/ui/component/blockButton/view.jsx
+++ b/src/ui/component/blockButton/view.jsx
@@ -7,6 +7,7 @@ import useHover from 'util/use-hover';
 
 type Props = {
   uri: string,
+  shortUrl: string,
   isSubscribed: boolean,
   toggleBlockChannel: (uri: string) => void,
   channelIsBlocked: boolean,
@@ -14,7 +15,7 @@ type Props = {
 };
 
 export default function BlockButton(props: Props) {
-  const { uri, toggleBlockChannel, channelIsBlocked, doToast } = props;
+  const { uri, shortUrl, toggleBlockChannel, channelIsBlocked, doToast } = props;
 
   const blockRef = useRef();
   const isHovering = useHover(blockRef);
@@ -25,13 +26,13 @@ export default function BlockButton(props: Props) {
     <Button
       ref={blockRef}
       iconColor="red"
-      icon={blockedOverride ? ICONS.UNBLOCK : ICONS.BLOCK}
+      icon={ICONS.BLOCK}
       button={'alt'}
       label={blockedOverride || blockLabel}
       onClick={e => {
         e.stopPropagation();
         if (!channelIsBlocked) {
-          doToast({ message: `Blocked ${uri}`, linkText: 'Manage', linkTarget: `/${PAGES.BLOCKED}` });
+          doToast({ message: `Blocked ${shortUrl}`, linkText: 'Manage', linkTarget: `/${PAGES.BLOCKED}` });
         }
         toggleBlockChannel(uri);
       }}
--- a/src/ui/component/blockButton/view.jsx
+++ b/src/ui/component/blockButton/view.jsx
@@ -1,0 +1,40 @@
+// @flow
+import * as ICONS from 'constants/icons';
+import * as PAGES from 'constants/pages';
+import React, { useRef } from 'react';
+import Button from 'component/button';
+import useHover from 'util/use-hover';
+
+type Props = {
+  uri: string,
+  isSubscribed: boolean,
+  toggleBlockChannel: (uri: string) => void,
+  channelIsBlocked: boolean,
+  doToast: ({ message: string, linkText: string, linkTarget: string }) => void,
+};
+
+export default function BlockButton(props: Props) {
+  const { uri, toggleBlockChannel, channelIsBlocked, doToast } = props;
+
+  const blockRef = useRef();
+  const isHovering = useHover(blockRef);
+  const blockLabel = channelIsBlocked ? __('Blocked') : __('Block');
+  const blockedOverride = channelIsBlocked && isHovering && __('Unblock');
+
+  return (
+    <Button
+      ref={blockRef}
+      iconColor="red"
+      icon={blockedOverride ? ICONS.UNBLOCK : ICONS.BLOCK}
+      button={'alt'}
+      label={blockedOverride || blockLabel}
+      onClick={e => {
+        e.stopPropagation();
+        if (!channelIsBlocked) {
+          doToast({ message: `Blocked ${uri}`, linkText: 'Manage', linkTarget: `/${PAGES.BLOCKED}` });
+        }
+        toggleBlockChannel(uri);
+      }}
+    />
+  );
+}
--- a/src/ui/component/channelContent/view.jsx
+++ b/src/ui/component/channelContent/view.jsx
@@ -36,7 +36,7 @@ function ChannelContent(props: Props) {
 
       {!fetching && channelIsBlocked && (
         <div className="card--section">
-          <h2 className="card__content help">{__('You have blocked this channel content.')}</h2>
+          <h2 className="help">{__('You have blocked this channel content.')}</h2>
         </div>
       )}
 
--- a/src/ui/component/channelContent/view.jsx
+++ b/src/ui/component/channelContent/view.jsx
@@ -12,12 +12,13 @@ type Props = {
   fetching: boolean,
   params: { page: number },
   claimsInChannel: Array<StreamClaim>,
+  channelIsBlocked: boolean,
   channelIsMine: boolean,
   fetchClaims: (string, number) => void,
 };
 
 function ChannelContent(props: Props) {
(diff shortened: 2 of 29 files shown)


src/ui/component/blockButton/view.jsx
```diff
@@ -0,0 +1,40 @@
...
+  const blockRef = useRef();
+  const isHovering = useHover(blockRef);
+  const blockLabel = channelIsBlocked ? __('Blocked') : __('Block');
+  const blockedOverride = channelIsBlocked && isHovering && __('Unblock');
+
+  return (
+    <Button
+      ref={blockRef}
+      iconColor="red"
+      icon={blockedOverride ? ICONS.UNBLOCK : ICONS.BLOCK}
```
seanyesmunt: I think this should just use the `BLOCK` icon. The `UNBLOCK` icon is not very intuitive

src/ui/component/blockButton/view.jsx
```diff
@@ -0,0 +1,40 @@
...
+      button={'alt'}
+      label={blockedOverride || blockLabel}
+      onClick={e => {
+        e.stopPropagation();
+        if (!channelIsBlocked) {
+          doToast({ message: `Blocked ${uri}`, linkText: 'Manage', linkTarget: `/${PAGES.BLOCKED}` });
```
seanyesmunt: This should use `claim.short_url` instead of `uri`. There is a selector for it.

src/ui/component/channelContent/view.jsx
```diff
@@ -27,21 +28,30 @@ function ChannelContent(props: Props) {
...
+      {!fetching && !hasContent && !channelIsBlocked && (
         <div className="card--section">
           <h2 className="help">{__("This channel hasn't uploaded anything.")}</h2>
         </div>
       )}
 
-      {!channelIsMine && <HiddenNsfwClaims className="card__subtitle" uri={uri} />}
+      {!fetching && channelIsBlocked && (
+        <div className="card--section">
+          <h2 className="card__content help">{__('You have blocked this channel content.')}</h2>
```
seanyesmunt: This should just be `<p className="help">` `.card__content` doesn't exist anymore

src/ui/component/channelThumbnail/view.jsx
```diff
@@ -25,8 +26,8 @@ function ChannelThumbnail(props: Props) {
         [colorClassName]: !thumbnail,
       })}
     >
-      {!thumbnail && <img className="channel-thumbnail__default" src={thumbnailPreview || Gerbil} />}
-      {thumbnail && <img className="channel-thumbnail__custom" src={thumbnailPreview || thumbnail} />}
+      {(!thumbnail || obscure) && <img className="channel-thumbnail__default" src={thumbnailPreview || Gerbil} />}
+      {!obscure && thumbnail && <img className="channel-thumbnail__custom" src={thumbnailPreview || thumbnail} />}
```
seanyesmunt: For some reason, the color background behind the gerbil is lost on blocked channels. It's probably something small, but it should still use that if we are hiding the thumbnail.

src/ui/component/claimPreview/view.jsx
```diff
@@ -93,6 +102,14 @@ function ClaimPreview(props: Props) {
         (outpoint.txid === claim.txid && outpoint.nout === claim.nout)
     );
   }
+  // block stream claims
+  if (claim && !shouldHide && !showUserBlocked && blockedChannelUris.length && signingChannel) {
+    shouldHide = blockedChannelUris.some(blockedUri => blockedUri === signingChannel.permanent_url);
+  }
+  // block channel claims if we can't control for them in claim search
+  if (claim && isChannel && !shouldHide && !showUserBlocked && blockedChannelUris.length && isChannel) {
+    shouldHide = blockedChannelUris.some(blockedUri => blockedUri === claim.permanent_url);
```
seanyesmunt: Why do we need this if we use `not_channel_ids`?

seanyesmunt: Oh, I guess on the search/related content lists. Can you add a comment saying it's explicitly for that?

src/ui/component/subscribeButton/view.jsx
```diff
@@ -6,7 +6,7 @@ import { parseURI } from 'lbry-redux';
 import Button from 'component/button';
 import useHover from 'util/use-hover';
 
-type SubscribtionArgs = {
+type SubscriptionArgs = {
```
seanyesmunt: Nice

src/ui/page/channel/view.jsx
```diff
@@ -91,7 +114,8 @@ function ChannelPage(props: Props) {
             <Tab>{editing ? __('Editing Your Channel') : __('About')}</Tab>
             <div className="card__actions--inline">
               <ShareButton uri={uri} />
-              <SubscribeButton uri={permanentUrl} />
+              {!channelIsBlocked && <SubscribeButton uri={permanentUrl} />}
+              {!isSubscribed && <BlockButton uri={permanentUrl} />}
```
seanyesmunt: The subscribe button and block button should be wrapped in
```
{!claimIsMine && <Fragment> ... </Fragment>}
```

src/ui/page/listBlocked/view.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+type Props = {
+  uris: Array<string>,
+};
+
+function ListBlocked(props: Props) {
+  const { uris } = props;
+
+  return (
+    <Page notContained>
```
seanyesmunt: `notContained` isn't used anymore.

src/ui/page/listBlocked/view.jsx
```diff
@@ -0,0 +1,38 @@
...
+            persistedStorageKey="block-list-published"
+            uris={uris}
+            defaultSort
+            showHiddenByUser
+          />
+        </div>
+      ) : (
+        <div className="main--empty">
+          <section className="card card--section">
+            <header className="card__header">
```
seanyesmunt: `card__header` doesn't exist anymore

src/ui/page/listBlocked/view.jsx
```diff
@@ -0,0 +1,38 @@
...
+              <h2 className="card__title">{__('It looks like you have no blocked channels.')}</h2>
```
seanyesmunt: This is twitter's empty block messaging

```
**You aren’t blocking anyone**

When you block someone, that person won’t be able to follow or message you, and you won’t see notifications from them. Learn more
```

```
**You aren’t blocking any channels**

When you block a channel, all content from that channel will be hidden.
```

seanyesmunt: The ** is supposed to be a header.

src/ui/page/settings/view.jsx
```diff
@@ -278,6 +281,16 @@ class SettingsPage extends React.PureComponent<Props, State> {
...
             </section>
 
+            <section className="card card--section">
+              <h2 className="card__title">{__('Blocked Channels')}</h2>
+              <p className="card__subtitle card__help ">
+                {__('You have')} {userBlockedChannelsCount} {__('blocked')}{' '}
+                {userBlockedChannelsCount === 1 && __('channel')}
+                {userBlockedChannelsCount !== 1 && __('channels')}.{' '}
+                <Button button="link" label={__('Manage')} navigate={`/$/${PAGES.BLOCKED}`} />
+              </p>
```
seanyesmunt: Let's just add this to the `Content Settings` right above it.

src/ui/component/claimPreview/view.jsx
```diff
@@ -144,16 +161,29 @@ function ClaimPreview(props: Props) {
...
       <div className="claim-preview-metadata">
         <div className="claim-preview-info">
           <div className="claim-preview-title">
             {claim ? <TruncatedText text={title || claim.name} lines={1} /> : <span>{__('Nothing here')}</span>}
           </div>
           {!hideActions && (
-            <div>
-              {isChannel && <SubscribeButton uri={uri.startsWith('lbry://') ? uri : `lbry://${uri}`} />}
-              {!isChannel && <FileProperties uri={uri} />}
+            <div className={'claim-preview-actions'}>
```
seanyesmunt: Does `card__actions` not work here? I think it should without needing to add the `claim-preview__button` class around the buttons either.

Definitely not a good name, but we should keep using the same classes for buttons until we change that to be something more generic.

src/ui/scss/component/_claim-list.scss
```diff
@@ -231,6 +231,14 @@ $border-color--dark: var(--dm-color-04);
   justify-content: space-between;
 }
 
+.claim-preview-actions {
+  display: flex;
+  justify-content: space-evenly;
+}
+.claim-preview__button {
+  margin-left: 2rem;
+}
```
seanyesmunt: I don't think we need these.

Issue #1192, opened 2018-03-27
Windows: Installer does not kill daemon process


tzarebczan:
## The Issue
When on 0.21.2 and running the 0.21.2 installer while LBRY is running, the installer tries to shutdown the lbrynet-daemon.exe process. Since two processes are spawn during app launch, the installer is able to kill the first, but not the 2nd because it's a child of a child. This becomes a concern during updates when we upgrade the daemon version because the user will have to kill the current daemon after app install and then re-install the app again so the new daemon is written into the file system. 

Startup process (per Igor):
App => Daemon process 1 => Daemon process 2

### Steps to reproduce
1. Run LBRY 0.21.2
2. Launch 0.21.2 installer process 
3. Notice that the daemon continues running and app restarts using that daemon process

### Expected behaviour
Installer should kill all daemon processes

### Actual behaviour
Daemon continues running

## System Configuration

 

- LBRY Daemon version: 0.19.1
- LBRY App version: 0.21.2
- LBRY Installation ID:
- Operating system: Windows

tzarebczan:
We should look into this before the redesign release, especially if we will be updating the daemon along with it. 

I'll try this on Mac to see if it's happening there.

IGassmann:
The solution has to be implemented on the daemon. The daemon has to kill its children process if it's being exited.

tzarebczan:
We are working with the daemon team to see if they can get the daemon to run as a single process. If this is the case, we can update the predesign version to the latest daemon. 

The upgrade procedure would be:
1) Don't autoupgrade - close LBRY completely and download latest installer from lbry.io/get
2) Run installer manually

or (most will opt for this because they will see the message)
1) Run autoupdate 
2) App restarts to incompatible daemon screen which allows them to quit daemon and shows them https://lbry.io/faq/incompatible-protocol-version which walks them through making sure the daemon is killed, downloading installer and re-running

tzarebczan:
Did a little more digging - went back to 0.21.2 to perform in-app upgrades and it looks like we have a different issue. When upgrading in app, you get 2 prompts to run the installer, one after another. If you wait for a few seconds after the first to click proceed on the 2nd, then you end up in a situation where LBRY is killed but the daemon is left running. Still not sure why there are 2 UAC prompts, but one  is run with --updated --force-run parameter and then a second time with --updated /S. I'll open a separate issue for this.

[...]

PR #2710, opened 2019-08-05
maint: electron builder update

tzarebczan:
This fixes https://github.com/lbryio/lbry-desktop/issues/1192, the bug where the SDK would not get terminated if the installer was run while the app was running. Also tested app build on Mac 10.13.6 (xcode 10) which is required now for code signing (https://github.com/electron-userland/electron-builder/releases/tag/v21.0.10). Per Julie, works fine with 10.12.5 OSX, which is good. Will confirm signing is working correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ matrix:
   include:
     - os: osx
       env: TARGET=mac
-      osx_image: xcode9.2
+      osx_image: xcode10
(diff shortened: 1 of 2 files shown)


tzarebczan:
Tested and it looks good to me! Had Julie try an older Mac to make sure it still worked on 10.12.4+.

PR #2720, opened 2019-08-08
mature tag first to display

jessopb:


seanyesmunt:
This should probably include any mature tags first. Ex: If there is a`porn` tag but no `mature` tag, the `porn` tag should be added. But this is good for now.

jessopb:
It's probably pretty easy to iterate through these
import { MATURE_TAGS } from 'lbry-redux';

Issue #1397, opened 2018-04-26
Add "Tip the creator" button on channel page


LavRadis:
## The Issue

To reward creator they subscribe to directly, users could simply use "Tip the creator" button inside subscription overview. The user could send a one-time donation in LBC.

tzarebczan:
Thanks for filing! It should be fairly easy to add a tip button similar to the ones on the claim pages. This tip would go directly to the creator's channel as a support, as opposed to their content. They can choose to leave it (to increase their effective bid) or claim it (be able to use it).  

This button should also be added on channel pages.

vv181:
@timurista Can I take over on the issue?

timurista:
@vv181 sure, sorry I haven't been able to get to this. No Problem.

PR #2722, opened 2019-08-10
add tip button component

vv181:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #1397


seanyesmunt:
Thank you for the pr!

Copying this from discord:

You should be able to support channels if they aren't your own. Check out how we do it here

https://github.com/lbryio/lbry-desktop/blob/master/src/ui/page/file/view.jsx#L250

If you create a new component, can you make it so we can pass a prop to choose if it's a support button or tip button?

Then we can have one component for everything

PR #2729, opened 2019-08-14
Random fixes

seanyesmunt:


src/ui/component/app/view.jsx
```diff
@@ -21,7 +22,7 @@ type Props = {
   language: string,
   theme: string,
   accessToken: ?string,
-  user: ?{ id: string, has_verified_email: boolean },
+  user: ?{ id: string, has_verified_email: boolean, is_reward_approved: boolean },
```
seanyesmunt: In before "there should be a User type"

seanyesmunt: Will be added...

sometime

src/ui/component/fileViewer/view.jsx
```diff
@@ -54,15 +58,24 @@ export default function FileViewer(props: Props) {
...
-  }
+  const previousUri = usePrevious(uri);
+  const previousIsReadyToPlay = usePrevious(isReadyToPlay);
+  const isNewView = uri && previousUri !== uri && isPlaying;
+  const wasntReadyButNowItIs = isReadyToPlay && !previousIsReadyToPlay;
+
+  useEffect(() => {
+    if (isNewView) {
+      setPlayTime(Date.now());
+    }
```
seanyesmunt: @tzarebczan brought up that we might not want to do play time (or keep play time separate) for non-streamable content.

src/ui/analytics.js
```diff
@@ -97,6 +103,30 @@ const analytics: Analytics = {
       });
     }
   },
+  emailProvidedEvent: () => {
+    if (analyticsEnabled && isProduction) {
```
kauffj: Making a `sendReactGAEvent` or equivalent function would eliminate these repeated if checks and could also help ensure the checks are not accidentally left out in the future.

src/ui/component/app/view.jsx
```diff
@@ -65,10 +69,22 @@ function App(props: Props) {
   }, [theme]);
 
   useEffect(() => {
-    if (userId) {
+    if (previousUserId === undefined && userId) {
```
kauffj: `=== undefined`? or just `!previousUserId` ?

kauffj: same question for below `undefined` checks

src/ui/component/fileViewer/view.jsx
kauffj: I'm not particularly concerned about this, I think it can be cleaned up in internal analytics.

src/ui/component/fileViewer/view.jsx
```diff
@@ -54,15 +58,24 @@ export default function FileViewer(props: Props) {
...
-    const newY = y + ui.deltaY;
-    setPosition({
-      x: newX,
-      y: newY,
-    });
-  }
+  const previousUri = usePrevious(uri);
+  const previousIsReadyToPlay = usePrevious(isReadyToPlay);
+  const isNewView = uri && previousUri !== uri && isPlaying;
+  const wasntReadyButNowItIs = isReadyToPlay && !previousIsReadyToPlay;
```
kauffj: :joy:

src/ui/modal/modalAutoGenerateThumbnail/index.js
```diff
@@ -1,11 +1,12 @@
 import { connect } from 'react-redux';
 import { doHideModal } from 'redux/actions/app';
 import { doToast, doUploadThumbnail } from 'lbry-redux';
+import fs from 'fs';
 import ModalAutoGenerateThumbnail from './view';
 
 const perform = dispatch => ({
   closeModal: () => dispatch(doHideModal()),
-  upload: buffer => dispatch(doUploadThumbnail(null, buffer)),
+  upload: buffer => dispatch(doUploadThumbnail(null, buffer, null, fs)),
```
kauffj: should these 3rd and 4th params be re-ordered? (meh)

src/ui/modal/modalConfirmThumbnailUpload/index.js
```diff
@@ -1,11 +1,12 @@
 import { connect } from 'react-redux';
 import { doHideModal } from 'redux/actions/app';
 import { doUploadThumbnail, doUpdatePublishForm } from 'lbry-redux';
+import fs from 'fs';
 import ModalConfirmThumbnailUpload from './view';
 
 const perform = dispatch => ({
   closeModal: () => dispatch(doHideModal()),
-  upload: path => dispatch(doUploadThumbnail(path)),
+  upload: path => dispatch(doUploadThumbnail(path, null, null, fs)),
```
kauffj: same question as above, if `fs` is regularly provided but other params are not, consider moving fs up in param order

when does JS get named parameters?!

src/ui/redux/actions/app.js
```diff
@@ -410,3 +418,18 @@ export function doToggleSearchExpanded() {
...
+    const state = getState();
+    const { txid, nout, claim_id: claimId } = makeSelectClaimForUri(uri)(state);
+    const claimIsMine = makeSelectClaimIsMine(uri)(state);
+    const outpoint = `${txid}:${nout}`;
+
+    if (claimIsMine) {
+      return;
+    }
+
+    analytics.apiLogView(uri, outpoint, claimId, timeToStart);
```
kauffj: could be:

`if (!claimIsMine) { logView() }`

src/ui/scss/component/_button.scss
```diff
@@ -94,6 +94,16 @@
   align-items: flex-start;
 }
 
+// Quick fix because this is a pain
+// There is something weird with wrapping buttons. Some places we want to wrap and others we want to ellips
```
kauffj: Good job noting why the rules are being broken so a future dev and/or future you can understand

src/ui/scss/init/_gui.scss
```diff
@@ -98,6 +98,11 @@ hr {
   }
 }
 
+img,
+a {
```
kauffj: Confident in no consequences to this rule being so universal?

src/ui/component/app/view.jsx
seanyesmunt: I guess this one specifically could be just be `!previousUserId` but the other ones need to check that they weren't `undefined` so the event isn't fired from the initial user fetch

src/ui/modal/modalAutoGenerateThumbnail/index.js
seanyesmunt: The third param is the android equivalent of `fs`

src/ui/scss/init/_gui.scss
seanyesmunt: 97% confidence

src/ui/analytics.js
seanyesmunt: Yeah I've been meaning to do this, just did and it's way nicer now.

PR #2738, opened 2019-08-15
update windows certificate

nikooo777:


tzarebczan:
I'd say let's save this for a patch / release next week. We wanna get it on some installs first to beat smart screen faster.

PR #2447, opened 2019-04-27
Enhance audio metadata access using music-metadata-browser

Borewit:
## PR Checklist

Please check all that apply to this PR using "x":

- [x] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)
- [ ] I have checked that this PR does not introduce a breaking change
- [ ] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting)
- [ ] Refactoring (no functional changes)
- [ ] Documentation changes
- [ ] Other

## Fixes

Issue Number: N/A

## Related

* #1824 Feature Request: It'll Be Nice To See Visualizations/Album Art When Playing Audio Files

## What is the current behavior?
Able to extract metadata from: MP3, FLAC, M4A

## What is the new behavior?
Able to extract metadata from: MP3, FLAC, M4A, WAV, Ogg (Vorbis, Opus), WavPack, DSDIFF, DSF, WMA (pretty much every audio format)

## Other information
[music-metadata-browser](https://github.com/Borewit/music-metadata-browser) is based on [music-metadata](https://github.com/borewit/music-metadata) which is the most advanced and [popular](https://npmcharts.com/compare/music-metadata,jsmediatags,musicmetadata,node-id3,music-metadata-browser,mp3-parser,browser-id3-writer,id3-parser,mp4-parser,parse-audio-metadata?start=600) metadata parser available today. Simular apps using music-metadata: [webtorrent-desktop](https://github.com/webtorrent/webtorrent-desktop), [webamp](https://webamp.org/), [volumio](https://volumio.org/), [mediacenterjs](https://github.com/jansmolders86/mediacenterjs) and many more.

I am looking forward to work with you. Please let know it you have specific challenges in this area I maybe able to help you with.

PR #2749, opened 2019-08-19
Remove unused dependency jsmediatags

Borewit:
I was checking if I could use [music-metadata-browser](https://github.com/Borewit/music-metadata-browser) with the audio-viewer, but I noticed it has been removed. 

Therefor this dependency is no longer required: jsmediatags

Related PR: #2447

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,7 +122,6 @@
     "hast-util-sanitize": "^1.1.2",
     "history": "^4.9.0",
     "husky": "^0.14.3",
-    "jsmediatags": "^3.8.1",
     "json-loader": "^0.5.4",
     "lbry-format": "https://github.com/lbryio/lbry-format.git",
     "lbry-redux": "lbryio/lbry-redux#1af092ce2cb507d9a41711b864874c0bd76935ae",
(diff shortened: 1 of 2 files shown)


kauffj:
Deleting code is better than adding it :)

Hope we see more from you @Borewit!

seanyesmunt:
Welcome back @Borewit !

I thought I just commented the audio player out, but I guess I removed it. It was still a ways away from working due to a number of bugs.

Would you be interested in working on a new audio player? I could bring back that old code, or set up a super simple one that could be added on to.

Borewit:
> Would you be interested in working on a new audio player? I could bring back that old code, or set up a super simple one that could be added on to.

I am not experienced with react, but I can give it a try.

I started with reviving the old code myself, just to get a bit in sync.

Borewit:
> I could bring back that old code, or set up a super simple one that could be added on to.

If you can setup the basic wiring either on top of [my branch](https://github.com/Borewit/lbry-desktop/tree/revive-old-audio-viewer), or fresh one, that would be great.

To capture the change, I opened issue #2757. If you can leave add some guidance @seanyesmunt to #2757, that would be very helpful.

PR #2812, opened 2019-08-28
does not autoplay nonfree

jessopb:


jessopb:
example test lbry://@oldmansaur#1/thinkingman#3

Issue #2798, opened 2019-08-27
Clicking enter in tag input (without typing) adds blank tag


tzarebczan:
Thanks to MH for spotting this one. 

If you are in the customize/publish page that has the inputs for tag search and don't type anything, clicking enter adds a blank tag.

zxawry:
a blank tag is also added if you type spaces only (unsure if people would want to follow space only tags) and if you add a tag with preceding spaces it escapes the duplicate check as well as being added into tags.

![](https://spee.ch/d/104ced03-dd22-4be9-9ea2-6c2eb253f7a7.png)

maybe input should be trimmed and checked for zero length before being added.

Issue #2642, opened 2019-07-18
sanitize tags before display/update


jessopb:
sdk allows tags with strange characters.
strip them in desktop on display and/or update.

PR #2817, opened 2019-08-29
fix blank tags when tags search field is empty

zxawry:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #2798 

new tags get added without being trimmed off spaces

## What is the new behavior?
tags search input gets trimmed yet allows multiple spaces between words

### DIFF
--- a/src/ui/component/publishForm/view.jsx
+++ b/src/ui/component/publishForm/view.jsx
@@ -140,9 +140,6 @@ function PublishForm(props: Props) {
             onSelect={newTag => {
               if (!tags.map(savedTag => savedTag.name).includes(newTag.name)) {
                 updatePublishForm({ tags: [...tags, newTag] });
-              } else {
-                // If it already exists and the user types it in, remove it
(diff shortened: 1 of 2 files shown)


zxawry:
Yes sure, but typing spaces only, suggests an empty tag too and if the input has trailing spaces it bypasses the duplicate check. so I used the second value to fix them.

zxawry:
Before I made the PR, I was getting repeated tag suggestions below the textbox. now I don't. :+1: 
I'm so sorry, perhaps I'm being too fussy about this small issue but I think tags (`item`) should also get trimmed here otherwise clicking add button submits untrimmed input as a tag: https://github.com/lbryio/lbry-desktop/blob/46a8a8635cc982f50e89b795b4ea09814e31ceb5/src/ui/component/tagsSearch/view.jsx#L92

tzarebczan:
You can claim https://github.com/lbryio/lbry-desktop/issues/2642

PR #2826, opened 2019-08-31
moves update banner to header

jessopb:
moving update banner to header makes it not push other content down which was causing positioning conflict between fileRenderInitiator and fileViewer boxes.


src/ui/scss/component/_content.scss
```diff
@@ -4,6 +4,12 @@
   top: var(--spacing-large);
 }
 
+.content__viewer--update-ready {
+  @extend .card;
+  position: absolute;
+  top: 6.4rem;
```
seanyesmunt: We should avoid using exact amounts like this whenever possible. If the text ever changes (font-size or wraps into two lines) this will no longer work. Maybe we could just disable the upgrade message on the file page for now?

PR #2861, opened 2019-09-06
Player state save WIP

jessopb:
Still working on setting volume at start
Includes / depends on the autoplay-first pr

### DIFF
--- a/src/ui/component/viewers/videoViewer/view.jsx
+++ b/src/ui/component/viewers/videoViewer/view.jsx
@@ -14,15 +14,43 @@ const VIDEO_JS_OPTIONS = {
 };
 
 type Props = {
+  volume: number,
+  position: number,
+  muted: boolean,
+  hasFileInfo: boolean,
+  changeVolume: number => void,
+  savePosition: (string, number) => void,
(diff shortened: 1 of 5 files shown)


src/ui/component/viewers/videoViewer/view.jsx
```diff
@@ -14,16 +15,81 @@ const VIDEO_JS_OPTIONS = {
...
+      currentVideo.addEventListener('seeked', doSeek);
+    }
+
+    // cleanup function:
+    return () => {
+      if (currentVideo) {
+        currentVideo.removeEventListener('ended', doEnded);
+        currentVideo.removeEventListener('pause', doPause);
+        currentVideo.removeEventListener('volumechange', doVolume);
+        currentVideo.removeEventListener('seeked', doSeek);
```
seanyesmunt: Could you just call `savePosition` here? Then every time it's taken off the page it saves. I think (?) that would do everything we need.

tzarebczan:
This would be nice to get in for the next patch if there isn't too much work left.

jessopb:
play next is done
volume is done.
will open a new one for position

PR #2862, opened 2019-09-06
Autoplay next

jessopb:
WIP only because currently choosing the first in related can result in a loop between 2 or 3 videos.

### DIFF
--- a/src/ui/component/viewers/videoViewer/view.jsx
+++ b/src/ui/component/viewers/videoViewer/view.jsx
@@ -14,15 +14,43 @@ const VIDEO_JS_OPTIONS = {
 };
 
 type Props = {
+  volume: number,
+  position: number,
+  muted: boolean,
+  hasFileInfo: boolean,
+  changeVolume: number => void,
+  savePosition: (string, number) => void,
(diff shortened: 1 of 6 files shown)


src/ui/component/viewers/videoViewer/view.jsx
```diff
@@ -14,16 +14,58 @@ const VIDEO_JS_OPTIONS = {
...
+    }
+    // cleanup function:
+    return () => {
+      if (currentVideo) {
+        currentVideo.removeEventListener('ended', doEnded);
+        currentVideo.removeEventListener('pause', doPause);
+        currentVideo.removeEventListener('volumechange', doVolume);
+      }
+    };
+  }, []);
```
seanyesmunt: Does the videoviewer always leave the page then get re-rendered again?

seanyesmunt:
This looks good. Nice work so far

jessopb:
Now we (mostly) only play unvisited links. However, we need to modify the doSearch to take some smarter options like mediaType.

seanyesmunt:
This looks fine. Closing this since it's contained in #2861

PR #2883, opened 2019-09-12
I18n for web, loaded from web, and token support

kauffj:
this technically works, but:

- I dislike the approach on web
- It splits web vs. desktop behavior
- It just generally feels hacky
- It still does not automatically push the translation files to Transifex (but this should be part of the build process)


kauffj:
@seanyesmunt, this is _potentially_ mergeable in that it is probably an advancement of the status quo. It has the following problems:

- I did not really test `<DateTime>` behavior before and after. It should be using the browser/OS locale here though, so I think it either should still work (or it did not work before).
- I simply had the desktop app mirror what the web app does. This is both: 1) a DRY violation and 2) sub-par to loading them from disk (but OTOH, does mean we get to keep the code the same).

It does:
- Remove `y18n`
- Make i18n stupid simple (if possibly too stupid)
- Load translations dynamically
- Add support for translations with tokens, both as strings or react components

kauffj:
@seanyesmunt officially requesting review on this, assuming we're fine with the timing consequences I outlined in chat.

Please be especially critical as to whether I'm following the current best practices regarding React/Redux state.

Additionally, I could not find a way to use the `<Spinner>` component inline in the setting component, let me know if I'm missing something there.

src/ui/component/i18nMessage/view.jsx
```diff
@@ -0,0 +1,35 @@
+// @flow
+import React from 'react';
+
+type Props = {
+  tokens: Object,
+  children: any,
```
seanyesmunt: Could you add some examples of what `children` will look like in a comment next to this?

src/ui/component/i18nMessage/view.jsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+export default function I18nMessage(props: Props) {
+  const message = __(props.children),
+    regexp = /%\w+%/g,
+    matchingGroups = message.match(regexp);
+
+  if (!matchingGroups) {
+    return <React.Fragment>{message}</React.Fragment>;
```
seanyesmunt: You can just do `return message`

src/ui/component/i18nMessage/view.jsx
seanyesmunt: I think this should be `children: string`

src/ui/component/rewardSummary/view.jsx
```diff
@@ -19,16 +20,15 @@ class RewardSummary extends React.Component<Props> {
...
+            <I18nMessage
+              tokens={{
+                credit_amount: <CreditAmount inheritStyle amount={unclaimedRewardAmount} precision={8} />,
+              }}
+            >
+              You have %credit_amount% in unclaimed rewards.
+            </I18nMessage>
           ) : (
-            __('You have no rewards available, please check')
+            <React.Fragment>{__('You have no rewards available, please check')}</React.Fragment>
```
seanyesmunt: `React.Fragment` is only needed if you are returning two or more "things" that aren't contained in a single wrapper, so it isn't needed here.

src/ui/component/walletSendTip/view.jsx
```diff
@@ -100,7 +101,6 @@ class WalletSendTip extends React.PureComponent<Props, State> {
                 label={__('Send')}
                 disabled={isPending || tipError || !tipAmount}
                 onClick={this.handleSendButtonClicked}
-                type="submit"
```
seanyesmunt: Removing this will prevent sending with the enter key

src/ui/redux/reducers/settings.js
```diff
@@ -1,20 +1,20 @@
...
+  isFetchingLanguage: false,
   daemonSettings: {},
   clientSettings: {
     // UX
     [SETTINGS.NEW_USER_ACKNOWLEDGED]: false,
     [SETTINGS.EMAIL_COLLECTION_ACKNOWLEDGED]: false,
 
     // UI
-    [SETTINGS.LANGUAGE]: 'en',
+    [SETTINGS.LANGUAGE]: window.localStorage.getItem(SETTINGS.LANGUAGE) || 'en',
```
seanyesmunt: Didn't make it long without `localStorage` in this file... 😛

static/index-web.html
```diff
@@ -7,6 +7,31 @@
 
   <body>
     <div id="app"></div>
-    <script type="text/javascript" src="/ui.js"></script>
+   <!-- <script src="/ui.js"></script> -->
```
seanyesmunt: old comment

src/ui/component/walletSendTip/view.jsx
seanyesmunt: Just saw it's a duplicate 🤦‍♂

PR #2887, opened 2019-09-16
basic search channel pages

seanyesmunt:
So i can search for stuff on channels

This is super basic, and probably should live in the component file, although after doing it this seems a lot simpler than running everything through redux (and a lot faster to write).

It does not include deep links to channel search queries. It was just something quick so I could use it.

Screen Shot 2019-09-16 at 1 21 32 AM


src/ui/page/channel/view.jsx
```diff
@@ -67,22 +71,47 @@ function ChannelPage(props: Props) {
...
     } else {
+      setSearchResults(null);
       search += `page=${page}`;
     }
 
     history.push(`${url}${search}`);
-  };
+  }
+
+  function handleSearch() {
```
kauffj: This should probably be in https://github.com/lbryio/lbry-redux/blob/master/src/redux/actions/search.js

seanyesmunt: I initially thought that and I think it makes sense to put it there eventually, but I'm starting to think small stuff like this is fine if live in the components. 

All it is is one lighthouse call

seanyesmunt:
I'm going to merge this as-is since I don't have time to do more work on it. I think this is fine for the first basic version of channel search.

PR #2892, opened 2019-09-16
Allow user to view and claim synced youtube channels

jessopb:
depends on
https://github.com/lbryio/lbry-redux/pull/199
https://github.com/lbryio/lbryinc/pull/57


jessopb:
Last thing is choosing whether and how to display any errors from the transfer process.

jessopb:
A couple things tz wants: to be able to refresh and see progress on claims transferred. This means another action to merge yt/transfer api response into the channels part of the user reducer to populate total transfered / total videos.

Then possibly if transfer is pending, the Claim button label changes to Refresh since claiming is already going on.

PR #2895, opened 2019-09-17
remove localStorage from settings

### LINKED ISSUE #2574: lbry.tv make settings persist in session
## The Issue
Settings like Dark mode should at least persist in the user session / between refreshes. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

seanyesmunt:
fixes #2574 

- Removes localStorage from the settings reducer
  - This will reset all users settings back to the defaults which kinda sucks, but it would be a pain to migrate them
- Cleans up some settings names/harcoded values


tzarebczan:
Not a fan of resetting all the settings, I'm sure some folks won't like that. Do we know if it happened? Maybe we can show a note in-app too.

seanyesmunt:
We won't know if they were just lost, but we could show some message in the app. I'm not as worried about that though.

PR #2900, opened 2019-09-18
makes claimRewards wait for view trigger

jessopb:
The suggested fix was to triggerview().then( () => claimRewards ). 
Assuming the reward is ready when the view response is received, this async/await should work.


src/ui/component/fileViewer/view.jsx
```diff
@@ -65,6 +65,10 @@ export default function FileViewer(props: Props) {
   const previousIsReadyToPlay = usePrevious(isReadyToPlay);
   const isNewView = uri && previousUri !== uri && isPlaying;
   const wasntReadyButNowItIs = isReadyToPlay && !previousIsReadyToPlay;
+  const logViewAndClaimRewards = async (uri, timeToStart) => {
+    await triggerAnalyticsView(uri, timeToStart);
```
seanyesmunt: This won't work because this function doesn't return a promise after the view event is complete
https://github.com/lbryio/lbry-desktop/blob/master/src/ui/analytics.js#L69

PR #2909, opened 2019-09-20
Paginate Transactions

jessopb:
I broke out the txList table from the txList. Now TransactionList handles placing the header and pagination, while TransactionListTable just displays the list it's given.
Converted from class to functional component.


seanyesmunt:
Closing in favor of #2912

PR #2934, opened 2019-09-28
stores and recalls volume

jessopb:


jessopb:
> This appears to re-mount the video component the first time you change the volume.
> 
> Changing it again after the first time works well, but it resets the video to the beginning after you change it initially

Ok, I removed muted and volume from the useEffect deps. On my test, I didn't experience any hiccups. Try again?

tzarebczan:
If there's time and this works, let's try to get it into the next release.

tzarebczan:
This LGTM!

PR #2945, opened 2019-10-01
fix: abandoning without file

tzarebczan:
This fixes not being able to abandon without first downloading the file.
I noticed the "go back" part is broken here and in the prod app...Sean can you see what's up there?


seanyesmunt:
You can use `doDeleteFileAndMaybeGoBack` to go back

tzarebczan:
That's what is calling doDelete....

seanyesmunt:
🤦‍♂

PR #2956, opened 2019-10-02
Rc fixes

seanyesmunt:


seanyesmunt:
I'm going to merge this since it's mostly small fixes, but would appreciate a look. I don't think it's necessary though.

PR #2958, opened 2019-10-02
Wording changes on wallet balance page

eggplantbren:
I had a go at improving the wording on the wallet balance page. You can reject this if you don't like it.

**Important: ** I don't know this language so check my syntax carefully. I also noticed that for one of the items, LBC wasn't in bold. I attempted to fix this by using the same syntax that was used for the bold ones, but I could be wrong in what I did.

### DIFF
--- a/src/ui/component/walletBalance/view.jsx
+++ b/src/ui/component/walletBalance/view.jsx
@@ -42,7 +42,7 @@ const WalletBalance = (props: Props) => {
                 <strong>
                   <CreditAmount badge={false} amount={rewardTotal} precision={8} />
                 </strong>{' '}
-                {__('Earned From Rewards')}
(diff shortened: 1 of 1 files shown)


tzarebczan:
I know @kauffj wanted to take a copy pass at this, assigned him for review.

tzarebczan:
"Your Publishes" - should be "Your Claims" as it includes channels and streams.

seanyesmunt:
IMO we should stick with publishes since new users probably don't know what a claim is.

kauffj:
Ended up providing notes in our chat, this is going to change but not exactly like this. Your feedback was useful and appreciated @eggplantbren.

Agreed that we should avoid using the word claim outside of technical docs or conversation.

Issue #907, opened 2017-12-28
Ability to search through Downloads / Publishes pages


tzarebczan:
## The Issue
As a user, I'd want to be able to search through my downloads and published files. Currently, if you are on any of those pages, the search function will just take you to a regular LBRY search. 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

astrolope:
Hello! 

I just forked and was looking through how the Wunderbar is rendered in relation to the downloads / published file lists. 

Would you only want the files to be searchable from their respective pages? Or, include the results in the auto-suggestion box. Also, should the rendered lists be filterable as a result of the search if you were on one of those pages?

tzarebczan:
Hey @astrolope, welcome to the LBRY GitHub! Have you run the current redesign version from master? The search now includes results from Downloads, but does not go into Publishes. There might be enough room after Downloads to also show search results from Publishes, or we could includes both downloads/publishes in the single column. Maybe @seanyesmunt has some thoughts here. 

Looking forward to your contributions. Check out https://lbry.io/faq/contributing :)

astrolope:
@tzarebczan, thanks for the welcome! Are you referring to the latest updates on the master branch correct? If so, yes. I ran a search for something I downloaded and it listed itself twice in the auto-suggestion box, so that must be the addition of the searchable downloads. Perhaps some text or creating a divider to indicate that it's an already downloaded/published file and available locally? If there was enough room I feel it could make sense to intermingle those respective suggestions.

tzarebczan:
@astrolope not sure if the duplicate was coming from downloads, it could be related to something else (the autocomplete results themselves). 

After entering your search term, select the first item (which is the search term -search) - this will take you to a search page - that's what I was referring to.

seanyesmunt:
Hey @astrolope!

All of the search redux code lives in `lbry-redux`. I think adding autocomplete suggestions from downloads would be great, and your solution sounds like it's going in the right direction. I know an additional change I would like to make is to concat all of the file "parts" (author, description, title) into one string and perform a regex search on it, which would allow partial matches to work. (That should probably be it's own issue though).

I think there are 3 parts to this issue:

1. Add published content to the search page. This would be in the same column as downloads
2. Add downloads/publishes to the autocomplete results in the search bar, we would need to limit them to some number of matches so you don't get 100 results from typing the letter "a" (not sure the best way to do that
[...]

PR #2969, opened 2019-10-03
File downloaded search

kauffj:
half of #907 (file downloads)

requires https://github.com/lbryio/lbry-redux/pull/214


seanyesmunt:
What is the expected behavior with pagination? Right now when you search for something generic, there can be more than one page, but clicking on the next page loses the search query.

seanyesmunt:
I think the only thing that needs to change is this line
https://github.com/lbryio/lbry-desktop/blob/master/src/ui/component/common/paginate.jsx#L30

kauffj:
https://github.com/lbryio/lbry-desktop/pull/2969/files#diff-56227b63889e385f7c6fdbe783ca60e2R30-R32 is the relevant change.

Issue #2029, opened 2018-10-11
Only animate GIF thumbnails on hover over


tzarebczan:
## The Issue
Due to the finding described in https://github.com/lbryio/lbry-desktop/issues/1848, we should see if there's a way to pause animation on GIF thumbnails unless a mouse is hovered over it.  This behavior would be similar to what is seen on other popular video sites. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

CoryWritesCode:
Mind if I take a shot?

hydrosquall:
Will have a PR for this up soon!

seanyesmunt:
Added!

seanyesmunt:
Thanks @hydrosquall !

PR #2986, opened 2019-10-06
[feature] Disable GIF animation unless user hovers

hydrosquall:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #2029 

- GIFs play all the time, making the CPU/GPU work hard on GIF intensive pages

## What is the new behavior?

- GIFs only play if the user's mouse hovers over them. Tested using the `lbry://@3DPrinting` address.

<a href="https://cl.ly/bc1d8285a927" target="_blank"></a>

## Other information

- I filed two feature requests to `freezeframe`, one to make the official react wrapper usable for our usecase (https://github.com/ctrl-freaks/freezeframe.js/issues/54), and the other to make sure that event listeners get cleaned up when the image is unmounted. https://github.com/ctrl-freaks/freezeframe.js/issues/55
- Two reads I found useful about how to solve this problem were 
  - https://dking.org/blog/2019/04/options-for-pausing-the-animation-of-gifs/ and 
  - https://stackoverflow.com/questions/5818003/stop-a-gif-animation-onload-on-mouseover-start-the-activation

Since we don't control the source of these GIFs as they are submitted by users, I felt that the solutions for requesting a separate static image, using CSS keyframes, or manually creating a canvas were not applicable, even though those other solutions would probably get even better performance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Ability to add multiple tags at once with commas ([#2833](https://github.com/lbryio/lbry-desktop/pull/2833))
+- Disable GIF animation unless user hovers ([#2986](https://github.com/lbryio/lbry-desktop/pull/2986))
 
 ### Changed
 
--- a/package.json
+++ b/package.json
@@ -123,6 +123,7 @@
     "flow-bin": "^0.97.0",
     "flow-typed": "^2.3.0",
     "formik": "^0.10.4",
+    "freezeframe": "^4.0.0-alpha.7",
     "hast-util-sanitize": "^1.1.2",
     "history": "^4.9.0",
     "husky": "^0.14.3",
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "electron-updater": "^4.1.2",
     "express": "^4.16.4",
     "if-env": "^1.0.4",
+    "imagesloaded": "^4.1.4",
     "keytar": "^4.4.1",
     "mysql": "^2.17.1",
     "nodemon": "^1.19.1"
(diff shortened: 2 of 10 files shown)


seanyesmunt:
Thank you for the PR!

I'm not sue if I want to add this dependency for this one feature. bundlephobia says it alone will add 400ms on 3g networks. Is there any way this could be achieved without adding freezeframe.js?

hydrosquall:
Hi @seanyesmunt ,

Thanks for the suggestion, I had forgotten that this wasn't just a desktop app that gets installed once, and that this code is also being use for the web application.

I took a look at the bundle results inside https://chrisbateman.github.io/webpack-visualizer/, and found that  ~36%  of the bundle size came from `core-js` (which is already a project dependency) and 36% came from `common-tags`, which could be eliminated in favor of using a regular template literal instead of a tagged template literal. 

With that motivation in mind, I extracted the areas of code + branches that we were using from freezeframe.js, and inlined them into the component folder. Each file has a URLs attributing the files back to the source. I checked on the desktop electron app, and it works the same as we had it before.

This lets us whittle dependencies down to just the `imagesloaded` package, which adds `34 ms` on emerging 3G.

https://bundlephobia.com/result?p=imagesloaded@4.1.4

Let me know if there are additional changes you'd like to see.

seanyesmunt:
Nice work. That seems a lot more reasonable. I'll give this a closer look later today!

seanyesmunt:
Hey sorry it took so long to review this. Just tested and it works great!

Can you add a change to this file? https://github.com/lbryio/lbry-desktop/blob/master/CHANGELOG.md

then it's ready to merge.

hydrosquall:
Thanks @seanyesmunt , I'll update the changelog :).

Hi @tzarebczan, thanks for the suggestion - I'll send an email after this is merged.

Regarding an indicator: the original freezeframe.js had an "arrow icon", see the third demo at the page below:

http://ctrl-freaks.github.io/freezeframe.js/

I think it can be incorporated, perhaps in a follow-up PR? Since we no longer depend on freezeframe directly, we can use any icon that we'd like (maybe something from the lbry brand?)

kauffj:
Great contribution @hydrosquall!

tzarebczan:
And don't forget to follow up about Hacktoberfest / [appreciation](https://lbry.com/faq/appreciation) :)

Issue #2677, opened 2019-07-26
ui/components/subscribeButton has glitchy visual behavior.


abueide:
## The Issue
When hovering over the heart icon of the unsubscribeButton, glitchy visual behavior happens.

### Steps to Reproduce
Hover over heart icon in a very specific pixel.

### Recorded Video of the Bug
https://beta.lbry.tv/bug/f

tzarebczan:
Thanks for opening the issue @abueide ! I was able to reproduce...,yeah, looks like it's in very particular spot. 

Can we show you some [appreciation](https://lbry.com/faq/appreciation) for the issue?

kauffj:
I love the usage of LBRY to report bugs on LBRY. It's LBRY all the way down!

@seanyesmunt Even with a resolution of this specific issue, I've found that it can create a feeling of "jumpiness" in other places. I really like the new behavior - maybe we need to fix the width?

abueide:
Haha, yeah no matter how I tried to make the gif, it was bigger than github/imgur's limit of 10/20 mb, so I thought why sacrifice quality might as well upload to lbry.

ICTman1076:
I think I might have a crack at this. My initial thought is to change the character spacing ever so slightly so that "unfollow" is slightly wider, and "following" is either the same width or slightly thinner, so that the two become equal width and result in this glitchy situation to be resolved. Having different character spacing may make it look worse, but it's worth a shot, right?

ICTman1076:
^ Sorry for the first PR, I made a mistake on that one. #2990 is the one that should be merged.

VeroniqueHuang:
Hi, I'd like to work on this if it's not fix yet !

seanyesmunt:
@VeroniqueHuang It's all yours!

SleepingFox88:
Hello, I am looking at fixing this issue. This is my first time developing with LBRY and I am having trouble seeing my changes take place.

I am currently trying to modify the string ``'Following'`` here

![](https://i.imgur.com/5DHWPle.png)

however after running ``yarn dev`` to compile the project I am not seeing that the text has changed. Any help is appreciated. I am currently using

https://spee.ch/7da73fc508ffc4ff8b2711e3c3950110430b0c5f/LBRYAppDesign.mp4

as a reference to start developing. If there is a more up to date reference I would be happy to look at it.

seanyesmunt:
Hey @SleepingFox88 

[...]

PR #2989, opened 2019-10-06
Fix glitchy follow button behaviour

ICTman1076:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #2677 

Hovering over a very particular part of the button caused the text to change, which made the element smaller, which made it so you were no longer hovering over it, causing it to change to bigger text, which made it so you were hovering over it, causing a feedback loop and creating a glitchy visual bug.

## What is the new behavior?

Now, the smaller text has a tiny bit of barely noticeable letter spacing, and the larger bit has had a tiny bit of letter spacing removed, to make the two states the same width, meaning the label remains the same size.

## Other information

As far as I'm aware, there's no other information required.


ICTman1076:
Aw, nuts, I forgot about when it's in the not-followed state - it expands the text then, which doesn't help. Let me try that again - sorry!

PR #2990, opened 2019-10-06
Fix glitchy follow button behaviour

ICTman1076:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #2677 

Hovering over a very particular part of the button caused the text to change, which made the element smaller, which made it so you were no longer hovering over it, causing it to change to bigger text, which made it so you were hovering over it, causing a feedback loop and creating a glitchy visual bug.

## What is the new behavior?

Now, the smaller text has a tiny bit of barely noticeable letter spacing, and the larger bit has had a tiny bit of letter spacing removed, to make the two states the same width, meaning the label remains the same size.

## Other information

I made another PR previous to this one which may seem similar, but was actually broken because it added character spacing where it wasn't appropriate. I also fine-tuned the numbers in this one so that the width of the element doesn't just look the same width, but is the exact same width (with 3 d.p. accuracy).


seanyesmunt:
I tested and it looks a lot nicer. It doesn't work for non-english languages, but this is still an improvement. Can you add an note in this file? Then it's ready to merge.
https://github.com/lbryio/lbry-desktop/blob/master/CHANGELOG.md

ICTman1076:
All done @seanyesmunt. I'll take a look into fine-tuning it for other languages, maybe even dynamically calculating the spacing if I can.

kauffj:
@seanyesmunt @ICTman1076 this solution will not work for international users. 

I believe a more general solution would be to render the button with all possible text, inspect the width of each render, then fix the width to the largest text.

ICTman1076:
I just went through each of the languages currently available, and the problem is that the largest possible text is more than double of the smallest possible text. That would mean that you could have up to a whole character's worth of letter-spacing in between each letter, which would not look good. If you want a solution that works for all languages, you'll either have to ditch letter-spacing and set the width to a fixed width (which would then introduce lots of blank space which also wouldn't look great), or work out the exact letter spacing needed for each language, rather than just one for all languages, or use react to calculate it dynamically. I don't know how to do the latter in react, but I feel that would be the best option.
[Here's a table of the widths of that element in hover and normal state for each language before this PR.](https://i.ibb.co/wMCKVtY/image.png) I don't know if that helps, but you may as well see it

ICTman1076:
Not sure why travis failed - there's been no code changes 🤔

seanyesmunt:
@kauffj I think that can end up looking worse than what we have currently, you can get large empty spaces with small text in the button. This at least fixes it for English, which is the majority of users.

tzarebczan:
Will follow up via email, thanks again for the contribution!

kauffj:
@seanyesmunt I don't want to start a riff over something so minor, but I'm pretty disinclined to merge this.

To clarify my previous answer, I don't mean to calculate the width for all languages, I mean for the user's currently set language, calculate the width for just those two words, and fix it to the larger of the two.

Alternatively, come up with another design for this altogether.

seanyesmunt:
@ICTman1076 Do you think you would be able to implement the solution that @kauffj has suggested?

ICTman1076:
I can definitely take a look at the very least - I'm not sure it'll look great with some of the languages with a bigger difference in width between the two states, but I'll see what happens

kauffj:
:clap: :clap: :clap: best idea yet @seanyesmunt!

ICTman1076:
I'll give it a shot - I'm quite new to react so this may end up never happening, or done completely wrong, but it's worth an attempt lol
(sorry for the delay, been busy recently)

seanyesmunt:
I'm going to close this. @ICTman1076 Feel free to re-open if you would like to keep working on it!

Issue #2935, opened 2019-09-29
Allow blocking of Anonymous content


eggplantbren:
A user in the Discord has requested the ability to block "Anonymous" content as if it was a channel. This seems like it wouldn't be hard to do, so I opened an issue for it.

tzarebczan:
Great idea. This should be possible on claim search with this option:

```--has_channel_signature         : (bool) claims with a channel signature (valid or invalid)```

tzarebczan:
Completed in https://github.com/lbryio/lbry-desktop/pull/3001 at a per claim basis, but will leave this open to move to the has_channel_signature option. This would most likely be done with https://github.com/lbryio/lbry-desktop/issues/2801 since adding that option removes channels from trending as well. Will discuss with SDK team if there's a better way around this.

Issue #2890, opened 2019-09-16
Blocked Mature Channels Are Hidden With Show Mature Off


FLStudioBits:
**Bug**
Blocked Mature Channels Are Hidden On The Manage Blocked Channels Page With Show Mature Off

**To Reproduce**
1. block a mature channel
2. turn show mature off
now the mature channel is also hidden on the manage blocked channels page

App    0.35.3
Daemon (lbrynet)    0.40.1

tzarebczan:
Thanks for finding and filing this @FLStudioBits ! Sent you some appreciation on Discord.

PR #3001, opened 2019-10-08
feat: add option to hide anonymous content

zxawry:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #2935 and probably #2890 

there is no option to block anonymous claims.

## What is the new behavior?
the option to hide anonymous claims is added to the settings page.

## Other information

this probably needs to improve, for example with the feature turned on, anonymous claims on library page are filtered too.


tzarebczan:
@zxawry , we should also pass this param to claim search when the option is enabled: https://github.com/lbryio/lbry-desktop/issues/2935#issuecomment-536555236

This will make sure trending/top/new don't display it from the get go, as opposed to blocking it and potentially running into pagination issues.

zxawry:
@tzarebczan I tried it but it filters channels as well, is it okay?

tzarebczan:
@zxawry not sure I understand. If you pass something like this `lbrynet claim search --order_by=trending_global --has_channel_signature` it should return claims with channels, and filter out anonymous ones. Is that what you tried? 

So pretty much we'd wanna tack on `has_channel_signature: true` to the claim search for trending with the option is enabled.

zxawry:
@tzarebczan, I mean channel claims e.g. `lbry://@history` that sometimes get listed in the discovery have no channel signature same as anonymous claims so they will be filtered too.

tzarebczan:
Oh, I see. I think that's okay for now - we could just make a note of that in the copy. Once we add more filtering options as part of https://github.com/lbryio/lbry-desktop/issues/2801,we can ignore this option if someone uses the All/Channels selector for channels vs claims. 

@seanyesmunt do you agree?

zxawry:
Is this considered appropriate to separate channel discovery from other claim types for the time being?
 
screenshot

tzarebczan:
I think those options would be somewhere else under an advanced / filter menu. 

For now for this pr, I think your solution might be fine, and we'll introduce the parameter I was referencing later.   Unless @seanyesmunt disagrees.

seanyesmunt:
I think the current solution is fine. I will update the copy for the setting help text then merge this.

tzarebczan:
@zxawry, thanks for yet another awesome PR!

Issue #2955, opened 2019-10-02
Auto-detect OS user language


kodxana:
**Is your feature request related to a problem? Please describe.**
Need manualy find and implement other Languages.

**Describe the solution you'd like**
App should pull informations from OS about user language, and try set translation automatic, or prompt user if he want use his native language or stay in English. If Transfix return no translation for that language it should set it to default English.

tzarebczan:
Great suggestion!

kodxana:
Suggestion how to get is locale https://github.com/sindresorhus/os-locale

peterjgrainger:
Can I have a go at this one?

tzarebczan:
@peterjgrainger welcome to LBRY! Sure, take a stab at it!

kauffj:
@peterjgrainger FYI language is being accessed before UI loads on both web (where this is probably necessary) and desktop (where it's probably not necessary but was more efficient from development perspective).

You can see this in `index-web.html` and `index-electron.html`. If you have any questions, don't hesitate to tag me!

Rupeshiya:
@kauffj Can I work on this?
I think simply can be done using https://github.com/sindresorhus/os-locale.

peterjgrainger:
@kauffj thanks for the offer of help will probably need it!

Just getting the project set up now

peterjgrainger:
@kauffj my thought is to use the `app.getLocale()` builtin function in electron https://github.com/electron/electron/blob/master/docs/api/app.md#appgetlocale

would the best place to put that be here? https://github.com/lbryio/lbry-desktop/blob/6471aa2140708f2ca3aaa5c15513ef4d02ceacbd/src/platforms/electron/index.js#L42

Also do you think it is a good approach to default to the system OS then give the choice to change it on the "Customize Your Homepage" discover view 
<img width="547" alt="image" src="https://user-images.githubusercontent.com/1332395/66585378-9df50d00-eb7e-11e9-894e-123f8e5a6ec7.png">

kauffj:
@peterjgrainger yes, `app.getLocale()` is the way to go. If we want this to work on both desktop and web, then there are probably two changes to make:

1. For desktop, use `app.getLocale()` to set the language _before_ rendering `index.html`. I believe you'll need to do this in `createWindow.js` somewhere between creating the window and the call to `loadURL`.
[...]

PR #3018, opened 2019-10-11
Use OS language to default localization

peterjgrainger:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #2955

When opening the app for the first time. The language is always English

## What is the new behavior?

The language is set to the primary OS language on first use.

After that, the logic stays the same. Use the stored language in the `window.localStorage`

### DIFF
--- a/src/platforms/electron/createWindow.js
+++ b/src/platforms/electron/createWindow.js
@@ -5,6 +5,7 @@ import windowStateKeeper from 'electron-window-state';
 
 import setupBarMenu from './menu/setupBarMenu';
 import * as PAGES from '../../ui/constants/pages';
+import SUPPORTED_LANGUAGES from '../../ui/constants/supported_languages';
 
 export default appState => {
   // Get primary display dimensions from Electron.
@@ -149,7 +150,9 @@ export default appState => {
 async function setLanguage(window) {
   const storedLanguage = await window.webContents.executeJavaScript("localStorage.getItem('language')");
   const lang = storedLanguage || app.getLocale().slice(0, 2) || 'en';
-  if (['pl', 'id', 'de'].includes(lang)) {
+
+  const supportedNonEnglish = Object.keys(SUPPORTED_LANGUAGES).filter(language => language !== 'en');
+  if (supportedNonEnglish.includes(lang)) {
     const response = await fetch('https://lbry.com/i18n/get/lbry-desktop/app-strings/' + lang + '.json');
     const json = await response.json();
     const messages = {};
--- a/src/platforms/electron/createWindow.js
+++ b/src/platforms/electron/createWindow.js
@@ -87,6 +87,8 @@ export default appState => {
   window.loadURL(rendererURL + deepLinkingURI);
   setupBarMenu();
 
+  setLanguage(window);
+
   window.on('close', event => {
     if (!appState.isQuitting && !appState.autoUpdateAccepted) {
       event.preventDefault();
@@ -143,3 +145,16 @@ export default appState => {
 
   return window;
 };
+
+async function setLanguage(window) {
+  const storedLanguage = await window.webContents.executeJavaScript("localStorage.getItem('language')");
+  const lang = storedLanguage || app.getLocale().slice(0, 2) || 'en';
+  if (['pl', 'id', 'de'].includes(lang)) {
+    const response = await fetch('https://lbry.com/i18n/get/lbry-desktop/app-strings/' + lang + '.json');
+    const json = await response.json();
+    const messages = {};
+    messages[lang] = json;
+    // Send message to render layer to update language.
+    window.webContents.send('language-update', messages, lang);
+  }
+}
--- a/src/platforms/electron/createWindow.js
+++ b/src/platforms/electron/createWindow.js
@@ -5,7 +5,7 @@ import windowStateKeeper from 'electron-window-state';
 
 import setupBarMenu from './menu/setupBarMenu';
 import * as PAGES from '../../ui/constants/pages';
-import SUPPORTED_LANGUAGES from '../../ui/constants/supported_languages';
+import setLanguage from './setLanguage';
 
 export default appState => {
   // Get primary display dimensions from Electron.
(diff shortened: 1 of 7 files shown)


peterjgrainger:
I'm new to the project (and electron) can whoever looks at this check I am on the right track?

src/platforms/electron/createWindow.js
```diff
@@ -143,3 +145,16 @@ export default appState => {
 
   return window;
 };
+
+async function setLanguage(window) {
+  const storedLanguage = await window.webContents.executeJavaScript("localStorage.getItem('language')");
+  const lang = storedLanguage || app.getLocale().slice(0, 2) || 'en';
+  if (['pl', 'id', 'de'].includes(lang)) {
+    const response = await fetch('https://lbry.com/i18n/get/lbry-desktop/app-strings/' + lang + '.json');
```
kauffj: good call to move language fetching here

src/platforms/electron/createWindow.js
```diff
@@ -87,6 +87,8 @@ export default appState => {
   window.loadURL(rendererURL + deepLinkingURI);
   setupBarMenu();
 
+  setLanguage(window);
```
kauffj: If this is an async function that happens after `loadURL`, couldn't that create a race condition where some strings are rendered in English before the translation files are loaded?

peterjgrainger: translations seem to happen on the fly. The original code was also Async. 

My main issue that I had was the `index.electron.html` file was loaded before the `CreateWindow.js` file

kauffj: Hmm I may be misunderstanding how this file works then. I have a meeting with @seanyesmunt momentarily, let me check with him and we'll get an answer to you.

src/platforms/electron/createWindow.js
```diff
@@ -143,3 +145,16 @@ export default appState => {
 
   return window;
 };
+
+async function setLanguage(window) {
+  const storedLanguage = await window.webContents.executeJavaScript("localStorage.getItem('language')");
```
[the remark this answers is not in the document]
peterjgrainger: Yeah, `BrowserWindow` doesn't have `localStorage` :(

src/platforms/electron/createWindow.js
kauffj: I think you need to move `setLanguage` before `loadURL` and you will be fine, since you have an `await` on language fetch. By calling it after, I do think there is a chance some translations will fail. You should be able to confirm/test this by arbitrarily delaying the call to `setLanguage` or the duration of the time to download the translation file.

tzarebczan:
@peterjgrainger, thanks for the PR! Looks like there are some additional comments for you to review, please take a peek.

src/platforms/electron/createWindow.js
peterjgrainger: Only async funcitons will wait for the function to finish. Putting it before loadURL won't do anything I don't think.

I could put it in `index.js`? That is an async function? I'm not sure that delaying startup is the best option though. I don't think the delay is worth it, but let me know if you disagree!

kauffj: @peterjgrainger it's possible I'm misunderstanding something. 

What I'm fairly confident of is that if `index.html` loads and runs `ui.js` _before_ the JSON translations are returned from lbry.com, then some translations could return in English.

Because you have an `await` in front of the translations download, I thought moving `setLanguage` up would be sufficient. Either way, the key is making sure the language resource is loaded and stored before `ui.js` runs.

This unfortunately does cause a ~100ms load hit, but avoiding it would be a substantial refactor that I don't think we can swallow right now (although if you're itching for more... :grin:).

peterjgrainger: > What I'm fairly confident of is that if index.html loads and runs ui.js before the JSON translations are returned from lbry.com and stored in window.i18n_messages, then some translations could return in English.

Yes agree completely. The translations  definitely will be loaded after the UI starts to render. Looking again at the index it already adds this delay. https://github.com/lbryio/lbry-desktop/blob/005aad6fdb83678c4e2093ee84da9c4a2848d11d/static/index-electron.html#L26. So at least I wouldn't be slowing things down. 

> Because you have an await in front of the translations download, I thought moving setLanguage up would be sufficient.

I'll have to relocate the code somewhere I can block the thread to stop `ui.js` running. I'll have to take a look.

> This unfortunately does cause a ~100ms load hit, but avoiding it would be a substantial refactor that I don't think we can swallow right now

I'll split up the logic so it gets the language from `localStorage` and if it's English then doesn't do the `fetch` (like the current logic)

> (although if you're itching for more... 😁).
Maybe we could speed this up by storing the translations in  `localStorage`? Maybe for the scope of this change I'll keep it simple :)

src/platforms/electron/createWindow.js
```diff
@@ -143,3 +145,16 @@ export default appState => {
...
+  const lang = storedLanguage || app.getLocale().slice(0, 2) || 'en';
+  if (['pl', 'id', 'de'].includes(lang)) {
```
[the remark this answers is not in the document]
peterjgrainger: @kauffj  can you look at the latest code and see if your happy with how I did this?

peterjgrainger:
@kauffj Can you have a look at my new edits. I had to put some logic back in the `index.electron.html` but I think it's the only way to do it without a major rewrite.

Let me know what you think?

I tested it with Polish and German on the web and the desktop and the translations come through when in `window.localStorage` or from the OS.

kauffj:
@peterjgrainger it passes the eyeball test. I suspect if we find any additional issues we can simply address them ourselves when we merge, which will possibly be tomorrow but is more likely to be next week.

Make sure to check out @tzarebczan's comments above and send us an email, we'd love to send you some LBC and a T-shirt.

peterjgrainger:
Thanks @kauffj it was fun!

kauffj:
Closing in favor of #3180 but the commit history on that PR should still credit @peterjgrainger. Thanks for your work on this Peter!

Issue #2703, opened 2019-07-31
Add message on mature tags page to say content is hidden because of users settings


seanyesmunt:
If I go to the `porn` tag the first time I open the app I will be (hopefully) greeted with 0 porn. This is good for people who don't want to see porn, but bad for people who want that no-good, vile content.

We should show a message at the top saying, `content is hidden on this page because your settings`.

We show this on the search/channel pages

max-arias:
I'd like to give this a shot! 🤚

max-arias:
@seanyesmunt Could you provide a little more context on how to repro this? Every time I seach for porn  I get SFW-related porn content, so I can't implement this.

tzarebczan:
@max-arias sorry for not replying earlier. We have a PR in progress for this already. Maybe try to find another hacktoberfest issue to tackle?

PR #3038, opened 2019-10-14
add message to say content is hidden according to settings (2703)

tombiju:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: 2703 (https://github.com/lbryio/lbry-desktop/issues/2703)

- No warning message that content might be hidden

## What is the new behavior?
- A message shows up in the UI near the top notifying the user that there settings might hide some content, and there is a link to go to the settings page and modify them.

- Although the ticket only asked for mature tags, the message is not too much in the users face so I thought it would be good to have it in the landing discover page, during tag searches, and during regular text searches:

image

image

image

### DIFF
--- a/src/ui/component/claimListDiscover/view.jsx
+++ b/src/ui/component/claimListDiscover/view.jsx
@@ -45,6 +45,7 @@ type Props = {
     [string]: Array<string>,
   },
   hiddenUris: Array<string>,
+  hiddenNsfwMessage?: Node,
 };
 
 function ClaimListDiscover(props: Props) {
@@ -61,6 +62,7 @@ function ClaimListDiscover(props: Props) {
     location,
     hiddenUris,
     hideCustomization,
+    hiddenNsfwMessage,
   } = props;
   const didNavigateForward = history.action === 'PUSH';
   const [page, setPage] = useState(1);
@@ -109,6 +111,7 @@ function ClaimListDiscover(props: Props) {
     (personalSort === SEARCH_SORT_CHANNELS && subscribedChannels.length) ||
     (personalSort === SEARCH_SORT_YOU && !!tags.length) ||
     personalSort === SEARCH_SORT_ALL;
+  const hasMatureTags = tags.some(t => MATURE_TAGS.includes(t));
   const claimSearchCacheQuery = createNormalizedClaimSearchKey(options);
   const uris = (hasContent && claimSearchByQuery[claimSearchCacheQuery]) || [];
   const shouldPerformSearch =
@@ -237,6 +240,7 @@ function ClaimListDiscover(props: Props) {
           ))}
         </FormField>
       )}
+      {hasMatureTags && hiddenNsfwMessage}
     </Fragment>
   );
 
--- a/src/ui/component/common/hidden-nsfw.jsx
+++ b/src/ui/component/common/hidden-nsfw.jsx
@@ -1,0 +1,23 @@
+// @flow
+import React from 'react';
+import Button from 'component/button';
+import * as ICONS from 'constants/icons';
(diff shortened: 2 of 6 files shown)


src/ui/component/common/hidden-nsfw.jsx
```diff
@@ -0,0 +1,21 @@
...
+  type?: string,
+};
+
+export default function HiddenNsfw(props: Props) {
+  const { type = 'page' } = props;
+
+  return (
+    <div className="section--padded section__subtitle">
+      <Icon icon={ICONS.EYE_OFF} />
+      {__(' Content might be hidden on this ' + type + ' because of your ')}
```
seanyesmunt: Please change this to:

```
<I18nMessage tokens={{ type, settings: <Button button="link" label={__('settings')} href="/$/settings" /> }}>
  Content may be hidden on this %type% because of your %settings%
</I18nMessage>
```

src/ui/component/claimList/view.jsx
```diff
@@ -92,6 +95,7 @@ export default function ClaimList(props: Props) {
         <div className={classnames('claim-list__header', { 'claim-list__header--small': type === 'small' })}>
           {header}
           {loading && <Spinner type="small" />}
+          {obscureNsfw && <HiddenNsfw type="page" />}
```
seanyesmunt: I'm not sure this should live in this component. I don't think this should be shown on the home page or your list of publishes. Probably only in `page/tags/view.jsx`, maybe we could pass a prop down from there?

I like the styling on this though!

src/ui/component/common/hidden-nsfw.jsx
tombiju: I updated this 👍

src/ui/component/claimList/view.jsx
tombiju: Moved it to the tags view! Also I edited the icons .scss to give a little bit of space between the icon and the message which are mashed together now that its in the I18nMessage component.

tzarebczan:
@tombiju thanks for the PR, your first! 

kauffj:
@seanyesmunt @tombiju maybe we could brainstorm some other ways of displaying this? I think adding this is a definitely a valuable addition, but I think it is noisy to see this frequently if you are actively and purposefully blocking channels. 

I would be inclined to put it at the bottom of the results, but with infinite scroll this is tricky.

If we cannot come up with a solution other than putting it in the top, I think we should reduce it to an icon with text on mouseover and/or make it collapsable to an icon so you don't have to keep seeing it.

seanyesmunt:
@kauffj This is only for viewing mature tag pages with the setting off. Blocked channels are completely hidden from the UI.

@tombiju This is closer but we should only show this setting if you are on the tag page and it's a mature tag. We have a list of the current tags we track here:
https://github.com/lbryio/lbry-redux/blob/master/src/constants/tags.js#L16

src/ui/component/claimListDiscover/view.jsx
```diff
@@ -109,6 +111,7 @@ function ClaimListDiscover(props: Props) {
     (personalSort === SEARCH_SORT_CHANNELS && subscribedChannels.length) ||
     (personalSort === SEARCH_SORT_YOU && !!tags.length) ||
     personalSort === SEARCH_SORT_ALL;
+  const hasMatureTags = tags.some(t => MATURE_TAGS.includes(t));
```
tombiju: @seanyesmunt @kauffj Added this in to see if the tag is mature, so the message will be shown accordingly and it shouldn't be too obtrusive anymore!

PR #3044, opened 2019-10-14
provides recheck button on verification screen

jessopb:
After receiving verification, user can click the "check again" button to continue doing what they were doing instead of reloading the app.


src/ui/component/userVerify/view.jsx
```diff
@@ -25,15 +26,16 @@ class UserVerify extends React.PureComponent<Props> {
...
     return (
       <React.Fragment>
         <section className="section__header">
           <h1 className="section__title--large">{__('Extra Verification Needed')}</h1>
           <p>
             {__(
-              "We weren't able to auto-approve you for rewards. Please complete one of the steps below to unlock them."
+              "We weren't able to auto-approve you for rewards. Please complete one of the steps below to unlock them. Then you can"
             )}{' '}
+            <Button onClick={() => fetchUser()} button="link" label={__('check again.')} /> {'Or'}{' '}
```
seanyesmunt: I think this can just say "Refresh". Then we don't need the change to the other text.

tzarebczan:
We should add it to this screen too: https://github.com/lbryio/lbry-desktop/blob/master/src/ui/page/rewards/view.jsx#L57

jessopb:
Added Tom's change.
I find the bottom table jumping up on refresh jarring now, though.

seanyesmunt:
@jessopb I'd be fine with that. I've noticed it can be a little jumpy so anything to improve that would be great

PR #3055, opened 2019-10-15
adds requiresAuth bool to button and enables it on selected buttons

jessopb:


jessopb:
> This looks good. Can you just change these to only be required when `IS_WEB === true`?

Done. Tested on tv and app.

PR #3058, opened 2019-10-16
Sync for everyone

seanyesmunt:
- Removes any balance related checks from the sync flow
- Adds sync toggle to settings page
  - Properly navigates to the auth page and redirects back if a password is required to sync
- Cleans up email entry screen


src/ui/component/splash/view.jsx
```diff
@@ -118,27 +118,29 @@ export default class SplashScreen extends React.PureComponent<Props, State> {
...
-      // If we cleared the error timeout due to a wallet being locked, make sure to start it back up
-      if (!this.timeout) {
-        this.adjustErrorTimeout();
-      }
-
-      Lbry.resolve({ urls: 'lbry://one' }).then(() => {
-        this.setState({ isRunning: true }, () => this.continueAppLaunch());
+      Lbry.wallet_status().then(walletStatus => {
+        if (walletStatus.is_locked) {
+          // Make sure there isn't another active modal (like INCOMPATIBLE_DAEMON)
```
kauffj: this is a fine solution but I think ideally this would instead be in the style of `main--auth-page` rather than a modal

PR #3070, opened 2019-10-17
WIP: wallet encrypt and decrypt with sync

tzarebczan:
There's a flow error you can hopefully fix up.

There seems to be an odd case when you have a synced wallet, then turn off sync, and then change the password. When you go back to turn it on, it will prompt for the old password and everything gets set back there. I'm thinking we ask to enable sync on the wallet encrypt page also.


seanyesmunt:
Closing this to avoid clutter. Will re-open when we start on this again.

PR #3092, opened 2019-10-20
fixes scrolling with long sidebars

jessopb:

- [x] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number:

You'd have to scroll to bottom of sidebar to load more content.
## What is the new behavior?
On scroll, fills content as long as main_wrapper class has significantly more offsetHeight than main class (100px)


src/ui/component/claimList/view.jsx
```diff
@@ -10,6 +11,7 @@ import usePersistedState from 'effects/use-persisted-state';
 
 const SORT_NEW = 'new';
 const SORT_OLD = 'old';
+const PADDING_ALLOWANCE = 100;
```
kauffj: is this value dependent on other elements on the page, or is this just used as a somewhat arbitrary threshold to initiate the scroll bottom behavior?

if dependent on elements on the page, please calculate their size dynamically

src/ui/component/claimList/view.jsx
```diff
@@ -61,8 +63,14 @@ export default function ClaimList(props: Props) {
...
         const x = document.querySelector(`.${MAIN_WRAPPER_CLASS}`);
+        const mc = document.querySelector(`.${MAIN_CLASS}`);
 
-        if (x && window.scrollY + window.innerHeight >= x.offsetHeight) {
+        if (
+          x &&
+          mc &&
+          (window.scrollY + window.innerHeight >= x.offsetHeight ||
+            x.offsetHeight - mc.offsetHeight > PADDING_ALLOWANCE)
+        ) {
```
kauffj: this looks like it can be combined with the inner `if` statement

src/ui/component/claimList/view.jsx
jessopb: @kauffj It looks like '.main-wrapper' is 28 top and bottom padding (+56) padding larger than '.main'. I could tighten that up if it's important but I don't trust it to be exactly 56 because some elements cause there to be decimals (64.39 or 64.40 for the header, not sure why that changes).

jessopb:
Will merge.

tzarebczan:
Tested and looks good to me. Eventually we may want to disconnect the sidebar/search results from each other...I think it's strange that you need to scroll back up to see the top of your sidebar, or that you need to scroll down search results to see the bottom.

PR #3105, opened 2019-10-22
Sync on lbry.tv

seanyesmunt:
Still requires more testing I would say. It is currently deployed to `dev.lbry.tv`.

### DIFF
--- a/src/ui/util/lbrytv.js
+++ b/src/ui/util/lbrytv.js
@@ -1,7 +1,5 @@
-const { LBRY_TV_API } = require('../../../config');
-
 function generateStreamUrl(claimName, claimId) {
-  const prefix = process.env.SDK_API_URL || LBRY_TV_API;
+  const prefix = process.env.SDK_API_URL;
   return `${prefix}/content/claims/${claimName}/${claimId}/stream`;
 }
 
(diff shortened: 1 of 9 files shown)


seanyesmunt:
@tzarebczan Can you give this a try?

src/ui/util/saved-passwords.js
```diff
@@ -37,16 +37,23 @@ function deleteCookie(name: string) {
...
     const password = value === undefined || value === null ? '' : value;
     sessionPassword = password;
 
     if (saveToDisk) {
       if (password) {
+        // @if TARGET='app'
         ipcRenderer.send('set-password', password);
+        // @endif
+        // @if TARGET='web'
+        setCookie('saved-password', password, 14);
```
kauffj: I should have caught this in an earlier review, but it was only reading this that triggered me. A lot of this file speaks generically of "saved password" or "saved passwords", but in reality this file only allows for the setting of a single password and a very specific one at that: your wallet password.

src/ui/util/lbrytv.js
```diff
@@ -1,5 +1,8 @@
+const { LBRY_TV_API } = require('../../../config');
+
 function generateStreamUrl(claimName, claimId) {
-  return `https://api.lbry.tv/content/claims/${claimName}/${claimId}/stream`;
+  const prefix = process.env.SDK_API_URL || LBRY_TV_API;
```
kauffj: What's the situation where setting `process.env.SDK_API_URL` is insufficient?

src/ui/util/saved-passwords.js
seanyesmunt: Also auth_token

Issue #3111, opened 2019-10-23
lbry.tv megaissue take 2


tzarebczan:
- [ ] hide Sync setting on Settings page

- [ ] fix check on daemonsettings for rewards page (can't verify after skipping initially, screen is greyed out). 

- [ ] Drive sign-in on Tips + Comments

- [ ] We can customize tags on homepage - this should be hidden right? 

- [ ] Need to click sign out twice sometimes for it to stick

PR #3129, opened 2019-10-28
lbry.tv changes in response to ui feedback

seanyesmunt:
Also fixed up a number of issues from #3111


src/ui/component/header/view.jsx
```diff
@@ -37,10 +37,12 @@ const Header = (props: Props) => {
     automaticDarkModeEnabled,
     hideBalance,
     email,
-    minimal,
+    authHeader,
     signOut,
   } = props;
   const authenticated = Boolean(email);
+  const homeButtonNavigationProps = authHeader ? { onClick: signOut } : { navigate: '/' };
+  const closeButtonNavigationProps = authHeader ? { onClick: signOut } : { onClick: () => history.goBack() };
```
seanyesmunt: I think it makes more sense to just log out and stay on the auth page, than redirect back to the page they were on. I think in most cases they will want to log back in, and when they do, the redirect will still exist to go back to the page they were on.

kauffj: Agreed that back is not correct for sign out. I think homepage is most common behavior on sign out.

src/ui/component/syncPassword/view.jsx
```diff
@@ -46,7 +48,18 @@ function SyncPassword(props: Props) {
             />
             <div className="card__actions">
               <Button type="submit" button="primary" label={__('Continue')} disabled={getSyncIsPending} />
+              <Button button="link" label={__('Cancel')} onClick={signOut} />
             </div>
+            <p className="help">
+              <I18nMessage
+                tokens={{
+                  help: <Button button="link" label={__('Help Guide')} />,
```
kauffj: lowercase

kauffj: and missing link...

src/ui/component/userEmailNew/view.jsx
```diff
@@ -26,20 +25,15 @@ function UserEmailNew(props: Props) {
...
-
-    // @if TARGET='web'
-    Lbryio.call('user_tag', 'edit', { add: 'lbrytv' });
-    // @endif
   }
 
   return (
     <React.Fragment>
-      <h1 className="section__title--large">{__('Welcome To LBRY')}</h1>
+      <h1 className="section__title--large">{__('Sign In To LBRY')}</h1>
```
kauffj: Sign in to LBRY or Sign In to LBRY

Typically in would not be capitalized but since it is the proper name of a page, it arguably could be and may look more natural

PR #3138, opened 2019-10-29
copyable embed tag

jessopb:
EmbedArea similar to copyable text.
Needs styling.


src/ui/component/embedArea/view.jsx
```diff
@@ -0,0 +1,63 @@
...
+        label={label}
+        value={embedText || ''}
+        ref={input}
+        readOnly
+        onFocus={onFocus}
+      />
+      <div className="card__actions card__actions--center">
+        <Button
+          icon={ICONS.COPY}
+          button="inverse"
```
seanyesmunt: This can just be `button="link"` to look like the facebook/twitter links above

jessopb: done

jessopb:
> This looks good. It doesn't seem to actually copy the link when I click the copy button though.

Ah, fixed that.

jessopb:
Still needs a solution for the best way to give it the proper input box border/focus behavior.

PR #3140, opened 2019-10-29
fixes flow issue in sidebar by dynamically building link key

j-collier:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation changes
- [x] Other - Please describe: Flow type correction 

## Fixes

Issue Number: #2024

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/ui/component/sideBar/view.jsx:71:22

Cannot create li element because in property key:
 • Either React.Element [1] is incompatible with string [2].
 • Or React.Element [1] is incompatible with number [3].

        src/ui/component/sideBar/view.jsx
    [1]  60│                   <>
         61│                     {__('Publishes')}
         62│                     <Spinner type="small" />
         63│                   </>
           :
         68│               ),
         69│             },
         70│           ].map(linkProps => (
         71│             <li key={linkProps.label}>
         72│               <Button {...linkProps} className="navigation-link" activeClass="navigation-link--active" />
         73│             </li>
         74│           ))}
         75│         </ul>
         76│

        /private/tmp/flow/flowlib_1ec7e4cc/react.js
 [2][3] 188│ declare type React$Key = string | number;
```

## What is the new behavior?

Does not pass a span as `li` key, which is the correct behavior for a react element

### DIFF
--- a/src/ui/component/sideBar/view.jsx
+++ b/src/ui/component/sideBar/view.jsx
@@ -68,7 +68,7 @@ function SideBar(props: Props) {
               ),
             },
           ].map(linkProps => (
-            <li key={linkProps.label}>
+            <li key={linkProps.navigate}>
               <Button {...linkProps} className="navigation-link" activeClass="navigation-link--active" />
             </li>
           ))}


tzarebczan:
Thanks for the PR! As last time, can we show you some [appreciation](https://lbry.com/faq/appreciation) once this is reviewed/merged?

src/ui/component/sideBar/view.jsx
```diff
@@ -68,7 +76,7 @@ function SideBar(props: Props) {
               ),
             },
           ].map(linkProps => (
-            <li key={linkProps.label}>
+            <li key={buildKey(linkProps.label)}>
```
seanyesmunt: We should avoid using `random` values for keys in React components. Each time it's rendered it will have a different key, which will cause issues if we ever add options to change the order of the menu.

I think we can just change this to `linkProps.navigate`?

j-collier: That's a great point. I'll make this change

PR #3146, opened 2019-10-30
fix(tagsSearch): Undefined tagsPassedIn prop prevents page from loading

hackily:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: N/A (Should I create an issue?)

If no tags are passed in, then the prop is undefined and an exception causes the entire page to error out, since a map operation cannot be done on an undefined variable.

## What is the new behavior?
Undefined prop defaults to empty array, and the page loads successfully.


seanyesmunt:
Thanks!

PR #3168, opened 2019-11-03
[WIP] auto launch setting

jessopb:
Adding autolaunch setting
Has a strange bug where reloading the page always enables autolaunch, but disabling and closing then starting the app allows it to stay disabled.


seanyesmunt:
@jessopb How difficult is it to auto launch in a minimized state? (not saying we should necessarily, just curious)

seanyesmunt:
Closing in favor of #3210

Issue #3150, opened 2019-10-30
Improve lbry.tv analytics tracking


kauffj:
Ensure lbry.tv traffic can be separated out from referrals that come from other community domains.

PR #3172, opened 2019-11-04
add analytics event for lbry:// links

seanyesmunt:


src/ui/index.jsx
```diff
@@ -141,6 +141,7 @@ ipcRenderer.on('open-uri-requested', (event, uri, newSession) => {
     } else if (isURIValid(uri)) {
       const formattedUri = formatLbryUriForWeb(uri);
       app.store.dispatch(push(formattedUri));
+      analytics.openUrlEvent(formattedUri);
```
seanyesmunt: @kauffj Not sure if this should be `formattedUri` which is something like `/one`, or `uri` which is `Lbry://one` (GA makes the first letter uppercase)

kauffj: Ambivalent, `formattedUri` is probably fine.

seanyesmunt:
Also added #3150 since it was quick

PR #3173, opened 2019-11-04
adds analytics to download button

jessopb:


src/ui/component/fileDownloadLink/view.jsx
```diff
@@ -28,6 +30,11 @@ function FileDownloadLink(props: Props) {
     return <span>{label}</span>;
   }
 
+  if (fileInfo && fileInfo.download_path && fileInfo.completed && clicked) {
+    triggerAnalyticsView(uri, 0);
```
seanyesmunt: I'm not sure if we should wait for the download to finish to fire this. If they leave the page before it downloads, it will never fire.

jessopb: Good point. Removed fileInfo.completed.

seanyesmunt: If we aren't tracking that, it can just be fired inside of the `onClick`

jessopb: Yes, Assuming we don't care if unsuccessful downloads are counted as views.

seanyesmunt: We could pass a callback that fires the analytics event here. Then it will happen after the `get` call is successful
https://github.com/lbryio/lbry-desktop/blob/master/src/ui/redux/actions/content.js#L185

PR #3180, opened 2019-11-05
Load language from OS

kauffj:
This improves #3018 to allow `ui.js` to load immediately without waiting for language files to load.


src/ui/redux/actions/settings.js
```diff
@@ -82,3 +83,30 @@ export function doSetDarkTime(value, options) {
...
   };
 }
+
+export function doSetLanguage(language) {
+  return dispatch => {
+    // this should match the behavior/logic in index-web.html
+    fetch('https://lbry.com/i18n/get/lbry-desktop/app-strings/' + language + '.json')
+      .then(r => r.json())
+      .then(j => {
+        window.i18n_messages[language] = j;
```
kauffj: @seanyesmunt is there a technique we can use to force/always trigger a complete re-render when `window.i18n_messages[language]` was not already set?

The only known time we are failing to re-render is when manually refreshing in app-mode.

seanyesmunt: You can set a value in redux here, then pass that value in as a prop to the top level `<App/>` component

src/ui/component/settingLanguage/view.jsx
```diff
@@ -3,44 +3,27 @@
 import React, { useState } from 'react';
 import { FormField } from 'component/common/form';
 import Spinner from 'component/spinner';
-import { SETTINGS } from 'lbry-redux';
+import SUPPORTED_LANGUAGES from '../../constants/supported_languages';
```
seanyesmunt: from 'constants/supported_languages'

PR #3197, opened 2019-11-07
refactor lbrytv web server

seanyesmunt:
#### Changes
- Forwards open.lbry.com to lbry.tv (retains claim/channel in url)
- Lbrytv is now mostly separated from the `lbry-desktop` repo

#### Todo
- [ ] embedd redirect
- [x] error handling


seanyesmunt:
Embeds are proving a little tricker than I thought. Everything else seems to work though so I think we can just merge this for now and finish up the rest in another PR

PR #3201, opened 2019-11-11
[WIP] feat: disable app updates by placing a file named upgradeDisabled in static

jessopb:


jessopb:
After initial review, we can ask @JoshStrobl to verify it accomplishes what he needs it to.

kauffj:
I thought when we discussed this the answer was to have `doCheckUpgradeAvailable` simply never detect upgrades when the flag/file was set. Did that approach not work out?

jessopb:
Settings page had a "Your app is up to date". I suppose if that's fine being untrue sometimes, we don't need the app state value.

jessopb:
#3206

JoshStrobl:
@kauffj Can confirm that shipping this file would be a perfect solution for disabling update checking, I can easily add supplementary files during packaging :+1:

kauffj:
Great! @JoshStrobl when you see this merged you should be able to modify your build/release process accordingly.

seanyesmunt:
Closing in favor of #3206

PR #3202, opened 2019-11-11
do not block text rendering while loading fonts

kauffj:
this was not tested, but may improve page paint speed a bit https://web.dev/font-display


kauffj:
https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display

Issue #3232, opened 2019-11-16
Confusing warning when deleting file but not abandoning claim in app 0.37.2-rc2


eggplantbren:
I started downloading a claim of mine, then changed my mind, so clicked the trash icon. I would expect that checking to delete the file but not abandon the claim would have the obvious behaviour that it has always had. However, it gave me this strange, concerning (and false?) message, so instead I used the command line to delete the file.

![image](https://user-images.githubusercontent.com/1578298/68985538-0de06c80-087c-11ea-9412-43fbff073aa2.png)

tzarebczan:
Thanks @eggplantbren! I noticed this recently too and have a fix for it.

tzarebczan:
Fixed in https://github.com/lbryio/lbry-desktop/pull/3234

PR #3234, opened 2019-11-17
Fix file delete messaging

tzarebczan:
Addresses https://github.com/lbryio/lbry-desktop/issues/3232


tzarebczan:
Can give that a shot...will keep it in mind for future prs.

PR #3236, opened 2019-11-18
autoLaunch on startup

jessopb:

### DIFF
--- a/electron/createWindow.js
+++ b/electron/createWindow.js
@@ -131,6 +131,7 @@ export default appState => {
     startMinimized ? window.hide() : window.show();
   });
 
+  // A backup incase https://github.com/electron/electron/issues/7779 happens
   window.webContents.once('dom-ready', () => {
     startMinimized && window.hide()
   })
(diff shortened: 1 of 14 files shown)


electron/createWindow.js
```diff
@@ -126,9 +128,13 @@ export default appState => {
   });
 
   window.once('ready-to-show', () => {
-    window.show();
+    startMinimized ? window.hide() : window.show();
   });
 
+  window.webContents.once('dom-ready', () => {
+    startMinimized && window.hide()
```
seanyesmunt: Is this needed? I would imagine once the `dom-ready` event is fired the window should already be hidden.

jessopb: It seemed like this is the one that makes it work on windows -reliably-.

seanyesmunt: Should this line be just removed https://github.com/lbryio/lbry-desktop/pull/3236/files#diff-dd110088e25f2da03c36e833bd1295a8R131

And placed here instead?

PR #3246, opened 2019-11-19
Skip captcha for previously signed in users

seanyesmunt:
Also added a refresh message if the captcha hasn't loaded after 2 seconds


seanyesmunt:
these changes are included in #3248

PR #3254, opened 2019-11-20
use SameSite=Lax instead of Strict so cookies aren't lost when opening links

seanyesmunt:


seanyesmunt:
Merging this since it fixes the issue on Firefox

cc @kauffj

PR #3307, opened 2019-12-02
When this is merged lbry.tv and the desktop app will handle open links better

seanyesmunt:
#### Changes
- Handle open.lbry.com links in the lbry.tv server
  - Includes links to pages like open.lbry.com/?rewards&value=test
  - Will redirect to lbry.tv/$/rewards?src=open&value=test
. 
- Better handle lbry://?rewards links in the app
  - Can now handle any number of query params
.
- A few changes from `uri` to `url`


ui/util/url.js
```diff
@@ -0,0 +1,68 @@
...
+
+  return false;
+};
+
+/*
+  Function that handles page redirects
+  ex: lbry://?rewards
+  ex: open.lbry.com/?rewards
+*/
+exports.formatCustomUrl = path => {
```
seanyesmunt: This is the only new thing I added to this file other than renaming. I don't really like the name of the function though.

kauffj: Agreed. `formatAppUrl`? `formatInAppUrl`? `formatAppPageUrl`?

ui/util/url.js
```diff
@@ -0,0 +1,68 @@
...
+  // Determine if we need to add a leading "/$/" for app pages
+  const APP_PAGE_REGEX = /(\?)([a-z]*)(.*)/;
```
kauffj: Assuming it's supported here, use `\p{L}` (Unicode letter capture group) rather than `[a-z]`. I'd also include numbers.

ui/util/url.js
```diff
@@ -0,0 +1,68 @@
...
+  const appPageMatches = APP_PAGE_REGEX.exec(path);
+
+  if (appPageMatches && appPageMatches.length) {
+    // Definitely an app page (or it's formatted like one)
+    const [, , page, queryString] = appPageMatches;
+    let actualUrl = '/$/' + page;
+
+    if (queryString) {
+      if (queryString.startsWith('?')) {
```
kauffj: do we still end up in this function with URLs containing two question marks? is this for malformed URLs / just-in-case, or is this specifically needed?

ui/util/url.js
```diff
@@ -0,0 +1,68 @@
+const LBRY_INC_DOMAINS = ['lbry.io', 'lbry.com', 'lbry.tv', 'lbry.tech', 'lbry.fund', 'spee.ch'];
```
kauffj: Make this an ENV variable (or configurable/overrideable in some way, though it's good if this is the default without explicit configuration).

Call it `WEB_DOMAINS`

ui/util/url.js
```diff
@@ -0,0 +1,68 @@
...
+  let webUrl = path.replace(/\\/g, '/');
+
+  if (webUrl[0] !== '/') {
+    webUrl = `/${webUrl}`;
+  }
+
+  return encodeURI(`file://${webUrl}`).replace(/[?#]/g, encodeURIComponent);
+};
+
+exports.isLBRYDomain = uri => {
```
kauffj: `isWebDomain`

ui/util/url.js
seanyesmunt: This isn't just web domains. We use it for triggering a warning when opening links that _aren't_ from one of these domains. Maybe `TRUSTED_DOMAINS`?

ui/util/url.js
seanyesmunt: I like `formatInAppUrl`

ui/util/url.js
seanyesmunt: Do you have an example url structure in mind I can test?

ui/util/url.js
kauffj: I fixed this problem in #3312 :grin:

ui/util/url.js
kauffj: I'm possibly confused, but wouldn't this only execute if `formatInAppUrl` received a `path` that contained two `?`?

ui/util/url.js
seanyesmunt: @kauffj Wouldn't we only want letters here? When would there be an app page with non letters?

PR #3311, opened 2019-12-03
change file viewer background color

kauffj:
This makes @eggplantbren's article https://lbry.tv/@BrendonBrewer:3/trending:a readable from lbry.tv

While this fixes that case, it also makes me wonder why this rule existed as it did. Are there cases that `<DocumentViewer>` should have a dark background?


seanyesmunt:
@kauffj It should keep the dark background on dark mode, since the text is white. I pushed up a commit.

PR #3312, opened 2019-12-03
We're adults

kauffj:
- Few major sites do this, and our entire ethos is treating our users like adults.
- Malicious claims can still be flagged and filtered


kauffj:
@seanyesmunt I screwed up and included the same changes from #3311, but if you just merge #3311 first I believe it will come in correctly and cleanly.

PR #3315, opened 2019-12-03
update splash % call to use headers_synchronization_progress (for sdk v38+)

jessopb:


ui/component/splash/view.jsx
```diff
@@ -146,19 +146,10 @@ export default class SplashScreen extends React.PureComponent<Props, State> {
       });
 
       return;
-    } else if (blockchainHeaders) {
-      const blockChainHeaders = blockchainHeaders;
```
seanyesmunt: nice

PR #3321, opened 2019-12-05
add youtube welcome message

seanyesmunt:
copy TBD
https://lbryians.slack.com/archives/CMJQ4JB1N/p1575523422005200


ui/component/youtubeWelcome/view.jsx
```diff
@@ -0,0 +1,24 @@
+// @flow
```
kauffj: Not necessary to do this now, but it looks like it'd be pretty easy to make this a generic referrer component, where a domain maps to a title, message, and button label. @finer9 something to keep in mind for the future.

PR #3322, opened 2019-12-05
Add open in app link when coming from open.lbry.com

seanyesmunt:
Screen Shot 2019-12-05 at 11 52 06 AM

#### Changes
- Hides `src=open` from url
- Users coming from open.lbry.com will see this new prompt
- Only shown for users on a desktop computer or android phones
- Web only users will not see this message again when they close it.

#### Notes
I think this is just step 1 for this feature. Additional stuff could be remembering your choice and opening the app automatically, redirecting to the playstore/lbry.com/get if you don't have the app installed, but both of those seemed to have a few edge cases so I'd rather just get this in now to see if it does anything.

I didn't end up doing the mobile style since it would be thrown away and I think it stands out enough to be seen as is.

Also not sure what the copy should be.

### DIFF
--- a/ui/component/openInAppLink/view.jsx
+++ b/ui/component/openInAppLink/view.jsx
@@ -1,17 +1,16 @@
 // @flow
-import * as ICONS from 'constants/icons';
-import * as PAGES from 'constants/pages';
 import React from 'react';
-import Button from 'component/button';
 import { withRouter } from 'react-router';
 import userPersistedState from 'effects/use-persisted-state';
+import { formatWebUrlIntoLbryUrl } from 'util/url';
+import Nag from 'component/common/nag';
 
 const userAgent = navigator.userAgent.toLowerCase();
 const isAndroid = userAgent.includes('android');
 const isDesktop = typeof window.orientation === 'undefined';
 
 type Props = {
-  history: { replace: string => void },
+  history: { replace: string => void, push: string => void },
   location: { search: string, pathname: string },
   uri: string,
   user: ?User,
@@ -35,15 +34,7 @@ function OpenInAppLink(props: Props) {
   let appLink = uri;
 
   if (!appLink) {
-    // If there is no uri, the user is on an internal page
-    // pathname will either be "/" or "/$/{page}"
-    const path = pathname.startsWith('/$/') ? pathname.slice(3) : pathname.slice(1);
-    appLink = `lbry://?${path || PAGES.DISCOVER}`;
-
-    if (search) {
-      // We already have a leading "?" for the query param on internal pages
-      appLink += search.replace('?', '&');
-    }
+    appLink = formatWebUrlIntoLbryUrl(pathname, search);
   }
 
   function handleClose() {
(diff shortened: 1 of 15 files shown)


ui/component/openInAppLink/view.jsx
```diff
@@ -0,0 +1,81 @@
...
+import * as ICONS from 'constants/icons';
+import * as PAGES from 'constants/pages';
+import React from 'react';
+import Button from 'component/button';
+import { withRouter } from 'react-router';
+import userPersistedState from 'effects/use-persisted-state';
+
+const userAgent = navigator.userAgent.toLowerCase();
+const isAndroid = userAgent.includes('android');
+const isDesktop = typeof window.orientation === 'undefined';
```
kauffj: good to know javascript is still dumb

ui/component/openInAppLink/view.jsx
```diff
@@ -0,0 +1,81 @@
...
+    user && !user.device_types.some(usedDevice => usedDevice === 'mobile' || usedDevice === 'desktop');
+  const [hideNagForGood, setHideNagForGood] = userPersistedState('open-in-app-nag', false);
+  const [showNag, setShowNag] = React.useState(initialShouldShowNag);
+  let appLink = uri;
+
+  if (!appLink) {
+    // If there is no uri, the user is on an internal page
+    // pathname will either be "/" or "/$/{page}"
+    const path = pathname.startsWith('/$/') ? pathname.slice(3) : pathname.slice(1);
+    appLink = `lbry://?${path || PAGES.DISCOVER}`;
```
kauffj: I think this should probably be a function for converting a web url and/or web path into a LBRY url in https://github.com/lbryio/lbry-desktop/blob/master/ui/util/url.js

ui/component/openInAppLink/view.jsx
```diff
@@ -0,0 +1,81 @@
...
+      replace(newUrl);
+    }
+  }, [search, pathname, replace]);
+
+  if (!showNag || hideNagForGood) {
+    return null;
+  }
+
+  return (
+    <div className="nag">
```
kauffj: I'm concerned that this introduces a new element. Is this the _only_ time we're going to nag? When else will we nag? Should this be called something else?

I looked at Material, which we're still relatively close to, and nothing quite fits. [Banners](https://material.io/components/banners/) are kind of similar, but primarily used for problem/"must address" states, which I'm not sure this is.

seanyesmunt: I'm not sure. I mainly called it that because that's what discord calls theirs internally. I can potentially see us adding more nags for something like "update your app".

seanyesmunt:
@kauffj re-requesting review after moving the upgrade prompt to a `<Nag />` too.

ui/component/openInAppLink/view.jsx
```diff
@@ -0,0 +1,66 @@
...
+      replace(newUrl);
+    }
+  }, [search, pathname, replace]);
+
+  if (!showNag || hideNagForGood) {
+    return null;
+  }
+
+  return (
+    <Nag message={__('Want even more freedom?')} actionText={__('Use The App')} href={appLink} onClose={handleClose} />
```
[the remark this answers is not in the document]
kauffj: Or "Use the app" if we're not generally capitalizing, the other button says "Install now", so "Use the app" would match that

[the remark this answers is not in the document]
seanyesmunt: Good catch

PR #3305, opened 2019-11-30
Create channels cleanly & adds icons to claim-list alt-controls

osilkin98:
## Added a 'New Channel' Button

image

## Created a dedicated Channel Creation Page
image

PR #3329, opened 2019-12-06
Create channels cleanly (finishing up #3305)

seanyesmunt:


tzarebczan:
I think the options here should match the channel edit page, otherwise, users need to go edit the channel anyway.

osilkin98:
So you want to wait until those two things have been put together?

seanyesmunt:
That will require significant work, and can be added later.

PR #3353, opened 2019-12-10
When this is merged, dates on claim previews will be the same as youtube

seanyesmunt:
Tested a few channels and it seems like all the dates are the same as they are listed on youtube.


seanyesmunt:
I made it so it rounds down for months and days too. The only difference with this and youtube is that they show weeks, but I think I prefer seeing the number of days until it switches over to "1 month"

PR #3357, opened 2019-12-10
When this is merged free text files will be autoplayed

seanyesmunt:


ui/component/walletBackup/view.jsx
```diff
@@ -81,7 +81,7 @@ class WalletBackup extends React.PureComponent<Props, State> {
       return;
     }
 
-    this.showSuccessMessage(__('Saved zip archive to ' + outputPath));
+    this.showSuccessMessage(__('Saved zip archive to %outputPath%', { outputPath }));
```
kauffj: guess this one never happened much

ui/redux/selectors/content.js
```diff
@@ -117,3 +118,12 @@ export const makeSelectShouldObscurePreview = (uri: string) =>
       return isClaimMature && !showMatureContent;
     }
   );
+
+export const makeSelectCanAutoplay = (uri: string) =>
+  createSelector(
+    makeSelectMediaTypeForUri(uri),
+    mediaType => {
+      const canAutoPlay = ['audio', 'video', 'image', 'text', 'document'].includes(mediaType);
```
kauffj: we'll regret the lack of constants for these one day, but that day is not today

PR #3358, opened 2019-12-10
When this is merged lbry.tv will have better static asset caching

seanyesmunt:


lbrytv/middleware/cache-control.js
```diff
@@ -0,0 +1,15 @@
+const THREE_MONTHS_IN_SECONDS = 7776000;
+
+async function redirectMiddleware(ctx, next) {
+  const {
+    request: { url },
+  } = ctx;
+
+  if (url.includes('.png') || url.includes('font.css') || url.includes('.woff')) {
```
kauffj: This check should be stronger so that it only catches the website assets. Remember that `file.png` is a prefectly valid claim name.

If we're not moving assets to a CDN off of the webserver, let's put all assets in a clearly named path so we can perform a strong check.

kauffj: Also, if asset names are unique to the asset (i.e. rather than font.css, we call it font-v5.css, or font-abc123hash.css), then the cache can be extremely long (essentially indefinite).

static/index-web.html
```diff
@@ -3,6 +3,9 @@
   <head>
     <meta charset="utf-8" />
     <script src="/ui.js" async></script>
+    <link rel="icon" type="image/png" href="favicon.png" />
+    <link rel="stylesheet" href="/font/font.css" />
```
kauffj: Add the `<link rel="preload"...>` for the font files themselves.

kauffj: And then if you're doing that, the font CSS doesn't need to be loaded independently, right?

seanyesmunt: I think the font css still needs to be loaded because it setups up the `@font-face` properties. But it should be fine with just the `preload` value

> his example is trivial, as the browser probably discovers the <link rel="stylesheet"> and <script> elements in the same chunk of HTML as the preloads, but the benefits can be seen much more clearly the later resources are discovered and the larger they are. For example:
>
> Resources that are pointed to from inside CSS, like fonts or images.
https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content

seanyesmunt:
@kauffj pushed up another commit. I'll change the electron html if you think this looks good.

The main change is that now all images/fonts/js are loaded through `public/` `public/img` and `public/font`. And I hardcoded the list of every asset we serve in the cache-control middleware.

static/index-web.html
```diff
@@ -2,7 +2,66 @@
 <html>
   <head>
     <meta charset="utf-8" />
-    <script src="/ui.js" async></script>
+    <script src="/public/ui.js" async></script>
+    <link rel="icon" type="image/png" href="/public/favicon.png" />
```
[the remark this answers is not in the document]
kauffj: (presumably if we want to experiment, this can be pushed and we can check in analytics for page requests, so if you deploy this please set a note or reminder to check analytics a few days after)

kauffj:
@seanyesmunt I approved this because it has no issues to block merging, but please at least address the ui.js issue before moving past asset optimizations.

static/index-web.html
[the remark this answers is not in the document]
seanyesmunt: It looks like it's just needed for IE. I can add an additional icon for that.

PR #3359, opened 2019-12-11
[WIP] Include tip with comment

osilkin98:
Includes an option to add a tip to a comment before posting 
Also removes 'Comments' section title to make the design look neater

- converse ability in the tipping 
- tips are coupled with comments


tzarebczan:
Planning to add some indication to the comment that it included a tip?

kauffj:
Let :ship:

seanyesmunt:
@osilkin98 is this ready for review? Or does it require more work?

osilkin98:
This needs SDK-side work to support the desired functionality.

seanyesmunt:
Why can’t we just send a tip after commenting? The functionality can be separated even if it’s bundled in the ui

osilkin98:
Jeremy wanted them to be coupled, though I would be perfectly fine with
that solution for the time being.

seanyesmunt:
is that something you plan to work on soon sdk side?

osilkin98:
Yes

kauffj:
There may have been some confusion in what I meant by coupling. It needs to be presented to user as a unitary function and _ideally_ both tip and comment would happen atomically.

However, I think it would be okay to send the tip, let that succeed, and then submit the comment.

Issue #3374, opened 2019-12-13
video length and tag cut off


tzarebczan:
lbry://@FredBouchal#e

![image (19)](https://user-images.githubusercontent.com/8120721/70838271-76682e00-1dd5-11ea-868f-548204386be8.png)

myungwoo-Y:
Hi, I interested in this issue. 
Can I work on this issue?

tzarebczan:
Sure, take a look! Also, see https://lbry.tech/contribute

Ruk33:
Hello, I would like to help in this issue although it seems it has already been solved. The duration of the video now appears on the video thumbnail, is this correct?:

![image](https://user-images.githubusercontent.com/1719111/90048389-d6d3a600-dca9-11ea-82c1-3c29a4d8feec.png)

Regards.

PR #3376, opened 2019-12-15
fix video length and tag cut off 

myungwoo-Y:
## PR Checklist

- [x] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number:  #3374

When the screen is reduced, the playing time is gone.
Containers with playing time have a width without considering the image next to it.

ezgif com-video-to-gif

## What is the new behavior?

Container considered the image width for the maximum width.

ezgif com-video-to-gif (1)

## Other information

The issue of tags has not yet been resolved. I'll try to solve it.

### DIFF
--- a/ui/scss/component/_claim-list.scss
+++ b/ui/scss/component/_claim-list.scss
@@ -133,6 +133,7 @@
   display: flex;
   flex-direction: column;
   flex: 1;
+  width: calc(100% - 12rem);
 }
 
 .claim-preview-info {


ui/scss/component/_claim-list.scss
```diff
@@ -133,6 +133,7 @@
   display: flex;
   flex-direction: column;
   flex: 1;
+  width: calc(100% - 12rem);
```
seanyesmunt: Thanks for the PR! Where does the 12rem come from?

myungwoo-Y: Hi! 14rem is the fixed size of the thumbnail defined in _claim-list.scss.

https://github.com/lbryio/lbry-desktop/blob/b50ec8fab1ba0c3d7b09cb7fc8272bf41130b178/ui/scss/component/_claim-list.scss#L81-L83

If I subtract 14 rem, the space was too big, so I subtract 12rem.

seanyesmunt: I think this should be possible without hardcoding some limit. That could cause issues when we add a better mobile view, or for larger claim previews.

myungwoo-Y: I think I need to use javascript to avoid hardcoding. Do you think it is a good idea?

seanyesmunt: Not necessarily. Maybe check how much space the tags take up and then see if we should only render 3 tags instead of 4? (Or something like that).

It might be too much for the homepage to handle when you scroll down a lot and have a ton of tags on the page, but I'm all for trying it!

kauffj:
@myungwoo-Y please re-open if you decide to work on this

you can also join our [chat](https://chat.lbry.com) to talk to other devs in real time

PR #3386, opened 2019-12-16
correct and suggest uris for publish on search

jessopb:


ui/page/search/view.jsx
```diff
@@ -40,14 +44,16 @@ export default function SearchPage(props: Props) {
...
-            {isValid && (
-              <header className="search__header">
-                <ClaimUri uri={uri} />
-                <div className="card">
-                  <ClaimPreview uri={uri} type="large" placeholder="publish" />
-                </div>
-              </header>
-            )}
+            <header className="search__header">
+              <ClaimUri uri={Boolean(isModifiedUriValid) && normalizedModifiedUri} />
```
kauffj: I assume this isn't rendered if it's false? might be slightly cleaner/obvious to do

`{ isModifiedUriValid && <ClaimUri uri={normalizedModifiedUri} /> }`

PR #3412, opened 2019-12-20
Fix search suggested publish

jessopb:


ui/component/claimPreview/view.jsx
```diff
@@ -175,6 +175,10 @@ const ClaimPreview = forwardRef<any, {}>((props: Props, ref: any) => {
     );
   }
 
+  if (placeholder === 'publish' && !claim && uri.startsWith('lbry://@')) {
+    return null;
```
seanyesmunt: We could still render the preview and open the channel create modal when they click on it.

PR #3432, opened 2020-01-02
Refactor navigation UI

seanyesmunt:
#### Notes
- A lot of the changes come from having to create several components
- ClaimListDiscover is a little simpler, and now requires more props to be passed in from wherever it's being used
- Will revert all the link changes back to `/` and get rid of the redirect on the server when we have an actual home page
- I didn't add the persistent filters on these new pages. That will take a little more work and I didn't want to make this PR even bigger.


ui/page/channelsFollowing/view.jsx
```diff
@@ -0,0 +1,69 @@
...
+    if (!hasSubsribedChannels) {
+      doFetchRecommendedSubscriptions();
+    }
+  }, [doFetchRecommendedSubscriptions, hasSubsribedChannels]);
+
+  return (
+    <Page>
+      {showTab ? (
+        <ClaimList
+          headerLabel={__('Find Channels To Follow')}
```
kauffj: Find Channels to Follow

ui/page/channelsFollowingManage/view.jsx
```diff
@@ -0,0 +1,55 @@
...
+    let url = `/$/${PAGES.CHANNELS_FOLLOWING_MANAGE}`;
+    if (!viewingSuggestedSubs) {
+      url += '?view=discover';
+    }
+
+    history.push(url);
+  }
+
+  useEffect(() => {
+    doFetchRecommendedSubscriptions();
```
kauffj: If you want to kill this and replace it with `claim_search` for top channels by bid or whatever I would be all for it.

lbrytv/middleware/redirect.js
```diff
@@ -13,6 +13,11 @@ async function redirectMiddleware(ctx, next) {
     return;
   }
 
+  if (path === '/') {
```
kauffj: conditional by user auth state until new homepage is done? this won't make much sense for unauth users

ui/analytics.js
```diff
@@ -46,7 +46,7 @@ type LogPublishParams = {
 let analyticsEnabled: boolean = true;
```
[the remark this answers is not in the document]
seanyesmunt: Ah that's a good idea.

lbrytv/middleware/redirect.js
seanyesmunt: That's harder to do unless we call internal-apis with the auth token. I figured we could still keep the url (even if it's a little confusing) but still show the signed out view

PR #3441, opened 2020-01-04
Fixes #3383 by formatting lbry links differently

### LINKED ISSUE #3383: Fix channel links in comments on lbry.tv
Comments that have a channel in them open the desktop app, instead of linking to that page on lbry.tv

dalhill:
## PR Checklist

  - **Read as: I clicked a bunch of links on the desktop and web apps, please review carefully**

## PR Type

- [X] Bugfix

## Fixes

Issue Number: #3383

Please see the image below as a reference for the inspection images to follow.
example

When using @(some_user) in a comment, both the desktop app and lbry.tv use the following style.
both-before

This works on the desktop app but prompts the lbry.tv user to open up the desktop app which is not the desired behavior.

## What is the new behavior?
The desktop app continues to open these links in the desktop app, though the link path is different.
desktop-app-after

The lbry.tv app now opens these links in the user's browser.
web-app-after

## Other information

I've got three concerns.
1. With the added code, I don't know if "SimpleLink" is fitting anymore.
2. I've got a hunch that there is a more elegant solution, but I'll post this for now to favor results over perfection. I welcome all feedback.
3. A separate issue exists. When @(some_user) is specified in a comment, it doesn't give the full path to the user. In my example above, @(mh) is trying to reply to the user @(OzFlor:367e8ca0bcd353126212947982ae0aee29a0d880). However, when using @(OzFlor), the user is routed to [OzFlor](https://lbry.tv/@OzFlor:7) which is a channel for a different user. In my opinion this is a separate issue from #3383 but I figured I'd bring it up here.

### DIFF
--- a/ui/component/common/markdown-preview-internal.jsx
+++ b/ui/component/common/markdown-preview-internal.jsx
@@ -36,7 +36,7 @@ const SimpleLink = (props: SimpleLinkProps) => {
   let { href } = props;
   if (href && href.startsWith('lbry://')) {
     href = formatLbryUrlForWeb(href);
-    // using NavLink after format to handle "/" vs "#/"
+    // using Link after formatLbryUrl to handle "/" vs "#/"
     // for web and desktop scenarios respectively
     return (
       <Link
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Thanks for the fix @dalhill!!

PR #3457, opened 2020-01-07
enable downloads on lbry.tv

seanyesmunt:
#### Todo
- [x] Trigger view event for download click
- [x] ~Fix html/pdf viewer~ api side fix needed

### DIFF
--- a/ui/component/fileDownloadLink/view.jsx
+++ b/ui/component/fileDownloadLink/view.jsx
@@ -17,13 +17,42 @@ type Props = {
   openModal: (id: string, { path: string }) => void,
   pause: () => void,
   download: string => void,
+  triggerViewEvent: string => void,
+  costInfo: ?{ cost: string },
 };
 
 function FileDownloadLink(props: Props) {
-  const { fileInfo, downloading, loading, openModal, pause, claimIsMine, download, uri, claim } = props;
+  const {
+    fileInfo,
+    downloading,
+    loading,
+    openModal,
+    pause,
+    claimIsMine,
+    download,
+    uri,
+    claim,
+    triggerViewEvent,
+    costInfo,
+  } = props;
+  const cost = costInfo ? Number(costInfo.cost) : 0;
+  const isPaidContent = cost > 0;
   const { name, claim_id: claimId, value } = claim;
   const fileName = value && value.source && value.source.name;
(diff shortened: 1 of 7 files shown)


ui/component/fileDownloadLink/view.jsx
```diff
@@ -15,10 +17,42 @@ type Props = {
...
+  function handleDownload() {
+    // @if TARGET='app'
+    download(uri);
+    // @endif;
+    // @if TARGET='web'
+    triggerViewEvent(uri);
+    // @endif;
+  }
+
+  if (IS_WEB && isPaidContent) {
```
kauffj: When do we `IS_WEB` and when do we `@if TARGET='web'`?

ui/component/fileDownloadLink/view.jsx
```diff
@@ -41,19 +75,21 @@ function FileDownloadLink(props: Props) {
...
+
+  return (
+    <ToolTip label={IS_WEB ? __('Download') : __('Add to your library')}>
+      <Button
+        button="alt"
+        icon={ICONS.DOWNLOAD}
+        onClick={handleDownload}
+        // @if TARGET='web'
+        download={fileName}
+        href={downloadUrl}
```
kauffj: I'd like `downloadUrl` to only get set on lbry.tv

kauffj: (Yes, I understand it isn't used)

kauffj:
@seanyesmunt nm, I was wrong in my other review comment. No blocking feedback.

ui/util/lbrytv.js
```diff
@@ -9,5 +9,10 @@ function generateEmbedUrl(claimName, claimId) {
   return `${URL}/$/embed/${claimName}/${claimId}`;
 }
 
+function generateDownloadUrl(claimName, claimId, apiUrl) {
+  const streamUrl = generateStreamUrl(claimName, claimId, apiUrl);
```
kauffj: Can these also get wrapped similar to embed links for posterity/readability? E.g. lbry.tv/$/download/${claimName}/${claimId}

seanyesmunt: Yeah. Good idea

PR #3458, opened 2020-01-07
Hide cursor on fullscreen

tuxfoo:
## PR Checklist

- [ x] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)
- [ x] I have checked that this PR does not introduce a breaking change

## PR Type

- [ x] Bugfix

When a video is fullscreen the cursor stays there. Also it changes to a hand.

## What is the new behavior?
The cursor is now hidden on full screen video playback as long as the mouse is inactive. The cursor only changes from a pointer to the hand when over the playback menu or when hovering over the big play button in center of screen.

### DIFF
--- a/ui/scss/component/_file-render.scss
+++ b/ui/scss/component/_file-render.scss
@@ -35,9 +35,13 @@
     height: 100%;
     object-fit: contain;
   }
-
   video {
     cursor: pointer;
+  }
+  .video-js.vjs-fullscreen.vjs-user-inactive.vjs-playing {
+    video {
+      cursor: none;
+    }
   }
 }
 


ui/scss/component/_file-render.scss
```diff
@@ -35,10 +35,6 @@
     height: 100%;
     object-fit: contain;
   }
-
-  video {
```
seanyesmunt: we want to keep this style for non-fullscreen. Maybe you could change it do `default` when it's full-screen?

tuxfoo: Changed cursor to none on the following conditions
```
.vjs-fullscreen.vjs-user-inactive.vjs-playing
```

tzarebczan:
Thanks for the PR @tuxfoo ! Can you squash your commits too? 

After that, can we show you some [appreciation](https://LBRY.com/faq/appreciation)?

tuxfoo:
> Thanks for the PR @tuxfoo ! Can you squash your commits too?
> 
> After that, can we show you some [appreciation](https://LBRY.com/faq/appreciation)?

squashed the commits :)

kauffj:
Great software comes from the accumulation of lots of improvements like this. Thank you @tuxfoo!

Issue #3348, opened 2019-12-09
Explicitly selecting new publish should clear all publish fields


kauffj:
If you start an edit and abandon it, state is still there when selecting new

dalhill:
I've got this one figured out. I'll open a review for it tomorrow evening.

PR #3463, opened 2020-01-09
Choosing "New Publish" Now Clears the Publish Form

dalhill:
## PR Checklist

## PR Type

- [X] Bugfix

## Fixes

Issue Number: #3348

Font size of cancel button set to `--font-small`.
cancel

The publish form is **not** cleared by the following actions.

1. User clicks the "New Publish" drop-down item from the Navigation bar.
publish-navbar

2. User clicks the "Publish something new" button when they go to their publishes and have not published previously.
publish-button

3. User clicks the "New Publish" link when they go to their publishes and have already published content.
publish-link

## What is the new behavior?
Font size of cancel button set to `--font-large`.
cancel-updated

1. Has changed to just "Publish" and continues its behavior of not clearing the form.
publish-navbar-updated

2. Now clears form on click.
3. Now clears form on click.

### DIFF
--- a/static/app-strings.json
+++ b/static/app-strings.json
@@ -910,5 +910,7 @@
   "Did something go wrong? Have a look in your log file, or send it to %support_link%.": "Did something go wrong? Have a look in your log file, or send it to %support_link%.",
   "%amount% LBC": "%amount% LBC",
   "%amount% fee": "%amount% fee",
-  "1 file hidden due to your %content_viewing_preferences_link%": "1 file hidden due to your %content_viewing_preferences_link%"
+  "1 file hidden due to your %content_viewing_preferences_link%": "1 file hidden due to your %content_viewing_preferences_link%",
+  "Thumbnail": "Thumbnail",
+  "spee.ch": "spee.ch"
 }
--- a/ui/component/header/view.jsx
+++ b/ui/component/header/view.jsx
@@ -147,7 +147,7 @@ const Header = (props: Props) => {
                   <MenuList className="menu__list--header">
                     <MenuItem className="menu__link" onSelect={() => history.push(`/$/${PAGES.PUBLISH}`)}>
                       <Icon aria-hidden icon={ICONS.PUBLISH} />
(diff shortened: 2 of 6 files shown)


tzarebczan:
Thanks for the PR @dalhill ! Can you squash your commits? We'll get a review on this soon..

After that, can we show you some [appreciation](https://LBRY.com/faq/appreciation)?

dalhill:
@tzarebczan I squashed the 3 most recent commits. I should have been more mindful of the commits. If you want the commits further squashed let me know and I'll see what I can do.

kauffj:
@seanyesmunt this code looks fine but there is a big UX question around when users will expect the publish to clear.

@dalhill part of the idea of this behavior was that you should be able to start a publish, navigate elsewhere, and then return. 

Two alternate ideas:
- Track the last time publish state changed and do not auto clear if it was updated recently
- Show a more prominent reset/clear option when resuming

kauffj:
@dalhill Sean is off work for the rest of this week so we may have to pick this up next week

dalhill:
@kauffj That makes sense. Out of the two options I like "Show a more prominent reset/clear option when resuming". I think that having a timer may give users the impression of an inconsistent app. Thanks for the response and I'll wait for @seanyesmunt to return.

seanyesmunt:
One option is to change the menu dropdown to just say `Publish`, and then reset the publish state for every button that specifically says `New Publish`

tzarebczan:
Thanks for the PR @dalhill ! Once this is reviewed/merged, can we show you some [appreciaiton](https://lbry.com/faq/appreciation)?

dalhill:
In response to feedback from @seanyesmunt
1. I changed the navbar dropdown to just say "Publish" and it continues to not clear the form. 
2. The other two button/links will clear the form as they still say "new". 

In response to feedback from @kauffj 
1. I increased the font size of the cancel button at the top of the publish form. It's a fairy subtle change and may not be enough.

Issue #3306, opened 2019-12-01
When editing your channel, clicking a subscribed channel results in a display error.


kcseb:
**Bug**
When you are editing your channel and click a channel that you are subscribed to (on the right-hand side), it will result in some wonky behaviour, displaying whatever channel you clicked on instead of yours at the top. 

**To Reproduce**
1. Go to a channel that you have claimed
2. Click on the edit button
3. Click on a channel you are subscribed to. 

**Expected behavior**
This should prompt you to either save your changes or discard them, OR just simply take you to the channel that you clicked. 

**Screenshots**
GIF Showing the Bug: https://i.imgur.com/aV04lG5.gif

**System Configuration**

- LBRY Daemon version: 0.46.1
- LBRY App version: 0.37.2
- LBRY Installation ID: 2TxtRu7JxCpVxNF572y4tir3kiiH3ni5djXhQo9CTzY2V7vAPB94cEd3cBeSaaKDJu
- Operating system: Windows 10

ProfessorDey:
Can confirm this does occur on my EndeavourOS Linux install (Arch derivative), kernel 5.3.12-arch1-1.

seanyesmunt:
Thank you for the solid bug report. Should be simple to fix.

PR #3470, opened 2020-01-10
Don't Show Edit Screen for Other Channels

dalhill:
## PR Checklist

## PR Type

- [X] Bugfix

## Fixes

Issue Number: #3306

When a user is editing their own channel and clicks on a channel they're subscribed to, it looks like they're editing the channel they don't own. This appears to only happen if the other channel is already "cached" (if user has already clicked on this channel during app execution).

**Editing own channel**
self-edit-screen

**"Editing" other user's channel**
editing-other-channel

## What is the new behavior?
Once you click on a channel you don't own, you'll be taken out of editing.

## Other information
@kcseb mentions the possibility of prompting a user to save or discard their changes. If this is preferred over just navigating to the selected channel (current behavior), I'd be happy to add this.

### DIFF
--- a/ui/page/channel/view.jsx
+++ b/ui/page/channel/view.jsx
@@ -161,6 +161,12 @@ function ChannelPage(props: Props) {
     fetchSubCount(claimId);
   }, [uri, fetchSubCount, claimId]);
 
+  React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Thanks for the PR! 🎆

Issue #3435, opened 2020-01-03
image rotated upside down, preview shows correctly


tzarebczan:
See https://lbry.tv/@LilCryptita:4/IMG_5860:c

The unfurl preview shows correctly, but thumb + image are upside down.

tuxfoo:
seems to not honor EXIF metadata.

PR #3476, opened 2020-01-12
[WIP] Img exif

tuxfoo:
## PR Checklist

## PR Type

- [ x] Bugfix

## Fixes

Issue Number: #3435

EXIF data is ignored, this leads to photos been shown with the wrong orientation. This can be a problem as people take photos with their smart phones upside down and sideways.

## What is the new behavior?
The orientation of the photo is honoured using EXIF data.

## Other information
At the time of writing this PR this patch only works for the preview, I still need to implement this for the thumbnails.
I thought I would submit this now because I was wondering if it was okay to use the `blueimp-load-image` library.
https://github.com/blueimp/JavaScript-Load-Image

In the future, it might be a good idea if the client could remove such meta data on upload. Especially since it can contain Geo data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "auto-launch": "^5.0.5",
+    "blueimp-load-image": "^2.26.0",
     "electron-dl": "^1.11.0",
     "electron-log": "^2.2.12",
     "electron-notarize": "^0.1.1",
--- a/ui/component/viewers/imageViewer.jsx
+++ b/ui/component/viewers/imageViewer.jsx
@@ -1,17 +1,30 @@
 // @flow
 import React from 'react';
+import loadImage from 'blueimp-load-image';
 
 type Props = {
   source: string,
 };
 
+function loadPic(source) {
+  loadImage(
+    source,
+    function(img, data) {
+      if (img.type === 'error') {
+        console.error('Cannot load image');
+      } else {
+        document.getElementsByClassName('file-render__viewer')[0].appendChild(img);
+      }
+    },
+    {
+      orientation: true,
+    }
+  );
+}
+
 function ImageViewer(props: Props) {
   const { source } = props;
-  return (
-    <div className="file-render__viewer">
-      <img src={source} />
-    </div>
-  );
+  return <div className="file-render__viewer">{loadPic(source)}</div>;
 }
 
 export default ImageViewer;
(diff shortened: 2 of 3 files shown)


tzarebczan:
Thanks for the PR @tuxfoo ! We'll get a review on this soon. Once it's completed, can we show you some [appreciation](https://lbry.com/faq/appreciation) ?

tuxfoo:
> Thanks for the PR @tuxfoo ! We'll get a review on this soon. Once it's completed, can we show you some [appreciation](https://lbry.com/faq/appreciation) ?

So it is okay to use the blueimp image library? I will have to change img and background-image to canvas. I can attempt to apply the fix to the thumbnail on the claim-list as well.

ui/component/viewers/imageViewer.jsx
```diff
@@ -1,17 +1,30 @@
...
 };
 
+function loadPic(source) {
+  loadImage(
+    source,
+    function(img, data) {
+      if (img.type === 'error') {
+        console.error('Cannot load image');
+      } else {
+        document.getElementsByClassName('file-render__viewer')[0].appendChild(img);
```
seanyesmunt: You should move `loadPic` inside of the `ImageViewer` component so that it can use `refs` and dynamically inserting the image, instead of using `appendChild`

I think it's fine to use this library though.

package.json
```diff
@@ -44,6 +44,7 @@
   },
   "dependencies": {
     "auto-launch": "^5.0.5",
+    "blueimp-load-image": "^2.26.0",
```
seanyesmunt: This should be a `devDependency`

kauffj:
@tuxfoo come backkkkkkkkkkkkkk

jsigwart:
How do we keep from malicious EXIF data being uploaded? https://umbrella.cisco.com/blog/2019/07/24/picture-perfect-how-jpg-exif-data-hides-malware/

tuxfoo:
> @tuxfoo come backkkkkkkkkkkkkk
Sorry back to work this week. I need read up on using refs. 

> How do we keep from malicious EXIF data being uploaded? https://umbrella.cisco.com/blog/2019/07/24/picture-perfect-how-jpg-exif-data-hides-malware/

I think metadata of images should be removed on upload(after correcting the orientation).
Malicious metadata is not the only concern, images can contain geo data that could put the uploader at risk. 
However, I am not sure if it is possible to prevent people from uploading malicious content via the the command line.

kauffj:
@tuxfoo would still love to see this, but will probably close PR next week if no progress

malicious EXIF data can be considered out of scope here

PR #3485, opened 2020-01-13
add download link on screen for lbry.tv

seanyesmunt:


ui/component/fileViewerInitiator/view.jsx
```diff
@@ -48,20 +50,29 @@ export default function FileViewerInitiator(props: Props) {
...
-  function getMessage() {
-    if (IS_WEB && cost) {
-      return __('Paid Content Not Supported on lbry.tv');
+  function getTitle() {
+    let message = __('Unsupported File');
+    // @if TARGET='web'
+    if (cost) {
+      message = __('Paid Content Not Supported on lbry.tv');
+    } else {
+      message = __("We're not quite ready to display this file on lbry.tv yet");
```
seanyesmunt: @kauffj looking for copy suggestions. I like this, but not sure about it since other are shorter and in title case (which I think would be weird for this sentence)

ui/component/fileViewerInitiator/view.jsx
```diff
@@ -48,20 +50,29 @@ export default function FileViewerInitiator(props: Props) {
...
   const webStreamOnly = contentType === 'application/pdf' || mediaType === 'text';
   const supported = IS_WEB ? (!cost && isStreamable) || webStreamOnly || forceVideo : true;
+  const { name, claim_id: claimId, value } = claim;
+  const fileName = value && value.source && value.source.name;
+  const downloadUrl = `/$/download/${name}/${claimId}`;
 
-  function getMessage() {
-    if (IS_WEB && cost) {
-      return __('Paid Content Not Supported on lbry.tv');
+  function getTitle() {
```
kauffj: I think the messaging and presentation of this should switch more strongly. At this point if it can't be rendered there is a good chance downloading is simply intended.

Issue #3303, opened 2019-11-28
using arrowkeys when in profile content search give unexpected behaviour


ramonschepers:
**Bug**
so i found this around 12 hours in the desktop app, reported in the discord, but now i am on request posting it here.
bug description: when i use my arrow keys to move the cursor in a profile's searchbox for it's content, it swaps between the profile tabs instead, which to me should not be happening when one has the cursor active in a text input box (this case the profile searchbox)

**To Reproduce**
1) go to any user profile in desktop app
2) type something in the "content search" field for that user
3) use arrow keys to move the cursor left or right in the field
4) find out that it doesn't have the expected behaviour like i report now :)

**Expected behavior**
the profile tab switching shouldn't happen when using the arrow keys while one is typing in the profile content searchbox, and move the corsor porition.

**Screenshots**
none

**System Configuration**
app: 0.37.2
deamon: 0.46.1
os: Windows 10 build 10.0.18362-SP0
install id: (not really needed for this i think? (correct me if i am wrong!))

**Additional context**
thanks for making awesome happen.
ramon :)

tzarebczan:
Thanks for finding and filing this!

PR #3496, opened 2020-01-15
Navigate Channel Search with Arrow Keys

dalhill:
## PR Checklist

## PR Type

- [x] Bugfix

## Fixes

Issue Number: #3303

User cannot navigate text in a channel search field due to the left and right arrow keys changing tabs.

*Image displaying the search field*
search

## What is the new behavior?
User can navigate text in a channel search field using the left and right arrow keys.


seanyesmunt:
Nice one!

Issue #3490, opened 2020-01-14
Choosing Same File Twice in Publish Form


dalhill:
**Bug**
1. Unable to select the same file multiple times in a row from the publish form.
2. Unable to select the same thumbnail multiple times in a row from the publish form.

**To Reproduce**
This applies to both the upload file and the upload thumbnail
1. Navigate to publish form
2. Choose a file to publish
3. Clear the form
4. Choose the same file from step 2 (won't work)

**Expected behavior**
When following the steps above, a user should be able to successfully complete step 4.

**Screenshots**
![publish](https://user-images.githubusercontent.com/15717854/72317684-a66a5180-365f-11ea-8bdb-cebdda60d654.png)

**System Configuration**
![image](https://user-images.githubusercontent.com/15717854/72317601-6acf8780-365f-11ea-9e05-160ac2b1d25f.png)

tzarebczan:
Nice catch!

Yamboy1:
It looks like when you try to clear the file input with `clearPublish`, you don't actually clear the internal file input. If the file input is given the same input twice, it ignores the second one, and that's what is happening here

Yamboy1:
Should probably close this issue now @tzarebczan

tzarebczan:
Thanks !!

PR #3500, opened 2020-01-15
Clear the internal file selector after `clearPublish`

Yamboy1:
## PR Checklist

## PR Type

- [x] Bugfix

## Fixes

Issue Number: #3490 

If you select a file (on the publish screen), then clear the form and try to select it again, it won't select

## What is the new behavior?
You are able to select it on the second try now.

## Other information

I've checked that these changes don't break any other places that the FileSelector component is used.

### DIFF
--- a/ui/component/common/file-selector.jsx
+++ b/ui/component/common/file-selector.jsx
@@ -18,6 +18,14 @@ type Props = {
 };
 
 class FileSelector extends React.PureComponent<Props> {
+  componentDidUpdate(prevProps: Props) {
+    // If the form has just been cleared,
+    // clear the file input
+    if (prevProps.currentPath && !this.props.currentPath) {
+      this.fileInput.current.value = null;
+    }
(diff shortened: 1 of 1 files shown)


dalhill:
@Yamboy1 nice fix. There is also a similar problem with choosing the thumbnail if you follow these steps.

1. Select a thumbnail
2. Select "cancel" when the modal pops up (image below)
3. Select the same thumbnail from step 1

This leaves us in the state similar to the one you fixed for the file selection.

*Modal from step 2*
image

Yamboy1:
I'll see if I can add that in. Thanks @dalhill

Yamboy1:
In it's current state, I can't seem to find a way to do this that doesn't break the publish entirely, so I'm going to leave that out of this PR for now.

Feel free to make another issue from that tho

seanyesmunt:
Nice! 🙂

Issue #3507, opened 2020-01-17
Add "No Results" Message for Channel Search


dalhill:
**Bug**
Tell the user when there are no results within the channel search.

**To Reproduce**
1. Go to any channel
2. Search for something bogus, like "sjdflksjdfklsdfl"
3. Observe no content, and no message.

**Expected behavior**
Display a message to the user letting them know that there are no results.

**Screenshots**
![image](https://user-images.githubusercontent.com/15717854/72585246-eecd7d80-38b2-11ea-9a0a-01deaad66ddb.png)

**System Configuration**
![image](https://user-images.githubusercontent.com/15717854/72585310-3522dc80-38b3-11ea-8f9a-e1facb2ec385.png)

dalhill:
I've got the solution. Will open PR tomorrow.

PR #3513, opened 2020-01-19
Add "No Results" Message When Searching Channel Content

dalhill:
## PR Checklist

## PR Type
- [x] Bugfix

## Fixes
Issue Number: #3507

When no results are returned, the user just sees an empty section.
no-results-before

## What is the new behavior?
When no results are returned, the user will see a message saying "no results".
no-results

## Bonus
I noticed the styling of returned results was off when searching a channel. I dug around and saw that we could use ClaimList here, this fixes both the styling and the 'no results' issue.

*Current Search Styling*
results-no-style

*New Search Styling*
results-with-style2

### DIFF
--- a/ui/page/channel/view.jsx
+++ b/ui/page/channel/view.jsx
@@ -19,10 +19,10 @@ import * as ICONS from 'constants/icons';
 import classnames from 'classnames';
 import * as MODALS from 'constants/modal_types';
 import { Form, FormField } from 'component/common/form';
-import ClaimPreview from 'component/claimPreview';
 import Icon from 'component/common/icon';
 import HelpLink from 'component/common/help-link';
 import { DEBOUNCE_WAIT_DURATION_MS } from 'constants/search';
+import ClaimList from 'component/claimList';
 
 const PAGE_VIEW_QUERY = `view`;
 const ABOUT_PAGE = `about`;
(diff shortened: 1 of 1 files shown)


ui/page/channel/view.jsx
```diff
@@ -250,7 +250,15 @@ function ChannelPage(props: Props) {
...
           <TabPanel>
             {searchResults ? (
-              searchResults.map(url => <ClaimPreview key={url} uri={url} />)
+              <ClaimList
+                header={false}
+                headerAltControls={null}
+                id={`search-results-for-${claimId}`}
+                loading={false}
+                showHiddenByUser={false}
+                type={''}
```
dalhill: @seanyesmunt please review these parameters. For 'type' I saw that some options are 'small, 'large', 'inline', etc.. All of these options gave undesired results. What value should I use here if I just want the default behavior?

seanyesmunt: @dalhill you can just remove that line and it will use the default.

ui/page/channel/view.jsx
```diff
@@ -19,10 +19,10 @@ import * as ICONS from 'constants/icons';
 import classnames from 'classnames';
 import * as MODALS from 'constants/modal_types';
 import { Form, FormField } from 'component/common/form';
-import ClaimPreview from 'component/claimPreview';
 import Icon from 'component/common/icon';
 import HelpLink from 'component/common/help-link';
 import { DEBOUNCE_WAIT_DURATION_MS } from 'constants/search';
+import ClaimList from '../../component/claimList/view';
```
seanyesmunt: Please change this import to just use `'component/claimList'`

seanyesmunt:
Also thank you for fixing the styling! I have been meaning to do that.

PR #3519, opened 2020-01-21
fallback to keychain for auth_token if desktop users don't have an auth_token cookie

seanyesmunt:
I didn't realize older users were missing the auth_token cookie. So this just puts back old code, plus sets the auth token cookie so they will be set in the future.


seanyesmunt:
Confirmation from Tom that this fixes it. So i'm gonna merge.

PR #3532, opened 2020-01-23
loads script for alternate sidebar

jessopb:


ui/scss/component/_navigation.scss
```diff
@@ -15,20 +15,15 @@
 
 .navigation--placeholder {
   @extend .navigation;
-  padding: 2rem 1.5rem;
+  padding: var(--spacing-quarter) 0rem;
```
seanyesmunt: This can be removed since it isn't in a card anymore. The variable can be removed in `vars.scss` too

lbrytv/component/ad/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+          document.body.getElementsByTagName('style')[0].remove();
+      }
+    };
+  }, []);
+
+  return (
+    <>
+      <div><p>Ads</p></div>
+      <div>
+      <div id="rc-widget-0a74cf" data-rc-widget data-widget-host="habitat" data-endpoint="//trends.revcontent.com" data-widget-id="117427" />
```
seanyesmunt: The indentation of this file makes me think that your precommit hook didn't run. This should be indented

lbrytv/component/ad/index.jsx
```diff
@@ -0,0 +1,33 @@
+import React, { useEffect } from 'react';
```
seanyesmunt: I would rename this file to `lbrytv/component/ads.jsx`

PR #3534, opened 2020-01-23
send events when video buffers

jessopb:
will send the following:
sendGaEvent('Media', 'BufferTimestamp', claimId, currentTime); 
 * reports every buffering event and its timestamp

sendGaEvent('Media', 'Started', claimId, duration);
 * reports how delayed the start of a video is

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "electron-updater": "^4.1.2",
     "express": "^4.17.1",
     "if-env": "^1.0.4",
-    "keytar": "^4.4.1"
+    "keytar": "^4.4.1",
+    "videojs-event-tracking": "^1.0.1"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -51,8 +51,7 @@
     "electron-updater": "^4.1.2",
     "express": "^4.17.1",
     "if-env": "^1.0.4",
-    "keytar": "^4.4.1",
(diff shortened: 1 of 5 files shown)


jessopb:
It seems like it's best to use a plugin to identify buffering. 
You can listen for the waiting event, but that fires immediately after seek and play as well.

package.json
```diff
@@ -51,7 +51,8 @@
     "electron-updater": "^4.1.2",
     "express": "^4.17.1",
     "if-env": "^1.0.4",
-    "keytar": "^4.4.1"
+    "keytar": "^4.4.1",
+    "videojs-event-tracking": "^1.0.1"
```
seanyesmunt: This should be a `devDependency`

ui/component/viewers/videoViewer/view.jsx
```diff
@@ -185,6 +205,30 @@ function VideoViewer(props: Props) {
...
+    function doTrackingFirstPlay(e: Event, data: any) {
+      analytics.videoStartEvent(claimId, data.secondsToLoad);
+    }
+    function doError(e: Event) {
+      console.log('ERROR', e);
+    }
+    if (player) {
+      player.on('tracking:buffered', (e, d) => doTrackingBuffered(e, d));
+      player.on('tracking:firstplay', (e, d) => doTrackingFirstPlay(e, d));
+      player.on('error', e => doError(e));
```
seanyesmunt: lets remove this if we aren't going to do anything with it

Issue #3474, opened 2020-01-11
Navigate forward/backward multiple pages


seanlis:
**Is your feature request related to a problem? Please describe.**
Each time the forward/backward button is clicked, you move one page forward/backward. One might need to click multiple times to reach the desired page.

**Describe the solution you'd like**
As in chrome or firefox, when right-click the forward/backward buttons, display the browsing history from which you could select the page you want.

![image](https://user-images.githubusercontent.com/7328445/72200012-140b5780-347f-11ea-85f6-0aaed6640919.png)

tzarebczan:
Thanks for the feature suggestion!

dalhill:
I'll give this a shot

seanyesmunt:
Added in #3547!

PR #3547, opened 2020-01-25
Right Click to Navigate History

dalhill:
## PR Checklist

## PR Type
- [x] Feature

## Fixes
Issue Number: #3474

In order to navigate history, the user must do so one step at a time.

## What is the new behavior?
User is able to right click the navigation buttons to move forward or backward multiple steps.

backward

forward

## Other information
Possible improvements
- [x] Limit the size of the history. Showing only 12 historical entries on either side. This is consistent with chrome.
- [x] Configure the list to dropdown the same height as the other header buttons.
- [x] If there is no "forward" or "backward" action to be taken, disable the button.

### DIFF
--- a/ui/component/navigationButton/view.jsx
+++ b/ui/component/navigationButton/view.jsx
@@ -3,6 +3,9 @@ import React, { useState, useCallback } from 'react';
 import * as ICONS from 'constants/icons';
 import Button from 'component/button';
 
+// the maximum length of history to show per button
+const MAX_HISTORY_SIZE = 12;
+
 type Props = {
   isBackward: boolean,
   history: {
@@ -18,9 +21,17 @@ type Props = {
 };
 
 // determines which slice of entries should make up the back or forward button drop-downs (isBackward vs !isBackward respectively)
-const sliceEntries = (currentIndex, entries, historyLength, isBackward) => {
-  const l = isBackward ? 0 : currentIndex + 1;
-  const r = isBackward ? currentIndex : historyLength;
+const sliceEntries = (currentIndex, entries, historyLength, isBackward, maxSize) => {
+  let l = isBackward ? 0 : currentIndex + 1;
+  let r = isBackward ? currentIndex : historyLength;
+  const exceedsMax = maxSize < r - l;
+  if (!exceedsMax) {
+    return entries.slice(l, r);
+  } else if (isBackward) {
+    l = r - maxSize;
+  } else {
+    r = l + maxSize;
+  }
   return entries.slice(l, r);
 };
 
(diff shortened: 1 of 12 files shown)


ui/component/router/view.jsx
```diff
@@ -60,14 +62,50 @@ type Props = {
...
+    title,
   } = props;
+  const { channelName, streamName } = parseURI(uri);
+  const { entries } = history;
+  const entryIndex = history.index;
+
+  useEffect(() => {
+    if (typeof title !== 'undefined' && title !== '') {
+      document.title = title;
+    } else if (typeof streamName !== 'undefined' && streamName !== 'undefined' && streamName !== '') {
```
dalhill: BTW - streamName was coming as a string of value 'undefined'. I didn't look too far into this bug.

ui/store.js
```diff
@@ -96,7 +96,7 @@ const persistOptions = {
 
 let history;
 // @if TARGET='app'
-history = createHashHistory();
+history = createMemoryHistory();
```
seanyesmunt: What is the difference between these two?

ui/component/router/view.jsx
```diff
@@ -60,14 +62,52 @@ type Props = {
...
     isAuthenticated,
+    history,
+    uri,
+    title,
   } = props;
+  const { channelName, streamName } = parseURI(uri);
+  const { entries } = history;
+  const entryIndex = history.index;
+
+  useEffect(() => {
```
seanyesmunt: Why did you move this effect out of the show component?

ui/component/navigationButton/view.jsx
```diff
@@ -0,0 +1,75 @@
...
+          {entry.title}
+        </li>
+      );
+    },
+    [currentIndex, isBackward, go]
+  );
+  const slicedEntries = sliceEntries(currentIndex, entries, historyLength, isBackward);
+  return (
+    <div>
+      <Button
```
seanyesmunt: Are you able to determine if we can go back here or not? If so, it would be great to disable the back button and give it some disabled style so it's easier to tell that you can't

tzarebczan:
Thanks for the PR! Once this is fully reviewed/corrected/merged, please send us an email so we can show you some [appreciation](https://lbry.com/faq/appreciation)

ui/store.js
dalhill: createMemoryHistory gives us an array of entries so that we can reference the history stack.

*Excerpt from [history/docs/GettingStarted](https://github.com/ReactTraining/history/blob/master/docs/GettingStarted.md)*
> Additionally, createMemoryHistory provides history.index and history.entries properties that let you inspect the history stack.

ui/component/router/view.jsx
dalhill: I want this to fire on every page and not just the pages that render using the ShowPage component.

ui/component/navigationButton/view.jsx
dalhill: PR has been updated to disable button if no historical entries.

seanyesmunt: Awesome!

seanyesmunt:
Nice work @dalhill. This is really close.

I'll spend some time testing your branch this week just to make sure there aren't any issues on electron moving to memoryHistory, but I don't think there will be. I'll also push up a commit and do another pass at the styling.

dalhill:
Pushed to resolve conflicts

ui/component/router/view.jsx
seanyesmunt: I think it's when you call `parseURI` with an `undefined` uri. Only calling `parseURI` if uri exists should fix it.

PR #3551, opened 2020-01-27
app plays embed in iframe

jessopb:


ui/component/fileRender/view.jsx
```diff
@@ -80,12 +81,23 @@ class FileRender extends React.PureComponent<Props> {
...
+      embedUrl,
+      uri,
+    } = this.props;
     const fileType = fileName && path.extname(fileName).substring(1);
+    const streamUrl = embedUrl || streamingUrl;
 
     // Ideally the lbrytv api server would just replace the streaming_url returned by the sdk so we don't need this check
     // https://github.com/lbryio/lbrytv/issues/51
-    const source = IS_WEB ? generateStreamUrl(claim.name, claim.claim_id) : streamingUrl;
+    const source = IS_WEB ? generateStreamUrl(claim.name, claim.claim_id) : streamUrl;
```
seanyesmunt: Why is `streamUrl` including `embedUrl`? That should only be used if `IS_WEB` is true, which would never be the case here.

ui/component/fileRender/view.jsx
```diff
@@ -186,9 +198,16 @@ class FileRender extends React.PureComponent<Props> {
     // Return viewer
     return viewer || unsupported;
   }
-
   render() {
-    const { isText } = this.props;
+    const { isText, embedUrl } = this.props;
+
+    if (embedUrl) {
```
seanyesmunt: Instead of returning early with almost the same markup, just use classnames.

`classnames({ "file-render": !embedUrl, ...`

ui/page/embedWrapper/index.js
```diff
@@ -0,0 +1,26 @@
...
+import EmbedWrapperPage from './view';
+import { doResolveUri, makeSelectClaimForUri } from 'lbry-redux';
+import { generateStreamUrl } from 'util/lbrytv';
+
+const select = (state, props) => {
+  const PROTOCOL = 'lbry://';
+  const { match } = props;
+  const { params } = match;
+  const { claimName, claimId } = params;
+  const uri = PROTOCOL + claimName + (claimId ? `#${claimId}` : '');
```
seanyesmunt: use `buildURI` here

ui/redux/reducers/app.js
```diff
@@ -45,12 +45,14 @@ const defaultState: AppState = {
...
   modalProps: {},
   platform: process.platform,
-  upgradeSkipped: sessionStorage.getItem('upgradeSkipped') === 'true',
   daemonVersionMatched: null,
   daemonReady: false,
   hasSignature: false,
   badgeNumber: 0,
+  // @if TARGET='app'
+  upgradeSkipped: sessionStorage.getItem('upgradeSkipped') === 'true',
   volume: Number(sessionStorage.getItem('volume')) || 1,
```
seanyesmunt: volume should still be included for lbry.tv. You could also move it above the reducers and use a try catch.

PR #3569, opened 2020-01-28
fix: temp hack for future content

tzarebczan:
This at least would work on the new/trending + all content page until the SDK Fixes it.


ui/component/claimListDiscover/view.jsx
```diff
@@ -130,6 +130,13 @@ function ClaimListDiscover(props: Props) {
           .startOf('week')
           .unix()
       )}`;
+    } else {
+      // Hack for at least the New page until https://github.com/lbryio/lbry-sdk/issues/2591 is fixed
```
kauffj: Good job!

Issue #3494, opened 2020-01-14
Autoplay next does not respect block / mature settings


tzarebczan:
The related content may block a channel or mature content, but it autoplays it anyway. We should fix this before making bigger autoplay changes.

seanyesmunt:
Fixing this in #3570

Issue #3360, opened 2019-12-11
only autoplay free content, skip channels / paid content


tzarebczan:
https://lbry.tv/@lol:9/stopped-working:2

It tries to autoplay the channel and navigates to the page.

PR #3570, opened 2020-01-28
make sure the app  doesn't autoplay paid content, channels, or content from blocked channels

seanyesmunt:
#3494 
#3360

### DIFF
--- a/ui/redux/selectors/content.js
+++ b/ui/redux/selectors/content.js
@@ -107,11 +107,11 @@ export const makeSelectNextUnplayedRecommended = (uri: string) =>
 
           const claim = claimsByUri[recommendedUri];
           const channel = claim && claim.signing_channel;
-          if (channel && blockedChannels.find(blockedUri => blockedUri === channel.permanent_url)) {
+          if (channel && blockedChannels.some(blockedUri => blockedUri === channel.permanent_url)) {
             continue;
           }
 
-          if (!history.find(item => item.uri === recommendedForUri[i])) {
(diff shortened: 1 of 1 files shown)


ui/redux/selectors/content.js
```diff
@@ -77,11 +80,39 @@ export const makeSelectNextUnplayedRecommended = (uri: string) =>
...
+    selectAllCostInfoByUri,
+    selectBlockedChannels,
+    (
+      recommendedForUri: Array<string>,
+      history: Array<{ uri: string }>,
+      claimsByUri: { [string]: ?Claim },
+      costInfoByUri: { [string]: { cost: 0 | string } },
+      blockedChannels: Array<string>
+    ) => {
+      if (recommendedForUri) {
```
kauffj: Re-writing `for` loops into functional equivalents is the closest I get to zen meditation. Thank you for today's edition.

More importantly, you were using `find` when you probably want to use `some`.

```
return recommendedForUri.find(recommendedUri => {
  const { isChannel } = parseURI(recommendedUri);
  if (isChannel) {
    return false;
  }

  const costInfo = costInfoByUri[recommendedUri];
  if (!costInfo || costInfo.cost !== 0) {
    return false;
  }

  const claim = claimsByUri[recommendedUri];
  const channel = claim && claim.signing_channel;
  if (channel && blockedChannels.some(blockedUri => blockedUri === channel.permanent_url)) {
    return false;
  }

  return !history.some(item => item.uri === recommendedUri);
});
```

PR #3586, opened 2020-01-30
Feat block nsfw on search

jessopb:


jessopb:
not building on windows - @seanyesmunt is aware

PR #3596, opened 2020-01-31
actually disable autoplay in embeds

jessopb:


seanyesmunt:
Good to merge if this fixes it

Issue #3005, opened 2019-10-09
Epic - [More] Interactive Comments


osilkin98:
- [x] Implement URL embedding in comments
- [x] Implement comment section on user channels
- [x] Add Comment Thumbnails
- [ ] Implement media uploading within comments (pasting an image from clipboard into comments - will require work on SDK side)
- [ ] Polish up the styling in the comment section 
- [ ] Push notifications for new comments (will require work on SDK side)

btzr-io:
Some progress on the frontend side, markdown support for comments, still needs a few minor styles tweaks:

![comments-ui](https://user-images.githubusercontent.com/14793624/66710724-726b5180-ed3b-11e9-893e-dcda1f7a48ad.png)

seanyesmunt:
@osilkin98 I don't think channel messaging should be bundled in with this, since it's not actually comments

osilkin98:
Alright, I’ll make a separate issue for it.

PR #3600, opened 2020-02-01
Added the ability to reply to comments

Meantub:
## PR Checklist

## PR Type

It allows you to reply to comments

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #3005 

People can currently only comment on the top level.

## What is the new behavior?
This allows you to comment on other people's comments. This will allow for more discussion in the comment section rather than only at the content.

image


tzarebczan:
Thanks for the PR @Meantub , we'll get this reviewed. Once it is merged, can we show you some [appreciation](https://lbry.com/faq/appreciation) for the contribution?

Can you squash the commits too? 

Also, what level of replies do we show by default? We probably need to have some sort of "show more" after level 1.

seanyesmunt:
@Meantub Thank you for the PR! Can you rebase this branch so it's easier to tell which changes are yours?

ui/component/comment/view.jsx
```diff
@@ -79,6 +214,12 @@ function Comment(props: Props) {
             </div>
           )}
         </div>
+        <div>
+          {isReplying ? (
+            <CommentReply uri={uri} parentId={commentId} setReplying={setReplying} />
```
seanyesmunt: `isReplying && <Comment ... />` 😛 

Recursive component!

Meantub: CommentReply is the component that comes up to type in your reply, not an actual Comment.

seanyesmunt: Ah right. In that case, it should use `<CommentCreate />` since it's almost exactly the same.

seanyesmunt:
@Meantub You must have been branched off of @osilkin98's branch before his commits were squashed before merging. You'll have to reset all of those, then rebase again.

ui/component/comment/view.jsx
Meantub: The main difference is this cancel button
image

I would use CommentCreate if I can get that cancel button somehow.

Making a modification to CommentCreate to include Cancel button is a possibility but might be weird since we won't want there to be one for the top level comments.
image

seanyesmunt: Yep. It should only include the cancel button if it's a reply.

Meantub: @seanyesmunt So should I keep the CommentReply component?

Meantub:
Opening new pull request with fixed commits! [Here is the new one](https://github.com/lbryio/lbry-desktop/pull/3634)

PR #3611, opened 2020-02-03
Revert "Right Click to Navigate History"

seanyesmunt:
Reverts lbryio/lbry-desktop#3547

Reverting this because open.lbry.com links don't seem to work when the app is closed. Probably from the change to `memoryHistory`. Will investigate. @dalhill FYI


dalhill:
@seanyesmunt - I want to look into this but I'm not sure how to recreate this scenario.

1. Can you provide a failing link
2. How are you opening links when the app is closed?

tzarebczan:
You'd go to open.lbry.com/jiggytom and then use the open in app button at the bottom. If that doesn't work, try in a new session.

dalhill:
Thanks @tzarebczan

Definitely related to `createMemoryHistory`. Reading up on the [docs](https://github.com/ReactTraining/history/blob/master/docs/Blocking.md). It looks like we'll need to add a `getUserConfirmation` function. I'm still looking into this, just wanted to give an update.

PR #3634, opened 2020-02-05
Ability to reply to comments

Meantub:
## PR Checklist

## PR Type

It allows you to reply to comments

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #3005 

People can currently only comment on the top level.

## What is the new behavior?
This allows you to comment on other people's comments. This will allow for more discussion in the comment section rather than only at the content.

image

### DIFF
--- a/ui/component/comment/view.jsx
+++ b/ui/component/comment/view.jsx
@@ -10,11 +10,15 @@ import { Menu, MenuList, MenuButton, MenuItem } from '@reach/menu-button';
 import Icon from 'component/common/icon';
 import * as ICONS from 'constants/icons';
 import { FormField, Form } from 'component/common/form';
+import CommentReply from '../commentReply/index';
+import classnames from 'classnames';
 
 type Props = {
+  uri: string,
   author: ?string, // LBRY Channel Name, e.g. @channel
   authorUri: string, // full LBRY Channel URI: lbry://@channel#123...
   commentId: string, // sha256 digest identifying the comment
+  parentId: string, // sha256 digest identifying the parent of the comment
   message: string, // comment body
   timePosted: number, // Comment timestamp
   channel: ?Claim, // Channel Claim, retrieved to obtain thumbnail
@@ -33,6 +37,7 @@ const ESCAPE_KEY = 27;
 
 function Comment(props: Props) {
   const {
+    uri,
     author,
     authorUri,
     timePosted,
@@ -44,6 +49,7 @@ function Comment(props: Props) {
     channelIsBlocked,
     commentIsMine,
     commentId,
+    parentId,
     updateComment,
     deleteComment,
   } = props;
@@ -54,6 +60,9 @@ function Comment(props: Props) {
 
   // used for controlling the visibility of the menu icon
   const [mouseIsHovering, setMouseHover] = useState(false);
+
+  // used for controlling visibility of reply comment component
+  const [isReplying, setReplying] = useState(false);
 
   // to debounce subsequent requests
   const shouldFetch =
@@ -96,10 +105,15 @@ function Comment(props: Props) {
   function handleSubmit() {
     updateComment(commentId, editedMessage);
     setEditing(false);
+    setReplying(false);
   }
 
   function handleDeleteComment() {
     deleteComment(commentId);
+  }
+
+  function handleReply() {
+    setReplying(true);
   }
 
   function handleMouseOver() {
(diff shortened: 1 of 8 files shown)


seanyesmunt:
Will give this a review after we get our next release out

ui/component/comment/view.jsx
```diff
@@ -1,73 +1,206 @@
...
+import classnames from 'classnames';
 
 type Props = {
-  author: string,
-  authorUri: string,
-  message: string,
-  timePosted: number,
-  claim: ?Claim,
+  uri: string,
+  author: ?string, // LBRY Channel Name, e.g. @channel
```
seanyesmunt: These changes were done in a previous PR, they already exist on master
https://github.com/lbryio/lbry-desktop/blob/master/ui/component/comment/view.jsx#L15

Meantub: Alright let me know if I still need to do something, I rebased with master

seanyesmunt: Looks good! I'll give this a review/test later today.

ui/component/comment/view.jsx
```diff
@@ -149,6 +167,13 @@ function Comment(props: Props) {
                   <MenuItem className="comment__menu-option" onSelect={handleDeleteComment}>
                     {__('Delete')}
                   </MenuItem>
+                  {parentId === null ? (
```
seanyesmunt: You'll need to move the `commentIsMine` line down to inside the menu. Right now it looks like you would only be able to reply to your own comments.

seanyesmunt: Also instead of the ternary you can just do `parentId === null && <MenuItem...`

ui/component/commentReply/view.jsx
```diff
@@ -0,0 +1,88 @@
...
+  if (!commentingEnabled) {
+    return (
+      <I18nMessage tokens={{ sign_in_link: <Button button="link" requiresAuth label={__('sign in')} /> }}>
+        Please %sign_in_link% to comment.
+      </I18nMessage>
+    );
+  }
+
+  return (
+    <Form className="comment__reply-form" onSubmit={handleSubmit}>
```
seanyesmunt: I still think this component should be combined with `CommentCreate`. They are mostly the same.

ui/component/commentsList/view.jsx
```diff
@@ -29,12 +29,31 @@ function CommentList(props: Props) {
     fetchComments(uri);
   }, [fetchComments, uri]);
 
+  function sortByParent(arrayOfComments) {
```
seanyesmunt: We should move this logic into the redux selector, so the result of it can be cached, instead of resorting every time.

ui/component/commentReply/view.jsx
Meantub: Any proposed way that I should do that? If I were to combine it into the CommentCreate component I feel like it would get very cluttered in CommentCreate when we could just keep it a separate component. There are some styles and functionality specific to replying, it seems counterproductive to put them all in one class. I feel it is neater this [way](https://en.wikipedia.org/wiki/Single_responsibility_principle).

ui/component/comment/view.jsx
Meantub: Something breaks in the MenuList when I don't do ternary. Same thing with `<React.Fragment>` on the remaining commentIsMine elements (Edit and Delete). So if I use ternary with all of them it works, like this:
```js
<MenuList className="comment__menu-list">
  {isCommentMine() ? (
	  <MenuItem className="comment__menu-option" onSelect={handleDeleteComment}>
		{__('Delete')}
	  </MenuItem>
  ) : ('')}
  {isCommentMine() ? (
	<MenuItem className="comment__menu-option" onSelect={handleSetEditing}>
	  {__('Edit')}
	</MenuItem>
  ) : ('')}
  {isParent() ? (
	<MenuItem className="comment__menu-option" onSelect={handleReply}>
	  {__('Reply')}
	</MenuItem>
  ) : ('')}
</MenuList>

```

If I do this:
```js
<MenuList className="comment__menu-list">
  {isCommentMine() ? (
    <React.Fragment>
	<MenuItem className="comment__menu-option" onSelect={handleDeleteComment}>
	  {__('Delete')}
	</MenuItem>
	<MenuItem className="comment__menu-option" onSelect={handleSetEditing}>
	  {__('Edit')}
	</MenuItem>
    </React.Fragment>
  ) : ('')}
  {isParent() ? (
	<MenuItem className="comment__menu-option" onSelect={handleReply}>
	  {__('Reply')}
	</MenuItem>
  ) : ('')}
</MenuList>
```
It doesn't work.

I'll put a new commit for what works currently

seanyesmunt: Ah, I remember the menu list being weird about those. Sorry about that, the ternary should be fine.

ui/component/commentReply/view.jsx
seanyesmunt: The only big difference between the two is calling `replyComment` instead of `createComment`. You could pass in a single prop called `isReply` that adds the cancel button, and changes what function is called when a user submits the form.

seanyesmunt:
@Meantub any plans to finish this up? It would be great to have in the app!

seanyesmunt:
Closing for now. Feel free to re-open!

PR #3642, opened 2020-02-05
navigate back after delete if abandoned

jessopb:
For whatever reason, dispatching connected-react-router.goBack() didn't work in batched actions.


ui/redux/actions/file.js
```diff
@@ -63,21 +63,20 @@ export function doDeleteFileAndMaybeGoBack(uri, deleteFromComputer, abandonClaim
     const { outpoint } = makeSelectFileInfoForUri(uri)(state) || '';
     const { nout, txid } = makeSelectClaimForUri(uri)(state);
     const claimOutpoint = `${txid}:${nout}`;
-
+    console.log('DEL');
```
seanyesmunt: 👀

PR #3650, opened 2020-02-07
Right Click to Navigate History (Round 2)

dalhill:
## PR Checklist

## PR Type
- [x] Feature

## Fixes
Issue Number: #3474

In order to navigate history, the user must do so one step at a time.

## What is the new behavior?
User is able to right click the navigation buttons to move forward or backward multiple steps.

backward

forward

## Other information
Possible improvements
- [x] Limit the size of the history. Showing only 12 historical entries on either side. This is consistent with chrome.
- [x] Configure the list to dropdown the same height as the other header buttons.
- [x] If there is no "forward" or "backward" action to be taken, disable the button.

### DIFF
--- a/ui/store.js
+++ b/ui/store.js
@@ -7,7 +7,7 @@ import { createFilter, createBlacklistFilter } from 'redux-persist-transform-fil
 import localForage from 'localforage';
 import { createStore, applyMiddleware, compose } from 'redux';
 import thunk from 'redux-thunk';
-import { createHashHistory, createBrowserHistory } from 'history';
+import { createMemoryHistory, createBrowserHistory } from 'history';
 import { routerMiddleware } from 'connected-react-router';
 import createRootReducer from './reducers';
 import { buildSharedStateMiddleware, ACTIONS as LBRY_REDUX_ACTIONS } from 'lbry-redux';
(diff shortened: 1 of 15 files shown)


ui/store.js
```diff
@@ -96,7 +96,16 @@ const persistOptions = {
...
 // @if TARGET='app'
-history = createHashHistory();
+let initialEntry = '/';
+let hash = window.location.hash;
+if (hash) {
+  hash = hash.replace('#', '');
+  initialEntry = hash.startsWith('/') ? hash : '/' + hash;
+}
+history = createMemoryHistory({
+  initialEntries: [initialEntry],
```
dalhill: @seanyesmunt @tzarebczan After way too much jiggytom I found the root of the issue. It seems the default initialEntries were overriding the entry sent over by lbry.tv.

seanyesmunt: @dalhill This fixes opening `lbry://xxx` links?

dalhill: Yes, links like the following.

image

I've tested this on videos, channels, "settings" and the main page. The only weird one so far is the main page as this makes the app open up into the "discover" view, however, this doesn't seem to be a bug related to the code in this PR. 

If you know of any "types" of links that I've missed, let me know and I'll test those as well.

seanyesmunt: Awesome!!!

The only other thing I would test is if this contains the correct path:
https://github.com/lbryio/lbry-desktop/blob/master/ui/analytics.js#L250

It's the first initial page view event we send. You could just console log the url there.

dalhill: Yea this initial event was sending "/" instead of "/jiggytom". I pushed a change that fixes this.

seanyesmunt:
Awesome. Sorry I've been a little slow getting to this. We will be able to give this a closer look. On the last PR I added a commit to clean up the styling. Could you add that back?

dalhill:
@seanyesmunt no worries. I added the changes you made back into this PR. I did one last check and I'm seeing the following (slightly differs from original images)
new-history

Do we want to keep this or trim the second half out?

PR #3700, opened 2020-02-17
use claim_search for channel discovery

seanyesmunt:
#### Changes
- Adds a simple banner to the top of the homepage if you are not currently following any channels
- Dropped "manage channels" page
- Added new page to discover channels using the same layout as the home page


ui/scss/init/_gui.scss
```diff
@@ -257,3 +257,10 @@ a {
 .download-text {
   font-size: var(--font-xsmall);
 }
+
+.notice-message {
```
seanyesmunt: We don't have any CSS for a generic message at the top of a screen, it's pretty basic so I just added it here.

ui/page/home/view.jsx
```diff
@@ -193,6 +194,23 @@ function HomePage(props: Props) {
...
+              discover_channels_link: (
+                <Button
+                  button="link"
+                  navigate={`/$/${PAGES.CHANNELS_FOLLOWING_DISCOVER}`}
+                  label={__('Find new channels to follow')}
+                />
+              ),
+            }}
+          >
+            You aren't currently following any channels. %discover_channels_link%.
```
seanyesmunt: This should probably be different text.

ui/constants/pages.js
```diff
@@ -24,6 +24,7 @@ exports.TAGS_FOLLOWING = 'following/tags';
 exports.TAGS_FOLLOWING_MANAGE = 'following/tags/manage';
 exports.CHANNELS_FOLLOWING = 'following/channels';
 exports.CHANNELS_FOLLOWING_MANAGE = 'following/channels/manage';
+exports.CHANNELS_FOLLOWING_DISCOVER = 'following/channels/discover';
```
kauffj: tempted to change all following URLs as follows:

/following/tags -> /tags (it's not directly used, just to prefix other tags, and https://lbry.tv/$/tags is semi-broken)
/following/tags/manage -> /tags/mange (or /tags/edit)
/following/channels -> /following
/following/channels/discover -> /following/discover
/following/channels/manage -> (gone)

If not to the above, to something that ensures:
- Anything in the right sidebar has only one word in the path (i.e. always `/$/foo`, never `/$/foo/bar`)
- Nothing has more than two words in the path

ui/page/home/view.jsx
```diff
@@ -193,6 +194,26 @@ function HomePage(props: Props) {
 
   return (
     <Page>
+      {(authenticated || !IS_WEB) && !subscribedChannels.length && (
```
kauffj: could this check pass while channels are loading?

seanyesmunt: The subscribed channels are repopulated before this page is rendered.

ui/constants/pages.js
seanyesmunt: Will update this in my following PR

PR #3707, opened 2020-02-19
Feat privacy update

### LINKED ISSUE #3610: Obtrain tracking consent for EU users before tracking
RE: GDPR - apparently the FTC enforces this in the US.  We need a pop of some sort before connecting to segment and GA

https://twitter.com/ArcVRArthur/status/1223800412320534533?s=20

Tom edit:
Recently we've received some feedback that users should be able to opt-out of analytics before signing into the app.

We may need to break down GA vs internal analytics (+SDK analytics) into separate settings. A user may not want data sent to GA, but still okay with sending to LBRY for rewards.

jessopb:
Some testing and touchup left to do:
Emojis
Wallet custody messaging
Testing to Making sure it doesn't apply on lbrytv
Making sure startup analytics are disabled
Cross device sync

closes #3610


static/app-strings.json
```diff
@@ -955,7 +955,73 @@
...
+  "[5]: Required by our lawyers. Mostly stuff like don't break the law and you are responsible for how you use this app, but do read the whole thing.": "[5]: Required by our lawyers. Mostly stuff like don't break the law and you are responsible for how you use this app, but do read the whole thing.",
+  "[4]: No information will be sent directly to LBRY, Inc. or third-parties about your usage. Note that as peer-to-peer software, your IP address and potentially other system information can be sent to other users, though this information is not stored permanently.": "[4]: No information will be sent directly to LBRY, Inc. or third-parties about your usage. Note that as peer-to-peer software, your IP address and potentially other system information can be sent to other users, though this information is not stored permanently.",
```
seanyesmunt: You have some extra strings in here. I would just clear them all out.

ui/component/page/view.jsx
```diff
@@ -13,18 +13,19 @@ type Props = {
...
 
 function Page(props: Props) {
-  const { children, className, authPage = false, authenticated } = props;
+  const { children, className, authPage = false, authenticated, noHeader } = props;
   const obscureSideNavigation = IS_WEB ? !authenticated : false;
 
   return (
     <Fragment>
-      <Header authHeader={authPage} />
+      {!noHeader && <Header authHeader={authPage} />}
```
seanyesmunt: I think we should still keep the header so the LBRY logo is visible. I would just pass `authPage` to the `<Page />` component. (and possibly rename that value.

ui/component/privacyAgreement/view.jsx
```diff
@@ -0,0 +1,197 @@
...
+  history: { push: string => void },
+};
+
+function PrivacyAgreement(props: Props) {
+  const { setWelcomeVersion, setShareDataInternal, setShareDataThirdParty, history } = props;
+  const [share, setShare] = useState(undefined); // preload
+  const [agree, setAgree] = useState(false); // preload
+
+  function handleSubmit() {
+    if (share === 'free') {
```
seanyesmunt: These should be constants

ui/component/privacyAgreement/view.jsx
```diff
@@ -0,0 +1,197 @@
...
+                {/*<span className="emoji">😢</span>*/}
+                {__('No')}
+              </>
+            }
+            helper={
+              <I18nMessage
+                tokens={{
+                  terms: (
+                    <Button
+                      tabIndex="2"
```
seanyesmunt: I don't think these should all have the same `tabIndex`? That could mess with tabbing through with your keyboard

ui/component/privacyAgreement/view.jsx
```diff
@@ -0,0 +1,197 @@
...
+                  href="https://www.lbry.com/termsofservice"
+                  label={__('Terms of Service')}
+                />
+              ),
+            }}
+          >
+            Do you agree to the %terms%?
+          </I18nMessage>
+        </p>
+        <fieldset>
```
seanyesmunt: Do we need to include the TOS here?

ui/index.jsx
```diff
@@ -277,8 +277,16 @@ function AppWrapper() {
...
 
+  useEffect(() => {
+    if (persistDone) {
+      app.store.dispatch(doToggle3PAnalytics());
+    }
+  }, [persistDone]);
+
   useEffect(() => {
     if (readyToLaunch && persistDone) {
+      analytics.startupEvent();
```
seanyesmunt: This shouldn't be moved. We still want to start it immediately, so the time to start is still accurate. It should delay calling the `readyEvent`

ui/component/page/view.jsx
jessopb: Fine - my goal was to get rid of the [X]. I'll revisit.

ui/index.jsx
jessopb: Hm, this is code that runs before we have the persisted GA preference value from store. Maybe need another earlier place to persist it ?

jessopb: We don't want to send GA events until we know we have permission.

seanyesmunt: It can start counting before we know if users are sharing analytics or not. If they aren't, we can just discard that value

jessopb: startupEvent() does the actual sendGA()
startTime is still set at the top of the file.

readyEvent() includes the total start duration.
const appReadyTime = Date.now();
const timeToStart = appReadyTime - startTime;
analytics.readyEvent(timeToStart);

jessopb: I ended up putting it back and trust the localStorage.

ui/component/page/view.jsx
jessopb: Calling it fine for now.

ui/redux/actions/settings.js
```diff
@@ -65,27 +63,28 @@ export function doClearDaemonSetting(key) {
     });
   };
 }
-
-export function doSetDaemonSetting(key, value) {
+// if doPopulate is applying settings, we don't want to cause a loop; updateSharedPreference = false.
```
jessopb: update this comment

ui/redux/actions/app.js
```diff
@@ -450,11 +451,48 @@ export function doSignOut() {
...
+        const { settings, sharing_3P: sharing3P } = savedPreferences.value;
+        Object.entries(settings).forEach(([key, val]) => {
+          if (daemonSettings[key] !== val) {
+            dispatch(doSetDaemonSetting(key, val, false));
+          }
+        });
+        if (sharing3P !== undefined) {
+          doToggle3PAnalytics(sharing3P, true);
+        }
+        // also doToggle
```
seanyesmunt: what does this comment mean?

ui/analytics.js
```diff
@@ -185,7 +204,7 @@ const analytics: Analytics = {
 };
 
 function sendGaEvent(category, action, label, value) {
-  if (analyticsEnabled && isProduction) {
+  if (thirdPartyAnalyticsEnabled && true) {
```
seanyesmunt: `&& true`

ui/redux/actions/settings.js
```diff
@@ -104,12 +103,21 @@ export function doSaveCustomWalletServers(servers) {
...
+  return dispatch => {
+    dispatch({
+      type: ACTIONS.CLIENT_SETTING_CHANGED,
+      data: {
+        key,
+        value,
+      },
+    });
+
+    // if (Object.values(SHARED_PREFERENCES).includes(key)) {
```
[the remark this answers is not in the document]
jessopb: Wasn't tested yet. syncing client settings is closer, but may want some changes first.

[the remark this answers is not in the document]
jessopb: removing for now

Issue #3706, opened 2020-02-19
Feature Request: Slight Design Change for the LBRY Player Embed


MichaelTunnell:
I decided I wanted to embed a LBRY video on my site rather than using the YouTube embed but I found an issue. The default color of the Play Button background is using the LBRY brand color and I think it is problematic for the embed because it clashes with my sites.

I think the LBRY color is actually quite good overall as a brand choice but it doesnt work for embeds in my opinion because of the high likelihood of clashes with the sites it is embedded on.

However, I did come up with a way to have both LBRY branding and no clashing.

The default embed button should use a transparent black like this:

> .file-render__embed .video-js .vjs-big-play-button {
> 	background-color: rgba(0,0,0,0.6);
> }
> 

from here you could use the hover effect to apply LBRY branding color.

> .file-render__embed .video-js:hover .vjs-big-play-button {
>     background-color: rgba(var(--color-primary),0.6);
> }

GIF Demo of what the change will look like:
![lbry-embed-2020-02-18_23 01 46](https://user-images.githubusercontent.com/1056731/74803582-0e393b00-52a3-11ea-9b31-83439f8f3d33.gif)

kauffj:
@MichaelTunnell thank you! Looks like a fine change. To make a PR, you typically fork the repo then follow these steps: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork

We can also copy/paste this in, but we encourage PRs so you can receive the credit!

PR #3711, opened 2020-02-20
improved color scheme for embed player

MichaelTunnell:
This commit changes the colors of the play button on the embed video player to blend better with other websites. This relates to Issue #3706

## PR Checklist

## PR Type

- [x] Feature - this is a design change

Issue Number: #3706

uses LBRY branding always

## What is the new behavior?
neutral color for the play button and then LBRY branding on hover

## Other information
I would point out that the portion for the Hover color applying the opacity should work if the variable output is RGB but if it is HEX then it likely wont work. I didnt find the main variable location in my scan so its on an assumption.

If the variable color uses HEX then I would suggest putting it hardcoded like the following:

 background-color: rgba(37, 119, 97,0.6);

I am not sure if this is the best approach based on your structure and I also don't know if the way I added it was how you wanted the structure of the CSS to be but I tried to match it. :sunglasses:

### DIFF
--- a/ui/scss/component/_file-render.scss
+++ b/ui/scss/component/_file-render.scss
@@ -167,6 +167,7 @@
     .vjs-big-play-button {
       @extend .button--icon;
       @extend .button--play;
+      background-color: rgba(0,0,0,0.6);
       border: none;
       position: static;
       z-index: 2;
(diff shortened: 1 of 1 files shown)


seanyesmunt:
This is great! Thanks @MichaelTunnell

tzarebczan:
Thanks @MichaelTunnell , congrats on the first PR! Can we should you some [appreciation](https://lbry.com/faq/appreciation)?

MichaelTunnell:
I would gladly accept appreciation but I'd like to address the bug I introduced prior to that. :smiley: 

I created another Pull Request that fixes the issue I just found after deployment. It's not a big breaking bug or anything but I knew what needed fixing so I went ahead and made this [PR](https://github.com/lbryio/lbry-desktop/pull/3714) for that.

Issue #3695, opened 2020-02-15
Text overflow improvements on nav for small screens


auber262:
**Bug**
Version 0.42.0 Desktop
When the screen is reduced small or viewed on a Netbook, the 'Content' 'About' & 'Comments' are flow badly whilst a lot of space is attributed to the search box. Some better percentaging to the space given here is needed - less space to the search box would give enough to those nav bar link titles to show whole.

**To Reproduce**
1. Launch desktop app
2. Go to any of your channels
3. View on a small screen or reduce the window size to its minimum
4. See middle nav bar with search box flows badly

**Expected behavior**
Words of nav remain whole and search box width reduces

**Screenshots**
attached

**System Configuration**

App version 0.42.0
Daemon verson 0.59.2
Installation ID: 5DfuNZyLVmSwFt6GcXHEp872bHWCb44xsZUVnPzbLb2bPWpksaskWRsS21eB72AA3c
![small-screen-shot](https://user-images.githubusercontent.com/39855197/74594139-dcb63a80-502a-11ea-9394-82564ca56b28.JPG)

Operating system: Windows 10

**Additional context**

PR #3777, opened 2020-02-29
decrease size for search box on channel

ykris45:
## PR Checklist

- [x ] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)
- [ x] I have checked that this PR does not introduce a breaking change

## PR Type

- [x ] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/3695

Text Overflow on small screen in channel pages

## What is the new behavior?
Resize a channel search box to alternative fix this issues
## Other information
image

### DIFF
--- a/ui/scss/component/_wunderbar.scss
+++ b/ui/scss/component/_wunderbar.scss
@@ -24,7 +24,7 @@
   margin-right: 0;
 
   fieldset-section {
-    width: 15rem;
+    width: 14rem;
   }
 
   @media (max-width: $breakpoint-small) {


tzarebczan:
Thanks for the PR Chris! Sean is away for a few days so we'll get a review on it when he's back or maybe Jessop can take a look earlier.

kauffj:
@ykris45 thanks for this, but this solution is only a mild improvement for the problem (while also making it slightly worse for larger views). 

A he correct solution is to adjust the flexbox rules so that the input shrinks properly.

Most likely this is a change to `.tabs__list--channel-page` in `tabs.scss`, but I only gave a cursory look.

PR #3782, opened 2020-03-02
make transcoding work

jessopb:
transcoding works for mov to mp4
publishing in progress removes all but the top card and the bottom controls rather than forcing you to scroll up to see what's going on
start to make transcoding settings card in settings
add optimize checkbox to publishing
connect to supporting code in lbry-redux (lbry)

todo:
add ffmpeg availability and "find" button
get file size and suggest / warn about estimated time
prune support to be only for video, and only for desktop


ui/page/settings/view.jsx
```diff
@@ -259,7 +281,30 @@ class SettingsPage extends React.PureComponent<Props, State> {
                 </React.Fragment>
               }
             />
-
+            {/* @if TARGET='app' */}
```
seanyesmunt: Maybe this should be under "extra settings". We could toggle showing that if the file ends in certain extensions, or over a certain size?

ui/component/publishForm/view.jsx
```diff
@@ -143,9 +143,10 @@ function PublishForm(props: Props) {
   return (
     <Fragment>
       <PublishFile disabled={disabled || publishing} inProgress={isInProgress} />
-      <div className={classnames({ 'card--disabled': formDisabled })}>
-        <PublishText disabled={formDisabled} />
-        <Card actions={<SelectThumbnail />} />
+      {!publishing && (
```
seanyesmunt: Nice. I like this

seanyesmunt:
I think this looks good!

kauffj:
@seanyesmunt re-review requested of this

PR #3794, opened 2020-03-05
Publish messages

jessopb:
increase lbrytv limit until tv transcoding
warn about bitrate
warn about strange types
more reliably convey size limit message
update faq link for publishing


ibLeDy:
Suggestion: Specify approximate bitrate (instead of resolution, or alongside it) for the content to be streamed without buffering in normal conditions.

Dunno if it has ben said anywhere else but i'm suggesting it cause resolution always seems a worse indicator than bitrate.

PR #3830, opened 2020-03-12
Fix cookie security issue

seanyesmunt:
#### Changes
- Cookies are no longer set with `.lbry.tv` domain so they aren't accessible from player.lbry.tv, where html scripts are loaded and ran as is
  - Users who open player.lbry.tv links could lose their auth token (and funds)
- Deletes old .lbry.tv cookies and replaces them
- Now able to signout properly when running app in dev mode
- Perpetual cookie refresh if users have accessed site (the fix just happened to cause this - not sure if we should keep it?)

### DIFF
--- a/ui/index.jsx
+++ b/ui/index.jsx
@@ -16,22 +16,15 @@ import { Provider } from 'react-redux';
 import { doDaemonReady, doAutoUpdate, doOpenModal, doHideModal, doToggle3PAnalytics } from 'redux/actions/app';
 import { Lbry, doToast, isURIValid, setSearchApi, apiCall } from 'lbry-redux';
 import { doSetLanguage, doUpdateIsNightAsync } from 'redux/actions/settings';
-import {
-  doAuthenticate,
-  Lbryio,
-  rewards,
-  doBlackListedOutpointsSubscribe,
-  doFilteredOutpointsSubscribe,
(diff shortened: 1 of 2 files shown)


ui/index.jsx
```diff
@@ -94,6 +87,11 @@ if (process.env.SEARCH_API_URL) {
   setSearchApi(process.env.SEARCH_API_URL);
 }
 
+// Fix to make sure old users' cookies are set to the correct domain
+// This can be removed after March 11th, 2021
```
seanyesmunt: Not sure when it's ok to remove this

ui/util/saved-passwords.js
```diff
@@ -141,6 +146,16 @@ export const doSignOutCleanup = () => {
   });
 };
 
-export const testKeychain = () => {
-  // we should make sure it works on startup
+export const doCookieCleanup = () => {
+  const authToken = getAuthToken();
+  if (authToken) {
```
kauffj: This is the perpetual refresh I assume? I'm confused by why this is happening.

ui/index.jsx
kauffj: since this isn't a general migration framework, you could call this something like `doCookieCleanupMarch2020()`

PR #3831, opened 2020-03-12
got frustrated about no claim id on channels and went wild

kauffj:


kauffj:
@seanyesmunt @tzarebczan wanted total publishes displayed and this was an easy place to put it

I'm in favor of putting this somewhere else when there's time to be more thoughtful about it though

seanyesmunt:
I think it makes sense to put it there, at least for now. Checking it out

ui/component/channelAbout/view.jsx
```diff
@@ -54,10 +57,41 @@ function ChannelContent(props: Props) {
...
+              amount={parseFloat(claim.amount) + parseFloat(claim.meta.support_amount)}
+              precision={8}
+            />{' '}
+            (
+            <Button
+              button="link"
+              label={__('view all claims', {
+                normalized_uri: uri,
+              })}
+              navigate={`/$/${PAGES.TOP}?name=${claim.name}`}
```
seanyesmunt: I don't think this link does what you want. It shows all claims for that claim name, which will just be other channels with the same name?
http://lbry.tv/$/top?name=@kauffj

PR #3832, opened 2020-03-12
Pass `trending_algorithm` to GA

seanyesmunt:
Requires https://github.com/lbryio/lbryinc/pull/85

### DIFF
--- a/ui/analytics.js
+++ b/ui/analytics.js
@@ -202,6 +202,9 @@ const analytics: Analytics = {
   openUrlEvent: (url: string) => {
     sendGaEvent('Engagement', 'Open-Url', url);
   },
+  trendingAlgorithmEvent: (trendingAlgorithm: string) => {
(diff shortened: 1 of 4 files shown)


ui/analytics.js
```diff
@@ -202,6 +202,9 @@ const analytics: Analytics = {
   openUrlEvent: (url: string) => {
     sendGaEvent('Engagement', 'Open-Url', url);
   },
+  trendingAlgorithmEvent: (trendingAlgorithm: string) => {
+    sendGaEvent('Engagement', 'Trending-Algorithm', trendingAlgorithm);
```
seanyesmunt: Is this an `Engagement` event? Or should it be called something else?

kauffj: (I thought I asked above 4 days ago, just found it stuck pending :upside_down_face:)

finer9: I think an Engagement event category is fine.  We should be able to slice engagement stats by what algo they were seeing

PR #3833, opened 2020-03-12
add error nag when api performance is degraded

seanyesmunt:
Screen Shot 2020-03-12 at 1 43 03 PM

### DIFF
--- a/ui/component/app/view.jsx
+++ b/ui/component/app/view.jsx
@@ -1,5 +1,6 @@
 // @flow
 import * as PAGES from 'constants/pages';
+import { LBRY_TV_API } from 'config';
 import React, { useEffect, useRef, useState } from 'react';
 import classnames from 'classnames';
 import analytics from 'analytics';
@@ -97,6 +98,7 @@ function App(props: Props) {
   const appRef = useRef();
   const isEnhancedLayout = useKonamiListener();
   const [hasSignedIn, setHasSignedIn] = useState(false);
+  const [currentlyDegradedPerformance, setCurrentlyDegradedPerformance] = useState(false);
   const userId = user && user.id;
   const hasVerifiedEmail = user && user.has_verified_email;
   const isRewardApproved = user && user.is_reward_approved;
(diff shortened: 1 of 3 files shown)


ui/component/app/view.jsx
```diff
@@ -280,7 +297,24 @@ function App(props: Props) {
...
+      {currentlyDegradedPerformance && (
+        <Nag
+          type="error"
+          message={
+            <I18nMessage
+              tokens={{
+                more_information: <Button button="link" label={__('more')} href="https://status.lbry.com/" />,
+              }}
+            >
+              lbry.tv is currently experiencing issues. Try refreshing to fix it.
```
seanyesmunt: Not really a fan of this copy

ui/component/app/view.jsx
```diff
@@ -246,6 +248,21 @@ function App(props: Props) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [syncError, pathname]);
 
+  // @if TARGET='web'
```
kauffj: please take all component logic and move to a component in `lbrytv/component` (e.g. `NagDegradedPerformance`)

ui/component/app/view.jsx
```diff
@@ -246,6 +248,21 @@ function App(props: Props) {
...
+  useEffect(() => {
+    fetch(`${LBRY_TV_API}/internal/status`)
+      .then(response => response.json())
+      .then(status => {
+        if (status.general_state !== 'ok') {
+          throw Error();
+        }
+      })
+      .catch(err => {
+        setCurrentlyDegradedPerformance(true);
```
kauffj: Can we do this?

1. If `general_state` is `down` or API does not return a 200, show full screen gerbil error (no ability to interact)
1. If `general_state` is not `ok` or `down`, show degraded performance

This will both:
1. Improve UX when endpoint is down since if endpoint is down site is unlikely to be working at all
2. Allow Andrey/Niko to evolve their end without another update from us (return `down` when users shouldn't be interacting but their endpoint is up)

ui/component/app/view.jsx
kauffj: lbry.tv performance may be degraded. You can try to use it, or wait 5 minutes and refresh. Please no crush us.

seanyesmunt:
Merging this just to get it in. Will do another PR with the change to the gerbil based on specific states.

PR #3835, opened 2020-03-13
Feat channel delete warning

jessopb:


tzarebczan:
Since we're making this more difficult to delete accidentally, can we add the delete button to the channel page itself if it's quick. If not, can do it on a follow on PR.

ui/modal/modalRevokeClaim/view.jsx
```diff
@@ -50,7 +49,20 @@ class ModalRevokeClaim extends React.PureComponent<Props> {
...
       );
+    } else if (type === txnTypes.CHANNEL) {
+      return (
+        <React.Fragment>
+          <p>
+            {__(
+              "You're about to permanently delete a channel. Content published under this channel will be orphaned and their signing channel invalid. Content sync programs using this channel will fail."
+            )}
+          </p>
+          <p>{__('Are you sure? Type %name% to confirm that you wish to delete the channel.', { name })}</p>
```
seanyesmunt: the channel name should at least be bolded. Maybe slightly bigger?

PR #3836, opened 2020-03-13
Localized player urls for lbry.tv

seanyesmunt:


seanyesmunt:
@nikooo777 feel free to merge once you are able to monitor. I can open another PR later that adds the cookie before users watch their first video

Issue #3295, opened 2019-11-26
channel edit + thumb/cover upload don't error properly on apostrophe 


tzarebczan:
If you select a file with an apostrophe in the file name for cover/thumbnail, it will just say uploading... and not fail. It should fail like the publish page does.

tzarebczan:
This sounds like an easy one - it's a bad experience when users run into it because they have no idea why it fails.

Issue #3653, opened 2020-02-07
Display wallet balance anywhere credits are used


auber262:
**Is your feature request related to a problem? Please describe.**
In some instances, when credits are spent (tips, supports etc) the balance is not displayed and so you do not know how much you have when making a payment. If you want to support a claim with, say 200 LBC, you need to be able to see your wallet balance before, rather than rely on entering a figure only to get the error of 'you do not have enough credits'.

**Describe the solution you'd like**
In all places where credits are spent, for the balance to be displayed. This needs to be not only on the desktop app but also on lbry.tv - mobile view from phone particularly.

If you use lbry.tv on your phone and you go to support a claim, the dialog box is very limited.
![suggestion](https://user-images.githubusercontent.com/39855197/74046218-100f1e80-49c6-11ea-83ec-1bce5780e521.PNG)
![current](https://user-images.githubusercontent.com/39855197/74046222-10a7b500-49c6-11ea-99ff-6d42fe5cfbb5.PNG)

**Describe alternatives you've considered**
Toggling back and forth from wallet to page or claim / tip support page (which is a nightmare!)

**Additional context**
Image of current and suggestion of solution attached

tzarebczan:
Thanks for opening this, certainly needed in a few places where we use credits!

seanyesmunt:
Added this for mobile views

Issue #3392, opened 2019-12-18
web publish errors not caught


tzarebczan:
If you publish from the web and the proxy returns an error (i.e. insufficient funds), you don't get an error and the form gets stuck in limo (says publishing - requires cache to be cleared).

Issue #3296, opened 2019-11-26
signing out + wallet merging on sign in on Desktop


tzarebczan:
On Desktop, we should make it clearer in the app that when signing out, the local wallet data is still kept and will be merged with any new user's that signs in. I'll mention this on the account-sync FAQ but would be good to hint at it in the app somewhere. 

lbry.tv is not affected, but users wishing to change their email would need their accounts merged until we implement https://github.com/lbryio/lbry-desktop/issues/2630

tzarebczan:
A few people have run into this recently. It's a pain in the butt to clean up. We should make this clearer  and give them steps on how to remove the wallet if needed.

PR #3837, opened 2020-03-13
Bug roundup

seanyesmunt:
Fixes:
#3295 
#3653
#3392
#3296

### DIFF
--- a/ui/component/common/form-components/form-field.jsx
+++ b/ui/component/common/form-components/form-field.jsx
@@ -101,7 +101,9 @@ export class FormField extends React.PureComponent<Props> {
       } else if (type === 'select') {
         input = (
           <fieldset-section>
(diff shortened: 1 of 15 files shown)


ui/component/common/form-components/form-field.jsx
```diff
@@ -101,7 +101,9 @@ export class FormField extends React.PureComponent<Props> {
       } else if (type === 'select') {
         input = (
           <fieldset-section>
-            {label && <label htmlFor={name}>{label}</label>}
+            {(label || errorMessage) && (
```
kauffj: error message replacing label seems a bit weird?

seanyesmunt: I agree. It's mostly for inputs that are side by side with other inputs or buttons. Adding it to the top or bottom messes up the side by side right now in a number of cases.

PR #3838, opened 2020-03-13
fix selection on firefox

kauffj:
@seanyesmunt I spent a minute or two and couldn't think of or find a good reason why `user-select: none` was on all inputs. 

Do you know? It's not (at least anymore) for the copyable inputs. So I just removed it.


seanyesmunt:
I'm not sure. I don't know why it would be applied for labels either.

PR #3839, opened 2020-03-13
no play button on paused player

### LINKED ISSUE #3765: Video play button issues / remove play button from center of videoplayer when playback is paused.
Older suggestion from ideas-and-feedback:

er 01/20/2020
one suggestion: dont display that large green play button in the middle of videoplayer screen when it is started playing and then paused. sometimes one needs to pause playback in order to read/inspect something in this frame and this huge thing just is in a way. its plenty to toggle play/pause icons at bottom/left.

Another one from help:
Jason Khanlar (@jasonkhanlar) Today at 12:36 PM
uhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh, double you teee eff? How come in the LBRY.tv website, when I pause a video, there is a play icon covering the center of the video? That is so annoying and disruptive! I want to take a screenshot of something quickly, but now I need to edit the CSS to hide that icon!

Adding to it that "minor ui thing" or not, its in a users face every time he uses lbry. Each and every time he plays/pauses something. Wants to read some text in video frame, wants to screenshot... It should not be classified as minor.

kauffj:
fixes #3765

### DIFF
--- a/ui/component/fileViewerInitiator/view.jsx
+++ b/ui/component/fileViewerInitiator/view.jsx
@@ -3,7 +3,7 @@
 // The actual viewer for a file exists in TextViewer and FloatingViewer
 // They can't exist in one component because we need to handle/listen for the start of a new file view
 // while a file is currently being viewed
-import React, { useEffect, useCallback } from 'react';
+import React, { useEffect, useCallback, useState } from 'react';
 import classnames from 'classnames';
 import Button from 'component/button';
 import isUserTyping from 'util/detect-typing';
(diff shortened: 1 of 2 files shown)


seanyesmunt:
I think we should still keep this on mobile at least. Until we have a better solution to handle autoplay, I think the play button should stay.

Without it, there are several cases where the player looks "stuck" (no controls) while paused on the first frame of the video because the browser stopped the autoplay.

kauffj:
@seanyesmunt do you know if the class `vjs-has-started` is added when autoplay fails, or only when it works?

seanyesmunt:
not sure. i can test

seanyesmunt:
https://github.com/lbryio/lbry-desktop/commit/2d56518ffe1a95bac6d5667bb5d4032c9190b6b4

PR #3844, opened 2020-03-14
select tags before channels and filter channels by tag

jessopb:
User prompted to select tags after email verified.
After tags, user can select channels to follow. 
Back button added on channels.
Tags filter added to claim search.
Still a bit janky.

### DIFF
--- a/ui/component/claimListDiscover/view.jsx
+++ b/ui/component/claimListDiscover/view.jsx
@@ -32,7 +32,8 @@ type Props = {
   hiddenUris: Array<string>,
   hiddenNsfwMessage?: Node,
   channelIds?: Array<string>,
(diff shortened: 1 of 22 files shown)


ui/component/claimListDiscover/view.jsx
```diff
@@ -32,7 +32,8 @@ type Props = {
   hiddenUris: Array<string>,
   hiddenNsfwMessage?: Node,
   channelIds?: Array<string>,
-  tags: Array<string>,
+  tags: string, // these are just going to be string. pass a CSV if you want multi
```
seanyesmunt: what's a CSV?

jessopb: comma separated

PR #3845, opened 2020-03-16
make blog posts look a bit nicer

lyoshenka:
I updated markdown typography, made the text wider, and moved the view count lower down.

Let me know if you like where this is going and what else you want me to look at.

Before:

After:


ui/component/layoutWrapperText/view.jsx
```diff
@@ -46,25 +46,22 @@ function LayoutWrapperText(props: Props) {
           <h1 className="media__title-text">{title}</h1>
         </div>
 
-        <div className="media__subtitle--between">
-          <DateTime uri={uri} show={DateTime.SHOW_DATE} />
```
seanyesmunt: I think at least the date should stay so you know if what you are reading is recent/old. We added the view count too, just because we had the extra space from the adding the date.

ui/scss/component/_file-render.scss
```diff
@@ -52,6 +52,9 @@
   .markdown-preview {
     height: 100%;
     overflow: auto;
+    background-color: #fff;
+    padding: var(--spacing-xlarge);
+    margin-top: var(--spacing-large);
```
seanyesmunt: I don't think we should add this. No other news/blog site that I can think of has something like it. Is there a reason you wanted to add it? Also the `#fff` would need to be changed to a css variable so we can change the actual color for dark mode.

ui/scss/component/_main.scss
```diff
@@ -115,7 +115,7 @@
 
 .main__document-wrapper--markdown {
   @extend .main__document-wrapper;
-  width: 40em;
+  width: 60em;
```
seanyesmunt: I'm in favor of keeping this small. Most things I've read recommend 50 - 70 characters per line for maximum readability. We are already over that with 40em. The latest lbrytech blog post has 83 characters for the first word right now. With these changes, that is increased to 107 characters.

ui/scss/component/_markdown-preview.scss
```diff
@@ -1,4 +1,265 @@
 .markdown-preview {
```
seanyesmunt: Most styles in this file should probably be targeting generic html elements, instead of specifically inside the markdown editor, so that the rest of the app is consistent with the blogs.

I'd also like to avoid multiple different hardcoded padding values (em and px), as well as hard coded colors.

PR #3847, opened 2020-03-16
updated PR to fix issue found when merging #3809

seanyesmunt:


ui/component/userVerify/view.jsx
```diff
@@ -29,10 +29,9 @@ class UserVerify extends React.PureComponent<Props> {
   }
 
   render() {
-    const { errorMessage, isPending, verifyPhone, fetchUser, skipLink, onSkip } = this.props;
+    const { errorMessage, isPending, verifyPhone, fetchUser, onSkip } = this.props;
     const skipButtonProps = {
       onClick: onSkip,
-      navigate: skipLink || '/',
```
seanyesmunt: the issue was that the the app was navigating before we could write to localStorage. We didn't need to navigate here anyway because the `UserSignIn` component handles that already

PR #3851, opened 2020-03-17
add gerbil when apis are down and move components into lbrytv/

seanyesmunt:
#### Changes
`"degraded"` => show nag at bottom
`"down"` => show full screen gerbil
`anything else` => assume "ok"

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,13 +2,6 @@
   "parser": "babel-eslint",
   "extends": ["standard", "standard-jsx", "plugin:react/recommended", "plugin:flowtype/recommended"],
   "plugins": ["flowtype", "import", "react-hooks"],
-  "settings": {
-    "import/resolver": {
-      "webpack": {
-        "config": "webpack.base.config.js"
-      }
-    }
-  },
   "env": {
     "browser": true,
     "node": true
--- a/ui/component/app/view.jsx
+++ b/ui/component/app/view.jsx
@@ -1,6 +1,5 @@
 // @flow
 import * as PAGES from 'constants/pages';
-import { LBRY_TV_API } from 'config';
 import React, { useEffect, useRef, useState } from 'react';
 import classnames from 'classnames';
 import analytics from 'analytics';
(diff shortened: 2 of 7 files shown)


.eslintrc.json
```diff
@@ -2,13 +2,6 @@
   "parser": "babel-eslint",
   "extends": ["standard", "standard-jsx", "plugin:react/recommended", "plugin:flowtype/recommended"],
   "plugins": ["flowtype", "import", "react-hooks"],
-  "settings": {
```
seanyesmunt: This isn't needed anymore

ui/component/app/view.jsx
```diff
@@ -284,80 +275,41 @@ function App(props: Props) {
...
-            </I18nMessage>
-          }
-          actionText={__('Refresh')}
-          onClick={() => window.location.reload()}
-          onClose={() => setCurrentlyDegradedPerformance(false)}
+      {IS_WEB && lbryTvApiStatus === STATUS_DOWN ? (
+        <Yrbl
+          className="main--empty"
+          title={__('lbry.tv is currently down')}
+          subtitle={__('Something about fixing the squeeky wheel or something')}
```
seanyesmunt: @kauffj I forget what you said this label should be.

ui/component/app/view.jsx
```diff
@@ -120,14 +121,14 @@ function App(props: Props) {
   const rawReferrerParam = urlParams.get('r');
   const sanitizedReferrerParam = rawReferrerParam && rawReferrerParam.replace(':', '#');
   const wrapperElement = appRef.current;
+  const shouldHideNag = pathname.startsWith(`/$/${PAGES.EMBED}`) || pathname.startsWith(`/$/${PAGES.AUTH_VERIFY}`);
```
kauffj: yuck but okay

ideally this would be handled either inside of the top-level component the route points at or concurrently with route definitions

ui/component/app/view.jsx
kauffj: if I had something before I forget

how about:

"My wheel broke, but the good news is that someone from LBRY is working on it."

ui/component/app/view.jsx
seanyesmunt: Do you mean something in the `EmbedPage`/`AuthVerifyPage` that hides nags?

PR #3857, opened 2020-03-19
Add creator analytics dashboard

seanyesmunt:
Screenshot from 2020-03-18 13-13-19

_some of the css is a little questionable... but it works_


seanyesmunt:
I'm assuming this will change a lot as the api changes so I'm just going to merge this

PR #3862, opened 2020-03-19
display repost counts on file pages, link to all reposts

kauffj:
A few parts of this are minorly wonky but I felt like I was working with what was already there


kauffj:
requires https://github.com/lbryio/lbry-redux/pull/292

PR #3886, opened 2020-03-24
fix #3884

### LINKED ISSUE #3884: reposted_url discover param not working as expected
I can get to a reposted url from https://lbry.tv/@boneheadmedia:d/covid194lbry:7, but then this URL doesn't work on refresh / when sharing: https://lbry.tv/$/discover?reposted_uri=lbry%3A%2F%2F%40boneheadmedia%23d%2Fcovid194lbry%237

Should we just use reposted_claim_id instead? Or claim name + claim id?

kauffj:
fixes #3884


seanyesmunt:
doh https://github.com/lbryio/lbry-desktop/pull/3885/files

PR #3889, opened 2020-03-24
improve open in app link for existing LBRY users

seanyesmunt:
#### Changes
- Doesn't show nag for web only users
- Changes nag to `type="helpful"` so it's always on top of the default nag (analytics)
- Changes text to "It looks like you may have [LBRY Desktop | LBRY Android] installed"
- Will now always show nag with open links, even users have closed it before


lbrytv/component/openInAppLink/view.jsx
```diff
@@ -0,0 +1,85 @@
...
+    if (hasSrcParam) {
+      params.delete('src');
+      const newParams = params.toString();
+      const newUrl = `${pathname}?${newParams}`;
+      replace(newUrl);
+    }
+  }, [hasSrcParam, search, pathname, replace]);
+
+  React.useEffect(() => {
+    if (hasSrcParam && ((!showNag && isAndroidUser && isAndroidDevice) || (isDesktopUser && isDesktopDevice))) {
```
kauffj: so a `src` parameter always required? I was thinking to show these always...

PR #3907, opened 2020-03-27
Tom's small fixes

tzarebczan:
Fixes a couple known issues...
- Claim /channel/transaction list no longer run multiple times on startup, but claim searches still do (this must be a similar bug, but in another part of the code)
- Don't delete files on web
- Move forced player types to constant
- add 1.75 x playback rate
- Fix clear cache (no longer signs out, but instead clears session cache and reloads)

### DIFF
--- a/ui/component/viewers/videoViewer/view.jsx
+++ b/ui/component/viewers/videoViewer/view.jsx
@@ -7,6 +7,7 @@ import eventTracking from 'videojs-event-tracking';
 import isUserTyping from 'util/detect-typing';
 import analytics from 'analytics';
 import { EmbedContext } from 'page/embedWrapper/view';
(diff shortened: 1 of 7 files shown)


ui/component/viewers/videoViewer/view.jsx
```diff
@@ -133,16 +124,16 @@ function VideoViewer(props: Props) {
         currentVideo.removeEventListener('volumechange', doVolume);
       }
     };
-  }, []);
+  }, [changeMute, changeVolume, onEndedCB]);
```
seanyesmunt: What does this fix?

tzarebczan: It was an autosuggested fix from the dependencies..maybe I was too trigger happy :)

PR #3908, opened 2020-03-27
Improve share modal

seanyesmunt:
gif


ui/component/socialShare/view.jsx
```diff
@@ -4,61 +4,119 @@ import React from 'react';
...
+                className="share"
+                iconSize={24}
+                icon={ICONS.EMBED}
+                title={__('Embed this content')}
+                onClick={() => this.setState({ showEmbed: !showEmbed })}
+              />
+              <Button
+                className="share"
+                iconSize={24}
+                icon={ICONS.MORE}
```
kauffj: The fact that the more icon only shows a direct link is a little weird. 

I'm not even sure we want to give that link, since I'm not sure we're promising embedding via it.

tzarebczan: I changed it to include LBRY URL and Download - what do you think?

PR #3916, opened 2020-03-30
save optimize check to fix publish navigation file reloading

jessopb:
uses publish reducer instead of react hooks to store results from file checking across page navigations.


seanyesmunt:
This looks fine. I left a comment on the redux PR about the variable names.

PR #3941, opened 2020-04-02
fix tip select slider messaging and behavior

jessopb:
Made sure deleting by text field makes sense.
Made *some* message at the bottom at all times to prevent the box from jumping.


jessopb:
This could certainly use styling when it's an error, but triage.

PR #3943, opened 2020-04-02
style and limit height of errors on tip unlock modal

jessopb:
huge errors would expand the modal too much
loading... showing when it was done trying to load
Amount available 0 was showing just because it hadn't gotten the amount yet.

### DIFF
--- a/ui/component/supportsLiquidate/view.jsx
+++ b/ui/component/supportsLiquidate/view.jsx
@@ -107,60 +107,68 @@ const SupportsLiquidate = (props: Props) => {
         </>
       }
       body={
-        <>
-          <div className="section">
-            <I18nMessage
-              tokens={{
-                amount: (
-                  <strong>
-                    <CreditAmount badge={false} amount={Number(previewBalance)} precision={8} />
-                  </strong>
-                ),
-              }}
-            >
-              %amount% available to unlock
-            </I18nMessage>
-          </div>
-          <div className="section">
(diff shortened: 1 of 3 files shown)


ui/component/supportsLiquidate/view.jsx
```diff
@@ -107,60 +107,68 @@ const SupportsLiquidate = (props: Props) => {
...
-                  type="text"
-                  value={amount >= 0 ? amount || '' : previewBalance && previewBalance / 4}
-                  helper={message}
-                  onChange={e => handleChange(e.target.value)}
-                />
-              </Form>
-            )}
-          </div>
-        </>
+        !abandonClaimError && (
```
seanyesmunt, line 110: if you are hiding this, then it might make more sense to replace it with the error message.

Then the UI won't jump around as much

ui/component/supportsLiquidate/view.jsx
```diff
@@ -107,60 +107,68 @@ const SupportsLiquidate = (props: Props) => {
...
+          </>
+        )
       }
       actions={
         <React.Fragment>
           {abandonClaimError ? (
             <>
-              <div className="error-text">{__('%message%', { message: abandonClaimError })}</div>
-              <Button disabled={error} button="primary" onClick={handleClose} label={__('Done')} />
+              <div className="error-wrapper--no-overflow">
```
seanyesmunt, line 161: This should use `<ErrorText>`

PR #3971, opened 2020-04-08
fetch continent cookie on server-side and clean up video effects

### LINKED ISSUE #3856: continent cookie should be fetched/saved sooner
currently the localized player still defaults to NA during the first visit. The continent should be fetched and used right away instead.

### LINKED ISSUE #3961: Clean up view tracking
Also render the `<video>` element by default

seanyesmunt:
fixes #3856
fixes #3961

View events weren't firing because the streaming URL changed after the first render for new users, which caused the player to re-render and the listeners weren't added correctly. 

I thought I just needed to move the continent cookie stuff to the server, but the actual issue was that those listeners weren't being attached properly from the beginning.


seanyesmunt:
merging since view counts are sorta broken right now

PR #3981, opened 2020-04-11
Change text displayed for blocked channels

TigerxWood:
## PR Checklist

- [x ] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [] Refactoring (no functional changes)
- [ ] Documentation changes

## Fixes
Fix the text displayed in other languages
Issue Number:

With the current english string there is a problem with the translation in other languages. For example in other languages current english text would be: "You have blocked channels 10" instead of "You have 10 channels blocked." Also, what happen when there si no blocked channel? So, I think the simpler form is better in this situation.
## What is the new behavior?
Instead of "You have x blocked channles. Manage" will be displayed "Blocked channels: x Manage"

### DIFF
--- a/ui/page/settings/view.jsx
+++ b/ui/page/settings/view.jsx
@@ -471,8 +471,11 @@ class SettingsPage extends React.PureComponent<Props, State> {
                 title={__('Blocked Channels')}
                 actions={
                   <p>
-                    {__('Blocked Channels')}: {userBlockedChannelsCount} {' '}
-                    <Button button="link" label={__('Manage')} navigate={`/$/${PAGES.BLOCKED}`} />.
+                    {__('You have %count% blocked %channels%.', {
+                      count: userBlockedChannelsCount,
+                      channels: userBlockedChannelsCount === 1 ? __('channel') : __('channels'),
+                    })}{' '}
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Thanks for the PR @TigerxWood!

You are right this string isn't formatted very well, but we can just change it to this 
```
  __('You have %count% unblocked %channel%.', {
    count: userBlockedChannelsCount,
    channel: userBlockedChannelsCount === 1 ? __('channel') : __('channels'),
  });
```

Then the string for non-english languages can be changed more.

TigerxWood:
Pay atention it's blocked not unblocked. And still the problem it's the same. It's about the word blocked before channel or channels. In other languages would be correct: You have x channel(s) blocked, so I think my sugestion is the only way to not make problems.

tzarebczan:
Thanks for the suggestion and PR @TigerxWood ! Once we get this merged, can we show you some [appreciaiton](https://lbry.com/faq/appreciation)?

seanyesmunt:
@TigerxWood Sorry, I meant 'blocked'. I updated the string in this branch. Now other languages can choose if they want the word "blocked" before "channels".

PR #3990, opened 2020-04-13
slightly-broken player branding and new autoplay

kauffj:


kauffj:
@seanyesmunt there is one pending issue, that we cannot adjust mute or volume (see FIXME comment), without breaking the player. I can't tell why, and no console output. Otherwise ready.

tzarebczan:
- [ ] If you mute and go to another video, muted state doesn't persist. If you refresh and then go to another video then it does.

kauffj:
@tzarebczan fixed in https://github.com/lbryio/lbry-desktop/pull/3990/commits/2b2d376dea9a822407968fd5b7324b0a4809a16b

tzarebczan:
- [ ] claim / outpoint data is incorrect when switching from one video to the next. It keeps the old data and calls view api on original content instead of the new one. 

- [ ] with autoplay on, getting stuck almost every time on black loading screen on page refresh / cold load. Happened on Chrome mobile too.  Firefox desktop seemed to be better, mobile had issues. Turning autoplay off does fix it.

tzarebczan:
Also, since we're digging in this code - can we disable this popout button on desktop? Or make it work with out popout player? 
image

seanyesmunt:
created #4064 for the embed changes. Since that will require changing how we fetch the invite data I don't think that should be included in this.

PR #3999, opened 2020-04-15
only call email provided event for real sign ups

seanyesmunt:


ui/component/userEmailNew/view.jsx
```diff
@@ -87,7 +88,7 @@ function UserEmailNew(props: Props) {
   return (
     <div className="main__sign-up">
       <Card
-        title={__('Join lbry.tv')}
+        title={__('Register with lbry.tv')}
```
seanyesmunt, line 91: @kauffj 

"Register with lbry.tv" or "Register for lbry.tv"? Both of those seem weird to me.

PR #4025, opened 2020-04-20
Feat reenable repost settings

jessopb:


ui/redux/actions/settings.js
```diff
@@ -117,7 +118,25 @@ export function doSaveCustomWalletServers(servers) {
...
 export function doSetClientSetting(key, value) {
-  return dispatch => {
+  return (dispatch, getState) => {
+    const {
+      settings: { daemonSettings },
+    } = getState();
+    const { share_usage_data: shareSetting } = daemonSettings;
+    const isSharingData = shareSetting || IS_WEB;
+
+    if (key === SETTINGS.SHOW_REPOSTS && isSharingData) {
```
seanyesmunt, line 128: I think this should just live in the component onClick handler

Right now it looks like I would need to opt in to sharing data if I want to add the tag to remove those emails.

tzarebczan:
My default state on Desktop was off - it should be turned on to match current behavior. Should it be "hide reposts" and turned off? 

Also, turning them off also disables content emails for reposts and we may wanna mention that. I think its easier to if the setting is flipped.

tzarebczan:
- [ ] don't hide if reposted claim id is passed (i.e. show reposts)
- [ ] don't hide if reposts is passed explicitly in filter

PR #4030, opened 2020-04-21
Fix flow error.

jeffslofish:
Partial fix for #2024

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: Partial fix for #2024

Causes flow errors.

## What is the new behavior?
This particular flow error is removed.

## Other information
This code was causing a flow error because the object is sealed and then a new object property (not_tags) is added afterwards. To fix this I made an unsealed object, then added the properties as necessary.

I tried testing this code, but couldn't get this function to ever stop in the debugger. Is this actually dead code? How do I get this code to run?

### DIFF
--- a/ui/component/channelContent/index.js
+++ b/ui/component/channelContent/index.js
@@ -1,5 +1,4 @@
 import { connect } from 'react-redux';
-import { doFetchClaimsByChannel } from 'redux/actions/content';
 import { PAGE_SIZE } from 'constants/claim';
 import {
   makeSelectClaimsInChannelForPage,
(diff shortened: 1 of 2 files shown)


seanyesmunt:
Woohoo! I'll check these out after we release a new version of the app today. 🙂

seanyesmunt:
Good find. This is actually dead code. We moved the channel pages over to use the `<ClaimListDiscover />` component and it looks like this wasn't removed. Would you mind removing it?

This can also be removed
https://github.com/lbryio/lbry-desktop/blob/master/ui/component/channelContent/index.js#L31-L33

seanyesmunt:
Thank you!

PR #4049, opened 2020-04-23
Show elapsed/total instead of the countdown to the end in video player.

### LINKED ISSUE #3813: Video elapsed/duration in the player
video duration in the player is strange, i want to see elapsed/total instead of the countdown to the end (dont try to reinvent the wheel and be different at any cost (KISS))

jeffslofish:
Closes #3813

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation changes
Improves existing functionality.

## Fixes

Issue Number: #3813

Displays remaining time of video 

## What is the new behavior?
Displays the elapsed time along with the total time with a "/" in between

### DIFF
--- a/ui/scss/component/_file-render.scss
+++ b/ui/scss/component/_file-render.scss
@@ -170,6 +170,18 @@
     // Make sure no videojs message interferes with overlaying buttons
     pointer-events: none;
   }
+
+  .vjs-control-bar {
+    .vjs-remaining-time {
+      display: none;
+    }
+
+    .vjs-current-time,
+    .vjs-time-divider,
+    .vjs-duration {
+      display: flex;
+    }
+  }
 }
 
 .vjs-paused .vjs-big-play-button,


jeffslofish:
> Isn't there a js solution to this instead of CSS?

Perhaps, why do you prefer a js solution instead of CSS?

HarshKhandeparkar:
I would prefer js because it is more manageable/controllable and besides,
it will be more intuitive/obvious. If, in the future, someone tries to
change this, they'll be searching through multiple files.

seanyesmunt:
Nice one! Since this content is hidden with CSS by default, I think it makes sense to enable it with CSS.

tzarebczan:
Thanks again for another PR @jeffslofish! [Can we show you some appreciation](https://lbry.com/faq/appreciation)?

tzarebczan:
Any ideas on how to get this to work on mobile, or would it take up too much real estate?

jeffslofish:
@tzarebczan Yes, you can simply adjust the breakpoint max widths via javascript options. Right now when the active breakpoint is tiny, x-small, or small, then the time display and the speed control are hidden. The max width for the small breakpoint could be made smaller, but then slider control is made much smaller.

Here is what it looks like now at a screen width of 375px:
Screen Shot 2020-04-23 at 9 07 45 PM

and here is what it looks like if the time (and the speed) controls are allowed in by adjusting the breakpoint:
Screen Shot 2020-04-23 at 9 07 20 PM

I guess it is just a judgement call on which is better.

PR #4056, opened 2020-04-23
Edited strings about backup

eggplantbren:
This is my attempt at solving https://github.com/lbryio/lbry-desktop/issues/4039. I hope I found all the relevant strings, and that the new versions aren't too long.


eggplantbren:
Done, I think...not sure why that was considered a conflict.

PR #4059, opened 2020-04-24
paginate claim list

jessopb:


ui/page/fileListPublished/view.jsx
```diff
@@ -4,45 +4,86 @@ import Button from 'component/button';
...
-    fetchClaimListMine();
-  }, [checkPendingPublishes, fetchClaimListMine]);
+  }, [checkPendingPublishes]);
+
+  useEffect(() => {
+    if (paramsString && fetchClaimListMine) {
+      const params = JSON.parse(paramsString);
+      fetchClaimListMine(params.page, params.page_size);
+    }
+  }, [uploadCount, paramsString, fetchClaimListMine]);
```
jessopb, line 54: uploadCount is for lbry.tv. 
I could use two different @if TARGETS for two different useEffect hooks, or just leave it - it seems to work.

seanyesmunt: I think it's fine.

seanyesmunt:
This looks good. Will test it today.

Issue #4035, opened 2020-04-21
Add back 3d file viewer / fix cbz reader


seanyesmunt:
After the file page refactoring, there are issues displaying 3d files.

It seems like some wrapper class is preventing the 3d canvas from expanding in height

tzarebczan:
There's also bugs with the cbz reader: lbry://@injustice#3/01byInjustice-GodsAmongUs-Annual0012014digitalSonofUltron-Empirecbz#7 

It won't load the first time and then crashes when you leave the page. 

@btzr-io do you think you can work on these?

btzr-io:
> It won't load the first time and then crashes when you leave the page.

hmm I think I fixed that bug or a similar problem, I'll have to search for those commits 🙃

btzr-io:
> It won't load the first time

See: Force non-streamable content to wait for download #4074

btzr-io:
I fixed the 3d viewer styles, the same solution can be used for the comic viewer but the other errors are probably in the `Villain` side and I'll need to publish a new version once I figure out what is the error...

Anyway let me know if this still an issue, maybe I can fix it in the process: https://github.com/lbryio/lbry-desktop/issues/2777

tzarebczan:
These are now fixed, will open a new issue for the CBZ requiring download first.

Issue #4075, opened 2020-04-25
Incorrect render mode on some files


btzr-io:
**Bug**
Some files are reported as `Video` on the `renderMode` prop of `fileRender` component.
```
File extension: .stl
Content Type:  application/octet-stream
````

**To Reproduce**
1. Go to any url from the examples list.
2. See error...

**Expected behavior**
Use the correct viewer for this file, or just don't try to render.

**Examples**

https://lbry.tv/@GRAVITYSHAPES:f/free-3dprint-bowl-19cm-voronoi-style-no4-stl:8
https://lbry.tv/@GRAVITYSHAPES:f/stl:b

btzr-io:
@seanyesmunt Any reason to have `application/octet-stream` listed as `FORCE_CONTENT_TYPE_PLAYER`
https://github.com/lbryio/lbry-desktop/blob/eaca77d39430fa6f64af054ac3705c687935c18f/ui/constants/claim.js#L14

DispatchCommit:
This looks resolved by the 2 linked PR's that are merged.

PR #4076, opened 2020-04-25
Fix and enable 3D viewer

### LINKED ISSUE #4074: Force non-streaming content to wait for download
**Bug**

Some viewers will fail to render content the first time, because some files cant be streamed, you have to wait until the download is completed. 

**To Reproduce**
1. Go to a non-streamable file like `CAD` or `Comicbook`
2. Viewer will try to render the file but it will fail until `download completed`.

**Expected behavior**
Wait for `download completed`, then render the file from the download path not the streaming url.

btzr-io:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Changes
- Fix 3d viewer styles: see #4035
- Fix minor warnings from three.js
- Remove deprecated themes
- Fix #4074

## Other information

Test this if you can, just to make sure everything works as  expected.

### DIFF
--- a/static/app-strings.json
+++ b/static/app-strings.json
@@ -1191,5 +1191,6 @@
   "There are no stats for this channel. Make sure you are signed in with the correct email and have data sharing turned on.": "There are no stats for this channel. Make sure you are signed in with the correct email and have data sharing turned on.",
   "%follower_count% followers": "%follower_count% followers",
   "%lbc_received% LBC Earned": "%lbc_received% LBC Earned",
-  "Earnings may also include any LBC you've sent yourself or added as support. We are working on making this more accurate. Check your wallet page for the correct total balance.": "Earnings may also include any LBC you've sent yourself or added as support. We are working on making this more accurate. Check your wallet page for the correct total balance."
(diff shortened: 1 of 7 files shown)


seanyesmunt:
Thanks for starting on this @btzr-io 

When I try lbry://@mime#6/obj#6, it looks like it tries to render the video player instead of the 3d viewer.

I think https://github.com/lbryio/lbry-desktop/blob/master/ui/redux/selectors/content.js#L190-L192 needs to be moved to the top of the that function. I think `application/octet-stream` is catching for some 3d files.

tzarebczan:
Yea that metadata is before we had the proper one for 3d files. I'd worry less about that one and test with more recent stl /obj samples.

seanyesmunt:
Alright. It worked for other files so I think this is good.

btzr-io:
> When I try `lbry://@mime#6/obj#6`, it looks like it tries to render the video player instead of the 3d viewer.

This issue is not exclusive to 3D files, https://github.com/lbryio/lbry-desktop/issues/4075

I'm still not sure why the app forces `application/octet-stream` files to render as video:
https://github.com/lbryio/lbry-desktop/blob/eaca77d39430fa6f64af054ac3705c687935c18f/ui/constants/claim.js#L14

Issue #3466, opened 2020-01-09
Use channel name if title doesn't exist on support/tipping modal


tzarebczan:
![image](https://user-images.githubusercontent.com/8120721/72093278-13898a00-32e2-11ea-96f7-69f85105e6c6.png)

PR #4091, opened 2020-04-28
Use channelName if title empty on support/tipping modal

vv181:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #3466

### DIFF
--- a/ui/component/walletSendTip/view.jsx
+++ b/ui/component/walletSendTip/view.jsx
@@ -110,14 +110,16 @@ function WalletSendTip(props: Props) {
             <React.Fragment>
               {claimIsMine || isSupport
                 ? __(
-                    'This will increase the overall bid amount for ' +
-                      (title || '@' + channelName) +
-                      ', which will boost its ability to be discovered while active.'
+                    'This will increase the overall bid amount for %title%, which will boost its ability to be discovered while active.',
+                    {
+                      title: title || '@' + channelName,
+                    }
                   )
                 : __(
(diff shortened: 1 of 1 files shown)


ui/component/walletSendTip/view.jsx
```diff
@@ -97,12 +110,14 @@ function WalletSendTip(props: Props) {
             <React.Fragment>
               {claimIsMine || isSupport
                 ? __(
-                    'This will increase the overall bid amount for %title%, which will boost its ability to be discovered while active.',
-                    { title }
```
seanyesmunt, line 101: We need to keep the current structure of the string here so it works well with translations.

You should be able to keep everything the same, but also add the channel name after the title

```
...
{ title || `@${channelName}` }
```

vv181:
```
`This will increase the overall bid amount for ${title ? '%title%' : `%channelName%`}, which will boost its ability to be discovered while active.`,
{ title, channelName }
```

Is it the correct approach? 
I've tested it and it shows the localization string on the modal that has a title but not when using channelName.

```
...
{ title || `@${channelName}` }
```
Or using that is the correct one?

seanyesmunt: The second one is correct

seanyesmunt:
Fixed with #4091

Issue #2775, opened 2019-08-22
Save media position in viewers


tzarebczan:
For viewers that support volume and position, we should be saving these at least to local state (and eventually position can be stored in the KV wallet possibly).

seanyesmunt:
Volume is now saved.

jeffslofish:
I'm going to work on this issue now

PR #4098, opened 2020-04-29
Save media position to local storage.

jeffslofish:
Closes #2775

## PR Checklist

Please check all that apply to this PR using "x":

- [x] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)
- [x] I have checked that this PR does not introduce a breaking change
- [ ] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting)
- [ ] Refactoring (no functional changes)
- [ ] Documentation changes
- [ ] Other - Please describe:

## Fixes

Issue Number: #2775

## What is the current behavior?
The media player always starts at the beginning of the video

## What is the new behavior?
The position of the media player is saved to local storage and the next time that video is played, the video will start at the position it was last played at.

## Other information

The following is a screenshot of what the local storage will start to look like with this solution:
Screen Shot 2020-04-29 at 12 10 12 AM

PR #4104, opened 2020-04-30
Save media position in video viewer

jeffslofish:
PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #2775

Videos start playing at beginning of the video.

## What is the new behavior?
Video position is saved when being played. If a previously viewed video is played again, it starts at the saved location.


jeffslofish:
I abandoned my previous pull request (#4098) and created this clean one with the changes you wanted.

seanyesmunt:
Nice! I like the minimal changes. Do you notice any issues writing that value so often? If so, maybe it should only be written every x seconds or so? Or when the video stops?

jeffslofish:
I didn't notice any issue, but perhaps there would be on some devices. Let me see what I can do to make it write less often (hopefully just once)...

jeffslofish:
I updated the code to only save the media position on pause or dispose (in case they leave the page without stopping the video first) instead of every timeupdate event. This should help out performance and save the position in the majority of cases.

PR #4107, opened 2020-04-30
new og image

seanyesmunt:
I was slowly dying from the old og image with the blue search bar


static/index-web.html
```diff
@@ -69,7 +69,7 @@
     <meta property="og:title" content="lbry.tv" />
     <meta property="og:site_name" content="lbry.tv" />
     <meta property="og:description" content="Meet LBRY, a content platform owned by everyone (and no one)." />
-    <meta property="og:image" content="https://lbry.tv/v1-og.png" />
+    <meta property="og:image" content="https:/lbry.tv/v2-og.png" />
```
HarshKhandeparkar, line 72: `//` not `/`

Issue #3926, opened 2020-03-31
reselect channel as "replying as" when switching channels


tzarebczan:
If you click reply and then switch channels, the channel is not updated. Thanks to @er013 for the catch.

bitcoinbrisbane:
@tzarebczan confirm this test scenario: 

- Given a channel,
- And it has a comment,
- And I click reply,
- And I add a message,
- When I switch channels,
- Then the channel

jtwebman:
Going to give it a try to fix this as the first issue I try in this codebase.

tzarebczan:
@bitcoinbrisbane / @jtwebman thank you!!

bitcoinbrisbane:
> Going to give it a try to fix this as the first issue I try in this codebase.

How did you go?  I was looking at it too.

jtwebman:
The merge above https://github.com/lbryio/lbry-desktop/pull/4110 fixed it.

tzarebczan:
Woohoo!

PR #4110, opened 2020-05-01
Fix channel reply switching channels.

jtwebman:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #3926 

Currently, the usePersistedState acts just like a normal useState call and would be different in each component. So if you are on the channel comment page, hit reply on an existing comment, then change the top-level comment dropdown it doesn't change the same state in the reply comment. This doesn't seem to be correct as this seems to be a way to persist and share state globally.

## What is the new behavior?

The change was to make it a global state if you use the same key. Since it is storing in local storage if it exists was the desired outcome. It also passes a setValue function back that if a key is passed in will set all components currently mounted states not just one.

## Other information

I wasn't 100% on if you wanted to do it this way or use a context instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add fallback when images fail to load _community pr!_ ([#4019](https://github.com/lbryio/lbry-desktop/pull/4019))
 - Fetch new content when clicking LBRY logo while on homepage _community pr!_ ([#4031](https://github.com/lbryio/lbry-desktop/pull/4031))
 - Aligns text across browsers and desktop _community pr!_ ([#4050](https://github.com/lbryio/lbry-desktop/pull/4050))
-- Reselect channel as "replying as" when switching channels ([#3926](https://github.com/lbryio/lbry-desktop/issues/3926))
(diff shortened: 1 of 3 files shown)


seanyesmunt:
Thanks for the PR @jtwebman!

I'm not sure what the issue is, but when I tried to run your branch, my browser locked up. I think there is some infinite loop somewhere.

jtwebman:
@seanyesmunt Ok looking into it more

jtwebman:
@seanyesmunt I am running in the browser now with no issue. Can you tell me which browser you are using? I am running in the latest Chrome and all it is doing is calling wallet_balance call every few seconds. I see no increase in memory or anything like a loop when profiling it.

jtwebman:
@seanyesmunt Ok have the fix in. There is no reason to persist the file viewer element.

seanyesmunt:
@jtwebman Good find! You're definitely right, that shouldn't be persisted. Thanks for fixing that!

tzarebczan:
@jtwebman thanks so much for the PR!  Can we show you some [appreciation for the contribution](https://lbry.com/faq/appreciation)? 

Also, please sign up for the dev mailing list at lbry.tech to stay up to date with us!

PR #4142, opened 2020-05-06
Allow video sharing with start timestamp

### LINKED ISSUE #3122: Sharing videos with a start timestamp
Similar to youtube, we should be able to share a URL to a video at a certain timestamp within the video.

jeffslofish:
Closes #3122

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #3122

Sharing videos always starts at the beginning of the video using the share modal.

## What is the new behavior?
When the share model opens, the start time field is auto-populated with the current time of the video. The use may then check a checkbox to share the video starting at that point in time. Also, the time field is editable, so the user can adjust as they wish. 

The "LBRY Link", the "Embedded" link, as well as all the social sharing links will include a "t" query param in the sharable url (if the checkbox is checked) which will start the shared video at the desired time.


ui/component/socialShare/view.jsx
```diff
@@ -16,29 +19,56 @@ type Props = {
...
-  const { claim, title, referralCode, user, webShareable } = props;
+  const { claim, title, referralCode, user, webShareable, position } = props;
   const [showEmbed, setShowEmbed] = React.useState(false);
   const [showExtra, setShowExtra] = React.useState(false);
+  const [includeStartTime, setincludeStartTime]: [boolean, any] = React.useState(false);
+  const [startTime, setStartTime]: [string, any] = React.useState(secondsToHms(position));
+  const [startTimeSeconds, setStartTimeSeconds]: [number, any] = React.useState(Math.floor(position));
   const isMobile = useIsMobile();
 
-  if (!claim) {
```
seanyesmunt, line 27: Why move this? Right now it could fail below if `claim` doesn't exist 

here:
https://github.com/lbryio/lbry-desktop/pull/4142/files#diff-87e5aa61f17fe660285218396b512969R46

ui/component/socialShare/view.jsx
```diff
@@ -16,29 +19,56 @@ type Props = {
...
+  let claimId = '';
+
+  if (claim) {
+    canonicalUrl = claim.canonical_url;
+    permanentUrl = claim.permanent_url;
+    name = claim.name;
+    claimId = claim.claim_id;
   }
 
-  const { canonical_url: canonicalUrl, permanent_url: permanentUrl, name, claim_id: claimId } = claim;
```
seanyesmunt, line 31: I would like to keep the claim destructuring here

ui/scss/component/section.scss
```diff
@@ -138,3 +138,15 @@
 .section__actions--no-margin {
   margin-top: 0;
 }
+
+.section__start-at {
+  display: flex;
```
seanyesmunt, line 143: adding `align-items: center` will center everything

ui/util/url.js
```diff
@@ -70,3 +70,41 @@ exports.generateInitialUrl = hash => {
   }
   return url;
 };
+
+exports.generateLbryUrl = (canonicalUrl, permanentUrl) => {
```
seanyesmunt, line 74: This isn't really a lbry url. Maybe it should be called something like `generateLbryContentUrl`?

ui/component/socialShare/view.jsx
jeffslofish: I didn't move it originally, but when I tried to commit, it gave me an error saying that you can't return conditionally before calling React.useState(). So I moved it after all the React.useState() calls, and tried to make the code work if claim did happen to be null. I am not really happy with it myself, but I couldn't figure another way around the problem. This is also why I got rid of the claim destructuring, so the code wouldn't give an error if claim happened to be null. It looks like I missed one line of code that you pointed out. Can you suggest any better ways to handle the problem?

seanyesmunt: Ah right.

I don't think you need to `useState` for the urls.

You could just do `const encodedLbryURL = generateEncodedLBRYURL(...`
Same with the `dotCom` url

jeffslofish: I am using state for the URLs so that the user can change the time if they want and the URLs will update automatically. Do we want to lose that functionality?

seanyesmunt: We can still keep that functionality. If you move the URLs out of state, they will be re-created on every render. If you `setStartTime`, all of the other values can be built from that. Even the other time now that I"m looking it closer.
```
const [startTimeSeconds, setStartTimeSeconds]: [number, any] = React.useState(...)
const startTime = secondsToHms(startTimeSeconds);
const encodedUrl = generateEncodedUrl(...., includeStartTime, startTime)
const OpenDotLbryDotComUrl = generateOpenDotLbryDotComUrl(..., includeStartTime, startTime);
```

Then you can just call `setStartTimeSeconds` and everything else with update automatically. It should simplify this code quite a bit

jeffslofish:
Thanks for you help! 

All of the issues you mentioned should be resolved now, with much cleaner code =)

PR #4196, opened 2020-05-15
Tv web rebase

jessopb:
This still has a language choosing bug when switching from default (i.e. 'es') to english.
Tested in
yarn dev
yarn dev:web
yarn dev:web-server
yarn run:web-server

need to test desktop build


jessopb:
Notified @nikooo777  about the lbrytv to web folder rename and that it would affect ansible scripts. This may require coordination at merge?

ui/component/header/view.jsx
```diff
@@ -131,7 +132,7 @@ const Header = (props: Props) => {
         <div className="header__navigation">
           <Button
             className="header__navigation-item header__navigation-item--lbry header__navigation-item--button-mobile"
-            label={__('LBRY')}
+            label={__(LOGO_TITLE)}
```
seanyesmunt, line 135: This shouldn't be inside `__()`

web/stubs/electron.js
```diff
@@ -1,15 +1,18 @@
...
 export const remote = {
   dialog: {
     showOpenDialog: callable,
   },
   getCurrentWindow: callable,
   app: {
     getAppPath: callable,
-    getLocale: () => 'en',
+    getLocale: () => {
+      console.log('electron stub language');
```
seanyesmunt, line 13: console.log

web/.env.defaults
```diff
@@ -0,0 +1,15 @@
+# Copy this file to .env to make modifications
```
seanyesmunt, line 1: There are two of these? Is that correct?

The other one is up a level

config.js
```diff
@@ -1,13 +1,16 @@
+require('dotenv-defaults').config({silent: false});
```
seanyesmunt, line 1: I would add a comment here explaining that `.env.defaults` is the base level config

custom/homepage.example.js
```diff
@@ -0,0 +1,172 @@
...
+
+  // if (showPersonalizedChannels) rowData.push(RECENT_FROM_FOLLOWING);
+  // if (showPersonalizedTags && !showIndividualTags) rowData.push(TRENDING_FOR_TAGS);
+  // if (showPersonalizedTags && showIndividualTags) {
+  //   individualTagDataItems.forEach((item: RowDataItem) => {
+  //     rowData.push(item);
+  //   });
+  // }
+  // rowData.push(TOP_CONTENT_TODAY);
+  rowData.push(LATEST_FROM_LBRYCAST);
```
kauffj, line 164: I like this pattern, good call

web/.env.defaults
jessopb: Web packs from inside web folder. /web/.env is read for web.
App packs from root folder.

PR #4206, opened 2020-05-17
Make embeds work without sesion or local storage

### LINKED ISSUE #4017: don't use localstorage in Embed if not available
Brave, with shields, turned on blocks localstorage from being accessed via an iframe. Can we see if this happens and not read/store anything but still allow the embed? 

image

jeffslofish:
Closes #4017

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #4017

Embeds don't work in Brave with shields up because session and local storage is trying to be accessed.

## What is the new behavior?
Embeds work in Brave with shields because session and local storage access is avoided.

## Other information
I tested this by creating a new local web site with an embedded iframe from a share link. I then modified the host from lbry.tv to localhost:9090. Without my changes I got errors and the video wouldn't load, but with my changes, the errors went away and the video played.

Note on changes: the 'startup' sesionStorage item is only used on the app, so I made that code that sets this item only appear in the app code.

The second change where I move IS_WEB to the front of the || will short circuit the || if it is web and hence the window.localStorage.getItem() code will not run on the web.

When testing, I did still get this error:
```
createPersistoid.js?d203:97 Error storing data Error: No available storage method found.
    at driverPromiseLoop (localforage.js?69ae:2714)
```
but the video played fine, so I am assuming we can ignore it?

### DIFF
--- a/ui/index.jsx
+++ b/ui/index.jsx
@@ -275,7 +275,9 @@ function AppWrapper() {
 
   useEffect(() => {
     if (readyToLaunch && persistDone) {
+      // @if TARGET='app'
       sessionStorage.setItem('startup', true);
+      // @endif
       app.store.dispatch(doUpdateIsNightAsync());
       app.store.dispatch(doDaemonReady());
       app.store.dispatch(doBlackListedOutpointsSubscribe());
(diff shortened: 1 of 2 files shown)


seanyesmunt:
Thanks for fixing this @jeffslofish

seanyesmunt:
Can you add a changelog addition?

jeffslofish:
Done.

seanyesmunt:
Thank you!

Issue #4202, opened 2020-05-16
Odd "file is now" text appearing incorrectly in translated Upload Success dialog.


infinite-persistence:
![bug](https://user-images.githubusercontent.com/64950861/82120616-feee5e80-97b9-11ea-91de-ad135472b173.png)

For reference, the following is the translation.  The variable needed to be moved to the front, and I believe it was done correctly, so I don't think this is causing the substitution error.

![translation](https://user-images.githubusercontent.com/64950861/82120652-4e348f00-97ba-11ea-889e-404bd4569c0d.png)

----
```
App | 0.45.2
Daemon (lbrynet) | 0.73.1
Platform | Linux (Linux-5.3.0-51-generic-x86_64-with-LinuxMint-19.3-tricia)
```

infinite-persistence:
Ah, found the culprit -- the variable just wasn't localized with `__( )`. 

The fix would be trivial, but I think a better fix would be to not break apart the strings for the translators.  #4239, if approved, would fix this.

Issue #4173, opened 2020-05-11
Message in `I18nMessage` not translated correctly


infinite-persistence:
In the error message below, it appears in English, but the embedded link was translated.

![partial](https://user-images.githubusercontent.com/64950861/81545497-8f3c3600-93ab-11ea-95ce-7174b9e110f2.png)

The string is part of `app-string.json` and has been translated. But since it was used in a `I18nMessage`, I'm guessing the string substitution doesn't work there?

```
    <I18nMessage
      tokens={{
        refreshing_the_app_link: (
          <Button
            button="link"
            className="load-screen__button"
            label={__('refreshing the app')}
            onClick={this.refresh}
          />
        ),
      }}
    >
      There was an error. Try %refreshing_the_app_link% to fix it. If that doesn't work, try pressing
      Ctrl+R/Cmd+R.
    </I18nMessage>
```
I then tried encapsulating the string with `{__(' ')}` and it seems to load the translated text, but it ends up adding the translated string into `app-string.json` in my workspace at runtime.  I'm not really sure what's going on, so I'll just submit this first.  I'll be willing to help if you can give me pointers.

---
```
yarn dev
commit: 7596fd64 (master)
```

tzarebczan:
Thanks for spotting and opening the issue! Maybe @seanyesmunt can point you in the right direction.

infinite-persistence:
I think I found the culprit.  PR coming up...

PR #4215, opened 2020-05-18
Changed how 'Blocked channels' are displayed on settings and a typo

TigerxWood:
Problems solved (second attempt):
A. Small typo for <div className="card__title card__t itle--deprecated"> (extra space in the word **t itle**)

B. More details about problems with languages:
1. The word blocked has two forms in other languages (one for plural and one for singular) and therefore must be put together with channel and channels words. Currently it can be translated in only one form.

2. In the case there are no blocked channels there is no need to show the 0 value and is more elegant the message 'You do not have blocked channels'

3.  The link to manage the channels should not be visible in the case that there is no channel blocked, because there is no channel that need to be managed.

And about your statement: "Your change would make it harder for other languages to change this sentence." 

It's quite the opossite, it's very clear and easy. They just have to translate (with examples in Romanian):
'%count% %channels%. ',     (leave it as it is)
'You don't have' = 'Nu aveți'
'You have' = 'Aveți'
'blocked channel' = 'canal blocat'
'blocked channels' = 'canale blocat**e**'  (see the change for plural forms for the word blocked, not only for channels word?)

I hope this time I explained better the need for the changes.
Thanks

Messages displayed:
Case with 0 blocked channels: You don't have blocked channels. (No Manage button displayed)
Case with 1 blocked channel: You have 1 blocked channel. **Manage** (manage link)
Case with 2 or more blocked channels: You have 2 blocked channels. **Manage** (manage link)

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number:

### DIFF
--- a/ui/page/settings/view.jsx
+++ b/ui/page/settings/view.jsx
@@ -263,7 +263,7 @@ class SettingsPage extends React.PureComponent<Props, State> {
       <Page className="card-stack">
         {!IS_WEB && noDaemonSettings ? (
           <section className="card card--section">
-            <div className="card__title card__t itle--deprecated">{__('Failed to load settings.')}</div>
+            <div className="card__title card__title--deprecated">{__('Failed to load settings.')}</div>
           </section>
         ) : (
           <div>
@@ -480,13 +480,15 @@ class SettingsPage extends React.PureComponent<Props, State> {
               <Card
                 title={__('Blocked Channels')}
                 actions={
-                  <p>
-                    {__('You have %count% blocked %channels%.', {
-                      count: userBlockedChannelsCount,
-                      channels: userBlockedChannelsCount === 1 ? __('channel') : __('channels'),
-                    })}{' '}
-                    <Button button="link" label={__('Manage')} navigate={`/$/${PAGES.BLOCKED}`} />
-                  </p>
+                <p>
+                  <React.Fragment>
+                    {__('%count% %channels%. ', {
+                      count: userBlockedChannelsCount === 0 ? __("You don't have") : __('You have') + ' ' + userBlockedChannelsCount + ' ',
+                      channels: userBlockedChannelsCount === 1 ? __('blocked channel') : __('blocked channels'),
+                    })}
+				            {<Button button="link" label={userBlockedChannelsCount === 0 ? null : __('Manage')} navigate={`/$/${PAGES.BLOCKED}`} /> }
+					       </React.Fragment>
+				        </p>
                 }
               />
             )}


seanyesmunt:
@tiger5226 I will merge the typo fix, but not the other fix. Please stop opening up PR's to change that string. Your change would require languages to translate multiple half-strings, instead of one single sentence.

TigerxWood:
Did you read my explanations? I give you more details why this is wrong as it is now.

TigerxWood:
And you sent the message to the wrong guy. I am TigerxWood not tiger5226.
The first time, some weeks ago, whe I maded the first change I knew there will be these problems and I suggested a more generic way, to just put "Blocked channels: %countchannels%", but you changed that and entered in more details with singular, plural forms. As we know the devil is in details. So, now if you want that details, then you need to make it right for all languages.

seanyesmunt:
Whoops.

I'm sorry I haven't given this as much time as I should. You are right, this should work for all languages, even if it makes the code more complex. We can merge this. Thank you for staying persistent!

TigerxWood:
Well, thank you, I almost worried about your decisions. That's why my name is tiger, because I have eyes of tiger and I don't back up when I know I'm right. Have a good day and I hope to see you soon with new bugs resolved.
Currently there are many half translated strings all over the app and lbry.tv, like those from #4202 and #4173 and I try to investigate them all because that make the application look bad, like a half finished app and not so proffesional.

infinite-persistence:
Hey guys,

I'm confused as it seems to be going backwards.  Isn't concatenating strings a bad approach, as it assumes all languages follow the same word structure?  Japanese (when supported) will not work due to this split as the order is almost always different.  I'm sure there are other affected languages as well.

      "you have 6 blocked channels" --> "6つのブロックされたチャンネルがあります"

Another odd things is that `"%count% %channels%. ": "%count% %channels%. "` will now appear in the next translation batch. I'm pretty sure translators won't know what to do with it and will just paste the same thing.  Since it was noted to be left as is, it feels like a waste of time for the translators.

The old `"You have %count% blocked %channels%.": "You have %count% blocked %channels%."` is a better solution as they can move the variables around.  

**Solution**
If the issue being solved here is about the plural form, the proper solution is to just have two strings, with _%channels%_ removed.
`"You have %count% blocked channels": "You have %count% blocked channels"`
`"You have %count% blocked channel": "You have %count% blocked channel"`
If a particular language doesn't have plural forms (e.g. ms), then both strings will simply be translated to the same value.  This will cover all cases, from a translator's point of view.

TigerxWood:
Yes, it seems some languages put the number first. I fixed the problem. Now all languages should be fine and easy to translate.

Issue #3440, opened 2020-01-03
Support proper i18n pluralization


kauffj:
We should use `__n` for numeric messages to support unlimited, language-specific messages.

More: https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html

marko-lorentz:
I saw that the number of places where `if`-statements are used in the code to pick the correct plural form is increasing. Having `if` in every piece of source code is a mess that makes it less readable and introduces new errors. It also assumes that the developer understands the plural rules of all language as he must add the correct `if`-blocks (for Russian this would be 4, for Arabic its up to 6).
All this should go to a central function that does the desired replacements automatically. In the source code, only a single method call should remain.

I would recommend to adapt the ICU recommendations for plural and gender oriented message formatting as far as possible. 

- The ICU format for JSON is (AFAIK partially) supported by Transifex, so there will be tool support for the translators.
- In the source code, [react-intl](https://github.com/formatjs/react-intl) or [Globalize](https://github.com/globalizejs/globalize) might be used to format the messages. These can be hidden behind a facade method like `__()`.

More information can be found [here](https://medium.com/i18n-and-l10n-resources-for-developers/the-missing-guide-to-the-icu-message-format-d7f8efc50bab)

Suisse00:
WARNING: No real experience with i18 and languages I work with only contains one plural form

On the code side it could be only one method with ONE specific translation key that, behind the scene, change according with the number.

Eg. __n('You have %n messages', n)

TLDR: Behind the scene it will look for the key 'You have %n messages' + `[${n}]`
   I added a kind of [n] at the end just to recognise such string, could append only the number

If n is 1 then it will look for 'You have %n messages[1]', if this is 3 'You have %n messages[3]', ...

Downsides:
- Programmers will need to flag potentiel plural keys (either to add them to a special file that will generate all translation keys for them, or they will have to create (not a good idea) X translation key
- Either programmers add some i18 logic per language to select up to N translation key (eg. french and english have only 2 plural format, >2 is the same as 2) or translators will need to copy the string into other translation keys (uh :( )
- Idealy translator shouldn't have to copy nor see plurals form that don't exist (like 3,4,5,6 that are the same as 2 for french/english)

marko-lorentz:
@Suisse00 The downsides you mention are nearly completely covered by the ICU model:

[...]

PR #4239, opened 2020-05-24
Fix localization concatenation + Cleanup unused strings

### LINKED ISSUE #4202: Odd "file is now" text appearing incorrectly in translated Upload Success dialog.

infinite-persistence:
This is a draft proposal before I go ahead to fully implement the changes.

Proposed changes:
1. Change all instances of concatenated strings into a single strings so that translators can handle them as a single entity.  Not all languages follow the same word structure as English.
2. Remove unused strings for (a) tidiness (b) save translators from working on unused strings.

See the commit messages for further details.

The last 2 commits are examples of the cleanup work;  if this PR is wanted, then more of these commits will come.

----
Fixes #4202 and others.


static/app-strings.json
```diff
@@ -77,7 +77,7 @@
   "Turkish": "Turkish",
   "Polish": "Polish",
   "Malay": "Malay",
-  "By continuing, you accept the": "By continuing, you accept the",
+  "By continuing, you accept the": "---obsolete--do-not-use---",
```
kauffj, line 80: just delete these

ui/i18n.js
```diff
@@ -3,8 +3,11 @@ let fs = require('fs');
 // @endif
 
 const isProduction = process.env.NODE_ENV === 'production';
+const obsoleteStrIdValue = '---obsolete--do-not-use---';
```
kauffj, line 6: drop this

infinite-persistence:
Ah, we can just safely delete strings.  Ok, will do.

infinite-persistence:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes
Issue Number: Fixes #4202 and others + general cleanup.

- There exists split sentences that is hard to translate in some languages.
- The are lots of obsolete strings that translators are wasting time over.

## What is the new behavior?
- Fewer occurrences of partially-translated strings.
- Less work for new translators (lesser unused strings).
- No more split sentences.

I still leave some unused strings around as they look useful for future use (e.g. those about encrypting wallets, etc.)

ui/component/userVerify/view.jsx
```diff
@@ -44,13 +44,13 @@ class UserVerify extends React.PureComponent<Props> {
                 rewards_program: (
                   <Button button="link" label={__('Rewards Program')} href="https://lbry.com/faq/rewards" />
                 ),
+                Refresh: <Button onClick={() => fetchUser()} button="link" label={__('Refresh')} />,
+                Skip: <Button {...skipButtonProps} button="link" label={__('Skip')} />,
               }}
             >
               If you'd like to participate our %rewards_program% to earn credits, please complete one of the steps below
```
TigerxWood, line 51: I think you should remove the LF (end of line) after word *below *, to be translatable with the string from app-strings.json file

ui/component/userVerify/view.jsx
```diff
@@ -73,7 +73,7 @@ class UserVerify extends React.PureComponent<Props> {
                 />
                 <p className="help">
                   {__('Standard messaging rates apply. LBRY will not text or call you otherwise. Having trouble?')}{' '}
-                  <Button button="link" href="https://lbry.com/faq/phone" label={__('Read more.')} />
+                  <Button button="link" href="https://lbry.com/faq/phone" label={__('Read more')} />.
```
TigerxWood, line 76: Why not make it 'Read More' (for consistency), and remove the duplicate 'Read more ' from strings?

ui/component/userVerify/view.jsx
infinite-persistence: Yeah, I noticed that with a few other sentences as well.  I agree that it makes string-searching hard(er).

I left it as is since there were few other similar instances, so I don't know if it was intentional or not.

infinite-persistence:
After all of this, I found https://github.com/lbryio/lbry-desktop/issues/3440.  I guess this could be a pre-cursor to #3440, but would be obsolete when that happens.

ui/component/userVerify/view.jsx
infinite-persistence: for future reuse e.g. 
```
<token:button:%read_more%>
"%read_more% or refresh the page to fix it."
```

TigerxWood: OK. then make it here 'Read more.' (It's a sentence) and leave also the 'Read more' in string list.

TigerxWood: But where did you find  this? I can't find it.
<token:button:%read_more%>
"%read_more% or refresh the page to fix it."

Also I searched for this strings below and I can't find them anywhere.
 "%duration% minute ago": "%duration% minute ago",
  "%duration% seconds ago": "%duration% seconds ago",
  "%duration% second ago": "%duration% second ago",
  "%duration%  ago": "%duration%  ago",

infinite-persistence: As I said, "future, e.g.".

As for %duration%, the string ID is generated on runtime.  Search for `%duration%`, or delete it on your local runtime and see it being automatically added.

TigerxWood: OK, also there are some strings like "Earn %rewards_link% for inviting your friends. Read our %referral_faq_link% to learn more." from "ui\component\inviteNew\view.js" that for some reasons are not translated. Everything seems fine in code, the text is already there in Transifex translated, but on the app interface is not translated. Maybe you can spot what could be the cause. I can't find anything wrong.

infinite-persistence: Ah, I've fixed that bug earlier this week, and Sean has merged into the next RC. There was a mistake in the `<i18nMessage>` that only translated the tokens and not the entire sentence.

Issue #508, opened 2017-08-27
Airplay support MacOS / Chomecast support


tumispro:
I understand this is not at all a priority, but it would be really cool if Apple Airplay functionality could be added to the Mac app. 
I understand an iOS app is underway as well, so native Airplay support on there is perhaps an option as well in the future :-)

tzarebczan:
Don't forget Chromecast support :)

eriknelson:
+1,must have feature for me.

alyssaoc:
open new issue for exploration

seanyesmunt:
Chromecast support seems like it should be possible. There is an [issue](https://github.com/electron/electron/issues/869) bake cast functionality directly into Eelectron, but it doesn't seem like that is going anywhere.

There are a couple options, https://github.com/thibauts/node-castv2-client, https://github.com/GPMDP/electron-chromecast

It looks like `node-castv2` has more support, and there is an example electron application using it
https://github.com/annexare/Chromecast

Most likely it will only work for a few media types, but I think this will definitely be possible.

We probably won't add airplay support, this is baked into macOS so I'm not sure what this feature would consist of anyway.

seanyesmunt:
My recommendation is to use `node-castv2` since there is an example application. We should be able to use that code without changing to much.

tzarebczan:
More and moe folks asking about this...perfect for lbry.tv too.

seanyesmunt:
We are now using video.js. This plugin should work:
https://github.com/silvermine/videojs-chromecast

mikadoplus:
Hi, already someone working on this ? I could work on it if it's ok, even though it's not a top priority.

seanyesmunt:
No one is working on this @mikadoplus. It's all yours! I'd love to have this in the app

mikadoplus:
Thanks @seanyesmunt, just not sure about Airplay, will continue to investigate on it, but Chromecast shouldn't be an issue. Let's dive into this ^^.

seanyesmunt:
No worries, just ignore Airplay for now. Chromecast support can be added first.

kauffj:
You will be a hero to the community if you add this @mikadoplus! This is one of the most requested features we haven't had the bandwidth for yet

mikadoplus:
Just teasing...

[...]

PR #4252, opened 2020-05-26
Chromecast integration (experimental)

mikadoplus:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/508

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "lbry",
-  "version": "0.46.0-rc.4",
+  "version": "0.46.0-rc.5",
   "description": "A browser for the LBRY network, a digital marketplace controlled by its users.",
   "keywords": [
     "lbry"


seanyesmunt:
Awesome! I'll review/test this later today or tomorrow.

seanyesmunt:
This looks like a great start! I'm ok using your changes in your fork if they are absolutely needed.

If we do, I would rather you publish that to NPM.

Can you also make sure this doesn't cause the web build to fail? When I run `yarn dev:web` if fails with a couple errors.

seanyesmunt:
I'm seeing this when I try to cast to my TV. It also doesn't actually stop the video that is playing when I click the "cast" button.

Screen Shot 2020-05-28 at 1 32 54 PM

mikadoplus:
Hello @seanyesmunt, no big deal for the errors on yarn dev:web, I saw them, and should work, indeed they already work on my side, but somehow the packages weren't up to date when I created the PR.
For the other error, that is what I said on my first comment. The problem here is that videos are played on a localhost URL with a specific port. So I assume Chromecast doesn't see it since it is playing on your mac or pc. I've tried with the web version "CDN Urls" and they worked. The main video stops and it's casted.

I'll investigate on this a bit further by tonight or tomorrow.

I think also that the error handling could be more intuitive for the user.

I see also on your screen some weird design glitches, I do not have, will investigate on that to.

seanyesmunt:
@mikadoplus Ah that makes sense. I'll try not on localhost!

mikadoplus:
Hi @seanyesmunt I've update the my last commit, the errors executing warn dev:web. Are now corrected. I've also improved the performance and user experience a bit. Let me know what do you think about this one.

Were you able to play non localhost videos ?

seanyesmunt:
I haven't tested that yet. Will do today. Can you rebase this PR so it doesn't include all of the extra commits?

mikadoplus:
Hi @seanyesmunt I just noticed I messed up this PR, didn't notice I've forced a push out here. Sorry about that. I should have created a new branch instead of continue commiting to the master one. I've pushed a new commit, but don't want to recreate a new PR if there is an other way to roll back this one. Maybe I'm not enough familiar with git but what would be the best way to do it ?

PR #4263, opened 2020-05-28
Fix a typo declare boolean

TigerxWood:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number:

Text is not translated because of the end of line that broke the string
## What is the new behavior?
Text will be translated

### DIFF
--- a/ui/component/claimInsufficientCredits/view.jsx
+++ b/ui/component/claimInsufficientCredits/view.jsx
@@ -7,7 +7,7 @@ type Props = {
   uri: string,
   fileInfo: FileListItem,
   isInsufficientCredits: boolean,
-  claimWasPurchased: boolea,
+  claimWasPurchased: boolean,
 };
 
 function ClaimInsufficientCredits(props: Props) {
@@ -24,8 +24,7 @@ function ClaimInsufficientCredits(props: Props) {
           reward_link: <Button button="link" navigate="/$/rewards" label={__('Rewards')} />,
         }}
       >
-        The publisher has chosen to charge LBC to view this content. Your balance is currently too low to view it. Check
-        out %reward_link% for free LBC or send more LBC to your wallet.
+        The publisher has chosen to charge LBC to view this content. Your balance is currently too low to view it. Check out %reward_link% for free LBC or send more LBC to your wallet.
       </I18nMessage>
     </div>
   );
--- a/ui/component/claimInsufficientCredits/view.jsx
+++ b/ui/component/claimInsufficientCredits/view.jsx
@@ -24,7 +24,8 @@ function ClaimInsufficientCredits(props: Props) {
           reward_link: <Button button="link" navigate="/$/rewards" label={__('Rewards')} />,
         }}
       >
-        The publisher has chosen to charge LBC to view this content. Your balance is currently too low to view it. Check out %reward_link% for free LBC or send more LBC to your wallet.
+        The publisher has chosen to charge LBC to view this content. Your balance is currently too low to view it. Check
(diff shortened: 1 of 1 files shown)


ui/component/claimInsufficientCredits/view.jsx
```diff
@@ -24,8 +24,7 @@ function ClaimInsufficientCredits(props: Props) {
           reward_link: <Button button="link" navigate="/$/rewards" label={__('Rewards')} />,
         }}
       >
-        The publisher has chosen to charge LBC to view this content. Your balance is currently too low to view it. Check
```
seanyesmunt, line 27: Is there an actual issue with this line? It is split into two lines by our autoformatter

TigerxWood: All the texts that have a LF in them are not translated ... it seems

infinite-persistence: - I think that is not true. LF has no impact.
- The translation fix for `<i18nMessage>` is already available in `master`, or you can try it in lbry.tv.
- As for the specific text above, it is not in Transifex yet.

TigerxWood: Of course this one it isn't included yet, but is similar with other that are. From my observations all the strings that were not translated had these difference in them. It's not about that prroblem with partial translation that you repaired. Their translation is missing completly. Also, I think it's not a good ideea to enforce an end of line in today responsive design pattern when are so many resolutions to display the same message. And with this occasion I just found another 3 situations like this, like "Sending information to third parties (e.g. Google Analytics or Mixpanel) allows us to use detailed⏎ analytical reports to improve all aspects of LBRY." (after 'detailed' word do you see th end of line?) and in app-strings.jsx with "Sending information to third parties (e.g. Google Analytics or Mixpanel) allows us to use detailed\n                analytical reports to improve all aspects of LBRY."  Well, kind of too many spaces. Do you think this will be translated? And if it is, they will look good in the same time? I will put them in the change list.
And don't think it's the fault of autoformatter, because there are many other long strings, formated also on 2 lines, but that aren't split by an hard LF by pressing Enter key.

infinite-persistence: We are mixing arguments here.

I believe Sean's question in this PR is specific to that `<i18nMessage>` entry, which the LF has no effect.  So, your change above doesn't do anything other than to make it easier for translators to find the text.  I was just responding to this point.

On the other hand, your new argument for the line break in `Sending information to third parties (e.g. Google Analytics or Mixpanel) allows us to use detailed\n` looks valid, but isn't what we are talking about here (probably in another of your PRs?), so I've leave that to the team.

TigerxWood: Well, as I said the effect is that because of that LF the text will be not mached with the string from app-strings.jsx and is a risky bet if it's there anyway, because some tranlators will be confused if to use it in that particular position or not. As you said earlier, not every language is the same as format and as lenght. So, in any circumstance inserting a forced LF I think is a bad decision.

seanyesmunt: @TigerxWood I closed a lot of the PR's you opened that just got rid of the newline on strings. This is added by our code formatted. If there is an actual error with newlines in strings, it should be fixed in this file

https://github.com/lbryio/lbry-desktop/blob/master/ui/i18n.js

I'm still not sure this is an issue though. The text wrapping to another line in the code will not affect how it's displayed in the app.

PR #4299, opened 2020-06-02
Make strings from fileDetails translatable

TigerxWood:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number:

### DIFF
--- a/ui/component/fileDetails/view.jsx
+++ b/ui/component/fileDetails/view.jsx
@@ -37,7 +37,7 @@ class FileDetails extends PureComponent<Props> {
     // Create path from name so the folder opens on click.
     if (fileInfo && fileInfo.blobs_completed >= 1 && fileInfo.download_path === null) {
       downloadPath = `${fileInfo.download_directory}/${fileInfo.file_name}`;
-      downloadNote = 'This file may have been streamed, moved or deleted';
+      downloadNote = __('This file may have been streamed, moved or deleted');
     }
 
     return (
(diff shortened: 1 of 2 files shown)


ui/component/fileDetails/view.jsx
```diff
@@ -113,11 +113,11 @@ class FileDetails extends PureComponent<Props> {
 }
 // move this with other helper functions when we re-use it
 function formatBytes(bytes, decimals = 2) {
-  if (bytes === 0) return '0 Bytes';
+  if (bytes === 0) return __('0 Bytes');
 
   const k = 1024;
   const dm = decimals < 0 ? 0 : decimals;
-  const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+  const sizes = [__('Bytes'), __('KB'), __('MB'), __('GB'), __('TB'), __('PB'), __('EB'), __('ZB'), __('YB')];
```
seanyesmunt, line 120: Maybe `Bytes` should be wrapped in `__()`, but I don't think the rest of these should be

TigerxWood: Well, it seems they should. For example im my language or in French MB is translated as MO (from mega octet) so KB, MB, GB etc will me KO, MO, GO etc.

seanyesmunt: Ok. I didn't know that! I think we should keep it as is. Can you rebase this PR? Then it's ready to merge.

PR #4301, opened 2020-06-02
Changes that closes #4192 issue

### LINKED ISSUE #4192: [French] Following menu/page title not great
### What is to be done?
Hi!
You use the translation key `Following` as for the menu option AND page title to list videos from your subscriptions.

- In french, if we are using the same word, it would need to be plural in this specific case. Which we can't because it would also change the button in a channel that allow you to follow/unfollow.

  Alternatively, we could use another word like `Subscriptions`, which seems to already make more sense in this particular case.

- Using `Following` also is also misused/would be misunderstood:
  - The french equivalent, in this context, is mainly used to mean "X that are following me". In fact, you mean the opposite, channels that I do follow.
  - The other thing that botter me (in english as well) is that we may think it is the page that will ultimately manage our subscriptions. Which is way off, it will list videos from our subscriptions.

With all those points it could make more sense to use `Subscriptions videos` or `Following channels` (which look weird to me in english) if we want to use the same word.

**NOTE:** I know this is an issue, I don't mean to have the right answer and I'm no way linguistics. Feel free to suggest or/and comment.

#### :mag_right: Files To Look Into:
For the menu item: https://github.com/lbryio/lbry-desktop/blob/master/ui/component/sideNavigation/view.jsx#L106

For the page title: https://github.com/lbryio/lbry-desktop/blob/master/ui/page/channelsFollowing/view.jsx#L28

### Suggested labels
help wanted, level 0, area: subscriptions (I think, otherwise/and channel), good first issue

Possibly: needs: exploration

TigerxWood:
Changed option in the search box about Followed tags and Followed channels in the side navigation bar (because is used the same string as in the status of a channel, we need to change that to be able to translate it different in other languages, also in English is more correct in this way)

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #4192


seanyesmunt:
we plan to stick with "following" instead of "subscriptions" for now so I am going to close this.

TigerxWood:
Well, then we must use different names for the text for "Following" from the channel page and another text for the sidebar panel. Currently is the same text and can't be translated separate. Any suggestion? Maybe in the side panel we can use "Followed channels" instead of Subscriptions and "Followed tags" instead of "My tags"? What do you think?

PR #4305, opened 2020-06-02
strip ga calls

jessopb:
Removes ga calls from app


ui/analytics.js
```diff
@@ -92,24 +86,18 @@ const analytics: Analytics = {
...
       });
     }
   },
   setUser: userId => {
-    if (thirdPartyAnalyticsEnabled && userId) {
-      ReactGA.set({
-        userId,
-      });
-
+    // SEND USERID TO MATOMO?
```
seanyesmunt, line 96: Should we do this?

PR #4339, opened 2020-06-07
force https thumbs in use-get-thumbnail

jessopb:


seanyesmunt:
I think this should go in the FileThumbnail component. And ChannelThumbnail

PR #4340, opened 2020-06-07
Closes issue #4192 (second attempt)

TigerxWood:
Because we must use different names for the text for "Following" from the channel page and another text for the sidebar panel and another for searh list options. Currently is the same text everywhere and can't be translated separate.
*And most interestingly*:
- the message that a new user is receiving after first channel **subscription** is "Awesome! You just subscribed to your first channel." with the label "Subscriptions 101" as you can see in the file ui\modal\modalFirstSubscription\viewjs.
- everywhere in the app (in over 20 files) is used the word subscriptions and subscribed as in:
 (import { selectSubscriptions } from 'redux/selectors/subscriptions';)
(subscriptions: selectSubscriptions(state),)
(subscriptions: subscriptionsReducer,)
- many commented lines in the code like:
     //   Populate a channels unread subscriptions or update the type
     // Remove multiple files (or all) from a channels unread subscriptions
     // If no channel is passed in, remove all unread subscriptions from all channels
    // if the user isn't sharing data, keep the subscriptions entirely in the app
   // They are sharing data, we can store their subscriptions in our internal database
    Lbryio.call('subscription', 'new', {
      channel_name: subscription.channelName, etc.
- even the files from redux (actions, reducers, selectors) folder are named subscriptions.js
- ui\store.js uses: const subscriptionsFilter = createFilter('subscriptions', ['subscriptions']); and const whiteListedReducers = [...   'subscriptions',];
and most interesting:
const sharedStateFilters = {
  tags: { source: 'tags', property: 'followedTags' },
  subscriptions: {
    source: 'subscriptions',
    property: 'subscriptions',
    transform: function(value) {
      return value.map(({ uri }) => uri);
    },
where we can see the use of the "subscriptions" and "followed tags" string also.

So with so many examples in the code, why not use the same term in the program interface too? But the main reason is that we need to separate the instances where the "following" word is used.

An alternative for using "Subscriptions" is that we can use "Followed channels", but I think is too long. So do you have any real reason for not using Subscriptions? After all, we must use other word in side panel instead of the "Following" which is inappropriate at least in other languages.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number:
#4192


seanyesmunt:
@TigerxWood Just because we use certain words in our code, doesn't mean we need to use the same words in the UI.

I'm going to close this. We do not plan to switch back to "subscribers" for the time being.

TigerxWood:
Again, did you read carefully what I said? That code examples was just as a side note to add an additional motive for using the word subscription. Personally I am not as attached about Subscription word, I choose it because was suggested in the opened issue by somebody. 
The main problem instead is, that we need to use 3 different words in those 3 different places: content page, side panel and filter in the search box. Choose what words you want, but we need to give a solution to this. Closing it I don't think is a solution without any proposed alternative. After all, it's an opened issue by someone else and a problem  for translation for many other languages, including my language. I saw this problem when I was translating it, but instead of opening an issue I tried to make a compromise, not very good it seems. When a friend asked me what means that word in the side panel I explained him why is this, because is the same word used in the content page to show you that you are following a channel and I was waiting to see if other languages have the same problem. And it seems they have even opened an issue about that before me.
So, my alternate solution to subscriptions, as I said above, are "Following channels" (in side panel), "Following" (on page content, as it is now) and "Followed" (in the search filter, where it's about filtering the followed content). If you have another suggestions I am glad to hear them, but we need to use 3 different words in those places.
And by the way, the same problem is with the word "Repost" from page content and filter from the search option, where should be another word, for example Reposted (which is more apropriate even in English, because it's about filtering only reposted content).

And if you want to understand why I am so invested in updating the strings (there are still more that are missing) and having a good translatable interface is because on one hand, a interface partially translated make an impresion in other languages of an application half baked, hard to understand, and on the other hand, and the main reason, is that many people from other languages, including myself, want to translate the main site, especially the FAQ section. Where, as we can see, are used many examples with screen captures in them. So, we can't make screen capture with half translated strings or innaproriate translations. It's not enough just to translate the application, just like any classic application. To understand fully what this application is about you need more informations that you can find it only on the main site, which should also be translated to make the app really international. That milestone I want to be reached as soons as possible. 

When I started to translate the site, soon I realised that I can't go further without a basic good translatable interface and I had to intervene in the source code to repair these problems. So, now instead of going forward with my translation, I am moving in circles, by fishing missing strings and trying to explain why they should be that or that. I am the main proofreader and translator in my language for the Proton applications. And I can assure you that this application is top notch in terms of professional translation in my language. Almost everyday, they update the strings for translations, even if is just because of a missing dot after a sentence. And I love their dedication for this. And I want that for LBRY also. To have a perfect translated interface that make a professional impresion in as many languages as possible. That's why is called Library after all, right? To be available to everyone.

I am sorry for the long message I know you are busy with the code, I just wanted you to understand my point of view that as exciting may be some new features to the app, they are not so appealing if the users can't understand them in their native language. And because in English some words can be used in different places with different meanings, in other languages the nuances make a big difference.

I hope that you will reconsider this closed issue #4192 and help me further to achieve my goal to make all strings translatable as soon as possible.
Have a great day

infinite-persistence:
If I understood both of their requests correctly, I believe it is something I've encountered before in my other projects.  

I think what they are saying is that while an English word can be re-used safely in different contexts, other languages might not.  The problem then lies in how we are using the English database (if memory serves) --> we assume the _key_ and _value_ is identical.

One possible fix to 
(1) use the English database the same way as other languages, and
(2) treat _key_ as a string ID.  In the string ID, we can add some "context" help in the _key_ to make it unique. (In my other project, we use an enum as the ID).

The end result would be something like this:

**English (all resolves to the same value):**
```
"Following !!Used in sidebar!!": "Following",
"Following !!Used in bottom bar!!": "Following",
"Following !!Used in other bar!!": "Following",
```
**French**
```
"Following !!Used in sidebar!!": "Abcdefg",
"Following !!Used in bottom bar!!": "Xcdhfue",
"Following !!Used in other bar!!": "Whatever",
```
**Client side:**
```
 label={__('Following !!Used in sidebar!!'} 
````

Changes needed:
1. Make English use the _value_, just like in other languages. 
3. The Transifex process probably needs to change a bit to accommodate this.

It's not a trivial change, but not too complex either.

TigerxWood:
Well I don't think the English is so poor in words and even in this case I think using the same English word in different contexts is inapropriate. And instead of complicating the source code we just need to use the right words as it need. Because as a programmer, maybe for you is very clear what it means, but is not necessarly the same for a user, especially for a new one. Even Youtube use the Subscriptions word for subscribed channels and Subscribe for the button to subscribe to a channel. So, I don't know why do we need to reinvent the wheel when there are already consecrated words and additionally want to attract more youtubers into the LBRY?

TigerxWood:
Ohh, I forgot. Many youtubers import the same videos from the Youtube or upload the same video to LBRY and Youtube and somewhere in the video they say "Hit that Subscribe button below to subscribe to my channel" So my question is: what subscribe button? I don't see any. So in this situation he looks like he don't know what he is talking, because there is no subscribe button, right? In this case we need to mirror exactly, where is the case, the words used in the Youtube interface.

PR #4341, opened 2020-06-08
Fix Repost input field being off-screen

### LINKED ISSUE #3697: Repost error not shown in mobile view
When using lbry.tv, going to a post and selecting to repost something, if the URL is not changed, the error handling and messaging are out of position and you cannot clear the page or move forward.

Suggest some better on-page instructions to the user and some mobile UI improvements to make messages and buttons accessible.

Screenshot attached.
Screenshot_20200216-073231

infinite-persistence:
## PR Checklist

## PR Type
- [x] Bugfix

## Fixes
Fixes #3697 

## What is the current behavior? What is the new behavior?
image

## Other information
See commit messages for rationale and decisions.

One minor "side effect" is that it pushes all the elements down when there is a need to wrap, but I think it's an ok compromise for a rare corner-case.

### DIFF
--- a/ui/scss/component/_comments.scss
+++ b/ui/scss/component/_comments.scss
@@ -57,6 +57,8 @@
 
 .comment__message {
   white-space: pre-line;
+  word-wrap: break-word;
+  word-break: break-all;
   margin-top: var(--spacing-s);
 }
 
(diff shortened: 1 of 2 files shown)


seanyesmunt:
Nice!

Issue #4358, opened 2020-06-09
The mute icon sometimes does not reflect actual state.


infinite-persistence:
**Bug**
The mute icon in the player does not reflect the actual state for this scenario.  It's very minor, but would probably impact the "tap to unmute" feature request.

**To Reproduce**
1. Open a video and mute it.
2. Click `Back` and close the floating player.
3. Open another video, and click to play (I had "autoplay" set to off). It should then start as muted, and if not, repeat 1 and 2 again.
4. Press 'm' to unmute.  You'll hear the sound, but icon stays as "muted".  You might need to wiggle the mouse to keep the controls visible.

If you enter into a video claim with audio played directly, then the 'm' button is in sync with the icon when pressed.

**Expected behavior**
Icons need to reflect actual state.

**System Configuration**
`0.46.1 yarn build`, but also seen in lbry.tv today.

PR #4365, opened 2020-06-10
Show "unmute" option on videos automatically muted by browser.

### LINKED ISSUE #4325: Show "unmute" option on videos automatically muted by browser
Some users may not notice that the video is muted automatically and the overlay would alert them to this. Youtube does something similar.

infinite-persistence:
## PR Type
- [x] Feature -- Fixes #4325 `Show "unmute" option on videos automatically muted by browser`

Per 4325:
>_Some users may not notice that the video is muted automatically and the overlay would alert them to this. Youtube does something similar._

## What is the new behavior?
image

## Other information
**Implementation:**
- The code is placed in `VideoJs` instead of `VideoViewer` as we need to control the video itself. It's more self-contained here, rather than trying to pass refs around between parent and child.
(_actually, I don't know how to control the player from the parent. I'd be curious to know how to do it in React as a side discussion_)
- useState cannot be used as it will cause a re-render when the hint it clicked and dismissed. The DOM is used to hide the button.

**Dev testing notes:**
[x] When manually unmute using regular button, the hint should go away.
[ ] When manually unmute using 'm', the hint should go away.
          This is actually #4358 `The mute icon sometimes does not reflect actual state.`.  It's either a videojs bug, or an issue with the redux dispatch.  I'm not sure how to address this corner-case. `player.muted()` is always `true` in this scenario.
[x] Should not appear if video failed to load.
[x] Don't show the button when in full screen (following Twitch)
[x] Test with "autoplay" on and off
[x] Test on floating player
[x] Test with entering via "back" vs. direct entry.
[x] Test with dev:web
[x] When video ends, hide it if still present.
[x] Add proper icon
[x] Hint sometimes doesn't appear on fresh lbry.tv. lbryTV: Doesn't appear when random-autoplay next clip. (fixed by changing to `player.one('play')` instead since `player.on('loadstart')` isn't the right event to use for this.)

### DIFF
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -7,7 +7,6 @@ import videojs from 'video.js/dist/alt/video.core.novtt.min.js';
 import 'video.js/dist/alt/video-js-cdn.min.css';
 import eventTracking from 'videojs-event-tracking';
 import isUserTyping from 'util/detect-typing';
-import isDev from 'electron-is-dev';
 
 export type Player = {
   on: (string, (any) => void) => void,
@@ -107,7 +106,7 @@ export default React.memo<Props>(function VideoJs(props: Props) {
       if (newState !== curState) {
         tapToUnmuteRef.current.style.visibility = newState ? 'visible' : 'hidden';
       }
-    } else if (isDev) {
+    } else if (process.env.NODE_ENV === 'development') {
       throw new Error('[videojs.jsx] Empty video ref should not happen');
     }
   }
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -114,12 +114,15 @@ export default React.memo<Props>(function VideoJs(props: Props) {
   function unmuteAndHideHint() {
     if (player) {
       player.muted(false);
+      if (player.volume() === 0) {
+        player.volume(1.0);
+      }
     }
     showTapToUnmute(false);
   }
 
   function onInitialPlay() {
-    if (player && player.muted()) {
+    if (player && (player.muted() || player.volume() === 0)) {
       // The css starts as "hidden". We make it visible here without
       // re-rendering the whole thing.
       showTapToUnmute(true);
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -1,13 +1,17 @@
 // @flow
 import React, { useEffect, useRef } from 'react';
+import Button from 'component/button';
+import * as ICONS from 'constants/icons';
 import classnames from 'classnames';
 import videojs from 'video.js/dist/alt/video.core.novtt.min.js';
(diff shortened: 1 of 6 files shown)


ui/component/viewers/videoViewer/internal/videojs.jsx
```diff
@@ -145,6 +193,11 @@ export default React.memo<Props>(function VideoJs(props: Props) {
 
       player = videojs(el, videoJsOptions, () => {
         if (player) {
+          player.one('play', onInitialPlay);
```
seanyesmunt, line 196: I think this needs to be `on`

ui/component/viewers/videoViewer/internal/videojs.jsx
```diff
@@ -93,6 +97,50 @@ export default React.memo<Props>(function VideoJs(props: Props) {
...
+  function showTapToUnmute(newState: boolean) {
+    // Use the DOM to control the state of the button to prevent re-renders.
+    // The button only needs to appear once per session.
+    if (tapToUnmuteRef.current) {
+      const curState = tapToUnmuteRef.current.style.visibility === 'visible';
+      if (newState !== curState) {
+        tapToUnmuteRef.current.style.visibility = newState ? 'visible' : 'hidden';
+      }
+    } else if (isDev) {
+      throw new Error('[videojs.jsx] Empty video ref should not happen');
```
seanyesmunt, line 111: Instead of `electron-is-dev`, use `const isProduction = process.env.NODE_ENV === "production"`

ui/component/viewers/videoViewer/internal/videojs.jsx
infinite-persistence: We only need to run this one time, so `one` is used.  If `on` is used, the button will re-appear every time "Play" is pressed while sound is muted.

seanyesmunt: Interesting. I didn't know videojs supported that.

I tried several videos that started muted and didn't see the button. Can you double-check that it's working?

infinite-persistence: Yeah, I just rebased to confirm the `one` before a few minutes ago and the button is working.

infinite-persistence: Tried a fresh clone on another linux VM and it's ok too.  Let me try on Windows.

infinite-persistence: Ok, if I clear my Electron settings and never touch the videojs player buttons, then I can replicate the issue.  Once I toggled the mute button, it will work from then on.  See if that works for you to test the other aspects.

This is again due to #4358 not reflecting the actual state of the audio.  Let me dig further.  
By the way, in `handleKeyDown` of `VideoJs`, do you know why the `HTMLVideoElement` was used when the `Player` is available? This is my wild guess on what's causing the mismatch.

infinite-persistence: I think I've figured it out.  The Mute and Volume are independent controls and we need to handle two types of "mute":
- Volume icon is pressed to mute
- Volume slider moved to 0%
Fix commited.

PR #4371, opened 2020-06-11
Add ability to support language subsets, with only CT and CS enabled for now

### LINKED ISSUE #4225: Add ability to support language subsets i18n
We have Chinese in our language selection (zh) but a community member really wants to add Taiwanese (zh_tw) to the app and gave a good description how it differs and is a legit request. Right now we don't support the languages on Transifex that have the underscore in their language code.

infinite-persistence:
**CT** - Chinese Traditional, **CS** - Chinese Simplified

## PR Type
- [x] Feature -- Fixes #4225 `Add ability to support language subsets i18n`

Per Julie:
> _We have Chinese in our language selection (zh) but a community member really wants to add Taiwanese (zh_tw) to the app and gave a good description how it differs and is a legit request. Right now we don't support the languages on Transifex that have the underscore in their language code._

I agree.  All translated products almost always support these 2 distinct languages.
If LBRY wants to support it, then this PR attempts to enable that.  

## What is the new behavior?
- "Chinese" will be split into "Chinese (Traditional)" and "Chinese (Simplified").
- It will be easier to split other common subsets like `pt-BR/pt-PT` and `en-US/en-GB` if there is demand and translator availability.

## Remaining Work Required
I don't have access nor the knowledge for the Transifex side:
- [ ] Create `zh-TW.json` and `zh-CN.json` in `https://lbry.com/i18n/get/lbry-desktop/app-strings/`
_It seems like the current Chinese translation is mixed.  But I'm not 100% sure, so we'll need to ask the translator.  If it's not mixed, we can just rename to the new format._

## Code changes (approach)
I don't think we need to support all subsets, so we should selectively choose what we want to support.  Given that, a whiltelist array `SUPPORTED_SUB_LANGUAGE_CODES` is added.  If the user's OS locale is a subset but not in this list, it will be "consolidated" (i.e. 'en-US' --> 'en'), which is the behavior of the existing version.

The change also involves using a string for the property name in order to handle the dashes.  I hope there is no rule in the coding standard that prohibits this.

## Notes for posterity
Note that if English subsets like `en-GB` is enabled in the future, the default `'en'` value used throughout the code (including in `redux.git`) needs to be changed to `'en-US'`.

## Testing notes
Anything I missed?
[x] Change OS locale to `zh-CN` and `zh-TW`.  Clear the Electron settings to really start from scratch.  The selection language in `Settings` should be the same as the locale.  The GUI will still be in English since the json doesn't exist.
[x] Test on Web
[x] Try hardcode `zh-CN.json` to `zh.json` to see if translation still works.
[x] Test the non-supported subset case. Switch the OS to `en-GB` to try.

### DIFF
--- a/electron/createWindow.js
+++ b/electron/createWindow.js
@@ -3,9 +3,22 @@ import { app, BrowserWindow, dialog, shell, screen } from 'electron';
 import isDev from 'electron-is-dev';
 import windowStateKeeper from 'electron-window-state';
 import SUPPORTED_LANGUAGES from 'constants/supported_languages';
+import { SUPPORTED_SUB_LANGUAGE_CODES, SUB_LANG_CODE_LEN } from 'constants/supported_sub_languages';
 
 import setupBarMenu from './menu/setupBarMenu';
 import * as PAGES from 'constants/pages';
+
+function GetAppLangCode() {
+  // 1. Gets the user locale.
+  // 2. Converts unsupported sub-languages to its primary (e.g. "en-GB" -> "en").
+  //    Note that the primary itself may or may not be a supported language
+  //    (up to clients to verify against SUPPORTED_LANGUAGES).
+  const langCode = app.getLocale();
+  if (langCode.length === SUB_LANG_CODE_LEN && !SUPPORTED_SUB_LANGUAGE_CODES.includes(langCode)) {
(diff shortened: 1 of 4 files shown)


kauffj:
@seanyesmunt please merge this

PR #4378, opened 2020-06-13
Limit description contents to 5k chars

### LINKED ISSUE #4369: Limit description contents to ~6K chars
If you create a description that's too long, the publish will fail with a claim script too large error. We can either handle that and tell the user it's too long, or impose a limit. A user mentioned 7K works, but it will vary on other areas like tags/title/name

infinite-persistence:
## PR Type
- [x] Bugfix -- Fixes #4369 `Limit description contents to ~6K chars`
> _If you create a description that's too long, the publish will fail with a claim script too large error. We can either handle that and tell the user it's too long, or impose a limit. A user mentioned 7K works, but it will vary on other areas like tags/title/name_

## What is the new behavior?
- Limit the `Description` form to 5K.  I choose 5k instead of the suggested 6k because
     (1) parity with YT
    (2) it's a nicer number
    (3) more buffer from the roughly-known 7k max
- Indirectly fixes a bug whether Basic and Advanced Editor are currently having different limits (2000 vs unlimited, respectively) and can cause truncation when switching between the two.

## Indirect side effects
- Advanced Editor is no longer having unlimited length.  `Comment` is now 2000, and `Description` is 5000, and `Default` (all other form fields) is 2000.

## Known issue(s)
- If the `Description` consists of entirely 5000 CJK characters, the transaction would still fail (doesn't solve the initial problem).  My guess is this is because CJKs are multibyte per character.
However, it should still be good to have a limit to cover the common cases.


seanyesmunt:
Nice one!

PR #4388, opened 2020-06-15
wider diff in header sizes for md posts

lyoshenka:


seanyesmunt:
I changed the sizes a bit so that h2s weren't bigger than the claim title h1.

PR #4392, opened 2020-06-16
Persist trending selectors between tags

### LINKED ISSUE #2679: Persist trending selectors between tags
## The Issue
When switching between tags, the selector defaults back to Trending even though you had another option already selected. I think it makes more sense to persist them. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

infinite-persistence:
## PR Type
- [x] Bugfix • Fixes #2679 `Persist trending selectors between tags`
> _When switching between tags, the selector defaults back to Trending even though you had another option already selected. I think it makes more sense to persist them._

Also, added persistence to `Channel` selection in `Publish` Page per discussion with Tom.

Skip the suggested change in the wallet for now until this method is reviewed.  There might be better ways to handle the history.

## Code changes
- `orderParamUser` will store the last user state persistently. The persistent state is also made unique for each page (i.e. "Your Tags" and "All Content" will have unique states).
- If the parent component passes in a specific order, that will be respected and will also become the new persisted value. One example is "Your Following", where it always starts at 'New'.
- Handling navigation history was a bit troublesome:
The test case:
   - Enter "Your Tags" (assume start at 'Trending')
   - Click 'New'
   - Click 'Top'
   - Back
   - Back (it should return to 'Trending')
As the top page history does not have any `?order=` value, we ended up with a no-op for the last Back. `orderParamEntry` is added to handle this, giving it a value to fall back to.

### DIFF
--- a/ui/component/claimListDiscover/view.jsx
+++ b/ui/component/claimListDiscover/view.jsx
@@ -2,6 +2,7 @@
 import type { Node } from 'react';
 import classnames from 'classnames';
 import React, { Fragment, useEffect, useState } from 'react';
+import usePersistedState from 'effects/use-persisted-state';
 import { withRouter } from 'react-router';
 import * as CS from 'constants/claim_search';
 import { createNormalizedClaimSearchKey, MATURE_TAGS } from 'lbry-redux';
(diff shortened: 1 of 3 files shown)


seanyesmunt:
This is awesome! Nice work!!

PR #4399, opened 2020-06-17
Tip-Modal: Add settings persistence and UX improvements.

### LINKED ISSUE #4394: add persistence to new support modal
Should save my tip/support preference and last amount (button) and custom amount.

infinite-persistence:
## PR Type
- [x] Improvement
1. Fixes #4394 `add persistence to new support modal`
   >_Should save my tip/support preference and last amount (button) and custom amount._
2. Should collapse the custom field when presets are selected.
   _Even though the Send button repeats the amount for Tip, it's still confusing to see both the Preset being active and Custom numeric field visible._

3. Missing localization tags.


seanyesmunt:
This is great. Definitely an improvement over the current modal. Thank you! 🙂

Issue #4367, opened 2020-06-10
Limit channel tags to reduce tag spam


jsigwart:
### Proposed: Allow 5 tags max for channels like we do for content publishes.

Inspired by this claim:
```{
  "@Coronavirus-covid19#3": {
    address: "bEQjFksKN5MynGazyeDnYGMndDpSmNqFG7",
    amount: "0.01",
    canonical_url: "lbry://@Coronavirus-covid19#3",
    claim_id: "3755d36cc5067abc8ae72dd4329f138b5fcb7f16",
    claim_op: "update",
    confirmations: 12795,
    has_signing_key: false,
    height: 766204,
    meta: {
      activation_height: 761547,
      claims_in_channel: 80,
      creation_height: 761547,
      creation_timestamp: 1589021873,
      effective_amount: "29.01",
      expiration_height: 2863947,
      is_controlling: true,
      reposted: 0,
      support_amount: "29.0",
      take_over_height: 761547,
      trending_global: 0,
      trending_group: 0,
      trending_local: 0,
      trending_mixed: 0
    },
    name: "@Coronavirus-covid19",
    normalized_name: "@coronavirus-covid19",
    nout: 0,
    permanent_url: "lbry://@Coronavirus-covid19#3755d36cc5067abc8ae72dd4329f138b5fcb7f16",
    short_url: "lbry://@Coronavirus-covid19#3",
    timestamp: 1589778063,
    txid: "2e504b218751a20e603d86ee0b2e55df3096e32591e6ac79c27f325c6a358523",
    type: "claim",
    value: {
      cover: {
        url: "https://spee.ch/b/a2bda4fb9ea0ed4d.jpg"
      },
      description: "The channel is dedicated to Coronovirus ... funny photos as well as formidable ...",
      public_key: "3056301006072a8648ce3d020106052b8104000a034200040bf434475175fbc6e0211d989484c5b090d45072d45d620a512a526b63ce1a0e21f89ce9e92dbd4e475a89b495d9dd3e37535160b3c6337c21cf0720ac9c5944",
      public_key_id: "bWJKGncGVf7814wSU3AZTNqE7Zai1124jr",
      tags: [
        "19th century",
        "4k",
        "activism",
        "adventure",
        "agnostic",
        "aliens",
        "altcoins",
        "amazon",
        "anarchy",
        "android",
        "animal",
        "animals",
        "animation",
        "anime",
        "apple",
        "arcade",
        "ark",
        "art",
        "artist",
[...]

PR #4403, opened 2020-06-18
Inf-scroll: (1) Reset to top (2) Fix "page vs. results" being out of sync

### LINKED ISSUE #2669: Page does not restore to top when navigating new tags
## The Issue
If you scroll down one of the trending lists and then click a tag, it goes to the tag page but not to the top of it. It may cut off the header and some results. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

infinite-persistence:
## PR Type
- [x] Bugfix
1. Fixes #2669 `Page does not restore to top when navigating new tags.`
   To replicate the issue, you will need to first visit the tag and scroll down, move to another tag, then re-visit the first tag. Querying new tags won't show the issue. 

2. "Encountered children with the same key" error (duplicate entries shown).
(Also reported by https://github.com/lbryio/lbry-desktop/issues/4367#issuecomment-645449206)

## Code changes
1. Ignore the history if it's a new query (i.e. explicitly clicked). The BACK history will still behave as normal (doesn't reset to top).

2. Previously, the `page` variable will continue to increment as you scroll and stay within the page (e.g. Trending vs New, or clicked another Tag). As you move between queries, we hit a scenario where `page` is significantly under or over the latest retrieved `claimSearchResult.length`. This messes up the rest of the code.

   Fix by correcting the value of `page` according to the current `claimSearchResult.length` when necessary.

## Notes
The new function introduced might be a bit inefficient.

### DIFF
--- a/ui/component/claimListDiscover/view.jsx
+++ b/ui/component/claimListDiscover/view.jsx
@@ -295,13 +295,32 @@ function ClaimListDiscover(props: Props) {
   const claimSearchCacheQuery = createNormalizedClaimSearchKey(options);
   const claimSearchResult = claimSearchByQuery[claimSearchCacheQuery];
 
+  const [prevOptions, setPrevOptions] = useState(options);
+
+  if (!isJustScrollingToNewPage(prevOptions, options)) {
+    setPrevOptions(options);
+
+    if (didNavigateForward) {
+      // --- Reset to top.
+      window.scrollTo(0, 0); // Prevents onScrollBottom() from re-hitting while waiting for doClaimQuery():
+      options.page = 1;
+      setPage(options.page);
+    } else {
+      // --- Update 'page' based on retrieved 'claimSearchResult'.
+      options.page = Math.ceil(claimSearchResult.length / CS.PAGE_SIZE);
+      if (options.page !== page) {
+        setPage(options.page);
+      }
+    }
+  }
+
   const shouldPerformSearch =
     claimSearchResult === undefined ||
     didNavigateForward ||
     (!loading &&
       claimSearchResult &&
       claimSearchResult.length &&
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Resetting to the top of the page when clicking on a tag works, but it seems like there is something weird happening with scroll restoration when clicking "back".

I think it only happens when clicking a tag, scroll restoration after navigating to the file page seems to work well. Can you take a look at that?

infinite-persistence:
I only found the following issue during my testing.  As it is also present in today's code, I thought of looking at that  separately later.
Is it the same as what you are seeing, or a totally different sequence? 

1. Reload
2. Click TagA, scroll down
3. Click TagB, scroll down
4. Click TagC, scroll down
5. Click `back` = TagC's previous scroll location retained ✅
6. Click `back` = TagB's previous scroll location retained ✅
7. Click `forward` = scroll location is lost.

I'll re-visit the above in this PR, and any other bad sequence that you know of.  Let me know if you can still remember the sequence.

infinite-persistence:
@seanyesmunt, I still couldn't find any glitches with `back`, although I did find other issues that are present in `0.46.2`.  I'm interested to look at those as well, but I think it's cleaner to start another PR for that.

Other issues found but also present in `0.46.2`:
1. `forward` doesn't restore the scrolling position (because it has been popped, so no data to restore with).
2. Clicking on the same tag 5 times results in 5 pushes, so there will be 5 `back` that are no-op.  (perhaps this is the behavior that you saw?)
3. 4408 `Inf-scroll: doClaimSearch() sometimes returns duplicates, causing scrolling problems.`

seanyesmunt:
Yeah I'm not sure exactly what I was seeing. I think this is still an improvement over what we currently have so I think it's good to merge.

PR #4419, opened 2020-06-22
Pause 'Autoplay' when scrolled off screen.

### LINKED ISSUE #4021: Don't Allow Autoplay when commenting / taking other actions
There has been a few times when I would be commenting or replying to a comment on someone's video and the video would autoplay to the next video while I am actively typing. When I go back, text isn't there anymore, so I have to retype. Pretty annoying from a user exeperience standpoint.

infinite-persistence:
## PR Type
- [x] Bug   »   Fixes #4021 `Don't Allow Autoplay when commenting / taking other actions`

## Changes
- When the autoplay overlay gets partially off-screen, the timer will be stopped. We can tweak how much down it needs to be scrolled.
- When timer is paused, the countdown resets.
- The pausing mechanism has no effect on the floating player (since the 'top' can never go negative). This behavior is the same as another big video platform.
- Pause when any `Modal` is present.

## Aside
- Also present in `0.46.2`: I noticed that with the first video viewed from a fresh session, the timer takes a while to appear. (Update: seems like just need to scroll down for "Related"to load at least once).

### DIFF
--- a/ui/component/autoplayCountdown/view.jsx
+++ b/ui/component/autoplayCountdown/view.jsx
@@ -31,6 +31,7 @@ function AutoplayCountdown(props: Props) {
 
   const [timer, setTimer] = React.useState(countdownTime);
   const [timerCanceled, setTimerCanceled] = React.useState(false);
+  const [timerPaused, setTimerPaused] = React.useState(false);
 
   let navigateUrl;
   if (nextTitle) {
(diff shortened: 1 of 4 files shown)


ui/component/autoplayCountdown/view.jsx
```diff
@@ -52,22 +57,53 @@ function AutoplayCountdown(props: Props) {
...
+      timeoutId = setTimeout(() => {
+        timeoutId = null;
+        fn(...args);
+      }, time);
+    }
+  }
+
+  React.useEffect(() => {
+    const handleScroll = debounce(e => {
+      const elm = document.querySelector('.autoplay-countdown');
```
seanyesmunt, line 76: Can you extract this class to a constant at the top of this file?

ui/component/autoplayCountdown/view.jsx
```diff
@@ -52,22 +57,53 @@ function AutoplayCountdown(props: Props) {
     }
   }, [navigateUrl, nextRecommendedUri, isFloating, doSetPlayingUri, doPlayUri]);
 
+  function debounce(fn, time) {
```
seanyesmunt, line 60: This already exists in `util/debounce.js`

ui/component/autoplayCountdown/view.jsx
infinite-persistence: Updated.  Hopefully I understood the comment correctly.

ui/component/autoplayCountdown/view.jsx
```diff
@@ -5,6 +5,10 @@ import UriIndicator from 'component/uriIndicator';
 import I18nMessage from 'component/i18nMessage';
 import { formatLbryUrlForWeb } from 'util/url';
 import { withRouter } from 'react-router';
+import debounce from 'util/debounce';
+
+const DEBOUNCE_SCROLL_HANDLER_MS = 150;
+const CLASSNAME_AUTOPLAY_COUNTDOWN = '.autoplay-countdown';
```
seanyesmunt, line 11: This should just be 
`const CLASSNAME_AUTOPLAY_COUNTDOWN = 'autoplay-countdown';`

Then below on line 104, it can be changed to
`<div className={CLASSNAME_AUTOPLAY_COUNTDOWN}>`

infinite-persistence: Ah, ok.  Re-did the "cleanup" commit.

PR #4433, opened 2020-06-24
Option to remove abandoned claims from Blocked Channels page

### LINKED ISSUE #3800: allow unfollow / unblock of abandoned or blocked content
- [ ] if you've blocked channels that have been DMCA's, you can't remove them since the follow button is hidden and they are hidden from the block list

- [ ] if you've blocked channels that have been removed, you can't unblock from since they don't resolve

infinite-persistence:
## Fixes:
Fixes #3800 `allow unfollow / unblock of abandoned or blocked content`

## Changes:
If the `ClaimPreview` is in the Blocked Page, and if the claim itself is invalid, show a button to remove it from the blocked list.

image

## Notes:
- I do not know if this change covers the DMCA part of #3800 or not since I don't see any special handling for DMCA'd claims in the code to emulate it.  I assume `!claim` does cover that scenario; if not, I'll might need LBRY to DMCA one of my claims for testing.
- This change involves altering the wallet, so extra testing would be good.
- Making `ClaimPreview` specifically check if it's in the `PAGES.BLOCKED` does feels a bit dirty, but I opted for this as it's cleaner compared to passing the info down 2 layers through `props`.
```
List Blocked
    ClaimList
        ClaimPreview
```

### DIFF
--- a/static/app-strings.json
+++ b/static/app-strings.json
@@ -1201,7 +1201,7 @@
   "%view_count% Views": "%view_count% Views",
   "Tap to unmute": "Tap to unmute",
   "Playing in %seconds_left% seconds...": "Playing in %seconds_left% seconds...",
-  "Autoplay timer paused.": "Autoplay timer paused.", 
+  "Autoplay timer paused.": "Autoplay timer paused.",
   "0 Bytes": "0 Bytes",
   "Bytes": "Bytes",
   "KB": "KB",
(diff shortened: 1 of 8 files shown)


seanyesmunt:
@infinite-persistence I think the actual bug here is that `<AbandonedChannelPreview />` isn't being rendered properly here. That component already has a "remove" button
https://github.com/lbryio/lbry-desktop/blob/master/ui/component/claimPreview/view.jsx#L213

infinite-persistence:
🤦‍♂️ I missed such a clearly-named component.

New version coming up ...

infinite-persistence:
New version with the correct component used:
image

PR #4438, opened 2020-06-27
Fix incorrect creator status shown when navigating between channels

### LINKED ISSUE #3750: refresh youtube creator status when switching between channels
If you navigate to @jiggytom, then go to a youtube channel like @veritasium, and then click back to jiggytom it shows youtube from the previous one.

infinite-persistence:
Fixes #3750 `refresh youtube creator status when switching between channels`


infinite-persistence:
Is it ok if I referenced the `Issue` instead of the `PR` in the changelog?

seanyesmunt:
@infinite-persistence I would prefer the PR, but not a big issue.

seanyesmunt:
Can you rebase this? Then it's good to go

infinite-persistence:
The rationale behind using `Issue` is to avoid a force-push or additional commits since the PR number is unknown at first, but I guess it's not a big deal.  PR it is :)

Rebasing:

seanyesmunt:
Thanks!

PR #4449, opened 2020-06-30
Fix "Notification" page button being incorrectly disabled by 0 blocked channels

### LINKED ISSUE #4447: Notifications and Blocked "Manage" button doesn't work with 0 blocked channels
`Notifications` and `Blocked` _Manage_ button doesn't work with 0 blocked channels.
- The former should work regardless of blocked channel count?
- The latter is technically correct, but no harm to just enter the page since it will show a _When you block a channel, all content from that channel will be hidden._ message which seems useful for beginners?

image

infinite-persistence:
- Fixes #4447 `Notifications and Blocked "Manage" button doesn't work with 0 blocked channels`
- I skipped the second bullet point -- no strong arguments for either way.

### DIFF
--- a/ui/page/settings/view.jsx
+++ b/ui/page/settings/view.jsx
@@ -492,7 +492,6 @@ class SettingsPage extends React.PureComponent<Props, State> {
                         button="secondary"
                         label={__('Manage')}
                         icon={ICONS.SETTINGS}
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Whoops. Thanks for fixing this!

PR #4458, opened 2020-07-01
Fix video transcode setting not reflected correctly (MP3 incorrectly transcoded to MP4)

### LINKED ISSUE #4332: Video transcode setting not reflected correctly (MP3 incorrectly transcoded to MP4)
**Bug**
If you previously transcoded an MP4 file and published, and later want to publish an MP3, the MP3 gets incorrectly transcoded to MP4 despite the checkbox showing unchecked.  

I think the setting isn't being reflected correctly, and it's because MP3 are supposed to gray out the checkbox anyway, so it didn't "update" the checkbox when the file was selected, but the publish process still uses that setting.

**To Reproduce**
1. I selected a video file in an odd format, and check "trancode".  I then proceeded to publish that.
2. Hours later (I believe I have closed or restarted the app several times by then), I wanted to publish an MP3.  I selected the music file, and the "transcode" option remains grayed out and unchecked as normal since it's music.  However, when I publish, I noticed it was transcoded and published as MP4.

To fix this, I had to:
1. Go to Publish again
2. Select an MP4 file so that the "trancode" won't get grayed out.  And as I suspected, the checkbox is checked.
3. Uncheck the checkbox.
4. Cancel the publish.
5. Now I'm able to publish music!

**Expected behavior**
The GUI and the setting should match.  

**System Configuration**
```
App 0.45.2
Daemon (lbrynet) 0.73.1
Platform Linux (Linux-5.3.0-51-generic-x86_64-with-LinuxMint-19.3-tricia)
```

infinite-persistence:
## Issue
Fixes #4332 `Video transcode setting not reflected correctly (MP3 incorrectly transcoded to MP4)`

2 issues here:
(1) The checkbox is mixing between user state and logic state.
(2) The variables (e.g. `optimize`, `isVid`, `filePath`, etc) will have values from the previous operation when you enter Publish Page, so GUI issues beyond Transcode can be also produced (e.g. transcode appearing for a image file)

## Changes
The "Transcode" checkbox state (checked vs. unchecked) will now reflect the user's desire and will be a persisted state. Whether or not this setting is used will be reflected by the checkbox's grayed-out state (i.e. it can be checked for non videos, but it will be grayed out).

### DIFF
--- a/ui/component/publishFile/view.jsx
+++ b/ui/component/publishFile/view.jsx
@@ -7,7 +7,8 @@ import Button from 'component/button';
 import Card from 'component/common/card';
 import { FormField } from 'component/common/form';
 import Spinner from 'component/spinner';
-import I18nMessage from '../i18nMessage';
+import I18nMessage from 'component/i18nMessage';
+import usePersistedState from 'effects/use-persisted-state';
 
 type Props = {
   name: ?string,
@@ -48,6 +49,8 @@ function PublishFile(props: Props) {
   const ffmpegAvail = ffmpegStatus.available;
   const [oversized, setOversized] = useState(false);
   const [currentFile, setCurrentFile] = useState(null);
+  const [optimizeAvail, setOptimizeAvail] = useState(false);
+  const [userOptimize, setUserOptimize] = usePersistedState('publish-file-user-optimize', false);
 
   const RECOMMENDED_BITRATE = 6000000;
   const TV_PUBLISH_SIZE_LIMIT: number = 1073741824;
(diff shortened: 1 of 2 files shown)


ui/component/publishFile/view.jsx
```diff
@@ -8,6 +8,7 @@ import Card from 'component/common/card';
 import { FormField } from 'component/common/form';
 import Spinner from 'component/spinner';
 import I18nMessage from '../i18nMessage';
+import usePersistedState from '../../effects/use-persisted-state';
```
seanyesmunt, line 11: Can you change these to use absolute import paths?

infinite-persistence: Fixed.

PR #4465, opened 2020-07-02
Inf-scroll: Change region to 'main' to ignore long tag/following list.

### LINKED ISSUE #2789: Infinite scrolling should activate at end of listed claims, not subscription list
This has been annoying me for a while because I have a long subscription list. Infinite scrolling shouldn't have anything to do with the subscription list. See attached video, I have to go to the end of the subscriptions for it to activate. After that it works well.

https://beta.lbry.tv/vokoscreen-2019-08-26_08-54-04/e
demonstrates the issue.

### LINKED ISSUE #2957: scroll restoration issues when tags/sub list longer than page
If you have lots of tags/subs or if the window size is small enough where the tags/subs fill up past the height, you run into scroll restoration problems with navigating pages / clicking home / etc.

infinite-persistence:
## Issue:
Fixes #2789 `Infinite scrolling should activate at end of listed claims, not subscription list`

I believe it also closes #2957 `scroll restoration issues when tags/sub list longer than page`.  Looks to be working, and any quirks with `Back` will be a common (existing) one that is unrelated to extended Tag/Following size.

## Remarks:
If you press `End` on the keyboard, you'll go to the bottom but only 1 extra page will be loaded.  If you scroll back up slowly, inf-scroll will continue to load one page at a time.  I think this is arguably a better behavior than trying to load all the way to the bottom in one-shot;  it's laggy enough as it is.

### DIFF
--- a/ui/component/claimList/view.jsx
+++ b/ui/component/claimList/view.jsx
@@ -1,5 +1,5 @@
 // @flow
-import { MAIN_WRAPPER_CLASS } from 'component/app/view';
+import { MAIN_CLASS } from 'component/page/view';
 import type { Node } from 'react';
 import React, { useEffect } from 'react';
 import classnames from 'classnames';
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Ha. Nice fix!

PR #4473, opened 2020-07-03
Add password reset link to settings page (4468)

### LINKED ISSUE #4468: Add password reset link to settings page
Use same password reset link as from the sign in page. If a user doesn't remember the old one, they can't change it if they are already logged in.

infinite-persistence:
## Issue
Fixes #4468 `Add password reset link to settings page`
>_Use same password reset link as from the sign in page. If a user doesn't remember the old one, they can't change it if they are already logged in._

## Screenshot
image


seanyesmunt:
I think instead of using this component inside of the "Account Password" card. It should redirect to a new page `/$/password/reset`. If it uses this same component, it should redirect properly after a successful reset if a `redirect` param is passed in the url.

infinite-persistence:
#### Attempt 2
See commit messages for the walk through.

I couldn't figure out the redirect thing, so I left it out.  But I think we probably shouldn't redirect back after the reset, otherwise there's no point in displaying the following advisory (they might miss it).  The password-reset in the main Sign In page also does not redirect or move away from this page.
image

ui/component/userPasswordReset/view.jsx
```diff
@@ -47,7 +48,11 @@ function UserPasswordReset(props: Props) {
     setEmail('');
     doClearPasswordEntry();
     doClearEmailEntry();
-    push(`/$/${PAGES.AUTH_SIGNIN}`);
+    if (restartAtSignInPage) {
```
seanyesmunt, line 51: Nice. This seems reasonable.

infinite-persistence:
Rebasing to solve conflict:

PR #4474, opened 2020-07-03
Add dialog to copy various types of links for a claim.

### LINKED ISSUE #4293: Better download link support
It's impossible right now to get a download URL for non-video content. 

- [ ] make it so you can right click a download button > copy download URL
- [ ] in share dialog, show download URL to copy for non-video content

infinite-persistence:
Fixes #4293 `Better download link support`

- [x] **make it so you can right click a download button > copy download URL**
   Given that this could be useful in mobile/web as well, I used a button/modal instead of the right-click approach.

- [ ] **in share dialog, show download URL to copy for non-video content**
   This is already implemented, albeit hidden in the "More..." expansion.

image

image


seanyesmunt:
Cool! I wonder if this should just be integrated with the Share modal.

I'll give your PR's a closer look on Monday. Have a good weekend!

infinite-persistence:
True, the new one is just a stripped-down version of what's already available in `Share`

seanyesmunt:
@infinite-persistence lets change the `...` icon on the share modal to the `copy` icon. Then also add the download link to that view too.

infinite-persistence:
The file-download link already exists in the Share Modal, so I only changed the icon.  I used the chain (link) icon as I think it makes more sense than the clipboard (can change if you disagree).

image

infinite-persistence:
Rebasing to latest master to resolve changelog conflict:

seanyesmunt:
Nice. I like the chain icon more!

PR #4482, opened 2020-07-06
Patch for "7839bbf2: Fix language-change not applied to all components immediately"

infinite-persistence:
Patch for `7839bbf2: Fix language-change not applied to all components immediately`

- Missed the "Account Password" card that is used for authenticated case.
- For cards that are only used in Settings, listen to the language change in "select" instead of passing it in as a prop. I think this is cleaner.


seanyesmunt:
Agreed this is cleaner.

PR #4483, opened 2020-07-06
Fix unmuted state lost or reverted when playing a new video.

### LINKED ISSUE #4460: unmuted state lost / reverted
Notice that my videos go back to muted state after using lbry.tv for a while or possibly in a new tab.  Not sure if the browser is causing this to happen or the mute persisted data is lost.

infinite-persistence:
## Issue
Fixes #4460 `unmuted state lost / reverted`

1. Play a video.
2. Press mute.
3. Drag the volume bar to unmute.
4. Play a new video --> the video starts muted.

## Fix
The `volumechange` handler was comparing against stale variables, so there are times where the state was not saved. Just save both `muted` and `volume` without additional gating (the gating is probably unnecessary in the first place, since we are in a onChange function).


infinite-persistence:
Adding changelog:

seanyesmunt:
I tried fixing the conflict from github.com but failed. @infinite-persistence Can you rebase this? You can push over my 'merge' commit.

infinite-persistence:
Rebased to equivalent spot.

infinite-persistence:
Rebasing again to latest master since there's another conflict in the changelog (might as well):

seanyesmunt:
Thank you.

PR #4489, opened 2020-07-07
Fix "Price=free" giving 0 results in Following query

### LINKED ISSUE #4477: Price="free" yields nothing
If "Anything" yields a bunch of free items, "Free" should yield the same thing, right?
https://youtu.be/HxOm51r8ERY

Seems to only affect `Following + New` query.

infinite-persistence:
## Issue
Fixes #4477

## Remarks
I didn't consult the documentation, but simply did a trial-and-error.
`"=0"` -- didn't work
`"<=0"` -- work

## Disclaimer
I have no idea what I'm doing, but it works.  I have a feeling this might be an SDK problem.

image


infinite-persistence:
Adding changelog (assuming fix is correct):

tzarebczan:
Hmm, mine is working on master with just =0.  On one of my accounts, I did see a query timeout...let me track that down. Maybe that's what you were seeing and not 0 results. I'm not sure how this fix would help...but maybe some strange sql quirk on the SDK.

infinite-persistence:
Closing this PR.  Continuing discussion in the original Issue.

PR #4493, opened 2020-07-08
Improve auto string addition to produce better diffs and ease merging.

infinite-persistence:
## Issue
- The diff for new strings are polluted by the need to add a comma to the previous entry.
- Having to re-add the newline at the end of file before commiting is a repetitive pain.

Before:
image

After:
image

## Caveats to this approach
- When manually adding strings, developers need to put it above the `--end--` entry. Hopefully it is obvious without having to put verbose comments like `^--- add new string before this line ---^`
- Translators will surely ask how to translate `--end--`.


seanyesmunt:
Nice. We can try with `--end--`, then change it to something more verbose if needed. Can you fix the conflict in the app strings file? 🤪

infinite-persistence:
Rebasing to fix conflict:

PR #4494, opened 2020-07-08
TipModal: Add final warning when making a permanent tip. 

### LINKED ISSUE #4492: Make support (tip case) dialog clearer
Make it clearer that tip being sent to creator when tip option is used. 

Potentially move other text to bottom of modal and add note below title.

infinite-persistence:
## Issue
Closes #4492 `Make support (tip case) dialog clearer`

## Changes:
- Move the "Buy More" button to the middle instead of isolated at the bottom
- Add warning sentence 

image

image

### DIFF
--- a/static/app-strings.json
+++ b/static/app-strings.json
@@ -1269,6 +1269,8 @@
   "Delete Your Channel": "Delete Your Channel",
   "Remove from Blocked List": "Remove from Blocked List",
   "Are you sure you want to remove this from the list?": "Are you sure you want to remove this from the list?",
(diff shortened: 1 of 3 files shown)


seanyesmunt:
@infinite-persistence I ended up change the title/subtitle here https://github.com/lbryio/lbry-desktop/pull/4499

I'd like to avoid adding additional text below the button since this is an already very complex form. I do really like how you added the buy button though. Can you update this PR to only include that?

infinite-persistence:
Removed the advisory message, leaving only the "Buy More" button changes intact:

PR #4511, opened 2020-07-09
only set matomo userid if hasverifiedemail

jessopb:


ui/component/app/view.jsx
```diff
@@ -231,6 +224,7 @@ function App(props: Props) {
   useEffect(() => {
     if (!hasSignedIn && hasVerifiedEmail) {
       signIn();
+      analytics.setUser(userId);
```
seanyesmunt, line 227: just move this inside of the `signIn` function. Right now the correct userId won't be included in this effect

PR #4513, opened 2020-07-10
Allow zooming on Desktop

### LINKED ISSUE #4501: Font size
I am visually handicapped. This mean that I always increase the font size to 200% - sometimes even more.

I have found no way to do that in LBRY.

Ideally I would like it to work as smoothly as in Firefox: CTRL + and CTRL - to zoom. CTRL scrollwheel also zooms.

But it would also be acceptable, if I could change the font size in settings.

As it is now, I cannot use the Desktop app - simply because the font size is too tiny for me to read. And I find that a silly limitation.

infinite-persistence:
## Issue
Closes #4501`Font size`
and Discord submission
image

## New behavior
The Desktop app can now zoom the same way as browsers:
- Zoom In: `Ctrl+=` or `Ctrl+numpadPlus` or `Ctrl+WheelUp`
- Zoom Out: `Ctrl+-` or `Ctrl+numpadMinus` or `Ctrl+WheelDown`
- Zoom Reset: `Ctrl+0` or `Ctrl+numpad0`

image

## Code changes
(1) Electron provides this functionality through the `zoomIn|zoomOut|resetZoom` roles in the Menu, so it would have been a quick job.

However, given that Electron currently does not support having multiple accelerators for one item, we can't add `Ctrl+WheelUp` to the mix and would have to implement our own handler and use `webFrame`.

Given that we need to add code anyways, we handle both keyboard and mouse cases through the same handler, hence the existence of `zoomWindow.js`.  It also provides the opportunity to few a few quirks with Electron's default implementation (e.g. stuck at both extremes)

(2) I recall there is another Issue for adding keyboard shortcuts.  Given that these shortcuts are universally used in browsers, they are probably "reserved", so shouldn't clash with that task.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Allow zooming on Desktop _community pr!_ ([#4513](https://github.com/lbryio/lbry-desktop/pull/4513))
+
 ### Changed
 
 ### Fixed
@@ -26,7 +28,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - New channel create/edit page ([#4445](https://github.com/lbryio/lbry-desktop/pull/4445))
 - Add dialog to copy various types of links for a claim _community pr!_ ([#4474](https://github.com/lbryio/lbry-desktop/pull/4474))
 - Add password reset link to settings page for logged in users _community pr!_ ([#4473](https://github.com/lbryio/lbry-desktop/pull/4473))
-- Allow zooming on Desktop _community pr!_ ([#4513](https://github.com/lbryio/lbry-desktop/pull/4513))
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


infinite-persistence:
Adding changlog:

ui/component/app/view.jsx
```diff
@@ -174,6 +177,52 @@ function App(props: Props) {
     return () => window.removeEventListener('keydown', handleKeyPress);
   }, []);
 
+  // Enable ctrl +/- zooming on Desktop.
+  // @if TARGET='app'
+  useEffect(() => {
```
seanyesmunt, line 182: Could you move this to `effects/use-zoom.js`?

Trying to avoid this file getting massive. Then you can just do
```
// @if TARGET='app'
useZoom();
// @endif
```

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - New channel create/edit page ([#4445](https://github.com/lbryio/lbry-desktop/pull/4445))
 - Add dialog to copy various types of links for a claim _community pr!_ ([#4474](https://github.com/lbryio/lbry-desktop/pull/4474))
 - Add password reset link to settings page for logged in users _community pr!_ ([#4473](https://github.com/lbryio/lbry-desktop/pull/4473))
+- Allow zooming on Desktop _community pr!_ ([#4513](https://github.com/lbryio/lbry-desktop/pull/4513))
```
seanyesmunt, line 21: Can you move this into the new unreleased section?

infinite-persistence:
- Refactored code into `effects\use-zom.js`
- Updated changelog

PR #4521, opened 2020-07-12
Inf-scroll: Fix glitch when results are exactly the page size.

### LINKED ISSUE #2675: Claim search with exactly 20 results keep fetching next page
## The Issue
If a page has exactly 20 results, it tries to load the next page constantly and causes lots of artifacts. 

a good example is the `premium trends` tag page, thanks @ceoger ! 

The bug is somewhere in https://github.com/lbryio/lbry-desktop/blob/master/src/ui/component/claimListDiscover/view.jsx#L108 or https://github.com/lbryio/lbry-desktop/blob/consistency/src/ui/component/claimList/view.jsx#L66

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

infinite-persistence:
## Issue
Fixes #2675 `Claim search with exactly 20 results keep fetching next page`
  The impact is milder now given recent fixes to inf-scroll, but still good to remove the glitch.

## Changes
Use `claimSearchResultLastPageReached` (I didn't know it existed back then).

The unnecessary "loading" refresh will still happen one time, but at least now it doesn't happen every time we scroll to the bottom (and incorrectly incrementing 'page').


infinite-persistence:
Adding changelog:

seanyesmunt:
@infinite-persistence do you know of a lbry.tv url that I can test this out at?

seanyesmunt:
Tested with `starfleet123456789` and it still seems to be in an infinite loop of requests.

seanyesmunt:
Nvm. Maybe I wasn't on the right branch? Seems to work great now.

PR #4523, opened 2020-07-13
Show "YT Creator" label in File Page as well.

### LINKED ISSUE #3964: Add Youtuber info to file page
We show Youtube channel info on the channel page, we should also show something similar on the file page near the channel.

infinite-persistence:
## Issue
Closes #3964`: Add Youtuber info to file page`

## Changes
(1) Refactor the YT-info query into `YoutubeBadge` component.
(2) For the File Page case, don't show "- last sync <date>" since the date has nothing to do with the file.

## Test
[x] YT Channel page
[x] YT Channel's claim page
[x] Non YT Channel page
[x] Non YT Channel's claim page
[x] No channel (anonymous) claim page
[x] Various `RENDER_MODES`

### DIFF
--- a/ui/component/youtubeBadge/view.jsx
+++ b/ui/component/youtubeBadge/view.jsx
@@ -1,0 +1,50 @@
+// @flow
+import * as ICONS from 'constants/icons';
+import * as React from 'react';
+import DateTime from 'component/dateTime';
+import Icon from 'component/common/icon';
+import { Lbryio } from 'lbryinc';
+
+type Props = {
+  channelClaimId: string,
+  includeSyncDate: boolean,
+};
+
+export default function YoutubeBadge(props: Props) {
+  const { channelClaimId, includeSyncDate = true } = props;
+
+  const [isVerified, setIsVerified] = React.useState();
+  const [lastYtSyncDate, setLastYtSyncDate] = React.useState();
+
+  React.useEffect(() => {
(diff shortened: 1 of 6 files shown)


ui/component/youtubeBadge/view.jsx
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      setIsVerified(false);
+      setLastYtSyncDate(undefined);
+    }
+  }, [channelClaimId]);
+
+  if (isVerified) {
+    const str =
+      includeSyncDate && lastYtSyncDate
+        ? __('Official YouTube Creator - Last updated %time_ago%', { time_ago: DateTime.getTimeAgoStr(lastYtSyncDate) })
```
seanyesmunt, line 39: Could you change this to `Last checked %time_ago%`? We've been talking about renaming this since "updated" can be confusing if there is no new content.

infinite-persistence:
Updated string to "last checked".
(Not sure why it isn't reflected in this page, but it's in the repo)

seanyesmunt:
I commented the badge out on the file page because it was the floating viewer wasn't pushed down properly on mobile. I would like to avoid using two lines for the lbry URL and the youtube badge. Maybe we could add an ellipsis on the lbry URL so they don't wrap to two lines? That would help on channel pages too

PR #4527, opened 2020-07-14
Fix channel file-search not available in mobile 

### LINKED ISSUE #4524: [mobile] in channel search not available
The search feature is not available on the channel page in mobile mode

infinite-persistence:
### Issue
Fixes #4524 `[mobile] in channel search not available`

### Changes
There are currently 3 wunderbars: (1) header (2) channel list (3) downloaded list.

The additional `wrapperStyle` in the header made it hard to find a solution that works for all 3, so I split 2 and 3 into `wunderbar__input--inline`.

### Issue
This button is usually placed beside the `Trending|New|Top` selector, so it should have the same bottom margin. If not, anything that flows below it (e.g. in mobile) will be touching it.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Fix 'transcoding' checkbox state when switching file types _community pr!_ ([#4529](https://github.com/lbryio/lbry-desktop/pull/4529))
+- Fix channel file-search not available in mobile _community pr!_ ([#4527](https://github.com/lbryio/lbry-desktop/pull/4527))
 
 ## [0.47.0] - [2020-07-13]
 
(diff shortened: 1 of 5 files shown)


infinite-persistence:
Adding changelog:

ui/scss/component/_button.scss
```diff
@@ -156,6 +156,10 @@ svg + .button__label,
   }
 }
 
+.button-toggle--more {
+  margin-bottom: var(--spacing-s);
```
seanyesmunt, line 160: This causes unequal spacing below the buttons in the header. They should have the same top/bottom margin. Maybe the search bar should go full-width on small screens? Then you could add a margin-top to it.

infinite-persistence: I've updated the PR with a different approach, where each element will have vertical spacing for better reflow.  See if that works.

seanyesmunt: Nice. Looks great!

PR #4529, opened 2020-07-14
Fix 'transcoding' checkbox state when switching file types.

infinite-persistence:
## Issue:
Currently, the checkbox grayed-out state is only correctly updated when changing from "no-file" to "file selected". If you have selected a video file and later changed to a text file, the checkbox doesn't gray out.

## Code Changes:
`fileVid` wasn't being updated for the case of non-videos; now it does.

### DIFF
--- a/ui/component/publishFile/view.jsx
+++ b/ui/component/publishFile/view.jsx
@@ -82,7 +82,7 @@ function PublishFile(props: Props) {
 
     setOptimizeAvail(isOptimizeAvail);
     updatePublishForm({ optimize: finalOptimizeState });
-  }, [filePath, isVid, ffmpegAvail, userOptimize]);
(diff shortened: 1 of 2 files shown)


ui/component/publishFile/view.jsx
```diff
@@ -216,6 +216,8 @@ function PublishFile(props: Props) {
       } else {
         updateFileInfo(0, file.size, isVideo);
       }
+    } else {
+      updateFileInfo(0, file.size, isVideo);
```
infinite-persistence, line 220: I'm not sure if `file.size` will cause any side effects for non-videos (should I just zero it?), but I believe the previous code just uses whatever the previous value was, so most likely it doesn't matter.

seanyesmunt: I think it's fine, as long as `file` isn't undefined. Which it doesn't look like it is

Issue #4540, opened 2020-07-15
lbry.tv: video cut-off without errors


infinite-persistence:
On some days, lbry.tv only partially loads a video.  No error messages.  The same video works fine on Desktop.
https://youtu.be/EXe9pt7O9Vc

infinite-persistence:
_Close_
Haven't seen it lately.

PR #4541, opened 2020-07-15
Add option to retry video stream on failure

### LINKED ISSUE #4475: Option to retry video stream on failure
There are various instances where a video may fail on start or mid way, and it would be nice to have a retry button on the page instead of having to refresh the page.

infinite-persistence:
## (1) Video: Stop loading circle when there's an error.
When there is an error in loading the video, the loading circle spins perpetually, giving false hope.  This commit removes the loading circle in that scenario.

## (2) Closes #4475 `Option to retry video stream on failure`
Added a `Retry` button to reload the video without re-rendering the app or even the page.  
See example of how it looks like: https://www.youtube.com/watch?v=C3_GRWWFDC8

Note that this doesn't cover #4540 `lbry.tv: video cut-off without errors`.  There's no error message for this case, plus refreshing will produce the same cut-off anyway (server issue?)

### DIFF
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -21,6 +21,7 @@ export type Player = {
   currentTime: (?number) => number,
   ended: () => boolean,
   error: () => any,
+  loadingSpinner: any,
 };
 
 type Props = {
@@ -136,6 +137,9 @@ export default React.memo<Props>(function VideoJs(props: Props) {
 
   function onError() {
     showTapToUnmute(false);
+    if (player && player.loadingSpinner) {
+      player.loadingSpinner.hide();
+    }
   }
 
   function onEnded() {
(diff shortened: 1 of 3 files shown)


ui/component/viewers/videoViewer/internal/videojs.jsx
```diff
@@ -118,29 +136,44 @@ export default React.memo<Props>(function VideoJs(props: Props) {
...
     if (player && !player.muted()) {
-      showTapToUnmute(false);
+      showTapButton(TAP.UNMUTE, false);
     }
   }
 
   function onError() {
-    showTapToUnmute(false);
+    showTapButton(TAP.UNMUTE, false);
+    showTapButton(TAP.RETRY, true);
```
seanyesmunt, line 167: Can you just set the other button to false when one is set to true?

infinite-persistence:
Consolidated the `showTapButton` call.
(not sure why the github page is not showing the change, but it has been pushed)

Issue #4454, opened 2020-06-30
URL / top claims overlap


tzarebczan:
![image](https://user-images.githubusercontent.com/8120721/86186664-a5db4f80-bb07-11ea-923d-5231e6f44516.png)

PR #4543, opened 2020-07-15
Changed YT label from "last updated" to "last checked" (part 2)

infinite-persistence:
Oops, missed out one file in the previous PR.

### DIFF
--- a/ui/component/youtubeBadge/view.jsx
+++ b/ui/component/youtubeBadge/view.jsx
@@ -39,7 +39,7 @@ export default function YoutubeBadge(props: Props) {
         ? __('Official YouTube Creator - Last checked %time_ago%', { time_ago: DateTime.getTimeAgoStr(lastYtSyncDate) })
         : __('Official YouTube Creator');
     return (
-      <div className="media__uri--right">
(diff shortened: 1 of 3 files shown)


infinite-persistence:
> _https://github.com/lbryio/lbry-desktop/pull/4523#issuecomment-659097018 @seanyesmunt: I commented the badge out on the file page because it was the floating viewer wasn't pushed down properly on mobile. I would like to avoid using two lines for the lbry URL and the youtube badge. Maybe we could add an ellipsis on the lbry URL so they don't wrap to two lines? That would help on channel pages too_


## Fix floating-player misalignment due to YT badge in mobile by hiding the badge.

At first, the suggestion to truncate the left URI to keep the entire thing single-lined was implemented, but during testing, I realized that the YT badge will take up the entire width anyway (especially on other languages). The URI is more important than the badge, and truncating both of them would be ugly.

Given the preference to not have 2 lines for this, we'll just hide the badge for the mobile case.

For the Channel Page case, let's make it single-line through #4454 `URL / top claims overlap` instead. We'll keep this commit/PR limited to fixing the YT badge.

Issue #4549, opened 2020-07-17
Mature content + navigating to URLs directly


tzarebczan:
If a user navigates to mature content directly, we currently show the page even if mature content setting is not enabled. With out recent mature content verification setting, we should hide the page until the user turns it on - we can prompt directly on the file page.

PR #4560, opened 2020-07-21
Block mature content when access directly from URL

infinite-persistence:
## Issue:
Fixes  [#4549 Mature content + navigating to URLs directly](https://github.com/lbryio/lbry-desktop/issues/4549)
> _If a user navigates to mature content directly, we currently show the page even if mature content setting is not enabled. With our recent mature content verification setting, we should hide the page until the user turns it on - we can prompt directly on the file page._

## Changes:
1. Show basic info of the claim like URL and Title, so that user knows which page is being blocked.
2. Initially, I had 2 separate "boxes" -- one for the title and another for the message. It felt a bit messy, so `FileTitle` was augmented to house everything in one connected box.

image


ui/component/fileTitle/view.jsx
```diff
@@ -39,14 +45,32 @@ function FileTitle(props: Props) {
         </React.Fragment>
       }
       actions={
-        <div>
-          <div className="section">
-            <FileActions uri={uri} />
+        isNsfwBlocked ? (
+          <div className="main--empty">
+            <h2 className="card__title card__title">
```
seanyesmunt, line 50: Since they are in the same card, this shouldn't use `card__title`. I think the generic `h2` will work.

ui/component/fileTitle/view.jsx
```diff
@@ -39,14 +45,32 @@ function FileTitle(props: Props) {
...
+              <Icon className="icon--hidden" icon={ICONS.EYE_OFF} />
+              {__('Mature content blocked')}
+            </h2>
+            <div>
+              <I18nMessage
+                tokens={{
+                  settings: <Button button="link" label={__('Content Settings')} navigate={`/$/${PAGES.SETTINGS}`} />,
+                }}
+              >
+                Change %settings%
```
seanyesmunt, line 60: "Change this in your content settings."

PR #4578, opened 2020-07-24
disable file remove/abandon while abandoning

jessopb:


ui/modal/modalRemoveFile/view.jsx
```diff
@@ -78,6 +79,7 @@ function ModalRemoveFile(props: Props) {
               <Button
                 button="primary"
                 label={__('OK')}
```
seanyesmunt, line 81: Please change the text to `Abandoning...` while `isAbandoning` is true

PR #4586, opened 2020-07-25
Fix 'Related' section lazy-load not working in some scenarios

infinite-persistence:
## Issue
1. Sometimes, the Related Section will load unnecessarily when a File Page is entered.  This happens more to Image claims, although I see it happening on Videos.  I think it's due to my slow connection causing the upper components to not fully inflate, so the Related Section will briefly appear on screen.
2. In the Autoplay case, if the `WaitUntilOnPage` has already opened the gates previously, the next video's Related Section will be loaded regardless of scroll position.

## Changes
See the commit messages for the walkthrough.

## Comments
Using the debounce method doesn't fix the first issue 100%, but it covers most cases with a decent connection speed.  In the future, I'm thinking of using a dummy placeholder component with typical Video height while waiting for Images to load.

### DIFF
--- a/ui/component/recommendedContent/view.jsx
+++ b/ui/component/recommendedContent/view.jsx
@@ -25,6 +25,7 @@ export default class RecommendedContent extends React.PureComponent<Props> {
     super();
 
     this.didSearch = undefined;
+    this.lastReset = undefined;
   }
 
   componentDidMount() {
@@ -36,6 +37,7 @@ export default class RecommendedContent extends React.PureComponent<Props> {
 
     if (uri !== prevProps.uri) {
       this.didSearch = false;
+      this.lastReset = Date.now();
     }
 
     if (claim && !this.didSearch) {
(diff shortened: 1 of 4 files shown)


ui/component/recommendedContent/view.jsx
```diff
@@ -68,14 +72,16 @@ export default class RecommendedContent extends React.PureComponent<Props> {
...
         body={
-          <ClaimList
-            isCardBody
-            type="small"
-            loading={isSearching}
-            uris={recommendedContent}
-            injectedItem={!isAuthenticated && IS_WEB && <Ads type="video" small />}
-            empty={__('No related content found')}
-          />
+          <WaitUntilOnPage lastUpdateDate={this.lastReset}>
```
seanyesmunt, line 75: Should this prop be added to the `WaitUntilOnPage` wrapper around `CommentList`?

infinite-persistence: Ah, missed it.  For some reason, I kept thinking that one is the Channel Comments.

PR #4614, opened 2020-07-31
prevent specific settings sync in config

jessopb:
Some lbry properties will not have control over some settings.
Make sure they can disable sync for users that visit them.


jessopb:
I'm wondering if instead certain settings we wish to fix should be actually set in .env rather than just ignored on sync.

seanyesmunt:
Maybe these should just use the `SIMPLE_SITE` config value to turn them all on or off?

.env.defaults
```diff
@@ -27,6 +27,8 @@ OG_IMAGE_URL=
 SITE_CANONICAL_URL=https://lbry.tv
 # LOCALE
 DEFAULT_LANGUAGE=en
+# Custom Settings
+# UNSYNCED_SETTINGS='theme dark_mode_times automatic_dark_mode_enabled'
```
seanyesmunt, line 31: I guess just keep this, but also include a comment pointing to where these names are defined

Issue #486, opened 2017-08-18
Add Confirmation before publishing file or creating channel (show fee estimate)


tzarebczan:
## The Issue
Currently, when a user clicks the Publish button to submit content there is no confirmation of their action but instead, it tells them that their publication is pending. Should we have some sort of confirmation box which would at least show the Content URL along with deposit amount? This came from a user suggestion and I thought it was worthy of at least bringing it up. 

This is possible today with the --preview feature on the SDK. 

## System Configuration

 

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

kauffj:
Yes. This would just be to generally improve the post publishing UX. Act like it is published as much as possible, but show some indicator estimating how long until others can see it.

tzarebczan:
We now have the --preview command which will show the fees / metadata that set before the tx is created.

Issue #344, opened 2017-07-10
Content shown as free, charged LBC


tzarebczan:
## The Issue
We've had a couple reports of users seeing the It's a Disaster video as Free, when in reality it costs 7-9 LBC (depending on LBC price).  When a user tries to download, they are still charged the LBC.  I've seen this happen on my own machine too, but not sure what exactly caused it.  When discussed within the team, it's most likely a caching issue. One time I did get an exchange error, so that may be related also.  When going to settings > Refresh the Cache, the price then shows correctly.

@6ea86b96 is aware of problem but we wanted to log it officially. 

### Steps to reproduce
1. Load LBRY
2. It's a Disaster shown as free
3. Download it, but get charged LBC

### Expected behaviour
Should not show as free, but even if it did, it should double check the price before downloading

### Actual behaviour
Content shown as free, but charged LBC when downloaded.

## System Configuration

 

- LBRY Daemon version: 0.13.1
- LBRY App version: 0.13
- LBRY Installation ID:
- Operating system: Windows

## Anything Else

```shell
2017-07-05 16:15:48,789 WARNING  lbrynet.lbrynet_daemon.ExchangeRateManager:69: There was a problem updating BTCLBC exchange rate information from lbry.io
2017-07-05 16:15:48,795 WARNING  lbrynet.lbrynet_daemon.ExchangeRateManager:69: There was a problem updating USDBTC exchange rate information from Coinbase via Google finance
```

## Screenshots

![disaster free](https://user-images.githubusercontent.com/8120721/28010805-159f2752-652e-11e7-9cbf-55adfd7e0288.PNG)

zestyr:
I was playing around with this a few days ago and when I refreshed the cache, it showed correctly, but if I restarted the app and refreshed again (or the other way around, don't remember), it showed as `free`. I'll see if I can reproduce this, but something is definitely going on here.

tzarebczan:
@hackrush01 saw this happen today on his discover page.

btzr-io:
i can't recreate the issue :disappointed:, 
can some one check if #465 can fix this?

tzarebczan:
@btzr-io I think there was also a larger formatting issue that was causing this, it should be fixed in 0.15 daemon

tzarebczan:
Closing this as it has not happened for a while. There's a different issue with price not being shown correctly for updated claims.

PR #4620, opened 2020-08-03
Add Publish Preview dialog

infinite-persistence:
_Requires a change in lbry-redux_ ([#344 doPublish: Add optional parameter for "preview-only" request.](https://github.com/lbryio/lbry-redux/pull/344))

## Feature
A no-scroll, single-glance dialog to preview the Publish data before really uploading.
Includes estimated transaction fees.

image

## Issue
Closes [#486 Add Confirmation before publishing file or creating channel (show fee estimate)](https://github.com/lbryio/lbry-desktop/issues/486)

## Changes
- See commit messages for walkthrough.

## Testing notes
Probably not useful, but hopefully can spark any ideas on missing test areas.
<details>
  <summary>Click to expand</summary>

Issues:
[x] Trancoding started even in preview; it should not. ==> force to "false" in preview-mode.
[x]    Test: trancoding still works after disabling in preview (ensure not permanently disabled)
[x] Tags that reflowed to second line need gaps.
[x] Long tags need to be truncated with ellipsis.
[x] Long tag touches border.
[ ] Warning: "each child in list should have unique key". ==> No longer reproducible.
[x] Warning: "whitespace text nodes cannot appears as child nodes in <tbody>" ==> Firefox didn't like the empty row at the bottom. Just removed it. We'll just do without the final separator line (which is probably what the theme intended anyways).

Todo:
[x] Include tags
[x] Include Transcode section
[x] Test web
[x]    The preview is not appearing. ==> `filePath` in web is a `WebFile` object. Need to handle it differently. Not sure how to get the full path in web, so we'll just show the filename for now.
[x]    The preview modal takes a while to appear in web. ==> Grayed out the button while previewing.
[x] Handle "license=other"
[x] Handle "license=copyrighted"
[x] Test: empty description
[x] Test on mobile screen
[x]    Issue: the table is overflowing on the right. ==> Using `nowrap` on the first column causes this. Just fix the label to 30% and truncate with ellipsis.
[x] Extreme case: What if screen height is super small? ==> it scrolls, so no problem.  Defeats the no-scroll intention, but at least it's not in an unusable state.
[x] Description is markdown. How to show? ==> just use `<MarkdownPreview>`
[x] What if Description is too long? ==> made row scrollable if exceeded 5-6 lines.
[x] What if `previewResponse` is null? ==> handled the null case, although not sure if it will ever happen.
[x] Test: Unplug cable ==> Preview doesn't seem to need internet connection, so nothing to handle.
[x] Test: Didn't break actual publish ==> Seems ok, as far as I can tell.  The same data is still sent to doPublish per original code.
[x] What if Editing instead of Publishing? ==> Disable preview for edits.
</details>


seanyesmunt:
This is great! I will probably add a few commits to this to change the styling a bit since we have already planned what this should look like somewhat, but that will be much easier now that all of the redux code is set up. I probably won't get to it until later this week though.

infinite-persistence:
Ah, it's part of a planned work.  Feel free to use as much or as little of it; I don't mind it being dropped entirely if it's not aligned to planned work 👍

seanyesmunt:
By planned I mean loosely discussed a couple of times. You can take a shot at it if you want but the main thing I would be interested in changing is making this less "technical" and more a simplified version of what someone would see on the file page.

infinite-persistence:
Rebasing to fix conflict with "in-app markdown post":

tzarebczan:
Should we allow users to skip confirmation if they don't wish to see it each time? Super excited for this, especially around the estimated fees!

infinite-persistence:
`Change:`  Added setting to never show the preview modal.  It can be re-enabled in the Advanced Settings.
The text and position are stubs since the UI is subject to change.
<details>
  <summary><b>Screenshots</b> (click to expand)</summary>

image
image
</details>

ui/scss/component/_table.scss
```diff
@@ -204,3 +204,29 @@ th {
     margin-left: var(--spacing-m);
   }
 }
+
+.table--publish-preview {
```
kauffj, line 208: this should not be a new style

table--file-details and table--publish-preview should probably be unified @seanyesmunt

ui/modal/modalPublishPreview/view.jsx
```diff
@@ -0,0 +1,178 @@
...
+              <>
+                <div className="section__actions">
+                  <Button autoFocus button="primary" label={__('Publish')} onClick={() => this.onConfirmed()} />
+                  <Button button="link" label={__('Cancel')} onClick={closeModal} />
+                </div>
+                <p className="help">{__('Once the transaction is sent, it cannot be reversed.')}</p>
+                <FormField
+                  type="checkbox"
+                  name="sync_toggle"
+                  label={__("Don't show previews anymore.")}
```
kauffj, line 165: needs better text

maybe "Publish without this preview", but I don't love it

ui/component/publishSettings/view.jsx
```diff
@@ -0,0 +1,31 @@
...
+  function handleChange() {
+    setEnablePublishPreview(!enablePublishPreview);
+  }
+
+  return (
+    <div>
+      <FormField
+        type="checkbox"
+        name="sync_toggle"
+        label={__('Review changes before uploading.')}
```
kauffj, line 23: this text should match or be similar to the text that allows toggling it at the bottom of the confirmation dialog

ui/modal/modalPublishPreview/view.jsx
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  createRow(label: string, value: any) {
+    // Probably need to do the colon differently when in RTL mode. Don't want it to be part of the string
+    // to reduce redundant strings for translation.
+    return (
+      <tr>
+        <td>
+          {label}
+          {label ? ':' : ''}
```
kauffj, line 61: colons are an unnecessary here, and generally should be avoided outside of visually indistinct text

infinite-persistence:
`Change`
- Use "Skip preview and confirmation" in both Modal and Settings for better correlation.
   - _Text is still subject to change until a better one comes along._
- Remove unnecessary colon in the Label column
- Enable "Publish Preview" on Edit Mode as well.
- Enable "Publish Preview" for Markdown.
   - _Depending on what the Preview Modal will be showing, we might need to pass the 'mode' over so that we can hide items like 'Description' when doing Markdown. As I'm not sure if the Description will be part of the modal or not, I skipped this for now._

infinite-persistence:
`Rebase`
Rebased to grab the lint fixes for _publishForm.jsx_

kauffj:
@infinite-persistence is there more work required here? Are you coming back to this?

infinite-persistence:
> @infinite-persistence is there more work required here? Are you coming back to this?

I think 2 things to do:
1. Rebase to fix any conflicts with the new MD Editor (probably not much)
2. GUI changes that Sean had in mind.

I mulled over item-2 for a while, but not sure how I should change it.  I kinda like the existing "single glance to see all entered info" version to double-check before posting.

seanyesmunt:
I'm fine merging this with the current styling, but disabling on Odysee. Lets wait until we release the next version of the app first though.

infinite-persistence:
`Final changes`
- Rebased to latest master.
- Simplified the preview by just showing important items:  _title_, _URL_, _deposit_, _price_, _transcode_ and _estimated_cost_.

`Issue`
- On Web, the preview takes forever to appear if the file is large. I believe the file is uploaded by `apiPublishCallViaWeb`.  I'm not sure if the file is needed to calculate the cost or not, nor was I able to figure out how to prevent the full upload.  
I don't think this is usable if we can't calculate the cost locally on the browser (the upload time is doubled).

Considering abandoning this...

seanyesmunt:
@infinite-persistence could we just not calculate the cost? I think it's still helpful to confirm all of the publish values.

infinite-persistence:
> _@infinite-persistence could we just not calculate the cost?_

Ah, given #486 [Add Confirmation before publishing file or creating channel (show fee estimate)](https://github.com/lbryio/lbry-desktop/issues/486), I thought that's the primary reason to add the dialog.
Ok, I'll remove the cost field;   solves the double-upload problem on Web.

seanyesmunt:
It would be nice to have, but if it's super slow I don't think it's necessary.

infinite-persistence:
Would "do it on Desktop only" be a good idea?

infinite-persistence:
`Changes`
- Disable 'tx-fee' field to handle the web delay problem.
- Undo the simplification -- all fields will be previewed.
- Rebranding update: "Publish" --> "Upload";  Use LBC icon.
- Hide irrelevant fields (`File`) in MD-POST mode.

tzarebczan:
I wonder if we can just send a temp file up (like we do with post publishing) on web for the preview, and then send the actual file during confirmation. The file itself is still needed, but it doesn't matter what file is sent as this does not affect the cost. The main factor for tx cost is the length of the URL

infinite-persistence:
> _I wonder if we can just send a temp file up (like we do with post publishing) on web for the preview, and then send the actual file during confirmation. The file itself is still needed, but it doesn't matter what file is sent as this does not affect the cost. The main factor for tx cost is the length of the URL_

It didn't work for me yesterday, but after seeing your comment, I gave it another shot and discover my syntax mistake.  It works!
It still takes a few seconds compared to Desktop, but not a very long wait.

Fix coming up...

infinite-persistence:
`Change`
- Send a dummy file on Web preview to fix the delay problem.

seanyesmunt:
Awesome!

PR #4624, opened 2020-08-03
use new buffer api

seanyesmunt:


ui/analytics.js
```diff
@@ -17,6 +17,8 @@ const devInternalApis = process.env.LBRY_API_URL;
 export const SHARE_INTERNAL = 'shareInternal';
 const SHARE_THIRD_PARTY = 'shareThirdParty';
 
+const BUFFER_API = 'https://collector-service.dev.lbry.tv';
```
seanyesmunt, line 20: will update this/move to config once I can successfully test it

PR #4649, opened 2020-08-11
Fix flow errors

btzr-io:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number:

### DIFF
--- a/flow-typed/web-file.js
+++ b/flow-typed/web-file.js
@@ -1,7 +1,6 @@
-declare type WebFile = {
-  name: string,
+// @flow
+
+declare type WebFile = File & {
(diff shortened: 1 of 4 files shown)


btzr-io:
`WebFile` type should inherit from `File` or `blob`, Idk how to do that or if that's possible.

seanyesmunt:
To inherit types you could do something like this:
```
delcare type WebFile = File & { ... }
```

Here is an example https://github.com/lbryio/lbry-redux/blob/master/flow-typed/Lbry.js#L119-L121

PR #4664, opened 2020-08-14
Publish: Make 'Channel' setting persistent.

infinite-persistence:
## Issue
Users are annoyed with the constant reset of the 'channel' setting in the Publish page.

## Changes
1. Revert the previous attempt in ff7b4092. The `usePersistedState` method is bad, as it will clash with the Redux value.  It only worked on initial entry, which was ok back then with the old version of Publish, but not with the new dynamic version of Publish (and it's probably partially broken too back then).
2. Implemented the persistence in Redux -- requires "https://github.com/lbryio/lbry-redux/pull/347".

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
     "imagesloaded": "^4.1.4",
     "json-loader": "^0.5.4",
     "lbry-format": "https://github.com/lbryio/lbry-format.git",
-    "lbry-redux": "lbryio/lbry-redux#7df96d47671308c09db0a28892dbbfd6486d81f4",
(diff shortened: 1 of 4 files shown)


kauffj:
:trophy:

seanyesmunt:
Nice. @infinite-persistence I just merged the lbry-redux PR. Can you update the commit hash in the package.json? Then this is ready to merge

infinite-persistence:
`Change`
Updated commit hash for _lbry-redux_.

kauffj:
Great software comes from lots of little improvements like this. Thanks @infinite-persistence!

PR #4682, opened 2020-08-20
Fix changelog

infinite-persistence:
- Added entry for the new "in-app editor"
- Fixed URL and sentence (made a guess on what it was supposed to be) for the file-property icon location change.
- Fixed missing parenthesis.


seanyesmunt:
Thank you!

PR #4701, opened 2020-08-25
String update for 0.47.2

infinite-persistence:
String changes from yesterday, and some old untranslated strings.


seanyesmunt:
Thank you!

PR #4716, opened 2020-08-28
Sync improvements

jessopb:


ui/component/app/view.jsx
```diff
@@ -257,7 +258,18 @@ function App(props: Props) {
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [hasVerifiedEmail, syncEnabled, checkSync]);
+  }, [checkSync, readyForSync, syncEnabled]);
+
+  // Currently we know someone is logging in or not when we get their user object {}
+  // We'll use this to determine when it's time to pull preferences
+  // This will no longer work if desktop users no longer get a user object from lbryinc
+  useEffect(() => {
+    if (user) {
+      if (typeof user === 'object' || typeof user === 'string') {
```
seanyesmunt, line 268: When would `user` be a string?

ui/page/settings/view.jsx
```diff
@@ -211,11 +211,20 @@ class SettingsPage extends React.PureComponent<Props, State> {
             <Card
               title={__('Sync')}
               subtitle={
-                walletEncrypted && !storedPassword
-                  ? __("To enable this feature, check 'Save Password' the next time you start the app.")
-                  : null
+                walletEncrypted && !storedPassword ? (
+                  storedPassword === '' ? (
+                    <>
+                      {__('Your wallet password cannot be empty string. Sync is disabled for now.')}
```
seanyesmunt, line 217: I don't think this is the message we should show. Why do we have to disable it for empty string?

ui/page/settings/view.jsx
```diff
@@ -211,11 +211,20 @@ class SettingsPage extends React.PureComponent<Props, State> {
...
+                      <HelpLink href="https://lbry.com/faq/account-sync" />
+                    </>
+                  ) : (
+                    __(
+                      "Encrypted wallet: To enable Sync, close LBRY completely and check 'Remember Password' during wallet unlock."
```
seanyesmunt, line 222: Remove `Encrypted wallet:`

`To enable sync, close the app completely, then check "Remember password" the next time you start it up`

ui/redux/actions/app.js
```diff
@@ -600,28 +600,33 @@ export function doGetAndPopulatePreferences() {
       const successState = getState();
       const daemonSettings = selectDaemonSettings(successState);
 
+      console.log('DGAP SUCCESS savedPrefs', savedPreferences);
```
seanyesmunt, line 603: console

ui/redux/reducers/settings.js
```diff
@@ -6,6 +6,7 @@ import { getSubsetFromKeysArray } from 'util/sync-settings';
 import { UNSYNCED_SETTINGS } from 'config';
 const { CLIENT_SYNC_KEYS } = SHARED_PREFERENCES;
 const settingsToIgnore = (UNSYNCED_SETTINGS && UNSYNCED_SETTINGS.trim().split(' ')) || [];
+// if (IS_WEB) settingsToIgnore.push(SETTINGS.ENABLE_SYNC);
```
seanyesmunt, line 9: extra comment

ui/page/settings/view.jsx
tzarebczan: It would prevent an encrypted wallet from getting decrypted via a sync apply + no password  (this sets a new password). Probably should disallow this on the SDK side and force with a flag. It shouldn't happen unless someone is in a really strange scenario, so I'm okay with removing it for now.

ui/component/app/view.jsx
jessopb: It's a bug in the code right now. I'll just make it if (user).

PR #4721, opened 2020-09-01
Web: Fix 'Download' not triggering until second attempt

### LINKED ISSUE #4669: Download doesn't trigger on web until 2nd attempt
I notice many times when I click download, it won't start until I click it a 2nd time.

infinite-persistence:
## Issue
Closes #4669: [Download doesn't trigger on web until 2nd attempt](https://github.com/lbryio/lbry-desktop/issues/4669)

The issue only happens when _Autoplay_ is disabled in the User Settings and the video hasn't been loaded when _Download_ is clicked.

The following code:
   `if (didClickDownloadButton && streamingUrl)`
didn't triggered because:
1. `streamingUrl` has not resolved yet when the Effect ran.
2. When it did resolve, the parent component was also notified and unmounted things, causing `didClickDownloadButton` to reset.

## Approach
- Change the state to the persisted version to deal with the unmounting.
- As a secondary precaution against any unexpected mismatch from the mounting/unmounting, the state was changed from `boolean` to `string`, with the compared value being the current `uri` when the Download button was clicked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Tip Modal: Don't do final submit when the intention is to create New Channel _community pr!_ ([#4629](https://github.com/lbryio/lbry-desktop/pull/4629))
 - Fix related + search results loading slowly ([#4657](https://github.com/lbryio/lbry-desktop/pull/4657))
 - Fix floating player being paused after dragging _community pr!_ ([#4710](https://github.com/lbryio/lbry-desktop/pull/4710))
+- Web: Fix 'Download' not triggering until second attempt _community pr!_ ([#4721](https://github.com/lbryio/lbry-desktop/pull/4721))
 
 ## [0.47.1] - [2020-07-23]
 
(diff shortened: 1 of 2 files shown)


seanyesmunt:
I think a potentially better fix would be to make sure this doesn't get removed from the page when while `streamingUrl` is being fetched. Then we don't need a new persisted value for every download you click.

If that's not possible, we should create a new effect `useSessionState` which is the same as `usePersistedState` except it uses `sessionStorage` instead of `localStorage` so the value dies when you leave the page.

infinite-persistence:
```
  const ActionWrapper = (props: { children: Node }) =>
    isMobile ? (
      <React.Fragment>{props.children}</React.Fragment>
    ) : (
      <div className="section__actions section__actions--no-margin">{props.children}</div>
    );
```
```
  <ActionWrapper>
    {!SIMPLE_SITE && <FileDownloadLink uri={uri} />}
```
Moving `{!SIMPLE_SITE && <FileDownloadLink uri={uri} />}`  outside of `<ActionWrapper>` prevents the unnecessary unmounting. 

- [ ] Experimenting with alternatives to get the equivalent output... (besides duplicating everything in a large `if-else` block)

infinite-persistence:
`Rework`
Avoid the unnecessary unmounting by not using a conditional section wrapper within a return statement. React probably couldn't do the diffs when the conditional is at a section level.

PR #4744, opened 2020-09-09
add logging of player point of presence

kauffj:


seanyesmunt:
Noticed a drop in events after this was merged. Reverting to see if the PR was the culprit.

PR #4750, opened 2020-09-10
FloatingPlayer: Stay within screen when window is resized 

### LINKED ISSUE #4741: Pop out player disappears or is cut off
The pop out player may either disappear off the viewing windows or be cut off if the Window size changes. 

Reproduce: 
1) Maximize window size, move popout player to somewhere on bottom left. Make window size smaller. 

Expected: popout player always be moved into viewing window.

infinite-persistence:
## Issue
Fixes #4741: [Pop out player disappears or is cut off](https://github.com/lbryio/lbry-desktop/issues/4741)

## Approach:
- When dragging stops, determine the location of the floating player wtih respect to the main window in terms of percentage.
- When window is resized, roughly re-position based on the stored percentage.


seanyesmunt:
This is great!

PR #4759, opened 2020-09-11
Check search string for web URL prefixes and remove

ioancole:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: 4758

404 when pasting URL from web browser into search bar, and some channel/stream URLs being reformatted before submitting, causing a search to be executed as opposed to directly navigating to the correct page.

## What is the new behavior?

App navigates to correct page

## Other information

This bug was introduced by a refactor to the wunderbar component by me a few days ago. It's annoying to have to check for all 3 of prod, dev and local environment url prefixes in the production code, so I attempted to re-write. This introduced the bug and the 3-way check has been re-introduced by this PR. It seems like we are stuck with this check for now.

Another problem was the logic deciding whether to hide the search page URI header... (see changes below) It has now been fixed.

*** Please test this branch before accepting the PR. As far as I can tell, the wunderbar is working correctly now. However I am new to this project and I imagine there are still some behaviours/test cases that I am not aware of (... this led to the bug originally being introducted in the initial PR earlier this week).

Thanks!

### DIFF
--- a/ui/component/wunderbar/view.jsx
+++ b/ui/component/wunderbar/view.jsx
@@ -104,12 +104,11 @@ class WunderBar extends React.PureComponent<Props, State> {
     updateSearchQuery(value);
   }
 
-  onSubmitWebUri(uri: string) {
+  onSubmitWebUri(uri: string, prefix: string) {
     // Allow copying a lbry.tv url and pasting it into the search bar
     const { doSearch, navigateToUri, updateSearchQuery } = this.props;
 
-    const slashPosition = uri.indexOf('/');
-    let query = uri.slice(slashPosition);
+    let query = uri.slice(prefix.length);
     query = query.replace(/:/g, '#');
     if (query.includes(SEARCH_PREFIX)) {
       query = query.slice(SEARCH_PREFIX.length);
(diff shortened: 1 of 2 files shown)


ui/component/wunderbar/view.jsx
```diff
@@ -127,7 +126,14 @@ class WunderBar extends React.PureComponent<Props, State> {
     const { navigateToUri, doSearch, doShowSnackBar } = this.props;
 
     if (suggestion.type === SEARCH_TYPES.SEARCH) {
-      doSearch(query);
```
seanyesmunt, line 130: I don't think this needs to change. With this change you can't just type something in and search, it will always try to go to the file page if the search term is a valid uri

seanyesmunt (changes requested):

ui/component/wunderbar/view.jsx
ioancole: Thanks, I have pushed the branch without this change.

seanyesmunt: It looks like this code is still here

ioancole: Sorry, it's now gone. Let me know if there are any more comments.

seanyesmunt:
Sorry for the delay. Just got back from vacation. Checking this out now

When I run your branch and type "test" into the search bar, it still navigates to lbry://test, instead of the search page

seanyesmunt:
After testing again this seems to work. I guess I was on the wrong branch. Thanks for the fix!

ioancole:
That's great! Thanks

PR #4764, opened 2020-09-15
FloatingPlayer: Run clamp-to-screen code when going into floating mode. 

infinite-persistence:
## Issue
The previous code only handled the clamping if the FP is already floating when the main window is resized.

1. In fresh Odysee session, the floating player is always clipped on the right.
2. If you resize the desktop while not floating, the FP could be clipped when you go into floating mode.

## Changes
- Factor out the clamping code.
- Add another effect to handle 'isFloating' changes.


seanyesmunt:
Works perfectly! Thanks for this. Great addition.

PR #4772, opened 2020-09-18
Missing translation in the Discover page.

infinite-persistence:
Added the translation macro in the component parameter.  Most of the strings are already in `app-string.json`.


infinite-persistence:
```html
<div key={title} className="claim-grid__wrapper">
```
I'm not sure if `key` needs to be identical to the label or not, or simple needs to be unique.  I didn't translate it.  FYI, just in case.

seanyesmunt:
The `key` just needs to be unique

PR #4784, opened 2020-09-22
Remove development strings + various fixes

infinite-persistence:
## Issue
- Prevent translators from translating useless strings.

## Notes
- While 'Mark all as read' is technically valid, I've removed it since Notifications are not available in Desktop. When it is enabled, then we should include all the notification-related strings (there are a bunch).

_If this is accepted, I suggest doing another push in Transifex after the merge to remove those experimental strings._


infinite-persistence:
If this is accepted, I suggest doing another push in Transifex after the merge to remove those experimental strings.

PR #4809, opened 2020-09-30
shift enter to comment

jessopb:


ui/component/commentCreate/view.jsx
```diff
@@ -65,13 +67,26 @@ export function CommentCreate(props: Props) {
   function handleCommentAck() {
     setCommentAck(true);
   }
+  const buttonref: ElementRef<any> = React.useRef();
```
seanyesmunt, line 70: Please move this to the top of the component, above any `useState` uses

seanyesmunt (changes requested):
One comment. Trying to figure out some component rules and I think it's helpful to have refs at the very top to quickly tell there are refs somewhere in a component.

PR #4811, opened 2020-09-30
String scrub from odysee/release merge

infinite-persistence:
- More missing strings
- Moved strings around to keep related strings together (easier for translators to guess the context)
- Removed translation macro for blank string


seanyesmunt:
@infinite-persistence Thank you for keeping this up to date.

PR #4818, opened 2020-10-01
Expandable: Fade out section when 'closed'

infinite-persistence:
## Issue
The abrupt cut-off of the expandable section bugs me (looks like a rendering glitch), especially when it cuts off in the middle of a line.

## Change
In addition to the existing 'More' button, we fade out the section to provide additional visual cues.

## Approach
This solution doesn't require the background color to be known, so it will work regardless where `<Expandable>` is used, or whatever color-scheme is chosen.

However, it does utilize non-standard css -- for older browsers, it should simply cut-off like the before.

image

## Test
- [x] Firefox (working)
- [x] Chrome (working)
- [x] Opera (working)
- [x] Edge (fails with style and grace)
- [ ] Internet Explorer (couldn't get lbry to load at all)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


seanyesmunt:
@infinite-persistence this is great!

PR #4835, opened 2020-10-02
Claim List channel items' thumb fix

alojzjakob:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

Channel thumbs are square dimensions in 99.9% cases. New claim listing UI (for example "Discover" screen) assumes all thumbs are for content in some aspect ratio. That is why channel items look strange.
## What is the new behavior?
Channel thumbs' backgrounds are set to contain size and transparent background, looks much better


seanyesmunt:
I would like to avoid doing this since it makes the tile styling very inconsistent for different channels. A better fix would be to somehow make the channel profile image on tiles square, instead of rectangular.

PR #4863, opened 2020-10-07
email resend requests limited to once per thirty seconds

jessopb:


ui/component/userEmailVerify/view.jsx
```diff
@@ -46,8 +51,14 @@ class UserEmailVerify extends React.PureComponent<Props> {
 
   handleResendVerificationEmail() {
     const { email, resendVerificationEmail, toast } = this.props;
-    resendVerificationEmail(email);
-    toast(__('New email sent.'));
+    if (!this.state.wait) {
```
seanyesmunt, line 54: If `wait` exists, we should show a separate snack bar saying "please wait a bit" or something.

jessopb: done

PR #4874, opened 2020-10-09
Support for multiple string context

### LINKED ISSUE #4796: i18n: Allow support for string overloading (multiple contexts)
## ?
Let me know what you guys think, and I can take a stab at changing the handful of reported cases.

## Issue
While a word in English can be re-used in several areas, the localized version in many languages require a different word due to difference in context and grammar (e.g. noun vs. verb, past vs. present tense, etc.)

## History
This has been briefly discussed in the past, and is still being brought up from time to time in the Translators Hives in Discord.  I believe if we perform verification on the translated strings, we'll uncover even more instances.

## Suggested fix for the short term
On a case-by-case basis (as reported by the Translators), split the affected strings into several string IDs as shown below:

**English (all resolves to the same value):**
```javascript
"SID_SUPPORT_FOR_CLAIM": "Support",
"SID_SUPPORT_FOR_HELP": "Support",
```
**Chinese**
```javascript
"SID_SUPPORT_FOR_CLAIM": "支持",
"SID_SUPPORT_FOR_HELP": "援助",
```
**Client side:**
```javascript
 label={__('SID_SUPPORT_FOR_CLAIM'} 
````
## Cons
- While the suggested fix allows better translation without major changes to the entire code base, we'll end up with a mixture of "inline string" and "string ID" usages in the code, which can be a bit messy.  But I think using all caps for the string ID should make it obvious on what is going on.

### PR DESCRIPTION
**Pull request by infinite-persistence**
## Issue
- Closes #4796 [i18n: Allow support for string overloading (multiple contexts)](https://github.com/lbryio/lbry-desktop/issues/4796)
- Closes # 4192 [French] Following menu/page title not great 

## Approach
- Minimal code and process change.
- Handle on a case-by-case basis when reported by translators.
- Split the affected key in the string json by appending the context.
- Translators need to be aware of the new format and not translate context itself. Code is added to detect bad translations and will revert to English.

Sample in json:
```
  "About --[About section in Help Page]--": "About",
  "About --[tab title in Channel Page]--":  "About",
```
Sample in client code:
```
  title={__('About --[About section in Help Page]--')}
```

- `--[ ]--` was chosen as it's unique enough (unlikely for real strings to use it) and hopefully not that distracting in the client code.
- `"str --[ ]--" : "str"` - the value for _str_ must be identical for now, as currently the fallback is using the _key_'s value minus the context, instead of _value_ directly.  It's mainly due to the development phase case as en.json won't have that entry in the server yet.
- In the key itself, spaces are allowed after the _str_ (i.e. before `--[`) for neatness. It will be trimmed by the system.

## First example
"About" is used in 3 places:
   - Channel Page
   - Help Page
   - Footer (in Odysee branch)

For Russian, the word "About" is "O" and is usually not used standalone, but requires something behind it. A translator said so, and seems to be the case in other sites as well.
   "O xxx"
   "O yyy"

## Other languages
For other languages that are not impacted, they can just clone the same translation for each of the split keys, just like in English.

## Possible enhancement in Transifex
I see that Transifex's API includes a `context` entry. It might be possible to move the context-metadata there during the upload, so translators will never see the `--[]--` messiness (it will be shown nicelly as "Context: xxx" in the Transifex GUI).

I'm not sure how to test the Transifex side, so I did not investigate further.


infinite-persistence:
## Comments
- If this is not desired, simply close it and I'll get the memo 👍 
- Without being able to push to Transifex, I couldn't test the concept fully on translated json files.  But based on simulation in my head, I think there's nothing left to handle.  Hopefully Transifex doesn't choke on the new _key_ format.

seanyesmunt:
Sorry I haven't gotten to this. I think it's a great idea! Will merge once the conflicts are fixed

infinite-persistence:
`Updates`
- Rebased to fix conflict
- Snuck in a few string updates

PR #4895, opened 2020-10-14
Feat sort pinned to top

jessopb:
https://github.com/lbryio/lbry-redux/pull/356 required


seanyesmunt (changes requested):
One comment. I pushed up a commit that tweaks the pin icon since it was messed up a bit

PR #4907, opened 2020-10-15
Feat language search

jessopb:
claimSearch by language
setting to constrain search to language in settings
channel update for language (primary, secondary)


jessopb:
https://github.com/lbryio/lbry-redux/pull/358

PR #4921, opened 2020-10-19
Feat sort pinned to top

jessopb:


ui/component/comment/index.js
```diff
@@ -6,33 +6,48 @@ import {
   makeSelectThumbnailForUri,
   makeSelectIsUriResolving,
   selectMyChannelClaims,
+  makeSelectMyChannelPermUrlForName,
```
seanyesmunt, line 9: The app crashes here before channels have finished loading
https://github.com/lbryio/lbry-redux/blob/master/src/redux/selectors/claims.js#L670

PR #4950, opened 2020-10-27
Fix long tag query-URL when changing between [Trending|New|Top]

### LINKED ISSUE #4393: Tag query URL goes on indefinitely as you change between [Trending|New|Top]
**To Reproduce**
1. lbry.tv
2. `Your Tags`
3. Click on one of your tags on the right bar.
4. Click `New`
5. Keep changing between `Trending`, `New` and `Top`.  URL keeps going longer.

image

The query still works, just that the URL gets weird.

infinite-persistence:
## Issue
Fixes #4393: [Tag query URL goes on indefinitely as you change between [Trending|New|Top]](https://github.com/lbryio/lbry-desktop/issues/4393)

## Change
Change from `append` to `set` when building the query to remove duplicates.

Flow hates null being not a string, so the FixMe was retained.

### DIFF
--- a/ui/component/claimListHeader/view.jsx
+++ b/ui/component/claimListHeader/view.jsx
@@ -140,8 +140,8 @@ function ClaimListHeader(props: Props) {
   function buildUrl(delta) {
     const newUrlParams = new URLSearchParams(location.search);
     CS.KEYS.forEach(k => {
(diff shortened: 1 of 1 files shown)


seanyesmunt:
@infinite-persistence thanks for the fix!

PR #4962, opened 2020-10-28
Restore accidental string deletion

infinite-persistence:
These strings are in use and have been translated by the translators. I think it got accidentally removed from the "send auth token with status call" commit.


infinite-persistence:
Sorry @seanyesmunt, my bad.  Those are duplicate-removals, as things got moved around, it seems.
Can you revert this?  Oops.

infinite-persistence:
Never mind, I have another PR coming up with new strings.  I'll fix the duplicates there.

PR #4967, opened 2020-10-30
Fix str-related issues with new Language setting

infinite-persistence:
```
  "Hrvatski": "Hrvatski",
  "Nederlands": "Nederlands",
  "Français": "Français",
  "Deutsch": "Deutsch",
  "Italiano": "Italiano",
  "Polski": "Polski",
  "Português": "Português",
  "Русский": "Русский",
  "Español": "Español",
  "Türkçe": "Türkçe",
  "Česky": "Česky",
```
**Remove translation macro on native language names.**
I believe we don't translate these since this is already the translated form?
(Well, there's a possibility that the original intention was to translate those languages into everyone's local language.  If that was the intention, I think we'll need to change those strings to the English version instead?  e.g. I don't know what Русский is to make the translation without googling)

**Use the contextual version of 'Post' instead of creating a new one.**
I couldn't find where this new 'Post' is being used, so I assume it's the same as the old one for now. If it appears in the list again, I'll ask around to put the context-metadata.

**Add 'Search only in this language by default'**

### DIFF
--- a/static/app-strings.json
+++ b/static/app-strings.json
@@ -1462,6 +1462,7 @@
   "None selected": "None selected",
   "Secondary Language": "Secondary Language",
   "Your other content language": "Your other content language",
+  "Search only in this language by default": "Search only in this language by default",
   "This link leads to an external website.": "This link leads to an external website.",
   "Hold on, we are setting up your account": "Hold on, we are setting up your account",
   "No Content Found": "No Content Found",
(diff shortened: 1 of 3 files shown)


infinite-persistence:
Hm ... something is still not right.  Moving back to draft ...

infinite-persistence:
Ah, found the other macro usage in the Search Filter

seanyesmunt:
Thanks. Just noticed these.

PR #4996, opened 2020-11-06
claim_search: Don't clear past page results if subsequent pages timeout.

### LINKED ISSUE #4609: If claim search on latter page fails, don't clear out all previous data
See https://lbry.tv/$/discover?t=imherelbry&t=imherelbry&t=imherelbry&t=imherelbry&t=imherelbry&t=imherelbry&content=video

infinite-persistence:
## Issue
Fixes #4609: [If claim search on latter page fails, don't clear out all previous data](https://github.com/lbryio/lbry-desktop/issues/4609)

## Change
This requires an accompanying change in `lbry-redux` (https://github.com/lbryio/lbry-redux/pull/365) that sets `claimSearchResultsLastPageReached` to `true` when the timeout occurs, and also not purging the previous pages.

## Test Case:
1. https://lbry.tv/$/discover?t=imherelbry&content=video&order=top&fresh=year
2. Scroll down 5 pages until 100 results are shown.
3. When querying the 6th page, a timeout occurs and the past results are gone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - Loss of subscriptions on startup ([#4882](https://github.com/lbryio/lbry-desktop/pull/4882))
+- Fix lost search results when a timeout occurs _community pr!_ ([#4996](https://github.com/lbryio/lbry-desktop/pull/4996))
 
 ## [0.48.2] - [2020-10-16]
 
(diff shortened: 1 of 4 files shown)


infinite-persistence:
`Bug`
`createNormalizedClaimSearchKey` ignores the freshness, so `claimSearchResultLastPageReached` could be stuck at the same value when freshness is changed, causing no query was submitted.

infinite-persistence:
`Changes`
Redo by using a separate flag to handle partial-page timeout, and also display the timeout msg.  See commit message for further details.

infinite-persistence:
`Change`

Reverted back to the simple version of just using an existing flag + not purging if results already exists. The only downside to this is that the "timeout" message will no longer appear, and it might look unresponsive when switching between various filters that all result in timeout.

PR #5003, opened 2020-11-09
custom homepages

jessopb:


ui/redux/reducers/settings.js
```diff
@@ -39,6 +42,7 @@ const defaultState = {
     [SETTINGS.LANGUAGE]: settingLanguage.find(language => SUPPORTED_LANGUAGES[language]),
     [SETTINGS.THEME]: __('dark'),
     [SETTINGS.THEMES]: [__('light'), __('dark')],
+    [SETTINGS.HOMEPAGE]: settingLanguage.find(language => homepageKeys.includes(language)),
```
seanyesmunt, line 45: What happens if there is no match?

jessopb: I was always going to call the default 'en'

PR #5012, opened 2020-11-10
pass domain to authenticate for appid

jessopb:


jessopb:
merge and bump after https://github.com/lbryio/lbryinc/pull/105

Issue #4913, opened 2020-10-16
Can't slide volume slider in pop out mode


tzarebczan:
It resorts to dragging the pop out window. You can however click inside the slider to change the volume.

tzarebczan:
This was fixed

PR #5034, opened 2020-11-13
FloatingPlayer: fix unusable volume slider in fp-mode

infinite-persistence:
## Issue
#4913: [Can't slide volume slider in pop out mode](https://github.com/lbryio/lbry-desktop/issues/4913)

## Approach
I first tried to handle this at the floating player level, but it was impossible to hack there due to how `react-draggable` works (it already moved the window before the "move" handler is called, so we can't do much).

Fix by overriding the dragging behavior of the videojs' `VolumeBar` class by not propagating the event further. It is odd that videojs didn't already do this, since it's `SeekBar` does stop the propagation.


seanyesmunt:
Nice fix!

PR #5056, opened 2020-11-17
Customize scrollbar to fit the theme colors

### LINKED ISSUE #4727: Sidebar's transient scrollbar doesn't fit in the Dark Theme
Having the default bright scrollbar for the Sidebar in the Dark Theme feels a bit jarring.  YT's dark theme uses a dark scrollbar.

Aside: It would be nice if the main scrollbar can be dark as well, but it's less distracting compared to the Sidebar's since the main scrollbar is on the far right.

infinite-persistence:
## Issue
Closes #4727: [Sidebar's transient scrollbar doesn't fit in the Dark Theme](https://github.com/lbryio/lbry-desktop/issues/4727)

## Approach
The 'webkit' API works on Chrome, but Firefox ignores it and uses it's own fancy scrollbar when 'webkit' is used (can't seem to change any properties). It's better than the current look, but still doesn't fit our theme, plus it makes the product look different on different browsers.

Firefox now supports the new 'scrollbar' API, so we can now tweak the scrollbar, but the API only provides limited attributes.

Don't wanna complicate things by using Javascript, so we'll use Firefox's limited attributes as the common denominator and just live with a plain, rectangle scroll bar BUT with better color to match the theme.

I think this color works for both lbry.tv and Odysee (both light and dark on each platform).


seanyesmunt:
This looks great! I think this should only happen on windows/linux though.

infinite-persistence:
@seanyesmunt, ah yes, forgot to mention that I didn't test on mac.
Is there an easy way to not include `_scrollbar.scss` globally when mac is detected?

infinite-persistence:
@seanyesmunt. Ah, I found it ... in `ui\index.jsx`

infinite-persistence:
@seanyesmunt  Hm ... it didn't work.  Was hoping to do something like the following, but `import` only works at the top level.
Is there an easy way to not include _scrollbar.scss globally when mac is detected?

```
// Import our app styles
// If a style is not necessary for the initial page load, it should be removed from `all.scss`
// and loaded dynamically in the component that consumes it
import 'scss/all.scss';

// Apply global scrollbar customization
const IS_MAC = process.platform === 'darwin';
if (IS_MAC) {
  import 'scss/component/_scrollbar.scss';
}
```

infinite-persistence:
`Update`
Assuming I understand the change-request correctly, the second commit covers it.  See commit message for details.

seanyesmunt:
Sorry I missed these messages. Was trying to get an app release out last week but have been having some build issues so this is going in too! Your solution looks great.

PR #5076, opened 2020-11-20
Prevent multiple embeds from creating new users

jessopb:
Store a lock value in session storage while authenticating to prevent multiple userIds from multiple embeds.


jessopb:
Should test desktop

seanyesmunt (approved):
Good to merge pending test on dev site

PR #5092, opened 2020-11-24
Reset scroll position on new Regular-Search query

### LINKED ISSUE #4783: New search query does not reset to the top
1. Search for anything.
2. Set to `100 results` and close the `Filter` expansion
3. Scroll down to somewhere in the middle.
4. Search for something else.  The list clears while you are typing, but restores to the previous scroll position after the search is done.

**_Should it reset the scroll position?_** (just in case it was intentional)

Actually, I think I was facing this issue with Tag Search with several pages of infinite scroll, which for this case is more troublesome as there is no Home key in Android to scroll all the way up.  But now I couldn't reproduce it on Desktop nor Mobile.  It's less annoying in Regular Search since there are only 100 results max to scroll up.

infinite-persistence:
## Issue
Fixes #4783: [New search query does not reset to the top](https://github.com/lbryio/lbry-desktop/issues/4783)

## Changes
Do the same things as Tag Search, which is to reset the scroll position on this scenario.


infinite-persistence:
`Reject`
While looking at several other scroll-related issues, I think this is not the best solution. I'm looking at an alternative that would solve the other issues as well, plus adding the ability to restore the scroll location when one navigates forward in history (currently, only backward history is supported).

PR #5119, opened 2020-12-01
Video: Mobile UI + overlay for keyboard shortcut feedback

### LINKED ISSUE #4889: Alt-Left shouldn't impact video
Brave Browser, with Floating Player enabled:
- Ok scenario: Play a video and press the browser's Back button.  The page goes back while the video continues to play correctly.
- Bad scenario: Play a video and press alt-left to go back.  The page goes back, but the video goes back n seconds (accidentally captured the left key action)

infinite-persistence:
## New
- Mobile version of the video play.  Includes:
   - Centered and larger `play/pause` button
   - Tap on left or right area of screen to seek +/- 10s
image
- Overlay for keyboard shortcut feedback:
   - Playback-rate changes
image
   - +/-10s seeking

## Approach
- `ui/component/videoViewer/internal/plugin` will contain the plugin files.  Changes done to these files will be minimal so ease any future patching from the original source.  The inline method is used to handle importing issues with Node modules.

## Note to reviewer
- [ ] In the last commit, I disabled the mobile UI for IOS and only enabled it for Android.  I believe the native UI was intentionally used before this, so I tried to retain the same behavior.  It's also because I don't have an iOS to actually confirm that it works, although the emulator in the browser seems to show that it does.
- [ ] For the overlays, I have not tried on a Mac/Safari/iOS -- again, the native player is used there, but should try it just in case something goes bad.

## Also
Closes #4889: Alt-Left shouldn't impact video


infinite-persistence:
`Update`
- Re-factored the method to add "one-off" overlays
- Added overlay for "seeked forward/backwards".

seanyesmunt:
Will look at this after the desktop release is out. Hopefully soon 😞

infinite-persistence:
`Update`
- Added mobile UI and updated the main description of this PR to reflect the sum of changes.

ui/component/viewers/videoViewer/internal/plugins/videojs-overlay/LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright Brightcove, Inc.
```
seanyesmunt, line 1: This one doesn't need to be included.

ui/component/viewers/videoViewer/internal/videojs.jsx
```diff
@@ -285,24 +293,25 @@ export default React.memo<Props>(function VideoJs(props: Props) {
...
       let rateIndex = videoPlaybackRates.findIndex(x => x === rate);
       if (rateIndex >= 0) {
-        rateIndex =
-          e.keyCode === PERIOD_KEYCODE
-            ? Math.min(rateIndex + 1, videoPlaybackRates.length - 1)
-            : Math.max(rateIndex - 1, 0);
+        rateIndex = isSpeedUp ? Math.min(rateIndex + 1, videoPlaybackRates.length - 1) : Math.max(rateIndex - 1, 0);
+        const nextRate = videoPlaybackRates[rateIndex];
 
-        player.userActive(true); // Bring up the ControlBar as GUI feedback.
```
seanyesmunt, line 304: I think it's still helpful to bring up the control bar when someone skips forward

seanyesmunt (changes requested):
Minor comments. Looks great

ui/component/viewers/videoViewer/internal/plugins/videojs-overlay/LICENSE
infinite-persistence: Will that be the one line, or the entire file?

infinite-persistence:
`Update`
- Remove the LICENSE file for both plugins.
- Bring up the control-bar when "seek" and "rate" is changed via keyboard.
- Updated CHANGELOG to place the item above 0.49.0

seanyesmunt:
I think we still need one of the licenses (don't remember which one). It said it had to be included with the code I think.

infinite-persistence:
`Update`
- Restored the MIT license file for `vjs-mobile-ui`
- Rebased to latest master.

PR #5140, opened 2020-12-05
Repost from search

jessopb:


jessopb:
Will rebase in the AM. Take a look.

jessopb:
Requires https://github.com/lbryio/lbry-redux/pull/369

PR #5150, opened 2020-12-08
top page selects channel tab if channel is winning

jessopb:


ui/component/searchTopClaim/view.jsx
```diff
@@ -14,21 +14,28 @@ type Props = {
...
+  const { doResolveUris, query = '', winningUri, hideLink = false, setChannelActive } = props;
   const uriFromQuery = `lbry://${query}`;
 
   let channelUriFromQuery;
+  let winningUriIsChannel;
   try {
     const { isChannel } = parseURI(uriFromQuery);
-
+    const { isChannel: winnerIsChannel } = parseURI(winningUri);
+    winningUriIsChannel = winnerIsChannel;
```
seanyesmunt, line 29: This can just use `isChannel`. It doesn't need to call `parseURI` twice

jessopb: I'll take another look - one is from the query, the other from the return of winningUri.

PR #5151, opened 2020-12-08
grayscale lbc when not winning on channels test

jessopb:
notsure


kauffj:
this is a very good idea but let's do this another way

PR #5156, opened 2020-12-09
Turn timestamps in a video markdown (description, comments) into links

### LINKED ISSUE #5103: Automatically turn timestamps in a video description into links
**Is your feature request related to a problem? Please describe.**
A number of YouTubers who cross-post to LBRY tend to use timestamps in their video descriptions. For example, [IvanOnTech](https://lbry.tv/@ivanontech:a/noo-way-bitcoin-dumps-faster-ready-for) or [TheCryptoLark](https://odysee.com/@TheCryptoLark:e/warning-bitcoin-dump-ethereum-pump:f). On YouTube, these timestamps are links, making it very simple to skip to whatever is interesting to the viewer. Whenever I watch these channels on LBRY, I'm frustrated that I have to find  the correct time in the video manually (especially frustrating on mobile, where it's basically impossible to find the correct timestamp, especially with longer videos).

**Describe the solution you'd like**
Timestamps in video descriptions would be automatically converted to links. Clicking a link would automatically skip to that time in the video.

**Describe alternatives you've considered**
There could be fancier ways to present these timestamps — e.g. in a separate section. However, that would remove the context in which these timestamps were added to the description, and that context could potentially be quite important.

infinite-persistence:
## Issue
Fixes  #5103 [Automatically turn timestamps in a video description into links ](https://github.com/lbryio/lbry-desktop/issues/5103)

## General behavior
- Works in `description` and `comments` in a video page.
- Invalid timestamps like `03:99` will be ignored.

## Notes
- In general, if the video player doesn't exist, the link will simply be a no-op (i.e. safe).  However, I do try to _not_ turn it into links when we know for sure it's not in a video page, e.g. in a Markdown Post.
- If _autoplay_ is turned off and video remains paused, and there will be no video instance (hence, a no-op as well).  I'm not sure if we should start playing it when the link is clicked. I think it's ok to just leave it paused (no-op).  I didn't want to duplicate code from `EmbedPlayButton` just for a corner-case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Turn timestamps in a video's description and comments into links _community pr!_ ([#5156](https://github.com/lbryio/lbry-desktop/pull/5156))
+
+### Changed
+
+### Fixed
+
+## [0.49.0] - [2020-12-10]
+
+### Added
+
 - Better search suggestions ([#5124](https://github.com/lbryio/lbry-desktop/pull/5124))
 - Inline video player in comments/markdown posts ([#4894](https://github.com/lbryio/lbry-desktop/pull/4894))
 - Better handling of winning claim + top page improvements ([#4944](https://github.com/lbryio/lbry-desktop/pull/4944))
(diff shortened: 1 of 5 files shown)


kauffj:
you are fantastic

seanyesmunt:
This is great!

seanyesmunt:
Reverting this because it breaks in Safari
```
SyntaxError: Invalid regular expression: invalid group specifier name
```

Didn't look too closely, not sure what the issue is.

infinite-persistence:
Ah, Safari not supporting "look back" regex yet.
Re-working the regex ...

PR #5199, opened 2020-12-17
Timestamp: Handle Safari + better parsing

infinite-persistence:
## Issue
- The previous regex in #5156 uses lookback (I think ES2018?) which Safari has yet to implement.
- There were a few bugs in the previous regex too, like missing out multiple timestamps in a line, or parsing "62:02" as "2:02" (although YT does this as well).
- The previous method searched too deep (matchAll) on each 'locate' call, which was wasteful.

## Known issue(s)
- `List` works, but `Bullet` does not.  The same thing happens for "@channel" parsing.  The `notInList` flag doesn't seem to work, or something is blocking it.

## Remark
I'm still trying to get BrowserStack to work on localhost instances to test on Safari.  Since a simpler regex is now used, plus I don't think there's any other browser-dependent code on this PR, it should work in Safari.  Can you help to double check?

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


seanyesmunt:
Works great in Safari!

PR #5212, opened 2020-12-18
Add channel staked indicator to UriIndicator

seanyesmunt:
Now channels can show off their street cred

This is disabled by default, will not be on lbry.tv


seanyesmunt:
@kauffj I'm assuming this will continue to be tweaked but I think this is releasable. It's live on nsa.lbry.tv.

The level logic is here
https://github.com/lbryio/lbry-desktop/blob/channel-bling/ui/component/channelStakedIndicator/view.jsx#L16-L35

jessopb (approved):
:+1:

PR #5216, opened 2020-12-21
i18n update

infinite-persistence:
Reported by translators

Aside:
`Do a thing` sounds weird.  Should we change it?  Wasn't sure if it was an intentional meme, so I added the string anyways.
image


seanyesmunt:
I updated that placeholder text.

PR #5231, opened 2020-12-23
Actual repost changes

jessopb:


ui/component/claimPreview/view.jsx
```diff
@@ -195,31 +206,37 @@ const ClaimPreview = forwardRef<any, {}>((props: Props, ref: any) => {
   }
 
   useEffect(() => {
-    if (isValid && !isResolvingUri && shouldFetch && uri) {
```
seanyesmunt, line 198: Why remove the check for `isResolvingUri`?

jessopb: It caused a weird infinite loop. I'll try again with it in..

jessopb: Apparently unable to reproduce - I put it back.

PR #5271, opened 2021-01-07
Add Chromecast support on cast-able browsers

infinite-persistence:
## Issue
Currently, we can only cast from Android Chrome (Mobile).  This PR extends it to browsers that support casting.  Goggle Chrome supports it by default, while other browsers will require some flag-tweaking or installing an extension.

## Notes
- The chromecast API ignores `http`, so `dev-server` alone won't work.  Needs `https`.  I used `ngrok` for testing.

## Screenshot
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "@reach/utils": "^0.12.1",
     "@sentry/browser": "^5.12.1",
     "@sentry/webpack-plugin": "^1.10.0",
+    "@silvermine/videojs-chromecast": "^1.2.2",
     "@types/three": "^0.93.1",
     "adm-zip": "^0.4.13",
     "async-exit-hook": "^2.0.1",
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -10,6 +10,11 @@ import * as OVERLAY from './overlays';
 import './plugins/videojs-mobile-ui/plugin';
 import isUserTyping from 'util/detect-typing';
 
+// @if TARGET='web'
+import '@silvermine/videojs-chromecast/dist/silvermine-videojs-chromecast.css';
+require('@silvermine/videojs-chromecast')(videojs);
+// @endif
+
 const isDev = process.env.NODE_ENV !== 'production';
 
 export type Player = {
@@ -31,6 +36,7 @@ export type Player = {
   userActive: (?boolean) => boolean,
   overlay: any => void,
   mobileUi: any => void,
+  chromecast: any => void,
   controlBar: {
     addChild: (string, any) => void,
   },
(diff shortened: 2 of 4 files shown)


tzarebczan:
Thanks for this! Have you looked into any of the videojs chromecast plugins? Maybe that could make it work on the desktop app...

infinite-persistence:
- I tried a few and silvermine's seems to be the most professional-looking, so I used that.
- There's another that replicates the mdns stuff that the browser is doing (that should make it work for Desktop), but I couldn't get it to work.  Also, it seems to bring in a lot of dependencies.
- I'll be looking at https://github.com/thibauts/node-castv2-client separately to see if that'll cover the missing pieces for Desktop.  I think some work is also needed to cast the global URL of the video instead of the `localhost`-based electron url (?).

_Disclaimer: I'm still wrapping my head around the chromecast system, so feel free to correct anything above :D_

tzarebczan:
Yes, it would definitely need to cast the web URL as the sdk listens on localhost by default.  Thank you for looking into this!

seanyesmunt:
I deployed this branch to https://nsa.lbry.tv

@tzarebczan Can you give it a try there?

package.json
```diff
@@ -46,6 +46,7 @@
     "postinstall:warning": "echo '\n\nWARNING\n\nNot all node modules were installed because NODE_ENV is set to \"production\".\nThis should only be set after installing dependencies with \"yarn\". The app will not work.\n\n'"
   },
   "dependencies": {
+    "@silvermine/videojs-chromecast": "^1.2.2",
```
seanyesmunt, line 49: This should be a `devDependency`

ui/component/viewers/videoViewer/internal/videojs.jsx
```diff
@@ -151,6 +157,9 @@ export default React.memo<Props>(function VideoJs(props: Props) {
     ],
     autoplay: false,
     poster: poster, // thumb looks bad in app, and if autoplay, flashing poster is annoying
+    // @if TARGET='web'
+    techOrder: ['chromecast', 'html5'],
```
seanyesmunt, line 161: What does `html5` mean here?

seanyesmunt (commented):
I think this looks ok. Left a couple comments

infinite-persistence:
`Update`
- Rebased to fix conflict with theater mode.
- Moved plugin to `devDependency`

ui/component/viewers/videoViewer/internal/videojs.jsx
infinite-persistence: It's the default "tech" for videojs, the one that wraps `<video>` I guess.  Other potential ones include Flash and Silverlight.
Per plugin documentation, it wants to be before the default tech.

infinite-persistence:
`Set as Draft`
Investigating desktop version...

infinite-persistence:
## Update
`Finalize web effort`
- Show the title and channel instead of "Playing on Default Media Receiver" on the TV screen. This change involves adding an extra prop -- I think it's fine though, but please double-check.
- Re-factored the code to not scatter `@if web` calls everywhere (confining them into one class). This should make it easier to revert/disable when needed.
- Updated the PR description to include other browsers. It'll work on other browser like Brave after tweaking some internal flags or installing the cast extension.

`Abandoning Desktop effort`
- I tried `nodecast` and others for the discovery part.  While I managed to connect to my Chromecast, I can't get playback to work properly.  My URL conversion didn't work, or perhaps I'm missing something. I'll drop this one for now.

------------
If the desire is to have Desktop support, feel free to close this PR :)

seanyesmunt:
That's alright. Thanks for investigating! I'm pretty swamped at the moment so I might not get to testing/reviewing until Monday but this is great! Excited to get it in 🙂

seanyesmunt:
Looks great! Nice work

PR #5276, opened 2021-01-08
Wallet balance UI refactor

seanyesmunt:
Just needs updated copy (and to update app-strings once new copy is figured out)

Screen Shot 2021-01-12 at 10 18 23 AM


tzarebczan:
Let's add inline available balance to places where credits are spent

PR #5305, opened 2021-01-13
encode share url parts

seanyesmunt:
Only encodes the claim/stream name this time, not the whole url


jessopb (approved):
Seems fine. Yay 1 url

PR #5355, opened 2021-01-22
g ads on master

jessopb:


jessopb:
this only shows up on SIMPLE_SITE

PR #5361, opened 2021-01-22
manually encode "'" because encodeURIComponent doesn't encode it

seanyesmunt:
This PR has made me realize we should probably combine `util/web.js` and `util/url.js` (or at least combine parts of them)


jessopb (approved):
works

Issue #5263, opened 2021-01-04
cannot scrub through content on IOS + Audio


tzarebczan:
See https://open.lbry.com/@ChristRescues:a/Biosecurity-Pharma-challenges-overcome:8 on ios, there's no scrub bar when you click the video

DispatchCommit:
this is caused by / duplicated by https://github.com/lbryio/lbry-desktop/issues/5323

#5344 should therefore resolve this

DispatchCommit:
@seanyesmunt can you verify if this issue is resolved?

PR #5367, opened 2021-01-23
Fix: Controls do not show on iOS

DispatchCommit:
This attempts to remove code that prevents any controls from appearing on iOS. This may not fully resolve the issue though, as there may be an issue with both native and custom controls appearing on iOS devices.
However, it's a lot easier to remove / disable when they appear, and very difficult to understand what is preventing the appearance when they do not show at all.

this pr does the following:
- remove `video-js-parent--ios` class which hid control bar via css.
- remove `nativeControlsForTouch` option in favor of allowing for default behavior.
- remove `IS_IOS` const as the video.js library already defines a value that could be used instead, and there are no longer any references to it.
- Add `videojs.browser.IS_IOS` back to videojs-mobile-ui plugin as we should aim to have a unified experience across all devices. (also it might fix the controls not appearing issue)
- remove styling for `.video-js-parent--ios` class as we no longer use this class
- move `videoJsOptions.muted = startMuted;` to where `videoJsOptions` is initially defined.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Code style update (formatting)
- [ ] Documentation changes

## Fixes

Issue Number: #5263 (maybe)

no controls on iOS

## What is the new behavior?

hopefully more controls on iOS. Possible too many controls.

## Other information

This may introduce a new bug. But that bug would be less troublesome than current behavior.
It is unclear why certain modifications were made to some settings and code by other devs, so this first pass attempts to simplify the problem first. 

It's possible that both native controls and custom video.js controls appear on iOS devices after this PR. 
If that occurs, debugging because much simpler in which dev tools can be used to trigger breakpoints for determining where in code the additional controls come from. 

This is easier to debug than trying to pinpoint which portion of these changes prevented or removed *all* the controls.

tldr: Testing on iOS hardware just needs to be done with these modifications to see if controls re-appear first, and I do not own an iPhone.

### DIFF
--- a/ui/component/viewers/videoViewer/internal/plugins/videojs-mobile-ui/plugin.js
+++ b/ui/component/viewers/videoViewer/internal/plugins/videojs-mobile-ui/plugin.js
@@ -151,8 +151,7 @@ const onPlayerReady = (player, options) => {
  *           Never shows if the endscreen plugin is present
  */
 const mobileUi = function(options) {
-  // if (videojs.browser.IS_ANDROID || videojs.browser.IS_IOS) {
-  if (videojs.browser.IS_ANDROID) {
+  if (videojs.browser.IS_ANDROID || videojs.browser.IS_IOS) {
     this.ready(() => {
       onPlayerReady(this, videojs.mergeOptions(defaults, options));
     });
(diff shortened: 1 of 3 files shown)


seanyesmunt:
@DispatchCommit Have you looked at this again?

seanyesmunt:
I'm still seeing this
image

seanyesmunt (changes requested):
see screenshot

Issue #5309, opened 2021-01-14
Support Timestamps in URL Query


ElectronEsq:
Typically when you use a [time bookmark](https://odysee.com/@DollarVigilante:b/2021-The-Great-Reset,-Lockdowns,-Genocide-and-Censorship:9?t=1584), it plays the content [automatically](https://youtu.be/McInmQt3eOM?t=6).
It would be nice if the LBRY framework also did this for a smother experience.

Example:

```js
$(document).ready(function () {
    if(window.location.href.contains("?t=") > -1) {
       // autoplay
    }
});

```

DispatchCommit:
may be resolved by: https://github.com/lbryio/lbry-desktop/pull/5372

tzarebczan:
Looks like it is, let us know if you see issues @ElectronEsq ! Thanks for filing btw.

ElectronEsq:
Has it been deployed yet? It's not working for me. I've tried in 3 browsers and incognito with no cache. Tried with Odysee and LBTY TV but you still have to press the play button.

tzarebczan:
Are you already playing the video? I think if it's paused, it won't work. Going to reopen to make sure we take that into account and start automatically. 


Test: https://lbry.tv/@jiggytom:e/test-yt-upload-1:1

ElectronEsq:
> Are you already playing the video? I think if it's paused, it won't work. Going to reopen to make sure we take that into account and start automatically.
> 
> Test: https://lbry.tv/@jiggytom:e/test-yt-upload-1:1

No videos autoplay with a timestamp link for me. You still have to press play.

**Example**:

This should auto play at 50s.
[https://odysee.com/@Electron:e/bonking-parrots:c?t=50](https://odysee.com/@Electron:e/bonking-parrots:c?t=50)

![image](https://user-images.githubusercontent.com/29914179/106091251-baec8f80-6177-11eb-98ab-f454aacc6c43.png)

ElectronEsq:
> Are you already playing the video? I think if it's paused, it won't work. Going to reopen to make sure we take that into account and start automatically.
> 
> Test: https://lbry.tv/@jiggytom:e/test-yt-upload-1:1

No videos autoplay with a timestamp link for me. You still have to press play.

**Example**:

This should auto play at 50s.
[https://odysee.com/@Electron:e/bonking-parrots:c?t=50](https://odysee.com/@Electron:e/bonking-parrots:c?t=50)

[...]

Issue #5328, opened 2021-01-19
videos play in background after floating player closed


tzarebczan:
Have had this happen a couple times, usually when I close the player quickly after it starts playing. The video will continue to play in the backround. If I go back to the video, it's a black window with no video.

DispatchCommit:
finally managed to get this one to reproduce.

```
instrument.ts:129 VIDEOJS: WARN: The element supplied is not included in the DOM
```

Checking for the global `window.player` variable reveals the player instance still exists and was not disposed. Attempting to dispose the player at this point fails too, so it seems the DOM is removed before video.js is able to call dispose.

DispatchCommit:
https://github.com/lbryio/lbry-desktop/blob/8552515e34f08d6af67ba6dd39d0e587b0497397/ui/component/fileRenderInline/view.jsx#L39

> code can currently double fire with videoViewer logic if a video is rendered by FileRenderInline
> (currently this never happens)

https://github.com/lbryio/lbry-desktop/commit/793f622d8d7ae0c333d4e20c0437b22e62014509 - "semi broken"

Should investigate the code around here.

PR #5372, opened 2021-01-23
Fix edge case issue where 2 videos players existed on page causing double audio

DispatchCommit:
This is a refactor of the video.js component's lifecycle hooks.

Existing behavior can cause strange behavior and may cause more than 1 video to play at the same time. This is because when any prop is updated for the VideoViewer component, it causes the entire video.js instance to be recreated. 

Additionally it relies on a callback from an asynchronous function to begin the creation of a new video.js instance which further leads to unstable behavior. For instance, when the player is switching to a new source URL it begin a fetch to get the content's headers. While waiting for this fetch to return, if we close the player, it will be removed from the DOM by a parent component (FileRender). Our fetch will then finish and execute the callback which now creates a new videojs instance on a dom reference that has been removed. The result is a "zombie" video player that does not exist on the page, has no references to it that are accessible from the outer scope, and has now begun playing a video that we should have closed out.

To solve this, I refactored the video.js component so that it only creates a new instance on mount, and only removes and disposes of the player on unmount by scoping the `useEffect` hook to an empty array. 

from: https://reactjs.org/docs/hooks-effect.html
> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the dependencies array always works.

There may still be better ways to approach this, and set values in react, but I'm not an expert in react yet, so open to suggestions and further improvements. However, this behavior should be more predictable and allow for this component to have additional properties set on it without interfering with the video. This also fixes the issues of double audio that I have been able to reproduce.

I dislike that a global `window` variable is needed to ensure we access the same instance of video js in our fetch callback, but it's just bad practice, not an issue that causes issues with the playback experience.

Overall, this should be more reliable when quickly switching between videos, and also reduce some load time slightly as it no longer needs to recreate the entire video player everytime the source changes.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #5328 

duplicated video.js players that cannot be controlled and play without an instance appearing on the page.

## What is the new behavior?

Only 1 video player will ever exist at a time on the page, and will only be recreated if unmounted.

## Other information

There may potentially be bugs related to the change of autoplay from `false` to `true` but I haven't run into any personally yet, and expect autoplay to be more consistent with this value set to `true`.

### DIFF
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -320,7 +320,6 @@ export default React.memo<Props>(function VideoJs(props: Props) {
   // Create the video DOM element and wrapper
   function createVideoPlayerDOM(container) {
     if (!container) {
-      console.error(`VideoPlayer Error: missing ref to container!`);
       return;
     }
 
@@ -339,14 +338,12 @@ export default React.memo<Props>(function VideoJs(props: Props) {
   // Initialize video.js
   function initializeVideoPlayer(el) {
     if (!el) {
-      console.error(`Failed to initialize video player: Missing element to attach to`);
       return;
     }
 
     player = videojs(el, videoJsOptions, () => {
       // this seems like a weird thing to have to check for here
       if (!player) {
-        console.error(`Failed to create videojs player!`);
         return;
       }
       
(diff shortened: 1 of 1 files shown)


DispatchCommit:
With autoplay now set to `true` this may also resolve #5309

seanyesmunt:
This seems to work well. Will give it a more thorough code review once it's out of draft.

@btzr-io I think this refactor will fix the double play issue that popped up with your chromecast integration.

seanyesmunt (approved):

PR #5380, opened 2021-01-25
move channel selection to top of publish form

seanyesmunt:
Maybe a little less confusing this way

### DIFF
--- a/static/app-strings.json
+++ b/static/app-strings.json
@@ -1567,5 +1567,6 @@
   "Delete or edit past content to spend": "Delete or edit past content to spend",
   "Delete supports to spend": "Delete supports to spend",
   "%lbc_amount% boosting content": "%lbc_amount% boosting content",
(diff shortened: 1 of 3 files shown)


static/app-strings.json
```diff
@@ -1567,5 +1567,6 @@
   "Delete or edit past content to spend": "Delete or edit past content to spend",
   "Delete supports to spend": "Delete supports to spend",
   "%lbc_amount% boosting content": "%lbc_amount% boosting content",
+  "Deposit cannot be higher than your available balance: %balance%": "Deposit cannot be higher than your available balance: %balance%",
```
DispatchCommit, line 1570: Why is this included in the PR?

DispatchCommit (approved):
LGTM, it's a nice improvement.

static/app-strings.json
seanyesmunt: Any time the desktop app is run locally, it populates this file with language strings if it's missing any.

This string wasn't added to the file so it was added automatically. (It should have been added when that string was originally added to the UI but I forgot)

PR #5381, opened 2021-01-26
Make dropdown menus actually dropdown

DispatchCommit:
I was annoyed that our dropdown menus didn't actually... dropdown.

odysee-dropdown-animation

So I fixed them:

lbry-dropdown-animation

The CSS change uses a translation instead of a scalar operation to translate the popover menu down 80px rather than expanding by scaling about the center of Y.

Translated 80px because that is the height of the top bar, and for some reason the popover component will affect the scroll position if it renders offscreen. That is, if we were to translate by 100px, it being partially offpage when it is first created will cause the page scroll up by 20px.

I looked into why it is that way, and it's because the popover menu is rendered as `position:absolute` and the position depends on the scroll location. Seems like this really should just be a `position:fixed`, but I don't want to open that Pandora's box potentially.

tldr: better dropdown animation.

### DIFF
--- a/ui/scss/component/_animation.scss
+++ b/ui/scss/component/_animation.scss
@@ -70,7 +70,7 @@
 
 @keyframes menu-animate-in {
   0% {
-    transform: translateY(-80px);
+    transform: translateY(var(--header-height));
     opacity: 0;
   }
   100% {
--- a/ui/scss/component/_animation.scss
+++ b/ui/scss/component/_animation.scss
@@ -70,7 +70,7 @@
 
 @keyframes menu-animate-in {
   0% {
-    transform: translateY(var(--header-height));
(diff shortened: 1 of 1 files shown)


ui/scss/component/_animation.scss
```diff
@@ -70,11 +70,11 @@
 
 @keyframes menu-animate-in {
   0% {
-    transform: scaleY(0);
+    transform: translateY(-80px);
```
seanyesmunt, line 73: This should use `var(--header-height)` (which is 80px)

seanyesmunt (changes requested):

ui/scss/component/_animation.scss
DispatchCommit: changed, thanks.

PR #5382, opened 2021-01-26
Use smooth scrolling

DispatchCommit:
I am not sure where this function is used / called (it might not be used even)

But I saw the comment in the code and this is an easy fix / feature using the `scrollTo` method with `behavior: 'smooth'`

If there are other locations where scrolling is done programmatically, it would be nice to copy paste this over to those areas to get smooth scrolling transitions by the browser (hardware accelerated too!).

`Element.scrollTo()`: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo

You can use this command to see how smooth scrolling is:  
`document.querySelector("html").scrollTo({top: 0, behavior: 'smooth'});`

### DIFF
--- a/ui/page/publish/view.jsx
+++ b/ui/page/publish/view.jsx
@@ -14,7 +14,10 @@ function PublishPage(props: Props) {
   function scrollToTop() {
     const mainContent = document.querySelector('main');
     if (mainContent) {
-      mainContent.scrollTop = 0; // It would be nice to animate this
+      mainContent.scrollTo({
+        top: 0,
+        behavior: 'smooth',
+      });
     }
   }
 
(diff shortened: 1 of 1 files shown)


seanyesmunt:
Nice! It looks like Flow thinks there is an issue with this though.

DispatchCommit:
@seanyesmunt Flow doesn't believe `scrollTo()` is a valid method for some reason.

seanyesmunt:
@DispatchCommit Maybe this will work? https://github.com/facebook/flow/issues/8107#issuecomment-680325729

seanyesmunt:
That didn't work so I just added a supression comment 🥴

PR #5385, opened 2021-01-26
Make zoom level persistent on Desktop

infinite-persistence:
## Change
Zoom on Desktop always starts with 1.0 -- this PR makes the setting persistent, matching the `web` behavior.

## ?
- Do I need to sync `DESKTOP_WINDOW_ZOOM` to `lbry-redux` if this is Desktop-only?


seanyesmunt:
I think that's ok. lbry-redux will most likely be merged into the lbry-desktop repo.

PR #5387, opened 2021-01-26
Zoomable image viewer in Markdown (posts and comments)

### LINKED ISSUE #4899: Ability to expand images in markdown posts for viewing
Images are usually resized to the same width as the paragraph in an MD Post.  In the recent `@lbry` post about the economy, I wanted to view the chart in greater detail.  Messing with the browser's zoom level and changing it back wasn't the best experience.

- It would be nice if clicking the image would invoke a temporary popup to view the image in a larger size, like what Discord does for attachments.
- Github simply opens the image in another tab, which is an ok temporary alternative, imo (not the best, though)

infinite-persistence:
## Issue
Closes #4899: [Ability to expand images in markdown posts for viewing](https://github.com/lbryio/lbry-desktop/issues/4899)

## Questionable
- I think it should be fine, but please double-check on my usage of `window.app.store.dispatch` and `import { doOpenModal } from 'redux/actions/app';` in a jsx file.

## Demo
- output

_p/s: Thanks to `btzr-io` for some pointers._

### DIFF
--- a/ui/component/common/markdown-preview.jsx
+++ b/ui/component/common/markdown-preview.jsx
@@ -12,6 +12,7 @@ import MarkdownLink from 'component/markdownLink';
 import defaultSchema from 'hast-util-sanitize/lib/github.json';
 import { formatedLinks, inlineLinks } from 'util/remark-lbry';
 import { formattedTimestamp, inlineTimestamp } from 'util/remark-timestamp';
+import ZoomableImage from 'component/zoomableImage';
 
 type SimpleTextProps = {
   children?: React.Node,
@@ -33,9 +34,15 @@ type MarkdownProps = {
   isMarkdownPost?: boolean,
 };
 
+// ****************************************************************************
+// ****************************************************************************
+
 const SimpleText = (props: SimpleTextProps) => {
   return <span>{props.children}</span>;
 };
+
+// ****************************************************************************
+// ****************************************************************************
 
 const SimpleLink = (props: SimpleLinkProps) => {
   const { title, children, href } = props;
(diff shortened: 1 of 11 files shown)


ui/component/common/markdown-preview.jsx
```diff
@@ -70,6 +78,25 @@ const SimpleLink = (props: SimpleLinkProps) => {
   return <a title={title}>{children}</a>;
 };
 
+// ****************************************************************************
+// ****************************************************************************
+
+function ZoomableImage({ ...imgProps }) {
+  const onClick = () => {
+    window.app.store.dispatch(
```
seanyesmunt, line 86: Could this just be turned into a regular component that connects to the store?

I would like to avoid using `window.app.store.dispatch`

seanyesmunt (commented):
One comment. Tested and it works great. This is really cool

infinite-persistence:
`Update`
- Move `ZoomableImage` into a regular component

ui/scss/component/_markdown-preview.scss
```diff
@@ -128,6 +128,10 @@
     }
   }
 
+  .img__zoomable {
+    cursor: pointer; // 'zoom-in' would be ideal, but browser-dependant.
```
DispatchCommit, line 132: This could be solved down the road by adding a hover overlay to the image that indicates zoom-in.

https://stackoverflow.com/questions/21086385/how-to-make-in-css-an-overlay-over-an-image

https://stackoverflow.com/questions/21086385/how-to-make-in-css-an-overlay-over-an-image

DispatchCommit (approved):
There may be some unintended behavior when an image is also used as a link: https://lbry.tv/@dispatch:5/markdown-tests:6

When you click the image, the link will open in a new tab, but when you return to the tab, the image is also expanded in the modal window. This could be resolved or improved at a later time though.

But overall, a very nice feature / improvement.

Issue #4027, opened 2020-04-21
Linting the source code is currently broken


jeffslofish:
**Bug**
According to https://github.com/lbryio/lbry-desktop/blob/master/CONTRIBUTING.md#lint we should be able to lint the code by running `$ yarn lint`. However, this currently produces an error because it is trying to lint starting at the top level "src/" directory, but there is no longer a top level "src/" directory. I can fix this issue if you let me know what directories you actually want linted.

**To Reproduce**
1. Run `$ yarn lint`

**Expected behavior**
The code should be linted.

**Screenshots**
![Screen Shot 2020-04-20 at 10 19 58 PM](https://user-images.githubusercontent.com/1240484/79828725-2e927e00-8356-11ea-8e90-e144087d408e.png)

**System Configuration**

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

**Additional context**

tzarebczan:
Thanks for filing this issue! I would say it's all the directories outside of the . ones.

jeffslofish:
Linting is still broken for me, but for a different reason. With eslint in package.json set to ^5.15.2, I get the following errors:

![image](https://user-images.githubusercontent.com/1240484/106059496-a345d600-60a7-11eb-8dfe-75b20b3facc5.png)

If I set the eslint version in package.json to the latest release (^7.18.0), then this error goes away. It appears that the older eslint has problems ignoring the directory if it ends in ".js". 

Can you reopen this issue, and I'll open a pull request to fix it?

seanyesmunt:
@jeffslofish Weird error. Just reopened!

jeffslofish:
Actually, I found a solution to this problem here: https://github.com/eslint/eslint/issues/10880. All you have to do is add single quotes around the glob pattern and then it will work properly. Pull request coming shortly.

PR #5398, opened 2021-01-27
stop linting my web build files please

DispatchCommit:
I never ran the linter locally when PR'ing because I would get some many linting errors I wouldn't even be able to scroll up far enough to see the relevant errors.

This is because the lint command will lint all `.js` and `.jsx` files in the `web/` directory... which includes `web/dist/**`.

This PR just adds `web/dist/**` to the `.eslintignore` file so that build files don't get linted, and brings some sanity to running `yarn lint`.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 node_modules/*
 ./node_modules/**
 **/node_modules/**
+web/dist/**


DispatchCommit:
Since this issue is still open, this could be considered a resolution for #4027

PR #5401, opened 2021-01-27
Update broken 'Retry' button after videojs refactoring.

infinite-persistence:
The button ended up not doing anything.  Just needed an update to match the refactoring.

### DIFF
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -158,16 +158,7 @@ class LbryVolumeBarClass extends videojs.getComponent(VIDEOJS_VOLUME_BAR_CLASS) 
 properties for this component should be kept to ONLY those that if changed should REQUIRE an entirely new videojs element
  */
 export default React.memo<Props>(function VideoJs(props: Props) {
-  const {
-    autoplay,
-    startMuted,
-    source,
-    sourceType,
-    poster,
-    isAudio,
-    onPlayerReady,
-    toggleVideoTheaterMode,
-  } = props;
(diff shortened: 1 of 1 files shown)


DispatchCommit (approved):
This is functional, but highlights an oversight of the `player` variable ending up as undefined. Seems like because of the re-render that is triggered by `reload`, the reference will be lost because `player` is not stateful.

I think to properly fix this, player should become a `useState` / `useRef` so it persists.

For a hotfix, this is fine and will work, but should fix properly.

ui/component/viewers/videoViewer/internal/videojs.jsx
```diff
@@ -447,29 +431,27 @@ export default React.memo<Props>(function VideoJs(props: Props) {
         type: type,
       });
     });
-  }, [source]);
+  }, [source, reload]);
 
   return (
-    reload && (
```
DispatchCommit, line 453: I never understood why this was here... the value of reload is initialized to a string (`initial`) and is updated to a `number`, I don't believe it would ever return falsy.

It's also a `useState` so modifying it's value triggers a re-render as anyways as well. Maybe if it's value was initialized to a falsy value I would understand it (i.e. it saves a few dom elements from being rendered), but in it's current state, I don't get it.

DispatchCommit:
I also opened https://github.com/lbryio/lbry-desktop/pull/5413 to address the concerns I stated above, and that PR is based off this branch.

infinite-persistence:
If 5413 is coming soon, I think we can skip this PR.

infinite-persistence:
`Reject`
Closing this in favour of 5413. We (I) can live with making full browser refresh for now.

PR #5413, opened 2021-01-27
Hotfix: Make VideoJs's internal reference stateful

DispatchCommit:
Discovered while reviewing https://github.com/lbryio/lbry-desktop/pull/5401

VideoJs's reference to itself was not stateful, and was possible for the component to lose it's state and the reference to itself due a rerender.

A hacky way around this was using `window.player` to store the reference globally, but that's bad practice. Instead we should maintain state within the component and persist state across renders.

This seems like a more "react-y" way of doing this.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 node_modules/*
 ./node_modules/**
 **/node_modules/**
+web/dist/**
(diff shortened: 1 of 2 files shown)


infinite-persistence:
It may or may not be relevant, but just wanted to highlight that `MarkdownLink` uses `window.player` as hacky way to gain access to the player.  Might want to include that in your refactoring.

DispatchCommit:
> MarkdownLink uses window.player as hacky way to gain access to the player

Oh boy, glad you mentioned that though as I almost removed that `window.player` statement. I'll trace that down and see if I can refactor that as well. 

Tightly coupling components through global window variables 😬, thanks for the heads up.

seanyesmunt (approved):

PR #5419, opened 2021-01-28
Add missing react lifecycle hooks to fileDrop

DispatchCommit:
(should) Correctly add missing react lifecycles hooks (`useCallback`) for handling file drag and drop.

Removes unnecessary FlowFixMe, adds additional code comments, handles all variable dependencies correctly (memoizing results).

Quick tests reveal behavior and functional remain the same, but now there is **1 less warning** when running `yarn lint` 😃 

Shouldn't functionally change any behavior.


seanyesmunt:
Great cleanup!

Issue #3930, opened 2020-04-01
Long commenter name (channel name) causes whole comment wrapper to be expanded


er013:
Looks like long channel name of comment author makes some unusable comment.

![2020-04-01-192903-import](https://user-images.githubusercontent.com/60119125/78163079-840ef580-7450-11ea-81ed-369eafe232a9.png)
![2020-04-01-193522-import](https://user-images.githubusercontent.com/60119125/78163085-883b1300-7450-11ea-80dc-503c1601cbf2.png)

sources:
lbry://@kauffj#f/014-illegal-walk-to-work-processed#8
lbry://@kauffj#f/walktowork013#9

infinite-persistence:
Similarly, a long comment without any spaces will cause the same problem.  A real-life example could be a long URL pasted into the comments section (I wanted to try it out, but seems like the comment system is dead currently). 
![image](https://user-images.githubusercontent.com/64950861/82920882-738b7f00-9faa-11ea-8c70-0986f315328f.png)
Japanese characters (without spaces) seems to flow, however.

DispatchCommit:
Looks like this should be resolved by #4331

DaniNz:
I'm working on this issue.

DaniNz:
Hi, check if this PR work for you guys: https://github.com/lbryio/lbry-desktop/pull/5431

PR #5431, opened 2021-01-29
"Fix Long commenter name issue"

DaniNz:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number:
https://github.com/lbryio/lbry-desktop/issues/3930

Long commenter name (channel name) causes the whole comment wrapper to be expanded

## What is the new behavior?
Long commenter name (channel name) no longer causes the whole comment wrapper to be expanded.

### DIFF
--- a/ui/scss/component/_button.scss
+++ b/ui/scss/component/_button.scss
@@ -242,6 +242,7 @@
 }
 
 .button__label {
+  max-width: 10rem;
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
--- a/ui/scss/component/_button.scss
+++ b/ui/scss/component/_button.scss
@@ -242,7 +242,6 @@
 }
 
 .button__label {
-  max-width: 10rem;
   white-space: nowrap;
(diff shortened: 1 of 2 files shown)


ui/scss/component/_button.scss
```diff
@@ -242,6 +242,7 @@
 }
 
 .button__label {
+  max-width: 10rem;
```
seanyesmunt, line 245: I would like to avoid adding this to the generic button style. Can you add it here instead?
https://github.com/lbryio/lbry-desktop/blob/master/ui/scss/component/_comments.scss#L195

seanyesmunt (changes requested):

ui/scss/component/_button.scss
DaniNz: Thank you @seanyesmunt , I've changed it, I hope it's ok.

seanyesmunt:
Tested and works great! Thank you!

PR #5455, opened 2021-02-01
videojs: make control bar slightly transparent.

### LINKED ISSUE #5392: better support for non-standard resolution videos
Notice these are cropped on our player:
https://lbry.tv/@DJCmusic:3/DJC---So-Let-There-Be-Light!-(inc.-lyrics):f

https://www.youtube.com/watch?v=ckjuZm_J20o

infinite-persistence:
## Issue

It's not actually cropped -- the control bar was opaque.

## Change
Use 0.80 opacity to somewhat see the full picture.

Using 0.5 or less would be even better, but due to the default tiny size of the control bar, the text becomes unreadable (it's readable if you zoom the browser to get larger text). Don't want to mess with sizing for now, so using 0.80 to get the best of both worlds.

image

### DIFF
--- a/ui/scss/component/_file-render.scss
+++ b/ui/scss/component/_file-render.scss
@@ -387,7 +387,7 @@
   }
 
   .vjs-control-bar {
-    background-color: #000000;
+    background-color: rgba(0, 0, 0, 0.8);
 
     .vjs-remaining-time {
       display: none;


seanyesmunt:
What browser/OS are you on? This is what I see.

Screen Shot 2021-02-01 at 1 53 00 PM

seanyesmunt:
Merging this but leaving that issue open

infinite-persistence:
> What browser/OS are you on? This is what I see.

I used Linux + [Desktop | Brave | Firefox]
The 0.24 image looks the same on YouTube.
Moving to 0:00 or 2:00 to see effect:
image

PR #5487, opened 2021-02-04
Active channel cleanup

seanyesmunt:
Did some refactoring so we can more easily use the `<ChannelSelector />` component

- This will make channel signing (required for new comment apis) much simpler
- Now has one active channel app wide, which isn't tied to comments
- Anonymous actions will happen by enabling the new `incognito` value which can be used to pause search/watch history when that is added back.


ui/component/comment/view.jsx
```diff
@@ -121,11 +110,6 @@ function Comment(props: Props) {
   } catch (e) {}
 
   useEffect(() => {
-    // If author was extracted from the URI, then it must be valid.
-    if (authorUri && author && !isResolvingUri && shouldFetch) {
-      resolveUri(authorUri);
```
seanyesmunt, line 126: This wasn't being used so I took it out.

ui/component/publishName/view.jsx
```diff
@@ -1,51 +1,41 @@
 // @flow
```
seanyesmunt, line 1: This was in the same component as PublishBid. I separated them to move the name field up to the top of the publish form

jessopb (commented):
Seems like the publish confirmation grabbing the channel?
It works on odysee.

ui/component/app/index.js
```diff
@@ -45,6 +58,8 @@ const perform = dispatch => ({
   getWalletSyncPref: () => dispatch(doGetWalletSyncPreference()),
   syncLoop: noInterval => dispatch(doSyncLoop(noInterval)),
   setReferrer: (referrer, doClaim) => dispatch(doUserSetReferrer(referrer, doClaim)),
+  setActiveChannelIfNotSet: () => dispatch(doSetActiveChannel()),
+  setIncognito: () => dispatch(doSetIncognito()),
```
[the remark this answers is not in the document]
seanyesmunt: It won't be labled in the UI as that. I thought it might be more useful as a variable name than `anonymous`.

seanyesmunt:
Fixed the issue with publish confirmation not picking it up. Merging once the build passes

PR #5512, opened 2021-02-11
List of string and site fixes

infinite-persistence:
- _Change upload limit to a variable to better handle site differences._
    `const TV_PUBLISH_SIZE_LIMIT_STR_GB = '2';`
    The above needs to be changed to 4 when merging into Odysee.

- _Wallet string cleanup and more_


seanyesmunt:
Thanks for this. You can set it to 4 on lbry.tv too

Issue #3660, opened 2020-02-09
search filters options partially broken


ramonschepers:
**Bug**
when i tried to use the search option a bit ago, i got to encounter a search bug with file types: it only being applied when i search cor content only, where it does not apply when i search for 'all' with for example only having selected media type "audio ", which still shows other types of content

**To Reproduce**
1) go to search on desktop app
2) search for a term (i used corona in my case)
3) open search filters
4) set searching for to everything (so not just users or content, but both)
5) set filetype to something else than all checkboxes (i tried only filtering by audio)
6) see the behaviour as linked in my screen from discord in screen1

**Expected behavior**
the search should be like the screen2 in screenshots, but the issue relies in the topmost option row it seems

some translation for the screenshots provided in case of needed:
zoeken naar = search for
bestanden / kanalen / alles = files / channels / everything(all)

**Screenshots**
screen1: 
https://cdn.discordapp.com/attachments/363087331475062785/675854179725606962/unknown.png
screen2: 
https://cdn.discordapp.com/attachments/363087331475062785/675854382755086366/unknown.png

**System Configuration**
deamon 0.54.1
app 0.41.0
os version: windows 10 x64

**Additional context**
small tip for a more stremlined search: making one filter section, thus merging the current 2 ones to: 
search for:
users, video, audio, pictures, text, other, downloads, your content.

yours, ramon.

tzarebczan:
Nice find, thank you! Looks like we don't pass them when the everything option is used, so they should be greyed out (when channels option is chosen also).

tzarebczan:
This recently came up as a source of confusion. We should hide or grey out those options on the search page.

Ruk33:
Thanks for the details in the bug report, it really helps! Will be working on it.

Ruk33:
Confirmed, the type parameter (`mediaType`) isn't being sent when using everything.

Ruk33:
There we go https://github.com/lbryio/lbry-desktop/pull/5516. Regards.

tzarebczan:
This is now fixed.

PR #5516, opened 2021-02-12
Fix mediaType filter parameter in search

Ruk33:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/3660

Media type filter isn't sent when Everything is selected.

## What is the new behavior?

Media type filter is sent when Everything is selected.

image

### DIFF
--- a/ui/component/searchOptions/view.jsx
+++ b/ui/component/searchOptions/view.jsx
@@ -14,12 +14,14 @@ type Props = {
 
 const SearchOptions = (props: Props) => {
   const { options, setSearchOption, expanded, toggleSearchExpanded } = props;
+  const stringifiedOptions = JSON.stringify(options);
   const resultCount = options[SEARCH_OPTIONS.RESULT_COUNT];
 
   const isFilteringByChannel = useMemo(() => {
-    const claimType = String(options[SEARCH_OPTIONS.CLAIM_TYPE] || '');
+    const jsonOptions = JSON.parse(stringifiedOptions);
+    const claimType = String(jsonOptions[SEARCH_OPTIONS.CLAIM_TYPE] || '');
     return claimType.includes(SEARCH_OPTIONS.INCLUDE_CHANNELS);
-  }, [options]);
+  }, [stringifiedOptions]);
 
   return (
     <div>
--- a/ui/component/searchOptions/view.jsx
+++ b/ui/component/searchOptions/view.jsx
@@ -1,7 +1,7 @@
 // @flow
 import { SEARCH_OPTIONS } from 'constants/search';
 import * as ICONS from 'constants/icons';
-import React from 'react';
+import React, { useMemo } from 'react';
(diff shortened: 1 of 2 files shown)


seanyesmunt (changes requested):
This isn't quite what is needed. Due to lighthouse limitations, we can't pass the `mediaType` param when "everything" is selected, because it will never return channels.

Instead, when "everything" is selected, clear out all of the "type" filters, and disable that row.

It's not the best, but it should make what's happening clearer

Ruk33:
Got it @seanyesmunt , thanks for the clarification. I've updated the PR. Regards.

ui/component/searchOptions/view.jsx
```diff
@@ -16,6 +16,11 @@ const SearchOptions = (props: Props) => {
   const { options, setSearchOption, expanded, toggleSearchExpanded } = props;
   const resultCount = options[SEARCH_OPTIONS.RESULT_COUNT];
 
+  const isFilteringByChannel = useMemo(() => {
+    const claimType = String(options[SEARCH_OPTIONS.CLAIM_TYPE] || '');
+    return claimType.includes(SEARCH_OPTIONS.INCLUDE_CHANNELS);
+  }, [options]);
```
seanyesmunt, line 22: To make sure this works properly, you need to stringify the object that you pass into a react hook

I usually do something like
```
const strinigifiedOptions = JSON.stringify(options)

React.useEffect(() => {
  const jsonOptions = JSON.parse(stringifiedOptions);
  ...
  }, [stringifiedOptions])
```

seanyesmunt (changes requested):
One more comment then it looks good to go

ui/component/searchOptions/view.jsx
Ruk33: I believe this is hidding a bug. We shouldn't need to `stringify` the object unless it's being mutated. Will update though but, I think this is a bug.

seanyesmunt: It's just how react hooks work. Objects are special. Even if you don't mutate the object it still needs to be stringified.

Ruk33: According to the documentation:

https://reactjs.org/docs/hooks-reference.html#usememo

> Pass a “create” function and an array of dependencies. useMemo will only recompute the memoized value when one of the dependencies has changed. This optimization helps to avoid expensive calculations on every render.

Since props are immutable, it means it will only check if it's filtering by channel when `props.options` change. This is why I was asking if `options` is immutable, because if it is, there is no need for `stringify`. If it isn't, we may need to dig deeper since it may cause issues.

Ruk33:
Ok @seanyesmunt I just sent the change. Thanks for the review. Please let me know if this is ok or if it needs anything else. Regards.

PR #5535, opened 2021-02-17
Video fixes

### LINKED ISSUE #5450: Video showing previous audio clip's thumbnail instead
1. Open https://odysee.com/@Musica_Clasica:a2/Dvorak--Sinfonia-n-8--Op-88--mp3:3. I used Brave Browser
(this issue is a little bit odd. Sometimes, it's not reproducible on Desktop, but reproducible on Desktop if I pick another pair of claims, and vice-versa for Web)
2. In the `Elements` dev tools, confirm that `<audio` is there.  If not, press Back to invoke the Floating Player, then Close (this forces a `dispose`), then return to this music file.
3. In the **wunderbar**, paste https://odysee.com/@veritasium:f/why-are-there-huge-colorful-pools-in-the:b  and press Enter.  Start the video without closing the Floating Player.
--> What's on screen is not the video but the previous thumbnail. The audio of the video is audible, though.

Notice that `<audio ...>` did not switch to `<video ...>`.  When changing to a small number of claims, sometimes `dispose` isn't called (probably React did not think the component needed to unmount).

Aside: I noticed in Discord that the playback issues have been attributed to certain routers blocking the CDN.  I don't think that's the case for me, as when I force it to dispose on every claim-change in my local build, the playback error does not happen (I think it's due to syncing issues between the async `fetch` for m3u8 header and videojs's caching of sources (suspiciously related error message -- github.com/videojs /video.js/issues/3751).  
So, along with this `<audio` problem, I wonder if we should just always dispose when the source changes.

image

### LINKED ISSUE #5513: Video-setting persistence broken
Playback Rate and Volume (and sometimes Theather Mode) setting is no longer persistent when switching to another video.

1. **Playback Rate**
    - Looking at the log, the saved value was correctly reloaded in `onPlayerReady`, but there is a random `ratechange` event (source unknown) that kept setting it back to 1x.
    - Issue is not present when reverted to `912489cc` (pre vjs refactoring)
    - _2021-02-26:_ Sometimes, the value looks to have persisted, but the actual speed was not.
2. **Volume**
    - This one's a little harder to reproduce, but the steps are:  
        - Lower the volume of a video to barely audible, but not muted.
        - Press _Back_, and **close** the player
        - Click on a new (uncached) video. Although the volume icon looks correct (not Max), the volume slider is at Max. Actual volume is also at Max.  Something is out of sync here.

Found at:  live site and latest `master` Desktop.

infinite-persistence:
## Issue
Closes #5450[: Video showing previous audio clip's thumbnail instead](https://github.com/lbryio/lbry-desktop/issues/5450)

## Notes
`createVideoPlayerDOM` depends on `isAudio`

## Issue
Closes #5513[: Video-setting persistence broken](https://github.com/lbryio/lbry-desktop/issues/5513)

## Notes
- Per videojs recommendation, the setting-restoration should be done after the video has been loaded, so the action was moved to `loadedmetadata`. This fixed the volume slider problem, and should have fixed the playbackRate too.
- For playbackRate, there is another special case where it gets reset to 1 (refer to comments in code).
- One caveat on moving to `loadedmetadata` is that the rate will be incorrect ("1") while the video is initially loading.

### DIFF
--- a/ui/component/viewers/videoViewer/view.jsx
+++ b/ui/component/viewers/videoViewer/view.jsx
@@ -19,8 +19,8 @@ const PLAY_TIMEOUT_LIMIT = 2000;
 
 type Props = {
   position: number,
-  changeVolume: number => void,
-  changeMute: boolean => void,
+  changeVolume: (number) => void,
+  changeMute: (boolean) => void,
   source: string,
   contentType: string,
   thumbnail: string,
@@ -36,9 +36,9 @@ type Props = {
   doAnalyticsBuffer: (string, any) => void,
   claimRewards: () => void,
   savePosition: (string, number) => void,
-  clearPosition: string => void,
+  clearPosition: (string) => void,
   toggleVideoTheaterMode: () => void,
-  setVideoPlaybackRate: number => void,
+  setVideoPlaybackRate: (number) => void,
 };
 
 /*
@@ -94,7 +94,7 @@ function VideoViewer(props: Props) {
   }, [uri, previousUri]);
 
   function doTrackingBuffered(e: Event, data: any) {
-    fetch(source, { method: 'HEAD' }).then(response => {
+    fetch(source, { method: 'HEAD' }).then((response) => {
       data.playerPoweredBy = response.headers.get('x-powered-by');
       doAnalyticsBuffer(uri, data);
     });
@@ -150,12 +150,14 @@ function VideoViewer(props: Props) {
       // https://blog.videojs.com/autoplay-best-practices-with-video-js/#Programmatic-Autoplay-and-Success-Failure-Detection
       if (shouldPlay) {
         const playPromise = player.play();
-        const timeoutPromise = new Promise((resolve, reject) => setTimeout(() => reject(PLAY_TIMEOUT_ERROR), PLAY_TIMEOUT_LIMIT));
+        const timeoutPromise = new Promise((resolve, reject) =>
+          setTimeout(() => reject(PLAY_TIMEOUT_ERROR), PLAY_TIMEOUT_LIMIT)
+        );
 
-        Promise.race([playPromise, timeoutPromise]).catch(error => {
+        Promise.race([playPromise, timeoutPromise]).catch((error) => {
           if (PLAY_TIMEOUT_ERROR) {
             const retryPlayPromise = player.play();
-            Promise.race([retryPlayPromise, timeoutPromise]).catch(error => {
+            Promise.race([retryPlayPromise, timeoutPromise]).catch((error) => {
               setIsLoading(false);
               setIsPlaying(false);
             });
@@ -167,6 +169,7 @@ function VideoViewer(props: Props) {
       }
 
       setIsLoading(shouldPlay); // if we are here outside of an embed, we're playing
+
       player.on('tracking:buffered', doTrackingBuffered);
       player.on('tracking:firstplay', doTrackingFirstPlay);
       player.on('ended', onEnded);
@@ -175,9 +178,8 @@ function VideoViewer(props: Props) {
         setIsPlaying(false);
         handlePosition(player);
       });
-      player.on('error', function() {
+      player.on('error', () => {
         const error = player.error();
-
         if (error) {
           analytics.sentryError('Video.js error', error);
         }
--- a/ui/component/viewers/videoViewer/view.jsx
+++ b/ui/component/viewers/videoViewer/view.jsx
@@ -98,11 +98,9 @@ function VideoViewer(props: Props) {
   useEffect(() => {
     vjsCallbackDataRef.current = {
       embedded: embedded,
-      muted: muted,
-      volume: volume,
       videoPlaybackRate: videoPlaybackRate,
     };
-  }, [embedded, muted, volume, videoPlaybackRate]);
+  }, [embedded, videoPlaybackRate]);
 
   function doTrackingBuffered(e: Event, data: any) {
     fetch(source, { method: 'HEAD' }).then((response) => {
@@ -148,10 +146,8 @@ function VideoViewer(props: Props) {
     }
   }
 
-  function restoreSavedSettings(player) {
+  function restorePlaybackRate(player) {
     if (!vjsCallbackDataRef.current.embedded) {
-      player.muted(vjsCallbackDataRef.current.muted);
-      player.volume(vjsCallbackDataRef.current.volume);
       player.playbackRate(vjsCallbackDataRef.current.videoPlaybackRate);
     }
   }
@@ -159,6 +155,9 @@ function VideoViewer(props: Props) {
   const onPlayerReady = useCallback(
     (player: Player) => {
       if (!embedded) {
+        player.muted(muted);
+        player.volume(volume);
+        player.playbackRate(videoPlaybackRate);
         addTheaterModeButton(player, toggleVideoTheaterMode);
       }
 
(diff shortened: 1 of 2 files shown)


seanyesmunt:
Thanks for this @infinite-persistence. I assigned @DispatchCommit since he has been working on the video player recently

DispatchCommit:
> Per videojs recommendation, the setting-restoration should be done after the video has been loaded
> action was moved to `loadedMetadata`

Do you have a reference for this? I'm curious as to why this would be the case. In my experience restoring state in `ready` worked best and solved issues like videos autoplaying at max volume for a brief moment before state is restored.

It may be even better to restore state both in `ready` *and* `loadedmetadata` though.

DispatchCommit:
Hmmm I trying to follow the event flow and I'm noticing some strange behavior with the `onPlayerReady` callback appearing to fire *too early*.

Changing:

```js
onPlayerReady(player);
```

to

```js
player.ready(() => onPlayerReady(player));
```

Seems to fix the volume bar also.. It shouldn't make a difference though if I am understanding https://docs.videojs.com/tutorial-setup.html#player-readiness correctly. I'll need to make a note of this and circle back to testing this.

My only reason in exploring this is that I don't think it's necessary to use a separate event handler for restoring volume and mute and I think it should be left in `onReady` as that is how setting the volume is shown in the documentation:
https://docs.videojs.com/tutorial-player-workflows.html#changing-the-volume-of-a-player

```js
var myPlayer = videojs('some-player-id');

myPlayer.src({type: 'video/mp4', src: 'http://www.example.com/path/to/video.mp4'});
myPlayer.ready(function() {
  // get
  var howLoudIsIt = myPlayer.volume();
  // set
  myPlayer.volume(0.5); // Set volume to half
});
```

ui/component/viewers/videoViewer/view.jsx
```diff
@@ -167,6 +185,8 @@ function VideoViewer(props: Props) {
       }
 
       setIsLoading(shouldPlay); // if we are here outside of an embed, we're playing
+
+      player.on('loadedmetadata', () => restoreSavedSettings(player));
```
DispatchCommit, line 189: In theory, using `ready` should work fine here (or `player.ready(() => restoreSavedSettings(player))`).

I'd prefer to set these values in `ready` if possible. I'm not sure if playback rate can be set in ready though, and may actually require this event handler regardless.

DispatchCommit (changes requested):
LGTM except the notes I mentioned above. If it's not possible and the issue still occurs, then we can sideline it for now and I'll come back to this another time and try to move it back to ready.

infinite-persistence:
> _Do you have a reference for this? I'm curious as to why this would be the case. In my experience restoring state in ready worked best and solved issues like videos autoplaying at max volume for a brief moment before state is restored.
It may be even better to restore state both in ready and loadedmetadata though._

I was referencing this: [github.com/videojs/ video.js/issues/5128] along with my local flow logs.
_Note: I broke up the URL with a space just so that github doesn't link this issue to their repo (reduce "mentioned" noise)_

I tried moving the restoration to after `player.play()` (i.e. at the bottom of `onPlayReady`, or equivalently, at `player.on('ready', ...)`, but it still got overridden. `loadedmetadata` is the earliest event I could find that happen after the reset.

Also tried the idea of leaving the original restoration there, _plus_ adding a redundant one at `loadedmetadata`. Because the connection to LBRY server in my region is slow (i.e. loading screen is longer), I was still able to see the reset, so I end up removing the original restoration.

ui/component/viewers/videoViewer/view.jsx
infinite-persistence: > _Do you have a reference for this? I'm curious as to why this would be the case. In my experience restoring state in ready worked best and solved issues like videos autoplaying at max volume for a brief moment before state is restored.
It may be even better to restore state both in ready and loadedmetadata though._

I was referencing this: [github.com/videojs/ video.js/issues/5128] along with my local flow logs.
_Note: I broke up the URL with a space just so that github doesn't link this issue to their repo (reduce "mentioned" noise)_

I tried moving the restoration to after `player.play()` (i.e. at the bottom of `onPlayReady`, or equivalently, at `player.on('ready', ...)`, but it still got overridden. `loadedmetadata` is the earliest event I could find that happen after the reset.

Also tried the idea of leaving the original restoration there, _plus_ adding a redundant one at `loadedmetadata`. Because the connection to LBRY server in my region is slow (i.e. loading screen is longer), I was still able to see the reset, so I end up removing the original restoration.

infinite-persistence: > If you're setting the rate before a source is loaded, use this.player.defaultPlaybackRate(savedRate)

Also tried this given that we update `player.src` in another effect, but didn't seem to work.

infinite-persistence:
I tried the suggestions, but I'm pretty sure now that another callback (`loadedmetadata`, or some other) is needed it because:
1. Per videojs link above, changing `src` has a tendency to reset the playback
2. `ready` won't be called in the case of `src` being changed without a `dispose`. (no issues when React decides to dispose)
3. The async header `fetch` can delay things, as shown below:
```
--- init -------------
fetch ...
fetch -ok
player.src()
player.hlsQualitySelector()
player.src()
ready
onPlayerReady
Restore:
  muted: false
  volume: 0.24390243902439024
  videoPlaybackRate: 1.1
  ratechange: 1.1 (skipped)
<<Video changed here>>
dispose
--- init -------------
fetch ...
ready
onPlayerReady
Restore:
   muted: false
   volume: 0.24390243902439024
   videoPlaybackRate: 1.1
ratechange: 1.1
fetch -ok        <<< too late >>>
player.src()
player.hlsQualitySelector()
player.src()
ratechange: 1    <<< oh no >>>
```



`Update`
- Rebased to latest master

seanyesmunt:
@DispatchCommit Can you take a look at the changes made?

infinite-persistence:
`Update`
- Corrected branch mixed with 5570.

DispatchCommit (approved):

ui/component/viewers/videoViewer/view.jsx
```diff
@@ -189,7 +213,12 @@ function VideoViewer(props: Props) {
...
       player.on('ratechange', () => {
-        if (player) {
+        const HAVE_NOTHING = 0; // https://docs.videojs.com/player#readyState
+        if (player && player.readyState() !== HAVE_NOTHING) {
+          // The playbackRate occasionally resets to 1, typically when loading a fresh video or when 'src' changes.
+          // Videojs says it's a browser quirk (https://github.com/videojs/video.js/issues/2516).
+          // [x] Don't update 'videoPlaybackRate' in this scenario.
+          // [ ] Ideally, the controlBar should be hidden to prevent users from changing the rate while loading.
           setVideoPlaybackRate(player.playbackRate());
         }
```
DispatchCommit, line 223: Can we change the scope of this PR to just address the improvement to the playback rate value?

I believe the issue related the the volume slider is caused elsewhere, and it'd be easier to isolate any regression if the PR is smaller. 

I have some refactoring done in the other lines modified that causes some conflicts as well.

DispatchCommit (changes requested):
On second pass, I'm uncertain as to whether this is the correct approach or whether it's going to complicate debugging event related issues. 

From what I've been investigating and working on, I think this might just cover up some underlying issues that get introduced before the event hooks you use that correct the problems. 

e.g. onReady sets values incorrectly, but you correct them in loadedMetadata. Best thing would be to fix earlier on in the lifecycle, but that's a can of worms currently.

I like the rate changes though.

infinite-persistence:
## Exclude Volume/Mute restoration
Okay, I understand.

## Restoring Playback-Rate
The behavior in `master` has changed a bit since Feb 26 (could be earlier). Now, the GUI reflects the restored value (say "x0.25"), but the actual video speed is "1", causing a mismatch with the GUI.  I no longer see the "1" change in `ratechange`.

I'm not sure if the comment/recommendation on not using `loadedmetadata` applies to playback-rate as well or not.  If it does, feel free to close this PR as this solution doesn't work without it.  The alternative would be to not make the setting persistence, hence it will match the reset that's happening from the `src` change.

Issue #5456, opened 2021-02-01
Fix whatever is broken with the video quality selector for hls streams


tzarebczan:
Sean said Xander thinks it's a race condition.

PR #5539, opened 2021-02-18
Add custom quality selector plugin

DispatchCommit:
Restores the quality selector to the site.

Uses a modified version of videojs-hls-quality-selector that detects source changes on the player and selectively hides or shows the quality selector dependent on if the file is an MP4 or M3U8 (HLS) file.

This is done by modifying the constructor for the plugin so that it is always active, and reactively hides or shows the control bar button based on data received from the loadedmetadata event which allows up to detect if a video is transcoded and has quality options or not.

This prevents us from having to re-create the entire video.js player when switching videos, and thus improves page speed slightly, and prevents creating extra garbage from repeatedly disposing the player.


DispatchCommit:
resolves #5456

DispatchCommit:
plugin forked from: https://github.com/chrisboustead/videojs-hls-quality-selector

seanyesmunt:
Closing in favor of https://github.com/lbryio/lbry-desktop/pull/5548

infinite-persistence:
Hi guys,  
Some minor (probably useless) observations:
1. When switching to another video _and_ if when `dispose` isn't invoke, there is an error message:
image
Not sure if it causes anything bad though, although I still don't see the button on the eevblog channel (used to see it there)
2. While the video is loading/buffering (mine takes a long time due to my region), the `AUTO` button is visible.  Not a big deal though, since the list, when clicked, is empty anyway.
image

Issue #4932, opened 2020-10-22
Add button to pop out via floating player


tzarebczan:
We removed the browser based pop-out button cause it would cause confusion between our pop out and the browser one, but now there's no way to force a pop out unless you navigate away. 

Other suggestion would be to automatically pop out when scrolling down in comments (in both comment replies and original)

DispatchCommit:
This could be done with an [intersection observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

To detect when >X% of the player has passed the top of the viewport, then pop out 
(and vice versa, pop *in* when scrolling back up far enough)

Ruk33:
Hello guys, I'll be working on this one.

Ruk33:
Hello guys, is this ok?

https://user-images.githubusercontent.com/1719111/107385710-e6556f80-6ad1-11eb-8a50-8e875a044c40.mp4

Regards.

tzarebczan:
That looks great! I wonder if we should force the pop-out player to the right side so it doesn't take over the comment area? @seanyesmunt thoughts?

Ruk33:
I have pushed a WIP branch in the meantime https://github.com/Ruk33/lbry-desktop/tree/issue-4932-floating-player-scroll. Not sure about a few implementation details though...

Ruk33:
Hello guys, just following this thread up, any other changes required or is this good? @tzarebczan @seanyesmunt . Regards!

PR #5544, opened 2021-02-18
WIP Floating player on scroll

kauffj:
see #4932 for context, @Ruk33 is looking for feedback on this

### DIFF
--- a/ui/page/file/view.jsx
+++ b/ui/page/file/view.jsx
@@ -17,9 +17,9 @@ type Props = {
   costInfo: ?{ includesData: boolean, cost: number },
   fileInfo: FileListItem,
   uri: string,
-  fetchFileInfo: string => void,
-  fetchCostInfo: string => void,
-  setViewed: string => void,
+  fetchFileInfo: (string) => void,
+  fetchCostInfo: (string) => void,
+  setViewed: (string) => void,
   renderMode: string,
   obscureNsfw: boolean,
   isMature: boolean,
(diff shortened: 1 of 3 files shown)


DispatchCommit (changes requested):
Listening to the scroll event can lead to performance issues.

At the minimum, the event listener should be a passive event listener:
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

But the *best* solution would be to use an intersection observer instead:
https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

Intersection observers is what bitwave uses to popout it's player on scroll:

https://github.com/bitwave-tv/bitwave/blob/40dc640fde12cf26aa320006ec259d4e66aeb6dc/pages/_watch/index.vue#L21-L30

https://github.com/bitwave-tv/bitwave/blob/40dc640fde12cf26aa320006ec259d4e66aeb6dc/pages/_watch/index.vue#L376-L378

ui/page/file/view.jsx
```diff
@@ -44,10 +46,32 @@ function FilePage(props: Props) {
...
+      doSetPlayingUri(uri, 'markdown');
+    }
+  }, [uri]);
+
+  React.useEffect(() => {
+    if (!floatingPlayerEnabled) {
+      return;
+    }
+
+    window.addEventListener('scroll', onScroll);
```
DispatchCommit, line 68: https://developers.google.com/web/updates/2016/06/passive-event-listeners

kauffj:
@Ruk33 do you want to come back to this?

Ruk33:
Hello @kauffj sure, I was finishing the release date input now I'll continue with this one. Thanks for following up!

seanyesmunt:
@Ruk33 I'm going to close this one while I am cleaning up the repo a bit. This would be a great feature to have though! Feel free to re-open this PR when you start working on it again.

PR #5556, opened 2021-02-19
decode uri before test /$/

jessopb:


jessopb:
tested on bob by replacing iframe url on page, seems to work

PR #5570, opened 2021-02-23
Restore video loading circle

### LINKED ISSUE #5554: Video: loading circle sometimes does not appear until 2nd click
This happens on Desktop, Live Site and local builds.

1. Enable "Autoplay" in settings.
2. Clicked on a video.
    - The expected behavior is that the video container would start the loading circle and eventually play the video.
3. Occasionally, the loading circle doesn't appear until either `(a)` I clicked on the large play button again `(b)` the player has gotten the data to play the video.
    - It's probably not noticeable if your region's server is fast, but for my slower region (i.e. long buffering), it looks odd that I need to click Play twice.

In the example below:
https://www.youtube.com/watch?v=Esr2-nBxFls
- I waited until 0:30 to to perform `(a)`.
- I then loaded another video, and patiently waited until 1:26 to get `(b)`, although it ended up buffering again (typical situation for me).

infinite-persistence:
## Issue
Closes #5554[: Video: loading circle sometimes does not appear until 2nd click](https://github.com/lbryio/lbry-desktop/issues/5554)

## What's happening
videojs behavior:
(a) A `src` change makes the Play button re-appear.
(b) An `onPlay` (or `play()`) makes the button go away.

Due to the `m3u8` header async fetch (i.e. return is potentially delayed), the initial `onPlay` (which cleared the button) that happened after user clicked Play gets negated by a potentially-delayed `src` change.  

image

## Approach
I believe in our use case, a change in `src` always means "change src and resume play". A `src` change always comes from a Play action, and we don't pause (don't need to anyway) the player during the change.

So, hiding the button after `src` is changed seems reasonable.

## New behavior
We now see the spinning loading circle when video is loading, as it was originally.

## Aside
Is there a non-async alternative to the header fetching?  Seems to be the root of many problems.  When the fetch is fast, we avoid the double `onPlay`
image

### DIFF
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -453,6 +453,11 @@ export default React.memo<Props>(function VideoJs(props: Props) {
         src: source,
         type: type,
       });
+
+      // PR #5570: Temp workaround to avoid double Play button until the next re-architecture.
+      if (!player.paused()) {
+        player.bigPlayButton.hide();
+      }
     });
   }, [source, reload]);
 


DispatchCommit:
I've been debating back and forth on how to handle and fix issues related to that async fetch call and haven't decided if it makes more sense to take that out of the internals and simply do the fetch *before* updating the `src` for the player or perform the check in the player code.

I think it makes more sense to separate the concerns and responsibilities and only update `src` after the fetch has completed in a parent component - then just pass `src` and `type` to the player directly so it only ever has to worry about standard video.js setup and event flows.

DispatchCommit (changes requested):
PR resolves issue with play button appearing while async fetch is waiting to return - however due to the change hiding the play button `onPlayerReady`, if the player is not allowed to autoplay, the play button will *not* appear.

Example: notice the play button appears on the left, but not the right. Neither player is in the `play` state currently.

image

Left is production, right is this PR. I have disabled all autoplaying on this browser, and as a result, the play button will not appear -- yet the player is currently paused (stopped).

I think this behavior further reinforces the idea that the fetch should be done prior as there are many edge-case issues with async fetch'ing internally.

I'm not sure it makes sense to address this issue within the component this way. You could workaround this by hidding the play button until after the fetch, then checking after the fetch if the player has started to show or hide the button, but that also circles back to the need to fix how the player autoplays.

DispatchCommit:
Ideally we would want to have this loading indicator show until the pre-fetch is completed I think...

image

infinite-persistence:
> _Ideally we would want to have this loading indicator show until the pre-fetch is completed I think..._

Yup, that is what the PR is trying to achieve -- it should show the loading screen above, followed by buffering loading circle, then play -- except that I missed out on the browser-level autoplay=off scenario,

I agree that the async fetch is the actual root of many issues.  If a fix is coming up soon, we don't need this PR;  else, I think having this workaround is better than today's version (the change is simple to revert).  Having to guess whether I did click play or not is getting annoying :)

----

`Update`
- Handle the case of the player being paused externally.

infinite-persistence:
`Update`
- Squashed commits and added comments in the code to ease the revert process during the upcoming re-architecture.

DispatchCommit:
Can #5535 be closed in favor of this since this PR contains the changes from that PR as well?

infinite-persistence:
> _Can #5535 be closed in favor of this since this PR contains the changes from that PR as well?_

@DispatchCommit, sorry, I ended up force-pushing the wrong branch and made it identical to 5535. Both PRs should be addressing different things.

`Update`
- Corrected the branch for this PR.

DispatchCommit (approved):
"better than nothing"

will properly address and fix in a future refactor of the pre-fetch

PR #5573, opened 2021-02-24
Lowered 'Channel Description' limit from 5000 to 3000 chars.

### LINKED ISSUE #5571: Limit description length on channel edit
Similar to the publish page, we should limit the description length. A too long of a length will make the creation/update fail with: 
```
 Desktop Error: "(1, 'the transaction was rejected by network rules.\\n\\nbad-txns-claimscriptname-toolarge (code 16)\\n[0100000001527f8ed1526102f582a5ac2185db165a2373870bea99d464bb15fd7620568e1a000000006b483045022100eb0
```

infinite-persistence:
## Issue
Closes #5571[: Limit description length on channel edit](https://github.com/lbryio/lbry-desktop/issues/5571)

## Notes
Similar to the previous value of '5000', this is a trial-and-error estimate.


tzarebczan:
If someone had more previously, does it still let them save it on edit. Probably should take that into account. Or maybe we can capture the error and only tell them to limit then.

Issue #5366, opened 2021-01-23
Clear values in the upload form after submission


DispatchCommit:
After uploading something to LBRY, the form does not get cleared when going to upload another file.

Fields such as title, description, thumbnail,  name, and deposit amount should be cleared between uploads.

I can see value in allowing users to set default values for new uploads (similar to youtube), but it's odd to return to the upload page and have it still show the same data from the last time I uploaded something.

DispatchCommit:
Also if you go to edit a claim, choose back, and then click upload, you will return to the "editing claim" page. 

You must click `cancel` to actually clear the form.

It makes sense to save edits in case someone accidentally closes the tab / navigates away and wishes to return, but clicking submit should also clear the form. 

The edit form and upload form arguably shouldn't share cache data also IMO, even if they do internally reuse the same components.

DispatchCommit:
Attempting to edit a claim after recently uploading content can also cause unexpected behavior such as this:

![image](https://user-images.githubusercontent.com/45262335/105577270-4788ef00-5d2d-11eb-9826-95b14679c956.png)


Where I am attempting to edit one claim (titled `XSS Test`) after having just uploaded a claim (titled: `Markdown XSS Tests`)

Closing the dialog in this case will then clear the form, and act like the normal upload page.

tzarebczan:
Originally filed under https://github.com/lbryio/lbry-desktop/issues/5139 , but we can keep this issue. The last comment sounds like a different issue with not updating the title from the edit.

Ruk33:
Hello guys, I can't reproduce the issue after submitting a claim.

When submitting, closing the modal clears the form correctly. I have tried:
- clicking on X
- clicking on close
- clicking on view uploads
- even clicking outside of the modal

I was able to reproduce the edit issue however. When the user edits, then tries to upload a claim you can see the previous state. Will be working on that.

Ruk33:
Ok, pushed a PR https://github.com/lbryio/lbry-desktop/pull/5578. Please let me know if this is ok guys. Thanks!
Regards.

tzarebczan:
Jessop also took care of this recently, shouldn't happen anymore!

PR #5578, opened 2021-02-24
Clear publish form when publishing from header

Ruk33:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/5366

- User edits a claim
- Doesn't finish and leaves the section
- Clicks on Publish
- The form isn't clean and remains with the previous edited claim 

## What is the new behavior?

The publish form now gets cleared when creating a new claim.

## Other information

NOTE: This PR depends on https://github.com/lbryio/lbry-redux/pull/385

### DIFF
--- a/ui/component/header/index.js
+++ b/ui/component/header/index.js
@@ -4,13 +4,14 @@ import { selectTotalBalance, selectBalance, formatCredits, SETTINGS } from 'lbry
 import { selectGetSyncErrorMessage } from 'redux/selectors/sync';
 import { selectUserVerifiedEmail, selectUserEmail, selectEmailToVerify, selectUser } from 'redux/selectors/user';
 import { doClearEmailEntry, doClearPasswordEntry } from 'redux/actions/user';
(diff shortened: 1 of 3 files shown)


seanyesmunt:
@Ruk33 I don't think this is the correct fix. We still want to allow starting a publish, then coming back to it later to finish it.

Ruk33:
Hello @seanyesmunt , thanks for the reply. Could you elaborate on the requirements? I'm facing a lot of possible scenarios that could be solved differently and I don't know how far we want to go. Thanks!

seanyesmunt:
@Ruk33 The first step would be to reproduce the bug from the original issue (maybe it is already resolved). When you have a successful publish, all of the publish fields should be cleared out.

seanyesmunt:
@Ruk33 I'm going to close this since it isn't quite what is needed. Feel free to re-open if you figure out what is going wrong with that bug.

PR #5579, opened 2021-02-25
Tell user to reduce 'Channel Description' when exceeded tx-limit.

infinite-persistence:
## Issue
Second attempt at #5571: `Limit description length on channel edit`

## Changes
- Undo #5573: restore limit back to 5000.
- Translate the `bad-txns-claimscriptsize-toolarge (code 16)` error into `Transaction limit reached. Try reducing the Description length.`

image

## Test method
- Temporarily changed `FF_MAX_CHARS_IN_DESCRIPTION` to 50k, then enter a 10k message,


seanyesmunt:
Should the description amount just be lowered more so this error won't happen?

infinite-persistence:
I vote for leaving at 5000 due to:
- matches a YT-sync'd video limit (prevent future editing issues?)
- couldn't actually replicate the limit in a real scenario (i.e. is it a super rare case?).  Seems like profile photos are uploaded separately (not part of tx?) and I also tried super long tags with the 5000-char scenario.

seanyesmunt:
That makes sense. I'll check this out more today

PR #5580, opened 2021-02-25
Restore "view" button when downloaded file is deleted

### LINKED ISSUE #4959: Deleting MD from downloads list causes spinning icon to run forever
1. lbry-desktop
2. Open https://lbry.tv/@lbry:3f/lbryeconomy2020:5
3. Click "Delete this from my computer"
4. **"Preparing your content" now spins forever.**

In the old version, the MD simply becomes the "preview" box with an eye icon and the thumbnail as the background.  That behavior made more sense.

### LINKED ISSUE #5077: player doesn't close when video deleted
this is really just a notice as we're on revision 10 already and I'm still having the issue after updating twice today
so I'm not certain it's known

but yeah, after deleting the video, the player continues running:
Screenshot_2020-11-20_08-33-15
and even plays again on re-download (experienced when downloading hangs and is deleted to restart)

and then when returning to the new posts, the player stays open there too:
Screenshot_2020-11-20_08-33-56

nothing is logged in the terminal when this happens.

infinite-persistence:
## Issue
Closes #4959: `Deleting MD from downloads list causes spinning icon to run forever`
Closes #5077: `player doesn't close when video deleted`

## New behavior
The "view" eye button now appears in this scenario, just like it was back in Oct 2020.

## Changes
The following is my understanding (please correct me if I'm wrong; been wondering about it for a while):
- `primaryUri` - the URI of the file page
- `playingUri` - the URI for the media that's currently playing. It's usually the same as `primaryUri`, but could be something else for the case of Floating Player.

The components were looking at `primary`, plus comparisons with `playing` weren't done correctly (it's an object, not string).

### DIFF
--- a/ui/component/fileRenderInline/index.js
+++ b/ui/component/fileRenderInline/index.js
@@ -1,14 +1,14 @@
 import { connect } from 'react-redux';
 import { makeSelectFileInfoForUri, makeSelectStreamingUrlForUri } from 'lbry-redux';
 import { doClaimEligiblePurchaseRewards } from 'redux/actions/rewards';
-import { makeSelectFileRenderModeForUri, selectPrimaryUri } from 'redux/selectors/content';
+import { makeSelectFileRenderModeForUri, makeSelectIsPlaying } from 'redux/selectors/content';
 import { withRouter } from 'react-router';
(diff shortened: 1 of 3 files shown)


seanyesmunt:
Thanks. Your description is correct.

PR #5584, opened 2021-02-26
videojs: Localize + add shortcut tooltip

### LINKED ISSUE #3041: Display/hint what keyboard shortcuts are available in videos
https://github.com/lbryio/lbry-desktop/pull/2999/files - we added keyboard shortcuts, but no one will know how to use them. We should do something similar to YouTube where there are tool tips over the actions. We may need to add an action overlay for skipping.

infinite-persistence:
## Issue
Partially closes #3041[: Display/hint what keyboard shortcuts are available in videos](https://github.com/lbryio/lbry-desktop/issues/3041)
(minus the suggested "all shortcuts" overlay -- saving that for another day)

## Notes
We override the text to:
1. Add keyboard shortcut to the tool-tip.
2. Override videojs' i18n and use our own (don't want to have 2 systems).

I was hoping the text can be set once during setup instead of using event listeners, but videojs itself updates the text to it's default strings at those listeners (thus resetting our effort), so we'll have to use listeners as well.

### DIFF
--- a/static/app-strings.json
+++ b/static/app-strings.json
@@ -1598,5 +1598,14 @@
   "Are you sure you want to view this content? Viewing will not unblock @%channel%": "Are you sure you want to view this content? Viewing will not unblock @%channel%",
   "Are you sure you want to view this content? Viewing will not unmute @%channel%": "Are you sure you want to view this content? Viewing will not unmute @%channel%",
   "View Content": "View Content",
+  "Mute (m)": "Mute (m)",
+  "Playback Rate (<, >)": "Playback Rate (<, >)",
+  "Fullscreen (f)": "Fullscreen (f)",
+  "Pause (space)": "Pause (space)",
+  "Play (space)": "Play (space)",
+  "Unmute (m)": "Unmute (m)",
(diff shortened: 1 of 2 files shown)


DispatchCommit:
This may lead to conflicts with accessibility / screenreader features (`aria-*` and `title` attributes) which video.js handles with their i18n.

So while having 2 i18n's may not seem ideal, it may be simpler than re-implementing the i18n system for video.js.

https://github.com/videojs/video.js/blob/master/docs/guides/languages.md

infinite-persistence:
> it may be simpler than re-implementing the i18n system for video.js.

- By "system", I just meant that we swap videojs' json with ours.  The rest of the code flow remains the same (videojs calls `controlText` too). `controlText` is the proper API to change text anyways.
- Looking at the final DOM, the output is equivalent (no changes to accessibility)

```
original -->  <json from videojs> -->  contructor or event handlers --> sets `controlText`
new      -->  <lbry json>         -->  event handlers --> sets `controlText`
```

image

DispatchCommit (approved):
LGTM

Issue #5384, opened 2021-01-26
Editing a comment that has an embed in the playing state, doesn't hide the player


DispatchCommit:
Post a comment with a link to a video so it embeds the video in the comment.
Click play so the video begins playing inline within the comment.
Click edit on the comment.

The player will still be visible and most likely be covering the text field and post / cancel buttons.

![image](https://user-images.githubusercontent.com/45262335/105796697-41209000-5f44-11eb-8896-61bac6a45ac0.png)

DaniNz:
Hello, I'm working on this issue.

DaniNz:
Here is the PR: https://github.com/lbryio/lbry-desktop/pull/5389

PR #5606, opened 2021-03-02
Don't close main player when editing comments

infinite-persistence:
## Issue
image
https://discord.com/channels/362322208485277697/363087331475062785/815132676293918730

## Related
#5384: Editing a comment that has an embed in the playing state, doesn't hide the player

## Changes
- Renamed `closeInlinePlayer` to `clearPlayingUri` to reflect it's actual function.
- Add additional check to see if the current video is actually inline before closing it.

Wanted to create an actual `closeInlinePlayer`, but let's wait until there are more usages.

## Note
- [ ]  I've tested the "main player" case (don't stop the video), but I couldn't test the "inline player" case (should close the video) as I couldn't create a comment today.  Kept receiving failure.


seanyesmunt (approved):
Looks good. Will merge after this is rebased

infinite-persistence:
`Changes`
- Rebased to master
- Added "mute and block" i18n update

PR #5617, opened 2021-03-04
Linux: Don't add extra call to associate 'lbry://'

### LINKED ISSUE #4729: LBRY desktop auto-sets itself as default application for opening html files in Linux file managers like nemo after logout and login or a cold/warm boot
**Bug**
When opening an HTML file in a Linux File Manager like nemo, LBRY is launched instead of the default browser or application set to open HTML files. The defaults can be reverted in an HTML file properties but after a logout and login or a cold/warm boot, LBRY desktop sets itself as the default application for opening HTML files. Reverting to the default browser only lasts as long as one's PC is on and is no logout has occurred.

**To Reproduce**
Steps to reproduce the behaviour:
1. Download and install the lbry-desktop deb package and install it in any Debian/Ubuntu based OS.
2. Login to lbry via the app. (I am not sure if this contributes to the cause of the bug).
3. Logout and log back in or perform a warm/cold boot.
4. Create an HTML file with sample HTML code and launch/open it in the File Manager or launch/open an available HTML file in a File Manager
5. If lbry-desktop is launched instead of the default browser, revert to the default browser via the file properties and launch the file.
6. Logout and log back in or perform a warm/cold boot.
7. Launch the HTML file in a File Manager.

**Expected behaviour**
The default browser or applications like editors or IDEs should be used to launch/open an HTML file instead of lbry-desktop which currently auto-sets itself as the default application for launching/opening an HTML file after a cold/warm boot or a logout and login.

**Screenshots**
Below are the screenshots of the default web browser set
default apps

and when right clicking an html file in nemo (Cinnamon DE File Manager) and opens with LBRY by default.
lbry default

**System Configuration**

- LBRY Daemon version: 0.79.1
- LBRY App version: 0.47.2
- LBRY Installation ID: 5nMz3Kv8dG4XhXqzeWBNt4dnYMYQpbFhWEuMQnhgkQxBsVuSRHQCwfwqLvGn6ru34N
- Operating system: Linux Mint 20 Cinnamon (Ubuntu 20.04 Base)

Full output of: `curl 'http://localhost:5279' --data '{"method":"version"}'`:
`{
  "jsonrpc": "2.0",
  "result": {
    "build": "release",
    "desktop": "X-Cinnamon",
    "distro": {
      "codename": "ulyana",
      "id": "linuxmint",
      "like": "ubuntu",
      "version": "20",
      "version_parts": {
        "build_number": "",
        "major": "20",
        "minor": ""
      }
    },
    "lbrynet_version": "0.79.1",
    "os_release": "5.4.0-45-generic",
    "os_system": "Linux",
    "platform": "Linux-5.4.0-45-generic-x86_64-with-LinuxMint-20-ulyana",
    "processor": "x86_64",
    "python_version": "3.7.8",
    "version": "0.79.1"
  }
}`

infinite-persistence:
## Issue
Potentially closes #4729[: LBRY desktop auto-sets itself as default application for opening html files in Linux file managers like nemo after logout and login or a cold/warm boot](https://github.com/lbryio/lbry-desktop/issues/4729)

- https://gitlab.freedesktop.org/xdg/xdg-utils/-/issues/180:
    - It appears that there is a piece of code in xdg that defaults the type to 'text/html' if the parameter is empty. I'm not sure if that's entirely xdg's fault, or it's the way Electron's `setAsDefaultProtocolClient` uses it.
    - `setAsDefaultProtocolClient` is called on app startup, causing the incorrect 'text/html' association every LBRY starts.

## Changes
In Linux Mint at least, it seems like the `protocols` entry in `electron-builder.json` is enough to associate the LBRY protocol to this app. The extra `setAsDefaultProtocolClient` seems unnecessary.

My test method:
With `setAsDefaultProtocolClient` removed for Linux,
- `<a href="lbry://xxx">` still opens the app.
- `<a href="lbry://xxx">` does not open the app if the `protocols` entry is removed from `electron-build.json` (confirming that it's doing it's job).

### DIFF
--- a/electron/index.js
+++ b/electron/index.js
@@ -46,7 +46,10 @@ let lbryFirst;
 
 const appState = {};
 
-app.setAsDefaultProtocolClient('lbry');
+if (process.platform !== 'linux') {
+  app.setAsDefaultProtocolClient('lbry');
+}
+
 app.name = 'LBRY';
 app.setAppUserModelId('io.lbry.LBRY');
 app.commandLine.appendSwitch('force-color-profile', 'srgb');


seanyesmunt:
Nice. @infinite-persistence Can you add an entry in the changelog with this? I'd like to start updating that with every PR.

OdinVex:
How about an option to disable lbry protocol entirely? I don't want associations.

PR #5618, opened 2021-03-04
Synchronize app-string with Odysee branch

infinite-persistence:
## Why
- After this PR, the next merge of `app-string.json` into Odysee will be easier (just take the version from `master` if git says there is a "conflict") **
- Future merges should have no more conflicts assuming changes are always done in `master` only.

## Required work during Odysee merge
- [ ] Add `SITE_HELP_EMAIL=help@odysee.com` in the Odysee branch's env file.
- [ ] Ensure the Odysee side is using the same variable'd string as in master for the following files (I think git should trigger a conflict, but just in case it doesn't:
   - `ui/component/userSignInPassword/view.jsx`
   - `ui/component/syncFatalError/view.jsx`
   - `ui/page/help/view.jsx`

## **
I believe the odysee-merge after this PR will yield a conflict for `app-string.json`.  Just take the one in `master` directly -- it'll be correct for both sides, since:
- `A channel is required to comment on lbry.tv` is not needed in odysee branch
- `Try refreshing to fix the issue. If that doesn't work, email help@lbry.com for support.`
- `Try refreshing to fix the issue. If that doesn't work, email help@odysee.com for support.` is not needed in odysee branch. Use the one with a variable from master.
- `No Results` (with caps) is not used on Odysee either. Take the master version (without this string). There is already a lower-case version.
- `You are currently following %followingCount% tag` was removed in Odysee, but it's still used in a component. Putting it back in would make the app-string file match master 100%.


infinite-persistence:
`Update - 92137bc`
- Rebased to fix new string conflicts.

PR #5621, opened 2021-03-04
livestream comments over websockets

seanyesmunt:


ui/constants/livestream.js
```diff
@@ -1,11 +1,11 @@
 //
-export const LIVE_STREAM_CHANNEL = '@SaltyCracker';
-export const LIVE_STREAM_CHANNEL_CLAIM_ID = 'ae12172e991e675ed842a0a4412245d8ee1eb398';
-export const BITWAVE_USERNAME = 'saltycracker1';
+// export const LIVE_STREAM_CHANNEL = '@SaltyCracker';
```
seanyesmunt, line 2: will update this before I merge

ui/component/commentCreate/view.jsx
```diff
@@ -12,7 +12,7 @@ import { useHistory } from 'react-router';
 import type { ElementRef } from 'react';
 import emoji from 'emoji-dictionary';
 
-const COMMENT_SLOW_MODE_SECONDS = 30;
+const COMMENT_SLOW_MODE_SECONDS = 0;
```
DispatchCommit, line 15: We should still have *some* lower limit. maybe 3-5 seconds for a smaller stream right now just to be safe?

DispatchCommit (changes requested):

PR #5531, opened 2021-02-17
Report claims via right-click

infinite-persistence:
## Issue
Closes #5500 [Allow right click + report feature on tiles ](https://github.com/lbryio/lbry-desktop/issues/5500)

## Changes
- Refactored context-menu code so that both Claim Preview and Claim Tile uses the same popup.
- Added ability to Report from the context menu.

_I left the commits separated to hopefully ease the review process. Feel free to squash if necessary._

Issue #5500, opened 2021-02-08
Allow right click + report feature on tiles


tzarebczan:
Users may come across untagged content and want to report it without opening the file itself.

PR #5623, opened 2021-03-04
ClaimPreview: use new MenuList instead of context-menu (works for both Desktop and Web)

infinite-persistence:
## Issue
- Second attempt at #5500: `Allow right click + report feature on tiles`
- Related: PR #5531

image

## Notes
Although the link can already be easily copied on web, the menu item is still retained for consistency between the platforms.


seanyesmunt:
awesome

infinite-persistence:
`Amend`
Separator:
- Handle runtime warning that `MenuList::onSelect` cannot be null or undefined (added empty handler)
- Also added `disabled` to prevent the separator from being "clickable"

seanyesmunt:
I just removed the `MenuButton` surrounding the `hr`

PR #5624, opened 2021-03-05
Add filtering on notifications page - 5193

### LINKED ISSUE #5193: Add filtering on notifications page
So I can easily see all comment notifications, uploads, etc

infinite-persistence:
## Issue
Closes #5193[ Add filtering on notifications page](https://github.com/lbryio/lbry-desktop/issues/5193)

## Notes
- Add a combo box, and simply filter out the notifications based on the combo box value.
- Selection state is persistent, but reset to All if there are unread ones when entering the page.
- Tell the user that "no notifications" could be due to the filter.
- Used "Followers" instead of "Subscribers". 
- Re-used the same CSS as "claim search filter" -- the white-over-pink is a little hard to read, though.

image


seanyesmunt (changes requested):
Please move the title outside of the card, similar to how the `/$/uploads` page looks. Then move the `select` to the left of the "mark all as read" button

infinite-persistence:
### Update
- Moved the title and actions out, similar to Uploads Page.
    -  Unlike the Uploads Page, I think we would want to display the action buttons on mobile as well (or at least show "Mark all as read").  With that, `claim-list__alt-controls--wrap` was added. Hopefully the mobile look doesn't deviate from style too much.

<details>
  <summary>Screenshots</summary>

## Normal:
image

## Medium:
image

## Small:
image
</details>

seanyesmunt:
Hmm. This might need some more UI work to deal with the smaller screens. I'll spend some time thinking more about how this should be displayed. Thanks for postign the screenshots.

seanyesmunt (approved):

PR #5625, opened 2021-03-05
Batch-resolves channel page query

### LINKED ISSUE #5597: batch resolves on channel search page
Looks like resolves are triggered one by one from lighthouse result

infinite-persistence:
## Issue
Closes #5597[ batch resolves on channel search page](https://github.com/lbryio/lbry-desktop/issues/5597)

## Changes
Do a batch-resolve immediate after getting the results and before setting the result variable, as the latter would result in the Claim* components resolving individually.

## Notes
I enabled `returnCachedClaims` -- I assumed that's a reasonable thing to do. I don't see other places use it, though, so highlighting it here.

### DIFF
--- a/ui/component/channelContent/index.js
+++ b/ui/component/channelContent/index.js
@@ -34,7 +34,7 @@ const select = (state, props) => {
 };
 
 const perform = (dispatch) => ({
-  doResolveUris: (uris) => dispatch(doResolveUris(uris)),
+  doResolveUris: (uris, returnCachedUris) => dispatch(doResolveUris(uris, returnCachedUris)),
 });
 
 export default withRouter(connect(select, perform)(ChannelPage));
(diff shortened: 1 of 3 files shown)


DispatchCommit (approved):
Can't comment on the `returnCachedClaims` part as I'm unfamiliar.

The rest of the code LGTM. I was going to try and compare the PR vs. Production performance for this, but I couldn't find any significant changes between performance graphs - and if there is a correlation that II'm not seeing, then more likely the overall network variations

image

infinite-persistence:
`Update`
- Rebased/resolved conflict with master.

ui/component/channelContent/index.js
```diff
@@ -32,4 +33,8 @@ const select = (state, props) => {
   };
 };
 
-export default withRouter(connect(select)(ChannelPage));
+const perform = (dispatch) => ({
+  doResolveUris: (uris) => dispatch(doResolveUris(uris)),
```
seanyesmunt, line 37: You need to forward the second argument here

seanyesmunt (changes requested):

ui/component/channelContent/index.js
seanyesmunt: I pushed up a commit to do that

seanyesmunt (approved):

Issue #5590, opened 2021-02-27
Hide hls-quality-selector by default


DispatchCommit:
Instead of allowing the hls-quality-selector to appear by default before being hidden when metadata loaded is fired, we should initially hide the button, and only optionally show the selector if multiple quality levels are available.

This would just require adding a call to `.hide()` when initializing.

PR #5632, opened 2021-03-08
hide quality selector by default

DispatchCommit:
Just hides the hls-quality-selector button from the UI until metadata is loaded that indicates multiple quality levels are available.

This will hide the "auto" button that currently appears while waiting for the video's pre-fetch to complete indicating whether the file in an mp4 or m3u8 file.


DispatchCommit:
Resolves #5590

PR #5643, opened 2021-03-09
vjs i18n: Fix "undefined" error due to hidden ctrls

infinite-persistence:
## Issue
The change in making the "Quality" component hidden by default exposed a minor error in `resolveCtrlText` for not handling hidden components.

Luckily, videojs cathes the error and became a no-op, so it didn't crash the app.

## Change
Add a try-catch block to handle any temp-removal of components in the future.

## Note
This is a touch up for PR #5584

### DIFF
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -253,44 +253,48 @@ export default React.memo<Props>(function VideoJs(props: Props) {
     // as the listener to update static texts.
     const player = playerRef.current;
     if (player) {
-      const controlBar = player.getChild('controlBar');
-      switch (e.type) {
-        case 'play':
-          controlBar.getChild('PlayToggle').controlText(__('Pause (space)'));
-          break;
-        case 'pause':
-          controlBar.getChild('PlayToggle').controlText(__('Play (space)'));
-          break;
-        case 'volumechange':
-          controlBar
-            .getChild('VolumePanel')
-            .getChild('MuteToggle')
-            .controlText(player.muted() || player.volume() === 0 ? __('Unmute (m)') : __('Mute (m)'));
-          break;
-        case 'fullscreenchange':
-          controlBar
-            .getChild('FullscreenToggle')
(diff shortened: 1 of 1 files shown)


DispatchCommit:
I can't help be feel like something is wrong with try...catch'ing blocks of code to this degree haha...

I think the undefined is caused by the HLS selector not inheriting from `ClickableComponent` like the other elements are.
Converting the plugin to use `https://docs.videojs.com/clickablecomponent#controlText` clickable component would likely be the proper solution and bring the quality selector's parent class up to date. But that requires updating many of the internal functions to use the clickablecomponent structure. 

Also there are localization tweaks needed for the quality selector in addition to just this that would allow for changing the word "auto" to other languages for example. Not sure how that should be handled.

infinite-persistence:
> _I think the undefined is caused by the HLS selector not inheriting from ClickableComponent like the other elements are.
Converting the plugin to use https://docs.videojs.com/clickablecomponent#controlText clickable component would likely be the proper solution and bring the quality selector's parent class up to date. But that requires updating many of the internal functions to use the clickablecomponent structure._

I see.  I didn't check what was chaned in the Quality selector.  The text-change used to work on initial PR.

infinite-persistence:
Hm ... the change was simply to `hide` qualit-selector, right?  Does that change the class structure?  When we didn't hide it, `controlText` was valid.

image
<sub>Here, I localized "Quality" to "Quality---"</sub>

DispatchCommit:
Yea it was basically a one-liner tbh.

hls-quality-selector uses `ConcreteButton` to create the button.
`ConcreteButton` extends `MenuButton`
`MenuButton` extends `Component`
`Component` defines `.hide()`: 
https://github.com/videojs/video.js/blob/3d676d68ca863795e2c6902bc9b495a0e8a7571a/src/js/component.js#L854

Though the docs say that `MenuButton` should also support `controlText` as a function that acts aas both a getter and a setter...
https://docs.videojs.com/menubutton#controlText

infinite-persistence:
`Close` as "no-good" in favor of a better solution.
Users not impacted at the moment, other than their console being polluted.

PR #5645, opened 2021-03-09
Disable video previews in comments/posts made by channels below a certain channel staked level

### LINKED ISSUE #5636: Disable video previews in comments/posts made by channels below a certain channel staked level
We would like to start disabling video/image previews in comments if the creator of the comment has < x LBC staked to their channel. This is the first feature that will be turned on based on LBC staked so some things will need to be moved around.

- [This function](https://github.com/lbryio/lbry-desktop/blob/master/ui/component/channelStakedIndicator/view.jsx#L17-L36) should be moved to a selector `makeSelectStakedLevelForChannelUri` and added to `redux/selectors/claims.js` in lbry-redux.

- MarkdownPreview should accept a new prop `displayClaimPreviews` that is defaulted to false, and only true when the creator channel is at level 4 or greater. 

- When a claim preview can't be displayed because of insufficient staked amount, just render a regular link with a tooltip. The tooltip can say something like "This channel isn't staking enough :LBC: for link previews".

- The UI for lbry.tv/desktop should not show anything about a channels "level". Levels are only for odysee.

infinite-persistence:
## Issue
Closes #5636[: Disable video previews in comments/posts made by channels below a certain channel staked level ](https://github.com/lbryio/lbry-desktop/issues/5636)

## Notes
- Added `makeSelectTotalStakedAmountForChannelUri` since the formula was repeated in many places. 
  - If this selector makes sense, I'll clean up the other instances to use this selector later.
- Placed `TRUSTED_CHANNEL_MIN_STAKED_LEVEL=4` in the env/config file for now.
- I changed the suggested `displayClaimPreviews` to `stakedLevel`, so that I need to only need to check against `>= TRUSTED_CHANNEL_MIN_STAKED_LEVEL` in one place (MarkdownPreview) instead of at every caller. The downside is that it puts more responsibility/logic in MarkdownPreview.
- Handled remaining usages of MarkdownPreview
  - Channel About
    - description: seems handy to have a preview, so I pulled in the "allowed" logic here as well.
    - email: shouldn't render preview at all (might also want to sanitize the input)
    - website: shouldn't render preview at all
  - Publish Preview:
    - description: there's not enough height anyway, so don't render preview.


seanyesmunt:
Nice work. I will check this out by tomorrow.

.env.defaults
```diff
@@ -30,6 +30,7 @@ YRBL_SAD_IMG_URL=https://cdn.lbryplayer.xyz/api/v3/streams/free/yrbl-sad/c2d9649
 ENABLE_COMMENT_REACTIONS=true
 ENABLE_FILE_REACTIONS=false
 ENABLE_CREATOR_REACTIONS=false
+TRUSTED_CHANNEL_MIN_STAKED_LEVEL=4
```
seanyesmunt, line 33: I think this should be specific to video comments. There will be more features like this that use different levels.

Maybe `CHANNEL_STAKED_LEVEL_VIDEO_COMMENTS=4`

Then in the future we can have 
`CHANNEL_STAKED_LEVEL_OTHER_FEATURE`

seanyesmunt: Please update the desktop config and set the level to 0
https://github.com/lbryio/lbry-desktop/blob/master/.github/workflows/deploy.yml#L76

seanyesmunt (changes requested):
Looks great. Just one comment.

.env.defaults
infinite-persistence: > I think this should be specific to video comments.

I blocked the videos in Post and Channel About as well.  Should I allow them?

infinite-persistence:
### Update
- Rebased 
- Changed variable name to `CHANNEL_STAKED_LEVEL_VIDEO_COMMENTS`
- Set it to 0 in `deploy.yml`

PR #5650, opened 2021-03-09
pre-roll ads

seanyesmunt:


seanyesmunt:
@DispatchCommit looking for feedback on this. Some of the real values are commented out to make it easier to test locally.

ui/component/viewers/videoViewer/view.jsx
```diff
@@ -217,16 +268,50 @@ function VideoViewer(props: Props) {
...
+            </I18nMessage>
+          </span>
+        </>
+      )}
+
+      {!isFetchingAd && (
+        <VideoJs
+          adUrl={adUrl}
+          source={adUrl || source}
+          sourceType={forcePlayer || adUrl ? 'video/mp4' : contentType}
```
seanyesmunt, line 306: Don't really like how I did the `source` and `sourceType` props

DispatchCommit:
https://github.com/videojs/videojs-contrib-ads

this library would probably solve some issues

CodeLikeCrazE:
If you are going to add ads, then at least make them able to be disabled by the user (and also if the user is signed in, just remove the ad, without a post to the iframe, because some people, like me, disable javascript in Google Ads).

CodeLikeCrazE:
Also for something as big as adding ads, there almost definitely be more than 2 people saying "yeah okay that's fine".

PR #5671, opened 2021-03-14
Show extra "Sign In" button on Desktop

### LINKED ISSUE #3989: [desktop] Better indication of being signed in / out on account header
Right now, without clicking into the account header, it's not easy to tell if you are signed into a lbry.tv account or not. 

Sometimes the app uses local data and gets signed out, but users are unaware of this. Maybe can be tackled with: https://github.com/lbryio/lbry-desktop/issues/3749

infinite-persistence:
## Issue
- Closes #3989[: [desktop] Better indication of being signed in / out on account header](https://github.com/lbryio/lbry-desktop/issues/3989)
- Also, I often got confused in development on whether I'm using `local` or `shared`.  Having something to indicate the logged-off status would be helpful.

## Notes
- Retained the ["Sign In" for Desktop, "Log In" for Web] labeling.

image
image

### DIFF
--- a/ui/component/header/view.jsx
+++ b/ui/component/header/view.jsx
@@ -162,8 +162,14 @@ const Header = (props: Props) => {
 
   const loginButtons = (
     <div className="header__auth-buttons">
-      <Button navigate={`/$/${PAGES.AUTH_SIGNIN}`} button="link" label={__('Log In')} className="mobile-hidden" />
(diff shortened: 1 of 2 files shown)


seanyesmunt:
I think #3989 is still a valid issue, but I don't think we should be make the sign in button more prominent in the desktop app. Generally we should be making the links/buttons to centralized features less noticeable on desktop.

I'm not sure what the solution to #3989 should be

PR #5682, opened 2021-03-16
Odysee string sync-up (mostly Livestream strings)

infinite-persistence:
Should be suitable for `0.50.0`, unless we don't want translators to see livestream-related strings yet.


seanyesmunt:
I would like to avoid bringing these livestream strings in. Most of them will be replaced in #5691

infinite-persistence:
### Update
- Narrowed down to just regular string update (removed Livestream strings)

PR #5699, opened 2021-03-17
add ENABLE_NO_SOURCE_CLAIMS to config

seanyesmunt:
this sets has_source:true if the variable is true to non-odysee apps won't be shown livestream claims

Not mergable until the new SDK is live on web


seanyesmunt:
-

PR #5703, opened 2021-03-18
i18n iapi

### LINKED ISSUE #5589: List of missing strings from internal repo
These are list of strings coming from internal repos.  As contributors can't ...
- check if they are enclosed by the i18n macro `__()`
- can't add the macro
- don't know the variables names, if exists (e.g. `%channel_name%`)
... I'll just list them here for now instead of adding the strings directly.

----
- [ ] `New content from %xxx%` in the Notifications page
image

- [ ] Not sure where this is from; probably needs capitalization too.
image

- [ ] Preferences
image

- [ ] `reward code not found`
image

- [ ] Strings in the Rewards section. Most of them include a "current count" variable.  Here are 2 examples:
    -  `Watch content and earn more Credits for each level unlocked! 10 views required for level 1 (Current Score: 0). Only up to 10 views per day count.`
    - `Follow your favorite creators and earn more Credits for each level unlocked! Follow 1 creators for level 1 (Current Score: 0).`

infinite-persistence:
## Issue
Closes  #5589: `List of missing strings from internal repo`
_Although the "variable'd strings" issue remain unresolved, we can close the issue since solving that would require messing with the repo._

## Notes
`"Favorite Five 🖐️"` -- not sure if this will work well with Transifex.  Let's see how it goes.


infinite-persistence:
If this goes in, we should make another Transifex push

PR #5711, opened 2021-03-19
Fix autoplay infinite loop

### LINKED ISSUE #3661: Autoplay + Related go into loops ( infinite ) sometimes
**Bug**
The related videos make the user go into play infinite loops if they have autoplay feature turned on.
This happens because the first related video is either the same video, or alternate secondary video that points back to the original one that you are watching.

**Expected behavior**

If the first related is autoplayed, I would expect my past watch history ( at least for this session) to be taken into account when deciding the first related video.

**Relation**

Video1 => has as First related Video2
Video1

Video2 => has as first related Video1
video2

Video3=> has as first related Video3
video3

**System Configuration**

- App | 0.41.0
- Daemon (lbrynet) | 0.55.0
- Platform | Windows (Windows-10-10.0.19041-SP0)

infinite-persistence:
## Issue
Closes #3661: [Autoplay + Related go into loops ( infinite ) sometimes](https://github.com/lbryio/lbry-desktop/issues/3661)

## GUI
Push the actual "next" item into the top of the list.

## History search
1. Skip if the next item is itself.
2. The URL stored in the history comes in various forms, so a direct comparison won't work.
    - There's also a weird case where the URL differs by just a little (p.09 vs p-09), but with the same claim ID:

lbry://vacuum-tube-computer-p.09-–-building#5212bc8bc63c373e2bf1ebc5b765595ed7b6514d
lbry://vacuum-tube-computer-p-09-–-building#5212bc8bc63c373e2bf1ebc5b765595ed7b6514d

Check the claim_id as well to cover cases like these.

### DIFF
--- a/ui/redux/selectors/content.js
+++ b/ui/redux/selectors/content.js
@@ -123,7 +123,23 @@ export const makeSelectNextUnplayedRecommended = (uri: string) =>
             continue;
           }
 
-          if (!history.some((item) => item.uri === recommendedForUri[i])) {
+          const recommendedUriInfo = parseURI(recommendedUri);
+          const recommendedUriShort = recommendedUriInfo.claimName + '#' + recommendedUriInfo.claimId.substring(0, 1);
+
+          if (claimsByUri[uri] && claimsByUri[uri].claim_id === recommendedUriInfo.claimId) {
+            // Skip myself (same claim ID)
+            continue;
+          }
+
+          if (
+            !history.some((h) => {
+              const directMatch = h.uri === recommendedForUri[i];
+              const shortUriMatch = h.uri.includes(recommendedUriShort);
+              const idMatch = claimsByUri[h.uri] && claimsByUri[h.uri].claim_id === recommendedUriInfo.claimId;
+
+              return directMatch || shortUriMatch || idMatch;
+            })
+          ) {
             return recommendedForUri[i];
           }
         }
--- a/ui/component/recommendedContent/index.js
+++ b/ui/component/recommendedContent/index.js
@@ -3,17 +3,19 @@ import { makeSelectClaimForUri, makeSelectClaimIsNsfw } from 'lbry-redux';
 import { doSearch } from 'redux/actions/search';
 import { makeSelectRecommendedContentForUri, selectIsSearching } from 'redux/selectors/search';
 import { selectUserVerifiedEmail } from 'redux/selectors/user';
+import { makeSelectNextUnplayedRecommended } from 'redux/selectors/content';
 import RecommendedVideos from './view';
 
 const select = (state, props) => ({
   claim: makeSelectClaimForUri(props.uri)(state),
   mature: makeSelectClaimIsNsfw(props.uri)(state),
   recommendedContent: makeSelectRecommendedContentForUri(props.uri)(state),
+  nextRecommendedUri: makeSelectNextUnplayedRecommended(props.uri)(state),
   isSearching: selectIsSearching(state),
   isAuthenticated: selectUserVerifiedEmail(state),
 });
 
-const perform = dispatch => ({
+const perform = (dispatch) => ({
   search: (query, options) => dispatch(doSearch(query, options)),
 });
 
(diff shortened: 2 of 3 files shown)


ui/redux/selectors/content.js
```diff
@@ -123,7 +123,21 @@ export const makeSelectNextUnplayedRecommended = (uri: string) =>
...
+          if (claimsByUri[uri] && claimsByUri[uri].claim_id === recommendedUriInfo.claimId) {
+            // Skip myself (same claim ID)
+            continue;
+          }
+
+          if (
+            !history.some((h) => {
+              const directMatch = h.uri === recommendedForUri[i];
+              const claimNameMatch = h.uri.includes(recommendedUriInfo.claimName);
+              const idMatch = claimsByUri[h.uri] && claimsByUri[h.uri].claim_id === recommendedUriInfo.claimId;
```
DispatchCommit, line 137: Shouldn't this check work in all cases? If so do we need to check for a name match?

ui/redux/selectors/content.js
```diff
@@ -123,7 +123,21 @@ export const makeSelectNextUnplayedRecommended = (uri: string) =>
...
+
+          if (claimsByUri[uri] && claimsByUri[uri].claim_id === recommendedUriInfo.claimId) {
+            // Skip myself (same claim ID)
+            continue;
+          }
+
+          if (
+            !history.some((h) => {
+              const directMatch = h.uri === recommendedForUri[i];
+              const claimNameMatch = h.uri.includes(recommendedUriInfo.claimName);
```
DispatchCommit, line 136: I don't like the idea of blindly ignoring claims with identical names. In theory, it should be fine as it's unlikely to occur, but it seems like a potential edge case that could be perceived as a bug at some point down the road in the case a creator  abstractly names all their videos the same thing.
Thinking back on it, I have a seen a few people that post identically posted videos when doing commentary or discussion videos, or sometimes just podcasts without episode numbers. 

just seems like this may lead to issues possibly.

ui/redux/selectors/content.js
```diff
@@ -123,7 +123,21 @@ export const makeSelectNextUnplayedRecommended = (uri: string) =>
             continue;
           }
 
-          if (!history.some((item) => item.uri === recommendedForUri[i])) {
+          const recommendedUriInfo = parseURI(recommendedUri);
+
+          if (claimsByUri[uri] && claimsByUri[uri].claim_id === recommendedUriInfo.claimId) {
+            // Skip myself (same claim ID)
```
DispatchCommit, line 129: I can potentially see someone complaining and thinking their videos are never recommended because they never their own videos autoplay. While it's likely you don't want to autoplay watch your own videos, I think a user would rather notice their video being suggested over thinking it's never suggested.

DispatchCommit (changes requested):
almost perfect.

ui/redux/selectors/content.js
infinite-persistence: `No change`

`claimsByUri` doesn't exist unless we resolve the entire history at startup (don't want to do that) -- it was added as a "try if possible" effort, primarily to handle odd URLs like

>_lbry://vacuum-tube-computer-p.09-–-building#5212bc8bc63c373e2bf1ebc5b765595ed7b6514d
lbry://vacuum-tube-computer-p-09-–-building#5212bc8bc63c373e2bf1ebc5b765595ed7b6514d_

ui/redux/selectors/content.js
infinite-persistence: `Fixed`

Compare using the somewhat unique short URI instead.

ui/redux/selectors/content.js
infinite-persistence: `Unrelated?`

This is not checking against "claims that I own", but solving the "video A recommended for video A" problem.
>_There's also a weird case where the URL differs by just a little (p.09 vs p-09), but with the same claim ID:
lbry://vacuum-tube-computer-p.09-–-building#5212bc8bc63c373e2bf1ebc5b765595ed7b6514d
lbry://vacuum-tube-computer-p-09-–-building#5212bc8bc63c373e2bf1ebc5b765595ed7b6514d_

infinite-persistence:
`Update`
- Rebased to master
- Fix potential name-clash

PR #5713, opened 2021-03-19
Support = [Boost | Tip] for both Desktop and Odysee

### LINKED ISSUE #5226: Figure out odysee-friendly way to re-enable supports
Conclusion is:

- Tipping and staking are always initiated via a single button labeled "Support". Default is always tip. Tip should not be selectable when the tip would come from same channel as the content was published under.
- Add high-level element, that lets you switch between Tip and Boost (stake). This should probably look like the tab switcher.
- Ensure button text and help text change when the mode switches.

infinite-persistence:
## Issue
Closes #5226: [Figure out odysee-friendly way to re-enable supports](https://github.com/lbryio/lbry-desktop/issues/5226)

## Merge
When merging into `odysee`, just take the `master` version -- it should work for both Desktop and Odysee.

## Screenshots
<details>
  <summary>Click to expand</summary>

1. Boosting own content
    - image

2. Boosting someone else's content
    - image
3. Tipping someone
    - image

  

</details>


seanyesmunt:
Nice.  I'll do a  UI pass of this in the next couple days.

seanyesmunt (commented):
I haven't looked at the UI yet, but I would like to keep using `support` for code variables, even if we are calling it `boost` in the UI

infinite-persistence:
`--Update--`
- Retained "support" as the variable name instead of changing it to "boost"
- Updated strins
- Rebased to master.

jessopb:
I rebased and force pushed this because it was my fault linter failed.

jessopb (approved):
Works great, of course.

PR #5716, opened 2021-03-19
Fix autoplay not working in floating mode

### LINKED ISSUE #5000: AutoPlay in Floating Player only works one time
1. Play a video and go into FP mode.
2. At the end of the first video, AutoPlay does play the next video automatically.
3. At the end of the second video, AutoPlay wasn't invoked.  In non-FP mode, AutoPlay continues to work beyond the second video.

infinite-persistence:
## Issue
Closes #5000: [AutoPlay in Floating Player only works one time](https://github.com/lbryio/lbry-desktop/issues/5000)

## Changes
Factor out the code that queries the recommendation. It needs to be called in two places:
(1) `RecommendedContent` - need to cover the case of floating player + visit another video page.
(2) When video is floating and autoplayed the next video.

## Notes to reviewer
I think there's no need for `React.useCallback` anymore, but please double-check in case it was needed for something else.

### DIFF
--- a/ui/redux/actions/search.js
+++ b/ui/redux/actions/search.js
@@ -1,6 +1,6 @@
 // @flow
 import * as ACTIONS from 'constants/action_types';
-import { buildURI, doResolveUris, batchActions, isURIValid } from 'lbry-redux';
+import { buildURI, doResolveUris, batchActions, isURIValid, makeSelectClaimForUri } from 'lbry-redux';
 import { makeSelectSearchUris, makeSelectQueryWithOptions, selectSearchValue } from 'redux/selectors/search';
 import handleFetchResponse from 'util/handle-fetch';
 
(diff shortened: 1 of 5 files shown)


DispatchCommit (commented):
passes my lookover. Moving the logic out of the component seems like the right approach.

I'll leave this on sean to test and approve, but I don't see any potential edge cases that might cause issues in the logic.

ui/redux/actions/search.js
```diff
@@ -113,4 +113,20 @@ export const doUpdateSearchOptions = (newOptions: SearchOptions, additionalOptio
   }
 };
 
+export const doGetRecommendedContent = (uri: string, mature: boolean) => (dispatch: Dispatch, getState: GetState) => {
```
seanyesmunt, line 116: `doFetchRecommendedContent`

seanyesmunt (changes requested):
This looks  good. I would like to use `doFetchXXX` since  that is the pattern we use in other actions.

infinite-persistence:
`-- Update --`
- Resolve conflict with master.
- Renamed: `doGetRecommendedContent` --> `doFetchRecommendedContent`

seanyesmunt (approved):

PR #5717, opened 2021-03-19
improve keyboard shortcuts

saltrafael:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Documentation changes

## Fixes

Issue Number:

Keyboard shortcuts are limited

## What is the new behavior?

Changed left and right arrow keys to move 5 in time, added J and L to move 10 and K for pause. Also added up and down keybinds for 5% volume change, now mimicking other platform's keyboard shortcuts

### DIFF
--- a/ui/component/viewers/videoViewer/internal/overlays.js
+++ b/ui/component/viewers/videoViewer/internal/overlays.js
@@ -97,3 +97,22 @@ export function showSeekedOverlay(player, duration, isForward) {
 
   showOneOffOverlay(player, OVERLAY_CLASS_SEEKED, overlayJsx, 'center');
 }
+
+/**
+ * Displays a transient "Volume Percentage" overlay.
+ *
+ * @param player The videojs instance.
+ * @param percentage how much % of volume active.
+ */
+export function showVolumeverlay(player, percentage) {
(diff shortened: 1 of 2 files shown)


DispatchCommit (approved):
Bless, looks great, nice quality of life improvement! :)

I pushed a fix for some of the linting errors - should pass test now.

if tests pass, it's good to merge.

infinite-persistence:
- [ ] **Swap seek duration with arrow keys?**
    - If we are going to add `j` and `l`, would it better to make them +/-10s instead, and lower the arrow keys (more commonly-used) to +/-5s?  This has been suggested in another issue and would match the other platform.

- [ ] **_Up/Down_ would also scroll the page.**
   - I wanted to do `up`/`down` for volume in the past, but the current approach would also scroll page down.  I think we are listening to the event at too high up of a component.

saltrafael:
> * [ ]  **_Up/Down_ would also scroll the page.**
>   
>   * I wanted to do `up`/`down` for volume in the past, but the current approach would also scroll page down.  I think we are listening to the event at too high up of a component.

Yeah I noticed that, but it works great on fullscreen mode!

PR #5719, opened 2021-03-20
Markdown editor: Remove character limit

### LINKED ISSUE #5687: Ensure post mode has no text limit
The post mode for desktop/lbry.tv (and soon to be odysee) should not have a character limit. Looks like is ~2k words right now.

infinite-persistence:
## Issue
Closes #5687[: Ensure post mode has no text limit](https://github.com/lbryio/lbry-desktop/issues/5687)

## Changes
- `type="markdown"` can now have unlimited length if clients don't define `textAreaMaxWidth`.
- The internal default limit of 2000 is narrowed down to `type=textarea`.

_also snuck in some strings_


kauffj:
sweet, thanks for this

PR #5741, opened 2021-03-24
use comment component for livestream comments

seanyesmunt:
This needs some more testing but seems to work.

So the creator can moderate livestream comments


DispatchCommit (approved):

jessopb:
it's currently under odysee on steve

PR #5742, opened 2021-03-24
Revamp search filters

### LINKED ISSUE #4945: Simplify / revamp search filters for odysee (and lbry.tv)
Most sites have at least some simple search filters that we should add back to Odysee. Channel vs videos would be one, and we can also add something to sort by newest/oldest (available today) and should be able to add a "last day/month/year/etc" param (switch between date/relevance) easily. 

Also, unleash the infinite scroll!

infinite-persistence:
## Issue
Closes #4945[: Simplify / revamp search filters for odysee (and lbry.tv)](https://github.com/lbryio/lbry-desktop/issues/4945)

<strike>_"Infinite scroll" will be attempted in another PR._</strike> _Added_

## Merging notes
- `ui/util/query-param.js`
- `ui/component/searchOptions/view.jsx`

When merging into Odysee, solve the conflict for the above 2 files by simply taking the `master` version directly.  It has been updated to handle `SIMPLE_SITE`.

## Changes
- Tweaked the `claim type = file | channel | everything` widget a little to hopefully make it more intuitive.
- Added "Sort By".
- Added "Upload Date" (hidden for now until it's available in Lighthouse).
- Fixed the affected files to handle both Desktop and Odysee.
- Tweaked the layout to be more condensed so that we can see some results as we change the filter.
- Added animations.
- Added "Exact match" option that helps user to surround with quotes (Tom-suggested).

## Differences with Odysee
This will happen automatically when merged to Odysee:
- "Other" (application types) will be hidden
- <strike>"Result Count" will be hidden</strike> No more "results count" for all platforms after adding infinite scroll.

<details>
  <summary>Click to expand...</summary>
  
image
</details>


jessopb:
Nice, bob is testing this.

infinite-persistence:
`Addition`
- Added infinite scroll

----
@jessopb, @tzarebczan,
The only item remaining is "Upload Date", which requires a Lighthouse update.  
I think it's in a merge-able state (assuming passed re-review on new infinite scroll), although waiting for Lighthouse is ok too.

infinite-persistence:
`Addition`
- Added "Upload Date".

No more known pending changes at this moment.

infinite-persistence:
`Set to Draft`
- Found a crashing issue on Recommendations, probably due to the changes to support Infinite Search (i.e. accumulating data  `&from=xx`).  Looking into it.
- But GUI is still good for review.

infinite-persistence:
`Update`
- Fix concat mistake
- Rebased to master

ui/page/search/index.js
```diff
@@ -26,13 +27,15 @@ const select = (state, props) => {
...
   const uris = makeSelectSearchUris(query)(state);
+  const isQueryMaxed = makeSelectIsSearchQueryMaxed(query)(state);
 
   return {
     isSearching: selectIsSearching(state),
     showNsfw: showMature,
     uris: uris,
     isAuthenticated: selectUserVerifiedEmail(state),
     searchOptions: selectSearchOptions(state),
+    isQueryMaxed: isQueryMaxed,
```
seanyesmunt, line 38: maybe something like `hasReachedMaxResultsLength`?

seanyesmunt (commented):
This looks great. Just one comment/question.

seanyesmunt:
I will do a styling pass and push up a commit. Then it can be merged.

ui/page/search/index.js
infinite-persistence: Sounds good.  Also, I'll rebase to fix the conflict

infinite-persistence:
`Update`
- Rename `isQueryMaxed -> hasReachedMaxResultsLength`

----

`Bug`
- [ ]  When changing "Sort By", the `from` field should be resetting to 0 but it is not.  We might only get the last few results if we continue from where we left off.

@seanyesmunt feel free to add the GUI pass commits, but hold off the merge due to the bug.

infinite-persistence:
@seanyesmunt,
- bug fixed (please review)
- style pass merged

PR #5750, opened 2021-03-24
Ls fixes

jessopb:


jessopb:
After publishing, and wiating for pending, pending livestream claim would disappear and be blank. this fixes that.
Also a couple changes to publish page.

DispatchCommit (approved):

PR #5761, opened 2021-03-25
Query filtered notifications instead

### LINKED ISSUE #5694: pass param for notitication drop down
Can pass the param to apis so that it only returns the relevant types, and actually return data. Right now we limit notifications to 50 so you won't get back any data if it's not in the last 50, but will increase back to 200 once I can finish some testing with larger resolves or we don't have to resolve channels/claims in notifications.

https://github.com/lbryio/internal-apis/commit/237c01715a2f01d30fdb70bec3690701358f45ea

We should group the daily watch ones under "Other". The filter value is "default"

the others are followers, comments, comment_replies and new_content respectively.

infinite-persistence:
## Issue
Closes #5694[ pass param for notitication drop down](https://github.com/lbryio/lbry-desktop/issues/5694)

## Merge Notes
The query wasn't working on older notifications, but Tom said he'll handle this later.  **Don't merge this until then**.

## Changes
See commit messages for walkthrough.


infinite-persistence:
Converted to draft until the server-side issue ("only filtering recent notifications") is handled.

infinite-persistence:
`Update`
- i18n
- Rebased to master

seanyesmunt (approved):
This is great. Nice work. This can be merged once it works for older notifications. @tzarebczan is that ready?

tzarebczan:
Yep, should be good to go now. Updated a bunch last night, and the only ones that need more updates are new content ones, but those are bountiful already. 

Planning to finish up the url changes in notifications so we don't have to resolve, and then we can bump up the limit too.

PR #5764, opened 2021-03-25
fix publish something empty

jessopb:
This fixes non-channel claims. It needs to alternately send you to channel create page, I suppose.


jessopb:
will rebase

jessopb:
on bob

Issue #5181, opened 2020-12-14
Bring back tag search


tzarebczan:
Many users complained about not having this option in search. We should bring it back, but in a clearer way.

PR #5789, opened 2021-03-28
Feature Concept: recommended bar extra options for ease of discovery

saltrafael:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Documentation changes

## Fixes

Issue Number:

Recommended bar could be improved to ease discovery in less clicks for the user

## What is the new behavior?

Added extra options for the recommended bar to allow for discovering more specific related content based on channel or tags, as shown in the pics:

 

 

Comparison to that one popular Odysee Alternative from Web2.0:

The main difference being that the odysee alternative platform doesn't allow to filter tags

## Other information

This is a concept I've been testing on the weekend, there are some styling issues like if the channel name is too big you can't expand. Also the autoplay will keep playing the recommended content from "All", it won't be affected if you are at the "More from" or tag tabs

### DIFF
--- a/ui/component/recommendedContent/view.jsx
+++ b/ui/component/recommendedContent/view.jsx
@@ -1,7 +1,6 @@
 // @flow
 import { SHOW_ADS, SIMPLE_SITE } from 'config';
 import React from 'react';
-import classnames from 'classnames';
 import ClaimList from 'component/claimList';
 import ClaimListDiscover from 'component/claimListDiscover';
 import Ads from 'web/component/ads';
@@ -27,14 +26,10 @@ type Props = {
 
 export default function RecommendedContent(props: Props) {
   const { uri, claim, search, mature, recommendedContent, nextRecommendedUri, isSearching, isAuthenticated } = props;
-  const [expanded, setExpanded] = React.useState(false);
   const [allRelated, setAllRelated] = React.useState(true);
   const [moreFrom, setMoreFrom] = React.useState(false);
-  const [tagList, setTagList] = React.useState(false);
   const signingChannel = claim && claim.signing_channel;
   const channelName = signingChannel ? signingChannel.name : null;
-  const claimValue = claim && claim.value;
-  const tags = claimValue ? claimValue.tags : null;
   const isMobile = useIsMobile();
   const isMedium = useIsMediumScreen();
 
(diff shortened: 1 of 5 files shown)


tzarebczan:
Wow, this is a great start and good place to add improvements! For tag searches, are you using claim search?

saltrafael:
> Wow, this is a great start and good place to add improvements! For tag searches, are you using claim search?

Thanks! I just used a claim list discover with the selected tag as filter, so it's the same as going to each tags page, but faster and with less clicks

tzarebczan:
Yes, that's awesome! Maybe it can default to trending sort, and then let users change new/top also. 

This would be a great one to add soon also: https://github.com/lbryio/lbry-desktop/issues/5181 (there's some other search improvements in another PR too).

saltrafael:
> Yes, that's awesome! Maybe it can default to trending sort, and then let users change new/top also.
> 
> This would be a great one to add soon also: #5181 (there's some other search improvements in another PR too).

You mean like this?

 

I had hid the filters so it doesn't look too different from before

kauffj:
this is a really cool idea but will need a style/UX pass from @seanyesmunt

ui/component/recommendedContent/view.jsx
```diff
@@ -24,6 +27,14 @@ type Props = {
 
 export default function RecommendedContent(props: Props) {
   const { uri, claim, search, mature, recommendedContent, nextRecommendedUri, isSearching, isAuthenticated } = props;
+  const [expanded, setExpanded] = React.useState(false);
+  const [allRelated, setAllRelated] = React.useState(true);
+  const [moreFrom, setMoreFrom] = React.useState(false);
```
seanyesmunt, line 32: for now lets just start with the "more from" toggle only.

seanyesmunt (changes requested):
This is great! For now lets just start with one category "more from". Then we don't have to worry about expanding the section for more categories. Once those are removed I will push up a commit to do some styling to this new section.

Thank you for the PR! This will be great to have :slightly_smiling_face:

saltrafael:
> This is great! For now lets just start with one category "more from". Then we don't have to worry about expanding the section for more categories. Once those are removed I will push up a commit to do some styling to this new section.
> 
> Thank you for the PR! This will be great to have 🙂

Done, and thanks for the feedback!!

seanyesmunt:
I added your commits to this branch so I could rebase
https://github.com/lbryio/lbry-desktop/pull/5847

Will merge once the build completes

Issue #5522, opened 2021-02-15
Use OS-locale to format date and time


StoffelNL:
It would be nice when a date and/or time is shown in the desktop-app that it is formatted in the same way as date/time in the operating system the app is running on.

edit: Maybe I should elaborate a little more: I'm in Europe, we don't use AM/PM and I have a hard time remembering which is which. That's why we use 24h-time notation in all written communication. By using the date/time-settings from the operating system people will automatically see the date and time formatted in the way they prefer.

PR #5820, opened 2021-04-03
Add 24-hour clock setting

stefansundin:
I am sure there are plenty of things that might be wrong here, so please let me know what changes you would like to see.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: I think this probably fixes https://github.com/lbryio/lbry-desktop/issues/5522.

Always shows 12-hour clock with AM/PM.

## What is the new behavior?

If the new `24-hour clock` checkbox in the settings is checked, the time is displayed in 24-hour format.

setting

video

transactions

Please let me know if there's another place that displays a date string with hours present.

## Other information

I don't think there's a foolproof way to detect 24-hour format preference from the browser/OS. https://stackoverflow.com/questions/27647918/detect-with-javascript-if-users-machine-is-using-12-hour-clock-am-pm-or-24-cl

Maybe there's a good way to do it in the Electron app, but I don't have much experience with that.


ui/redux/reducers/settings.js
```diff
@@ -46,6 +46,7 @@ const defaultState = {
     [SETTINGS.HIDE_BALANCE]: false,
     [SETTINGS.OS_NOTIFICATIONS_ENABLED]: true,
     [SETTINGS.AUTOMATIC_DARK_MODE_ENABLED]: false,
+    [SETTINGS.CLOCK_24H]: false,
```
stefansundin, line 49: I have no idea what this does, but the other settings were here so I added this one too.

tzarebczan:
Thanks for the PR, we'll get a review on it soon. Can you also squash the commits?

stefansundin:
> Thanks for the PR, we'll get a review on it soon. Can you also squash the commits?

I could do that, but first do you know about the "squash and merge" option? I think you'll find that it is better to squash using that, since 1. you get to edit the commit message, and 2. it preserves the commits in the PR which makes it easier to come back here and look at the changes separately at a later time. At least I think this is superior to manually squashing.

Please let me know if you still want me to squash, and please review it first so I know if there's something I should remove.

seanyesmunt:
We tend to do "Rebase and merge". I think it's fine for you to not squash these since they are simple enough and there are only a few.

ui/redux/reducers/settings.js
seanyesmunt: This is the default value for this setting, which looks good

seanyesmunt (approved):
This is great! Thanks for the PR

stefansundin:
Thanks for merging the PR. Looking forward to seeing it in production. :)

> We tend to do "Rebase and merge".

I think that technically, a single commit merged with "Rebase and merge" will have the exact same end result as many commits merged with "Squash and merge". In both cases there will not be a merge commit. Unless I'm missing something?

PR #5847, opened 2021-04-08
Reopening #5789 - "more from" in related side bar

seanyesmunt:
Created a new PR so I could push to it with a style commit + rebase

Original PR is #5789


saltrafael:
Wow this looks great, I really like the style but on dark mode it looks like this:



is the text in black intended ?

infinite-persistence:
There is also reflow issues on long channel names.  Also, when i18n comes in, `More from` itself could be wide in some languages.

image

PR #5854, opened 2021-04-08
update livestream eligibilty

seanyesmunt:
- add `CHANNEL_STAKED_LEVEL_LIVESTREAM` to config set to level 5
- use `odysee_live_enabled` tag on users insead of `experimental_ui`


tzarebczan:
this LGTM, let's ship it!

PR #5862, opened 2021-04-10
Attempt at Loop Button

saltrafael:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Documentation changes

## Fixes

Issue Number:

Need to manually replay videos

## What is the new behavior?

Since there are no playlists yet, added a Loop button so the same video can be auto replayed, useful if listening to music, tracklists etc on the background

## Other information
Needs a better icon and an indication that the Loop mode is active or not


tzarebczan:
Nice, great feature add and thanks for the PR! We'll get a review on this soon. 

After that, you know the drill...

ui/component/viewers/videoViewer/internal/videojs.jsx
```diff
@@ -299,7 +303,7 @@ export default React.memo<Props>(function VideoJs(props: Props) {
           // (2) We'll have to get 'makeSelectClientSetting(SETTINGS.VIDEO_THEATER_MODE)'
           // as a prop here so we can say "Theater mode|Default mode" instead of
           // "Toggle Theather mode".
-          controlBar.getChild('Button').controlText(__('Toggle Theater mode (t)'));
+          // controlBar.getChild('Button').controlText(__('Toggle Theater mode (t)'));
```
seanyesmunt, line 306: Why is this removed?

ui/scss/component/_button.scss
```diff
@@ -118,6 +118,26 @@
   }
 }
 
+.vjs-button--loop.vjs-button {
```
seanyesmunt, line 121: Instead of duplicating these styles, just add this class to the rule above
```
.vjs-button--theater-mode,
.vjs-button--loop {
  ...
  ```

seanyesmunt (changes requested):
I have a couple comments. Can you also change this to only show on audio files?

ui/component/viewers/videoViewer/internal/videojs.jsx
saltrafael: Was causing a clash with the new button, both were saying "Theater mode"

ui/scss/component/_button.scss
saltrafael: thanks, just kept a separate one for the different icons

saltrafael:
> I have a couple comments. Can you also change this to only show on audio files?

you think it's better to leave it exclusively to audio files only? I can try that

DispatchCommit:
Looping can be done by right clicking the player and choosing "loop" 

image

Also with playlists nearly finished, that will reduce any need to loop a single video. Also this would be better if done as a video.js plugin to better separate code concerns.

Playlist should be the better solution to this though.

saltrafael:
> Looping can be done by right clicking the player and choosing "loop"

This was more a solution to the desktop app, since right clicking doesn't show the same pop up as in browser

Issue #2514, opened 2019-05-26
Can't exit full-screen from embedded content with key F11


btzr-io:
## The Issue
Can't exit `full-screen` from embedded content (video player) with key `F11`.

> Note: The window exits full-screen mode but the video-player is still displayed on the full page.

### Steps to Reproduce
1)  Click the `full-screen` button from the video player.
2) Press `F11` to exit.

### Expected Behavior

When `F11` is press it should trigger the `exit-full-screen` event  from the  embedded content.

See:  https://github.com/electron/electron/issues/18188

![test-fullscreen](https://user-images.githubusercontent.com/14793624/58377220-85ba8380-7f38-11e9-8431-3001548fb761.gif)

PR #5874, opened 2021-04-13
Fullscreen hotkey fix

MaxKotlan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: 5865

CTRL+F is a commonly used hotkey for searching text on a page
LBRY currently has an "F" hotkey that will put the video player in fullscreen mode.
Unfortunately, it will activate fullscreen even if the user is trying to use the CTRL+F shortcut.
## What is the new behavior?
This PR makes it so the F hotkey only works if the control key/metakey are not also being pressed.
## Other information
Tested it on windows and it works. 
Haven't tested on OS X, but I think it should also work, since this also checks for the metakey (⌘ Command / ⊞ Windows )
and the shortcut is Command + F on OS X
[https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/ctrlKey](url)
[https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey](url)

### DIFF
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -81,7 +81,6 @@ const VIDEO_JS_OPTIONS: VideoJSOptions = {
   },
 };
 
-const F11_KEYCODE = 122;
 const SPACE_BAR_KEYCODE = 32;
 const SMALL_F_KEYCODE = 70;
 const SMALL_M_KEYCODE = 77;
@@ -185,7 +184,6 @@ export default React.memo<Props>(function VideoJs(props: Props) {
   } = props;
 
   const [reload, setReload] = useState('initial');
-
   const playerRef = useRef();
   const containerRef = useRef();
   const videoJsOptions = {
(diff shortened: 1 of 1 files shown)


infinite-persistence:
- It would be nice if we can clean up the shortcut-handling code for all single-key actions to ignore `ctrl`, `alt`, `cmd`, etc.
    - Currently, the issue will still happen with `Alt+F`, which is the shortcut to invoke Chrome's Menu.  That action still causes the video to go fullscreen.

MaxKotlan:
Good point. I'll add in the altkey and apply it to all single-key actions.

MaxKotlan:
@infinite-persistence 
I broke up the handleKeyDown function into a couple of new smaller functions:
_handleShiftKeyActions()_ and _handleSingleKeyActions()_.
Both handleShiftKeyActions() and handleSingleKeyActions() will exit if either alt, ctrl, or the meta key is pressed.
handleShiftKeyActions() will exit if shift is **not** being pressed
handleSingleKeyActions() will exit if shift is being pressed.

There is still some weird behavior with the hotkeys, but this behavior exists in the current version of lbry.

1. F11 hotkey is really buggy even though it uses the same code as the F hotkey. There seems to be a side effect that's also causing the video to go fullscreen (or something like that). 
2. The < and > hotkeys for video scrubbing require the user to press shift, even though there is nothing specified in the handleKeyDownFunction that would require that. The code for < and > is located in the exact same spot as the j and l hotkeys, and those work just fine without pressing shift. In fact, it's really confusing to me, because the code for all seeking actions is located in the handleSingleKeyActions() function, which disables shift key use, so I'm still not entirely sure how that is working. My guess is it's another side effect, but I'm just not sure. Maybe it's just the default player behavior.

In my testing, everything else was working normally, and the issues listed still exist in current lbry. This PR fixes the ctrl+f issue and it should have fixed any other issue related to other browser hotkeys.

MaxKotlan:
Actually never-mind on the 2nd issue. I was getting the speedup keys confused with my scrubbing keys. Everything should be fine.
Although I do wonder if there's a specific reason why the < and > hotkeys require the user to hold shift, while the rest of the hot keys don't. Is there a specific reason why the < and > hotkeys require the user to press shift?

infinite-persistence:
> _F11 hotkey is really buggy even though it uses the same code as the F hotkey. The F hotkey does not have this issue. There seems to be a side effect that's also causing the video to go fullscreen (or something like that)._

Yeah, F11 should be for "browser full screen" instead of "player full screen".  Seems like it was added to address #2514 [Can't exit full-screen from embedded content with key F11 ](https://github.com/lbryio/lbry-desktop/issues/2514), which looks like is already fixed in the version of Electron that we are using.  I think we can remove F11.

MaxKotlan:
After looking into the fullscreen issue a bit more... I realized that it was actually freaking out because the fullscreen of the video player was interfering with the fullscreen hotkey in chrome.
I added e.preventDefault(), which actually prevents chromes fullscreen behavior, but the video player enables handles enabling it.
Now the F hotkey and F11 work the same.

MaxKotlan:
Sorry, I realized that what I was doing was a bad idea. Removed the f11 hotkey instead like you said. F11 makes the browser go fullscreen, and f makes the video go fullscreen.

Is there anything else I need to do for this PR? Not sure what the contributing process is for this project.

seanyesmunt:
@MaxKotlan This is great! Can you just squash your commits? Then it's good to merge.

MaxKotlan:
@seanyesmunt Commits are now squashed!

Issue #5873, opened 2021-04-12
Unify swap and buy flow // swaps round #2


kauffj:
- Combine buy/swap into one flow
- Audit drivers to buy flow to incorporate above
- Swap history merged into transaction history (if not difficult)

tzarebczan:
Also:
- [ ] copy button on amount should only copy the amount, not the coin symbol

PR #5915, opened 2021-04-20
Swap: Don't copy the currency, only amount

infinite-persistence:
## Issue
Part of #5873[: Rounds 2 of LBC swaps](https://github.com/lbryio/lbry-desktop/issues/5873)

## Notes
It was an intended feature to include the currency -- I can paste the full string into my note book, while pasting into wallet apps like Exodus will automatically trim off the currency anyway.

Regardless, removed the 'feature' :D

## Approach
Updated `CopyableText` with `onCopy` prop that allows the client to override the full text selection.

### DIFF
--- a/ui/component/copyableText/view.jsx
+++ b/ui/component/copyableText/view.jsx
@@ -11,10 +11,11 @@ type Props = {
   label?: string,
   primaryButton?: boolean,
   name?: string,
+  onCopy?: (string) => string,
 };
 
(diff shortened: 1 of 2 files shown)


tzarebczan:
Thanks! I forget what app i was using, but it would not let me paste at all because of the string, so I found myself typing it in manually. Hoping this is where some of the underpayments are coming from.

Issue #5931, opened 2021-04-22
Fast actions will often be undone when there's a background update


infinite-persistence:
Impacted areas:

- [ ] **/$/settings/block_and_mute**
    - If you have a huge list of muted channels, and you try to unmute them one by one in that page, you'll occasionally see an entry that you've just unmuted became muted again (the button updated).
- [ ] **Manage Tags**
    - If you add or remove tags at a certain pace, you'll see tags appear/disappear (opposite of what you just did) on the screen. Hard to reproduce, but will happen.
- [ ] **Past Swaps**
    - Occasionally, you'll see an entry that you just removed 0.5s ago reappear because the wallet updated in the background.

There's potentially more.

tzarebczan:
Related: https://github.com/lbryio/lbry-desktop/issues/5904

tzarebczan:
This also happens when the first action you take is a button (i.e. follow) when clicking in a page, since that action triggers a sync + then the follow.

tzarebczan:
Noticed this creep up again when trying to change autoplay setting too quickly - it reverts the change and is poor UX.

keikari:
This also makes editing claim order in playlists quite annoying.

PR #5934, opened 2021-04-22
Muted Words Settings Page

### LINKED ISSUE #5833: Support creator controlled commenting settings (mainly muted word list for now)
See current list at https://github.com/lbryio/commentron/commit/1822ad2a34524746c98bc021160c44b4fc9412ef#diff-6f94b0243c2495a65018f234fb8fe042534b2804dea30e5631acef38f13ab902R4

- muted words
- min tip to comment (needs super chats)
- commenting enable/disable global 
- slow chat mode (need to add to commentron)

infinite-persistence:
## Demo
KP dev instance.

## Issue
Closes #5833 [Support creator controlled commenting settings (mainly muted word list for now) ](https://github.com/lbryio/lbry-desktop/issues/5833)

I believe only "Muted Words" is functional at this moment, as noted by the issue's title.
The rest, while set-able, has no effect, so we'll need to comment out the "General" and "Tip" cards before merging later.

## Notes to reviewer
- Double check the `fetchCommentsApi: Relay error messages to the client's 'catch' block.` commit.  This would fail if any clients are depending on `res.results` on the failure case, but I don't think there are any.
- I'm not really sure what are the units for "min amount" (LBC or satoshis?) and "slow mode gap" (seconds?) are.  Looking at how superchats are being used, I think they are satoshies (integer) and the GUI doesn't need to do any translation?

## GUI
- Added "Creator Settings" card in the Settings Page when user has a channel.
- "Creator Settings" currently only have the "Muted Words" card. 
- The toast will say `The comment contains contents that are blocked by %author%` when commenting with a blocked word.
- Tried to somewhat address `#5931 Fast actions will often be undone when there's a background update` by waiting for a 1s user-idle before syncing the GUI with the API data.

## API bugs
- [ ] [_Blocking_] Due to an empty string bug in the block list, the system will block every comment even if you cleared the blocked-words list.
    - There is a workaround:  create 2 or more blocked words, then delete 1 of them as the last action. Deleting all words will result in `[""]` instead of `[]`, causing the matching to always be true.
- [ ] `unable to marshall` error when changing any numerical setting.
- [x] Doesn't accept Unicode so it only works for English creators.
- [ ] There is a 3-character minimum for the blocked word. If we support Unicode, the limit won't make sense since bad words in Asian languages can be single-character.


infinite-persistence:
`Set to DRAFT`
- The ability to control other settings has been added, so I'd like to include those as well.
- There are new error messages that needs to be beautified ("minimum length of word is 3")

ui/component/app/view.jsx
```diff
@@ -249,6 +250,7 @@ function App(props: Props) {
 
     if (hasMyChannels) {
       fetchModBlockedList();
+      fetchCreatorSettings();
```
seanyesmunt, line 253: Does this need to be fetched on app startup? Could it just be fetched on the creator settings page?

ui/page/settingsCreator/view.jsx
```diff
@@ -0,0 +1,215 @@
...
+                  min={0}
+                  step={1}
+                  type="number"
+                  placeholder="1"
+                  value={minTipAmountComment}
+                  onChange={(e) => setSettings({ min_tip_amount_comment: e.target.value })}
+                />
+                <FormField
+                  name="min_tip_amount_super_chat"
+                  label={__('Minimum tip amount for superchats')}
```
seanyesmunt, line 199: ~superchats~ => `hyperchats`

ui/redux/actions/comments.js
```diff
@@ -309,6 +309,17 @@ export function doCommentCreate(
           toastMessage = __('Unable to comment. This channel has blocked you.');
         }
 
+        if (error) {
+          const BLOCKED_WORDS_ERR_MSG = 'the comment contents are blocked by';
+
+          if (error.message === 'channel is blocked by publisher') {
```
seanyesmunt, line 315: Now that `comment.Create` is going directly though commentron, can we access error codes instead of looking at the error message string?

ui/scss/component/_tags.scss
```diff
@@ -73,3 +73,10 @@
   margin: calc(2rem / 10) calc(2rem / 10);
   max-width: 20rem;
 }
+
+.tag--blocked-words {
+  .tag {
+    color: var(--color-text);
+    background-color: var(--color-gray-5);
```
seanyesmunt, line 80: These will need custom variables in light/dark.scss. Try a lighter gray like `--color-gray-2` for light mode.

seanyesmunt (commented):
This mostly looks good. A few comments. The styling looks a little off on the muted words.

infinite-persistence:
`UPDATES PER REVIEW`
- Removed settings-fetch at startup.
    - _I was thinking that we'll need to look at `comments_enabled` to put a reminder or something in the Comment section, but now I think  that's not necessary._
- Updated text to `hyperchats`.
- Fix "word tag" colors in both themes.
    - _I thought I've tested this, but now I recall that light mode was only tested in odysee, which the code is actually a bit different.  I think the Tag color for Desktop-Light is broken._

ui/redux/actions/comments.js
infinite-persistence: The code is the same for all errors `32000`, so I parsed the message instead.
```
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32000,
    "message": "words to block ? must be greater than length 3",
    "data": null
  },
  "id": 1
}
```
```
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32000,
    "message": "channel is blocked by publisher",
    "data": null
  },
  "id": 1
}
```
@tiger5226, is it possible to have unique codes?

jessopb: did you get an answer about error codes?

infinite-persistence: Haven't. No reply yet.

tiger5226: https://github.com/lbryio/lbry.go/issues/87 added an issue to the lbry.go package. The current errors package we use does not support error codes.

tzarebczan:
Tried to run locally and ran into this on startup: ```bundle.es.js?ea01:1516 Uncaught Error: Claim sequence must be a number.
    at parseURIModifier (bundle.es.js?ea01:1516)
    at parseURI (bundle.es.js?ea01:1418)
    at generateShareUrl (url.js?9533:101)
    at ClaimMenuList (view.jsx?ebf8:47)
    at renderWithHooks (react-dom.development.js?4b4a:14820)
    at mountIndeterminateComponent (react-dom.development.js?4b4a:17476)
    at beginWork (react-dom.development.js?4b4a:18625)
    at HTMLUnknownElement.callCallback (react-dom.development.js?4b4a:185)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?4b4a:237)
    at invokeGuardedCallback (react-dom.development.js?4b4a:290)```

Will blow out node modules and try again.

tzarebczan:
I see it's up on kp.odysee.com too, nevermind! Poked Mark about the commentron side fix, he'll get it up shortly.

tzarebczan:
Gave some feedback for Mark to fix the things you ran into also.  Comments should be enabled by default, gave him a heads up about that. 

- Add credits icon in min tip amounts
- hide control tags (this is under the experimental UI tag) - should probably be combined with the first setting - it disables comments directly on claim/channels based on the tag being set. We'd need to issue a channel update to do that.  Maybe one setting it a bit more temporary (turning off on comments server) and tag is more permanent. Not sure how to present this. The tag one would also disallow showing any previous comments across content, whereas the control setting only stops new ones. 
-for the comment tip setting, I think the intention was to force all comments to have tips in order to be posted - which would be an anti spam mechanism. That should be made clearer and ensure @tiger5226 enforces. That would apply to hyperchats as well. So that setting is more global and forces tips, whereas the other is a min when a tip is used. 

image

infinite-persistence:
@tzarebczan, this is ready for final review before shipping Muted Words.  Nothing new to test other than to look for corner-cases, I think.

## Changes/fixes:
- Hide the other features that are not ready.
- "Empty string" issue fixed in Commentron.
- Handle the case when bid is insufficient.
  - image

## Commentron Updates:
Feature not ready, but known issues fixed for now:
- `setting.CommentsEnabled` now should be defaulted to `true`.
- "unable to marshall" issue fiex

Issue #5933, opened 2021-04-22
Thumbnail lazy-load causes ChannelSelector icon to not update


infinite-persistence:
When you change the channel in a `ChannelSelector` (e.g. in Supports modal), the channel avatar no longer updates.

PR #5935, opened 2021-04-22
Disable thumbnail lazy-load for Channels

infinite-persistence:
## Issue
Quick fix for #5933 [Thumbnail lazy-load causes ChannelSelector icon to not update](https://github.com/lbryio/lbry-desktop/issues/5933)


seanyesmunt:
Nice find. Thanks

PR #5936, opened 2021-04-23
use redux for livestream claim setup

jessopb:
For livestreamsetup, rely on claim_search with --include_is_my_output to load the claims into myClaims, then select them out filtering for current channel, etc.


ui/page/livestreamSetup/view.jsx
```diff
@@ -104,112 +99,41 @@ export default function LivestreamSetupPage(props: Props) {
...
-  }, [activeChannelClaimStr, pendingLength, setSpin, checkLivestreams]);
+  }, [channelId, pendingLength, fetchNoSourceClaims]);
 
   return (
     <Page>
-      {(fetchingChannels || spin) && (
+      {fetchingChannels && (
         <div className="main--empty">
-          <Spinner delayed />
+          <Spinner />
```
seanyesmunt, line 136: These should still use `delayed`

ui/redux/actions/livestream.js
```diff
@@ -0,0 +1,33 @@
...
+import * as ACTIONS from 'constants/action_types';
+import { doClaimSearch } from 'lbry-redux';
+
+export const doFetchNoSourceClaims = (channelId: string) => async (dispatch: Dispatch, getState: GetState) => {
+  dispatch({
+    type: ACTIONS.FETCH_NO_SOURCE_CLAIMS_STARTED,
+    data: channelId,
+  });
+
+  const items = await dispatch(
```
seanyesmunt, line 11: This needs to be wrapped in a try/catch in case the claim_search fails

ui/redux/reducers/livestream.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+import * as ACTIONS from 'constants/action_types';
+import { handleActions } from 'util/redux-utils';
+
+const defaultState: LivestreamState = {
+  idsFetching: {},
```
seanyesmunt, line 6: `fetchingById`

seanyesmunt (changes requested):

seanyesmunt (approved):

Issue #5905, opened 2021-04-16
own comments show first in controversial / best


tzarebczan:
Seems confusing, especially when you made lots of comments.

PR #5939, opened 2021-04-24
own comments show first in controversial / best #5905

saltrafael:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #5905

### DIFF
--- a/ui/component/commentCreate/view.jsx
+++ b/ui/component/commentCreate/view.jsx
@@ -35,6 +35,7 @@ type Props = {
   toast: (string) => void,
   claimIsMine: boolean,
   sendTip: ({}, (any) => void, (any) => void) => void,
+  setjustCommented: (boolean) => void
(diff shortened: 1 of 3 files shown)


seanyesmunt:
I think this will need some additional work. With this solution, your comment may disappear immediately after posting it because it will go to the bottom of the list when using "Best" sorting.

I think ideally it would be added to the top of the results, then when you refresh it will be sorted properly

saltrafael:
> I think ideally it would be added to the top of the results, then when you refresh it will be sorted properly

How about this? The only problem now is if you made multiple comments they all move to the top at once, but once you refresh they all go to proper sorting, I will try to fix the multiple comments issue

saltrafael:
@seanyesmunt Check it out now

PR #5899, opened 2021-04-16
Re-add ability to export transactions 

infinite-persistence:
## Issue
Closes #4793 [Export Wallet History For Taxation Purposes](https://github.com/lbryio/lbry-desktop/issues/4793)

_Requires lbry-redux update_

## Screenshots
- Start:
    - image
- Fetching:
    - image
- In the unlikely-event that the fetched data couldn't be parsed:
    - image
- Fetched and file is ready:
    - image

PR #5950, opened 2021-04-26
(redux) doFetchTransactions: bump pageSize to 999999; remove doFetchSupport

infinite-persistence:
## Issue
Follow-up for #5899 Re-add ability to export transactions

## Note
I think this one was missed when merging in 5899. Oddly, I couldn't find the PR for the lbry-redux side in the Closed section. Either I forgot to create the PR, or something went wrong because I re-used an old branch name back then.


tzarebczan:
Jessop will be merging something with redux soon and will include this also.

PR #5969, opened 2021-04-28
Show and mark active live channels on homepage

### LINKED ISSUE #5841: Show active live channels on follow section + page
- make em stand out a bit more, move to front of list
- both following section on homepage and following page

infinite-persistence:
## Issue
Closes #5841: [Show active live channels on homepage + categories](https://github.com/lbryio/lbry-desktop/issues/5841)

## Screenshots
- Original:
    - image
- New:
    - image
- Extra (lbry tv):
    - image

## Discussion
1. I understand there is already an "eye" icon version of the live marker.  We can change to that if that's is more desired, but I prefer "Live" as it is less cluttering without the numerical value.

## Changes
See commit messages for walk through.

### DIFF
--- a/ui/effects/use-get-livestreams.js
+++ b/ui/effects/use-get-livestreams.js
@@ -1,0 +1,52 @@
+// @flow
+import React from 'react';
+import { BITWAVE_LIVE_API } from 'constants/livestream';
+
+/**
+ * Gets latest livestream info list. Returns null (instead of a blank object)
+ * when there are no active livestreams.
+ *
+ * @param refreshMs
+ * @returns {{livestreamMap: null, loading: boolean}}
+ */
+export default function useGetLivestreams(refreshMs: number) {
+  const [loading, setLoading] = React.useState(true);
+  const [livestreamMap, setLivestreamMap] = React.useState(null);
+
+  React.useEffect(() => {
+    function checkCurrentLivestreams() {
+      fetch(BITWAVE_LIVE_API)
+        .then((res) => res.json())
+        .then((res) => {
+          setLoading(false);
+          if (!res.data) {
+            setLivestreamMap(null);
+            return;
+          }
+
+          const livestreamMap = res.data.reduce((acc, curr) => {
+            acc[curr.claimId] = curr;
+            return acc;
+          }, {});
+
+          setLivestreamMap(livestreamMap);
+        })
+        .catch((err) => {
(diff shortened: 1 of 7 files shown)


ui/effects/use-get-livestreams.js
```diff
@@ -0,0 +1,52 @@
...
+      fetch(BITWAVE_LIVE_API)
+        .then((res) => res.json())
+        .then((res) => {
+          setLoading(false);
+          if (!res.data) {
+            setLivestreamMap(null);
+            return;
+          }
+
+          const livestreamMap = res.data.reduce((acc, curr) => {
```
seanyesmunt, line 27: Maybe this should have some limit so only livestreams with greater than x viewers are shown?

tzarebczan:
This works for the categories on the homepage, and each category too? 

Now that we are adding these, we could hive the livestream claims that aren't actually live (until we have the scheduling feature). That should definitely be the case on the category sections, but we can probably leave it on the followed section/page for now (since we'd need to filter out those claims, that may get trickier).

tzarebczan:
Actually, instead of filtering, we can bring back the --has_source to following list/page and then append the live followed ones similar to the other areas.

infinite-persistence:
> _This works for the categories on the homepage, and each category too?_
- Works for all categories as long as we're in the root (homepage):
    - image
- If you click on a category, no sorting or labelling is done. I figured it is confusing given it goes against the filters, plus there'll be an upcoming `/$/live` page anyway.

tzarebczan:
Hmm, it would be nice to get into categories - they can just go to the front of each, regardless of sort. I think we're still some time away from a dedicated live page.

infinite-persistence:
`Updates` -- _kp-instance updated as usual_
- `useGetLivestreams`: added "minimum viewers" filter.  But for now, still using `0 minimum viewers, no polling`.
- Filter out dormant livestreams from the homepage/categories.
- Extend the "append and sort active livestream" changes into category pages as well, essentially pinning livestreams to the top.  (However, I did see 1 odd instance where `claim_search` did not return 1 particular livestream when `Top` was selected)
    - image
    - image




`Notes`
- There will be a bunch of conflicts when merging into Odysee.  Fortunately, no conflict in "meaning", just conflict in code location -- so just take new changes from both sides.

kauffj:
:ship: :ship: :ship:

Issue #5985, opened 2021-04-30
Allow sending directly to a channel (or content?) address


kauffj:
Modify the `<WalletSend>` component to allow the user to enter a channel handle and/or LBRY URL to allow sending to, instead of just an address.

tzarebczan:
Do you want the SDK to support this or the desktop app to resolve. 

My only concern is for channels that are not yet claimed on YT sync, we don't have a process to transfer extra Lbc outside of tips.

saltrafael:
@tzarebczan should close this?

tzarebczan:
Yep, thanks for the poke!

PR #5990, opened 2021-05-01
allow sending directly to a channel or content address

saltrafael:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #5985

Just wallet address:

## What is the new behavior?

Wallet address + channel or content handle:

### DIFF
--- a/ui/page/send/view.jsx
+++ b/ui/page/send/view.jsx
@@ -103,7 +103,7 @@ export default function SendPage(props: Props) {
         setContentUri(``);
       }
     }
-  }, [enteredContent, setContentUri, setContentError, parseURI, isNameValid]);
+  }, [enteredContent, setContentUri, setContentError, contentError]);
 
   return (
     <Page
--- a/ui/page/send/view.jsx
+++ b/ui/page/send/view.jsx
@@ -11,9 +11,11 @@ type Props = {};
 export default function SendPage(props: Props) {
   const [isAddress, setIsAddress] = React.useState(true);
   const [contentUri, setContentUri] = React.useState('');
+  const [draftTransaction, setDraftTransaction] = React.useState({ address: '', amount: '' });
   const [enteredContent, setEnteredContentUri] = React.useState(undefined);
   const contentFirstRender = React.useRef(true);
   const [contentError, setContentError] = React.useState('');
+  const [confirmed, setConfirmed] = React.useState(false);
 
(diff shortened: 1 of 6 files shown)


jessopb:
Great feature! 
Check out the tip modal and maybe use two tabs: [ address ] [ lbry:// ] for the different modes,
Then it would be good to show a preview of the selected if you're in url mode as shown here:
https://odysee.com/$/repost?to=jsojsjjjjjj
Check out how this was done in https://github.com/lbryio/lbry-desktop/blob/master/ui/page/repost/view.jsx 

If you're feeling awesome, make pull down autocomplete suggestions :)

saltrafael:
@jessopb thanks for the feedback! something like this?

jessopb:
Great job - I think it looks a bit better with ClaimPreview type="small" in this case - it hides the tags and such. Will take a look at actual code today.

jessopb:
Nice.
 - it doesn't currently work yet - send stays grayed out.
 - It doesn't clear when switching between modes.
  - Might have to change the preview onClick so it doesn't navigate.
  - We are trying to move tab buttons to the card body rather than the title area (I know the tip modal isn't that way yet).
  - i found the input a bit buggy, either refusing to let me edit, etc
  - no error info is being shown during entry
  - feels weird to fill in part of the field, and click send, so the field is more of a search field than the actual entry. 
     -- maybe put a search field on top, the preview in the middle, and the amount and actual recipient url at the bottom?

saltrafael:
check it out now

jessopb:
I think I would use claim preview type small in the confirm too.
And the placeholder text is wrong.
Anything else you think has bad visual hierarchy or spacing, but looks pretty good.

saltrafael:
@jessopb What about this?



> Anything else you think has bad visual hierarchy or spacing, but looks pretty good.

I didn't understand this part

ui/page/send/view.jsx
```diff
@@ -103,7 +103,7 @@ export default function SendPage(props: Props) {
         setContentUri(``);
       }
     }
-  }, [enteredContent, setContentUri, setContentError, parseURI, isNameValid]);
```
tzarebczan: are you sure these two can be removed?

saltrafael: thanks, it was an accident

PR #5514, opened 2021-02-11
Add release date field

Ruk33:
## PR Checklist

Please check all that apply to this PR using "x":

- [x] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)
- [x] I have checked that this PR does not introduce a breaking change
- [ ] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting)
- [ ] Refactoring (no functional changes)
- [ ] Documentation changes
- [ ] Other - Please describe:

## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/2328

## What is the current behavior?

The release date is assigned automatically.

## What is the new behavior?

The user can pick a release date.

image

image

image

image

### NOTE
This PR requires https://github.com/lbryio/lbry-redux/pull/381 (change in lbry redux)

PR #6049, opened 2021-05-11
Continuation of "Add release date field #5514"

infinite-persistence:
## Original PR:
- #5514: [Add release date field ](https://github.com/lbryio/lbry-desktop/pull/5514)
    - I'm creating a new PR since I couldn't figure out how update someone else's fork (I know it's possible since Sean did update mine in the past).
    - The authorship of the original commits (squashed) is retained.

## Test
- kp.odysee.com

## Changes from original PR
- Upgrade to latest react-datetime-picker
- Handle dark theme and general style fixes.
- Fixed how "Edits" were being handled.  See accompanying PR in lbry-redux (https://github.com/lbryio/lbry-redux/pull/412)

## Notes
I think the calendar should only appear when the calendar button is clicked, and not when the numerical fields are clicked.  It's not easy to change without forking (`https://github.com/wojtekmaj/react-datetime-picker/issues/36`), so I'm leaving as is for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
     "imagesloaded": "^4.1.4",
     "json-loader": "^0.5.4",
     "lbry-format": "https://github.com/lbryio/lbry-format.git",
-    "lbry-redux": "lbryio/lbry-redux#6fc11454eb1b8f9e323c234669880721f12664ee",
+    "lbry-redux": "lbryio/lbry-redux#35dd7650fb235109d4a7ab44d1f6dde63dfdf2f7",
     "lbryinc": "lbryio/lbryinc#8f9a58bfc8312a65614fd7327661cdcc502c4e59",
     "lint-staged": "^7.0.2",
     "localforage": "^1.7.1",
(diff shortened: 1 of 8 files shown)


infinite-persistence:
> _This needs additional help text that shows when the release date is set in the future. Text is: "Setting a future timestamp does not guarantee your content is inaccessible". Assign PR to me to re-work text before shipping._

@kauffj, after talking to @tzarebczan, I ended up preventing "future" dates in the date picker.  Concerns includes:
- unclear how notifications would work with future dates (it should send immediately, we believe).
- need to tweak homepages and potentially other areas to do `< current time`.

kauffj:
Above works for me. @jessopb please review and merge.

tzarebczan:
I filed https://github.com/lbryio/lbry-desktop/issues/6054 for future content - we're accounting for it on apis on the initial publish, but not on edits (i.e. if you change the future date). The notification would come in at the time you set the release. 

Testing this out now otherwise...

PR #6051, opened 2021-05-12
Posts: restore "Tip unlock | Claim details" component

### LINKED ISSUE #5882: tip unlock + claim id detials missing from markdown posts view

infinite-persistence:
## Issue
Closes #5882: `tip unlock + claim id detials missing from markdown posts view`

## Test
kp.odysee.com

## Notes
The easiest solution would be to put `FileDescription` into posts, but I think that goes against the clean up of the Post layout, where the focus should be on the content. The faded style of the File Details section will be too distracting, plus we don't want the File Description anyway.

Fixed by:
- Make the existing "LBC amount" clickable to show credit details. An additional padlock will appear to the right if the content is yours and you have tips to unlock.
- Add an "info" icon to the left to show file details.

These "link" buttons are usually lit, but I dimmed it in this case to make them stand out less. Again, focusing on Post content instead of buttons.

image


tzarebczan:
This is great, thank you!

PR #6106, opened 2021-05-21
Additional pop up menu options 

### LINKED ISSUE #5782: Additional pop up menu options
On own claims:
- Edit (claims + channels)
- Delete claims
- View Channel Analytics

Others: 
- support
- (playlist stuff)

saltrafael:
## PR Checklist

## PR Type

- [X] Bugfix
- [X] Feature
- [ ] Documentation changes

## Fixes

Issue Number: Closes #5782

## What is the new behavior?

On own content:
 

## Other information

just missing playlist stuff for now

### DIFF
--- a/ui/component/claimMenuList/index.js
+++ b/ui/component/claimMenuList/index.js
@@ -2,35 +2,38 @@ import { connect } from 'react-redux';
 import {
   doCollectionEdit,
   makeSelectClaimForUri,
-  makeSelectClaimIsMine,
+  makeSelectFileInfoForUri,
+  doPrepareEdit,
   makeSelectCollectionForIdHasClaimUrl,
   makeSelectNameForCollectionId,
   makeSelectCollectionIsMine,
   COLLECTIONS_CONSTS,
 } from 'lbry-redux';
 import { makeSelectChannelIsMuted } from 'redux/selectors/blocked';
-import { doToggleMuteChannel } from 'redux/actions/blocked';
+import { doChannelMute, doChannelUnmute } from 'redux/actions/blocked';
+import { doSetActiveChannel, doSetIncognito, doOpenModal } from 'redux/actions/app';
 import { doCommentModBlock, doCommentModUnBlock } from 'redux/actions/comments';
 import { makeSelectChannelIsBlocked } from 'redux/selectors/comments';
-import { doOpenModal } from 'redux/actions/app';
 import { doToast } from 'redux/actions/notifications';
 import { makeSelectUserPropForProp } from 'redux/selectors/user';
-import ClaimPreview from './view';
 import * as USER from 'constants/user';
+import { makeSelectSigningIsMine } from 'redux/selectors/content';
+import { doChannelSubscribe, doChannelUnsubscribe } from 'redux/actions/subscriptions';
+import { makeSelectIsSubscribed } from 'redux/selectors/subscriptions';
+import ClaimPreview from './view';
+import fs from 'fs';
 
 const select = (state, props) => {
   const claim = makeSelectClaimForUri(props.uri)(state);
   const permanentUri = claim && claim.permanent_url;
   return {
     claim,
-    claimIsMine: props.channelIsMine
-      ? props.isRepost
-        ? makeSelectClaimIsMine(props.uri)(state)
-        : true
-      : makeSelectClaimIsMine(props.uri)(state),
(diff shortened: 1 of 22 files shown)


ui/component/claimMenuList/index.js
```diff
@@ -1,20 +1,41 @@
 import { connect } from 'react-redux';
-import { makeSelectClaimForUri, makeSelectClaimIsMine } from 'lbry-redux';
+import { makeSelectClaimForUri, selectMyActiveClaims, makeSelectFileInfoForUri, doPrepareEdit } from 'lbry-redux';
```
tzarebczan, line 2: I'm not sure we can use selectMyActiveClaims for this feature - this would require knowing all your claims, which we don't run anywhere on the web at least since it's an expensive operation, and could be lots of claims to store in redux. 

We should go back to using the "claim is mine" - this comes from the resolve/claim search result: ```      "is_my_output": true,
```

True means it's their claim.   This should work with both channels and claims.

saltrafael:
## Notable new changes:

-Reverted myActiveClaims

-Added some toasts for menu actions:



-Muting and Blocking toast will link to the /block_and_mute settings page, which is a way for the action to be undone since the content being muted or blocked disappears from the page:



-Switched up what Mute and Block messages display on toast, to clear confusion:

 

-Also changed up redux code a little to separate better Mute from Unmute, and Subscribe from Unsubscribe, like how it already works for Block and Unblock

ui/component/claimMenuList/index.js
saltrafael: yeah that makes sense, don't know why I thought it would be better

saltrafael:
### Explaining my final solution to claimIsMine:

I added a new selector called makeSelectSigningIsMine, which is similar to makeSelectClaimIsMine but instead it returns true when signing_channel is mine, which then solves the issue of not returning true on the search page:



And also prevents Reposts from returning true, as your claims, which would show on the menu pop up options like "edit" or "delete", and would break trying to edit or delete the reposted content

tzarebczan:
By search page do you mean the lighthouse results? We can fix that by adding the --include_is_my_output parameter to claim searches and resolves. Looks like we don't pass this in a few places that we should (i.e. homepages, top search result). 

I think that's the better way of doing it so it's consistent with the previous method we use for checking ownership. 

For claims that are your own, should also hide the mute/block button if it's not muted (allow unmute/unblock if it is).

saltrafael:
> I think that's the better way of doing it so it's consistent with the previous method we use for checking ownership.

Hmm I think so too, that one might be it then

> For claims that are your own, should also hide the mute/block button if it's not muted (allow unmute/unblock if it is).

Hide mute/block and show unmute/unblock? why not hide both? can you even mute/block your own claims?

tzarebczan:
Shouldn't allow to block/mute your own, but if someone did, should allow unmuting/blocking. 

Pushed on rebased version, before playlists. I can try to take a stab at rebasing on playlists later - unless you want to:? 

Some bugs: 
- channel analytics option doesn't load the channel on the analytics page (I see it in the URL, but loads my default).
- should probably hide options that are already visible on certain pages: 
image


Nice to have: 
- support reposts for deletion button (i.e. I have a repost on my channel page > I can menu > delete.

saltrafael:
> Shouldn't allow to block/mute your own, but if someone did, should allow unmuting/blocking.

✔️


> Pushed on rebased version, before playlists. I can try to take a stab at rebasing on playlists later - unless you want to:?

✔️Merged playlists changes

> Some bugs:
> 
> * channel analytics option doesn't load the channel on the analytics page (I see it in the URL, but loads my default).

✔️Fixed

> * should probably hide options that are already visible on certain pages:

✔️Now only shows options that are not already on screen


> 
> Nice to have:
> 
> * support reposts for deletion button (i.e. I have a repost on my channel page > I can menu > delete.

⌛ Added support for it but when trying to delete a repost (abandon claim) the claim only returns transaction_id for the original post, so the revoke fails. maybe could return --repost_transaction_id on claim resolve?

jessopb:
Looking better -
I think it would be better to use the query param that comes from the analytics button and use that to set the active channel on the dash component.
If you just want to ship what you have, strip the analytics item out. 
I don't think passing setActiveChanged into the SelectChannel component is the best way.

jessopb (approved):

Issue #5979, opened 2021-04-30
claim search runs twice sometimes and "refreshes" view


tzarebczan:
Been noticing this on the homepage and channel pages, but maybe happening on others too. This is on a refresh of the channel page: 
![image](https://user-images.githubusercontent.com/8120721/116705720-d8a9da80-a99a-11eb-9b81-66062477b638.png)

So after page load, you'll see a flicker and it reloads again. 

There's also a bug where if you scroll down to page 2 and back to page 1, then page 2 is shown on top. (not sure if related, but might be).

tzarebczan:
This happens after the block list is loaded. Should only happen on refresh/first load.

tzarebczan:
Other improvements here include:

- [ ] parallel processing of signing/comment server calls
- [ ] Only call for first channel? All channels should have the same block lists outside of a small percentage where accounts are merged + channels come from different wallets. We can eventually get them back in sync via some other tools on blocked/muted page.

infinite-persistence:
> _Only call for first channel? All channels should have the same block lists outside of a small percentage where accounts are merged + channels come from different wallets. We can eventually get them back in sync via some other tools on blocked/muted page._

The culprit seems to be `channel_list` consistently taking too long to return. `doFetchModBlockList` needs that data to proceed; meanwhile, the frontpage tiles are already resolved.

Any way to speed up `channel_list`❓

- ![image](https://user-images.githubusercontent.com/64950861/119636163-f3128f00-be46-11eb-896d-8cddbd873855.png)
<sub>_`channel_list` for a 10-channel account_</sub>

Side note:  given the current spammer mitigation, we can't just check the first channel -- we would have check until the first successful response.

----

I'm experimenting with `ClaimTileDiscover` using existing `uris` while waiting for `doClaimSearch` results.  At the expense of memory (stashing the uris), it seems to work ok:
- no re-draw or blinking if the results end up being the same
- minor tile movements when re-appending Livestreams or removing blocked content.

Even if we end up fixing `channel_list` to be faster, this would still be useful to reduce redraws.  Will test further.

PR #6123, opened 2021-05-26
Parallelize 'channel_sign' calls.

infinite-persistence:
## Issue
- The snackbar when blocking a channel can then take a while to appear.
- Was also suggested in [#5979 claim search runs twice sometimes and "refreshes" view ](https://github.com/lbryio/lbry-desktop/issues/5979), although it does not address the "twice" issue in the end.

## Before
- image

## After
- image
- 10-channel per account case (starts to depend on server speed, it seems):
  - image

### DIFF
--- a/ui/redux/actions/comments.js
+++ b/ui/redux/actions/comments.js
@@ -489,6 +489,22 @@ export function doCommentUpdate(comment_id: string, comment: string) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


tzarebczan:
This seems to be working as intended, but I wanna let @jessopb review. Can you fix the merge conflict in the meantime?

infinite-persistence:
`Update`
- Fixed conflict with latest master.

PR #6127, opened 2021-05-27
Fix broken Block List due to surfacing of error

infinite-persistence:
## Issue
The "Muted Words" PR elevated Comment-API errors so we can handle it (previously, it goes unnoticed). This broke the other Comment functions because the spammer mitigation might return an error, causing `Promise.all` to bail early.

## Fix
Changed to `Promise.allSettled` and reconstructed the results. This will make it equivalent to the code prior to the "Muted Words" PR.

## Potential issues
- [ ] ⚠️ `allSettled` is relatively new.  Safari, for example, requires version 13 (2019).  Should we polyfill?

----

@jessopb @tzarebczan, I had to bypass lint for this one.  Not sure how to add `allSettled` to lint.  Was trying to tweak the Bluebird file but didn't work -- not sure where else to tweak.

### DIFF
--- a/ui/redux/actions/comments.js
+++ b/ui/redux/actions/comments.js
@@ -537,7 +537,7 @@ export function doCommentModToggleBlock(channelUri: string, unblock: boolean = f
 
     const commentAction = unblock ? Comments.moderation_unblock : Comments.moderation_block;
 
-    return Promise.all(
+    return Promise.allSettled(
       channelSignatures.map((signatureData) =>
         commentAction({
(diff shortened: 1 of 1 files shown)


tzarebczan:
Good catch! What happens on older Safari if we don't polyfill? 

We'll be removing that spammer mitigation soon as it's not longer effective, but it's good to surface the errors regardless for better debugging.

tzarebczan:
Merging, plan to re-base on odysee tomorrow (so creator settings aren't live there yet).

Issue #5049, opened 2020-11-16
Move build process off of production server


seanyesmunt:
Currently the prod server is handicapped when building the app. Build the app on a separate server, then transfer the assets to the prod server

seanyesmunt:
This needs to be done soon

PR #6170, opened 2021-06-03
Clear out stale JS files during build

mayeaux:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation changes
- [X] Other - Please describe: Code improvement: deleting unused files that clog up disk

## Fixes

Issue Number:

Artifact JS files will be left after each build

## What is the new behavior?
Old JS files that aren't needed are deleted during build process. 

Could use some testing to make sure there are no bad side effects (aka if a build fails and there's no .JS file available which is the main possible bad scenario)

### DIFF
--- a/web/webpack.config.js
+++ b/web/webpack.config.js
@@ -75,6 +75,12 @@ if (fs.existsSync(ROBOTS_TXT_PATH)) {
   });
 }
 
+// clear the dist folder of existing js files before compilation
+let regex = /^.*\.(json|js|map)$/;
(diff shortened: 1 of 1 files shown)


neb-b:
Maybe this isn't the case with cloudflare caching but until https://github.com/lbryio/lbry-desktop/issues/5049 is resolved you might need to make sure that the most recent bundle.js is kept so that there is a js file to serve for users that go to the site for the first time while a new app is being built.

jessopb:
I think I'll just do this in bash in the rebuild script.

PR #6175, opened 2021-06-04
move pwa to web static, change desc

jessopb:


mayeaux:
Seems to work just fine

mayeaux (approved):

PR #6178, opened 2021-06-05
Don't reset content language on edit [+redux]

### LINKED ISSUE #6132: Don't reset content language to user language on edit
Publish page tries to set the upload language to the user langauge, and it does so on edit and overwrites the existing value.

saltrafael:
[Continuation of lbryio/lbry-redux/pull/414]

## Issue

Publish page tries to set the upload language to the user langauge, and it does so on edit and overwrites the existing value.
Closes #6132

## Change

selectPublishFormValues was trying to return "language" from formValues but would always return false, and set the language as default instead, should return "languages" instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
     "imagesloaded": "^4.1.4",
     "json-loader": "^0.5.4",
     "lbry-format": "https://github.com/lbryio/lbry-format.git",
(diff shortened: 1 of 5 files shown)


tzarebczan:
Please take a look at the redux comment. You'll want to change the main js file, and then run yarn build to generate the other.

infinite-persistence:
This is breaking `/$/upload` in master/lbry.tv.  Reverting for now.

PR #6183, opened 2021-06-08
Fix time to start metric

andybeletsky:
Time to start is multiplied by 1000, which was supposed to be a division if we're trying to get milliseconds, but metrics endpoint expects seconds anyway so there is no need for it.


jessopb:
will merge in a bit

Issue #6111, opened 2021-05-24
Channel avatars displayed in side menu


julianchandra:
Can we include the avatars next to the channel name?

![image](https://user-images.githubusercontent.com/67612037/119292633-e5bd9f00-bca4-11eb-8641-a5509e7935ae.png)

tzarebczan:
Duplicate of https://github.com/lbryio/lbry-desktop/issues/4512, closing so it's bumped up.

tzarebczan:
This is now live

PR #5712, opened 2021-03-19
Improve sidenav subscriptions appearance

saltrafael:
## PR Checklist

Please check all that apply to this PR using "x":

- [ ] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)
- [ ] I have checked that this PR does not introduce a breaking change
- [ ] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting)
- [ ] Refactoring (no functional changes)
- [ ] Documentation changes
- [ ] Other - Please describe:

## Fixes

Issue Number:

## What is the current behavior?

Currently the sidenav shows subscriptions only as their handle (@ lbry, @ odysee)

## What is the new behavior?

Changed it to show the channel thumbnail and title instead, to look better

PR #6193, opened 2021-06-09
Show channel thumbnails in side menu

mariosvlad:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/6111

Screenshot 2021-06-09 202653

## What is the new behavior?

Screenshot 2021-06-09 202453

## Other information

- The icons currently have size 1rem/16px but the thumbnails at that size seem too small to me so I made both 1.5rem/24px
- To batch-resolve the subscriptions (when they are needed), my implementation is a bit ugly so let me know if you have any recommendations on how to improve it


jessopb:
This looks good. 

image
It looks like on web or odysee mode, the channel level indicators are too big.

tzarebczan:
Hey @mariosvlad , thanks for taking a shot at this! Check out a previous attempt and feedback here: https://github.com/lbryio/lbry-desktop/pull/5712 too.

kauffj:
We discussed on the call today that the resolves needed to be batched, but it looks like this already does that? If so, then all that needs to happen is the channel level indicators shouldn't be shown and this can be merged.

mariosvlad:
This concern https://github.com/lbryio/lbry-desktop/pull/5712#issuecomment-815988748 that if you have a very large subscription list is still relevant with this implementation because on load it has to resolve all of them. It could be improved by only batch resolving the visible subscriptions (similar to image lazy load) but I'm not sure yet how this could be done (and how complex it would be)

jessopb (approved):
Nice

PR #6198, opened 2021-06-10
Handle comment + tip failure, don't re-support

### LINKED ISSUE #6063: Handle comment + tip failure, don't re-support
Right now if the comment fails to publish or validate the support, it will let you resend. When it does this, it re-supports again. We should save the result of the support (if successful) and send that with the comment on resubmit instead of trying to re-support.

saltrafael:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: Closes #6063

## What is the new behavior?

In case of failed comment, allows to re-submit the previous tip, without having to send another tip

_(tried to comment twice, first tip failed, re-submitted without tipping twice)_

### DIFF
--- a/ui/component/commentCreate/view.jsx
+++ b/ui/component/commentCreate/view.jsx
@@ -62,6 +62,8 @@ export function CommentCreate(props: Props) {
     location: { pathname },
   } = useHistory();
   const [isSubmitting, setIsSubmitting] = React.useState(false);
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks for the PR! Can test this today. Looks like linter is failing?

PR #6228, opened 2021-06-14
Code-split

infinite-persistence:
## Issue
Closes [#6150 Finish code-splitting](https://github.com/lbryio/lbry-desktop/issues/6150)

There's still some testing and bug fixes left, but I'd like a review now to see if this path will have major flaws.

## Test instance
https://kp.odysee.com

## Changes
- Split code (imports) into 2 main chunks + individual files.
  - `ui.js` reduced by 50%, allowing for faster download and earlier processing of the home page while the less-important components are being fetched.
  - Splitting code achieved 2 things:  (a) if the component is not mounted, then there's zero processing time used, (b) if the component will be mounted anyway, it will be delayed a bit to give priority to items in `ui.js` to load first.
- Lazy-load homepage categories.
- <strike>Don't block the homepage with a spinner while waiting for `user/me`.  Instead, put placeholders on the small number of sub-components that rely on that data.</strike>  Had to revert.

## Known issues
<strike>- 3rd-party CSS customizations broken
  - [x] Markdown editor
  - [x] Wunderbar-dropdown theme dark theme broken.  CSS-import order problem to override Reach's, I think.
  - [x] Videojs</strike>

## What's next:
- App:
  - Split CSS (next largest chunk)
  - Really lazy-load huge components, not just delaying it's import.
- Improve `claim_search`?


infinite-persistence:
- [ ] Ugh, "lazy loading homepage categories" seems like not working in the test instance, but ok locally.  Pointing to the same branch.

infinite-persistence:
`Update`
- Appended hash to the chunk files.
- Rebased to master to grab fixes for `WaitUntilOnPage`
- Re-do "category lazy-loading" -- added placeholders so the layout don't change much (better visually when scrolling).

CSS issues remain...

infinite-persistence:
**Update** (567e96c to 974e075):
- Reverted the 'user/me' bypass.  If a `resolve` was called without this, we currently don't re-resolve those claims after 'user/me' is retrieved, so the GUI is messed up.

infinite-persistence:
@DispatchCommit, do you have any good ideas on how to solve the CSS problem?  Not lazy-loading those components with 3rd-party libraries (videojs, simplemde, etc.) would be a quick fix, but it defeats the purpose of the PR as those are the exact items that takes the longest to init during boot up.

Currently looking at https://github.com/webpack-contrib/mini-css-extract-plugin, which will break apart the css to 1-per-bundle (I think).  Seems like things will get messier.

I'm considering dropping this until a better approach is found.

neb-b:
@infinite-persistence Does it work if you move the mde/video.js css imports to the top level app component? Not the best solution because those are still really big but that might fix it

infinite-persistence:
> _Does it work if you move the mde/video.js css imports to the top level app component? Not the best solution because those are still really big but that might fix it_

That worked for ReachUI and Videojs.  Thanks a lot for the tip!
Still had to revert lazy-loading MDE, though.  It's mixing mde and codemirror -- just couldn't get the order right.

----
Update:
- Add quick-fix for CSS until we revisit splitting CSS properly.
- Reverted lazy-loading SimpleMDE.  Gained back half a meg.

infinite-persistence:
## Current score
- Desktop improved from ~27 to ~50:
  - 

- Mobile didn't budge.  Should be due to slow `claim_search` affecting LCP.
    - .
      - The tile is the _Largest Contentful Paint_, and it is blocked by the slow `claim_search`.  In Desktop, there are other things to satisfy LCP.
        - 
      - Some javascript/css is blocking for mobile (_Total Blocking Time_).

Issue #6237, opened 2021-06-14
Analytics button from Channels page doesn't open analytics to right channel


tzarebczan:
It must not be reading the url paramaters properly.

PR #6276, opened 2021-06-18
Set 'activeChannel' before entering Dashboard

infinite-persistence:
## Issue
Closes [#6237: Analytics button from Channels page doesn't open analytics to right channel](https://github.com/lbryio/lbry-desktop/issues/6237)

## Notes
I assume we have changed to the `activeChannel` method, so we are dropping the direct URL support.  Saves us the need to update the URL when user changes from the selector after entering the page.

### DIFF
--- a/ui/page/channels/index.js
+++ b/ui/page/channels/index.js
@@ -5,18 +5,20 @@ import {
   doFetchChannelListMine,
   selectFetchingMyChannels,
 } from 'lbry-redux';
+import { doSetActiveChannel } from 'redux/actions/app';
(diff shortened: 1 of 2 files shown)


tzarebczan:
Will this work for the analytics button from the Accounts > channels page if it's not active?

infinite-persistence:
Yeah, because of `hideAnon` in `<ChannelSelector hideAnon`, it'll force a channel to be selected if none was.  
For no channels at all, it'll show yrbl.

PR #6284, opened 2021-06-19
Fix videos not switching when floating player is up

### LINKED ISSUE #6278: Video doesn't switch when floating player is up
In the video below, it's supposed to go back to the wrestling video on the second click, but it got stuck with the interview video:
https://lbry.tv/@ip-test-1:4/video-no-switch:2

Reverting eff948f7864aa7b16960db8798c0f46c5f7cde66 seems to fix it.

infinite-persistence:
## Issue
Closes #6278 Video doesn't switch when floating player is up

## Notes
Not sure why the double `src` call is needed, but it is.


tzarebczan:
Confirmed it fixes the issue + doesn't reintroduce the 20 second delay.

PR #6288, opened 2021-06-19
fix incorrect variable being used to determine view state for live view count

DispatchCommit:
`isLive` is being checked instead of `livestream`.


DispatchCommit:
#shipit

PR #6294, opened 2021-06-21
change tagline

drewhancock:


DispatchCommit (changes requested):
Tagline needs to be updated in `app-strings.json` as well. 

https://github.com/lbryio/lbry-desktop/blob/153185ecc553d94f883ff8c4402a97d4a9f3d34a/static/app-strings.json#L1684

jessopb:
I think, going forward, we'll just do {'this cave is not a natural formation'} instead of including the __(...).  There's no reason to submit a short lived popular english quote to translation.

PR #6303, opened 2021-06-21
test odysee theme

jessopb:
Proposing solution to give odysee and other brands per-platform themes.


jessopb:
We could change "themes" to /scss, then put themes inside that, and any other components that need custom scss can go in /scss too.

infinite-persistence (approved):
Looks good :+1: 

> _We could change "themes" to /scss, then put themes inside that, and any other components that need custom scss can go in /scss too._

I vote for this as well (flatter hierarchy), although either is fine.

Thinking out loud (non-issues, really):
- The merge-conflicts in the old world had the tiny benefit of alerting the merger that some colors need to be updated in Odysee (e.g. a variable was added -- need new custom color, otherwise will inherit Desktop's).  Tried to think of a way to retain this benefit when we briefly discussed this idea previously, but no solution.  Overthinking, I guess -- the changer will just need to pass on the info.
- Any additional `ms` penalty on startup for the double import? (feels insignificant from testing so far)
- `isOdysee` --> `IS_ODYSEE` or `BRAND='odysee'`?

PR #6306, opened 2021-06-22
Remove kiyui as maintainer for AUR package

kcseb:
Kiyui has made it evident that they're no longer on Arch and outright removed themselves as a maintainer for the AUR package.
See: https://aur.archlinux.org/packages/lbry-app-bin#comment-813900

Best wishes to em' in their journies!


DispatchCommit:
Best of luck Kiyui!

Issue #6311, opened 2021-06-22
autoplay blocked with infinite loading bug


tzarebczan:
Not sure if something changed on our end or in FF recently, but autoplay on + first interaction ends up in this state: 
![image](https://user-images.githubusercontent.com/8120721/122955927-8f4c8a80-d34e-11eb-8fe3-ae5b0ec15bef.png)

https://odysee.com/@StruggleNation:4/one-punch-man-1x12-the-strongest-hero:f

tzarebczan:
Happening on Brave too (incognito) and IOS devices apparently also.

PR #6313, opened 2021-06-22
fix autoplay for transcoded files

DispatchCommit:
Fixes issue where shouldPlay becomes undefined and will break autoplay if video is transcoded due to a new load request being started.

image


DispatchCommit:
Fixes #6311

Issue #6233, opened 2021-06-14
restore subtitles/CC button when transcoder fix is in


tzarebczan:
![image](https://user-images.githubusercontent.com/8120721/121912882-ee7f1f00-ccfe-11eb-86ca-603a9ddc0a7c.png)

This is on https://odysee.com/@OdyseeCafe:5/ODYSEE-LIVESTREAM-01:c

DispatchCommit:
This is the expected behavior given the metadata we receive from the playlist file of transcoded files. The player team will need to correct the m3u8 files and add appropriate metadata to properly fix this.

https://developer.apple.com/library/archive/qa/qa1801/_index.html
> iOS 7 now supports the TYPE attribute of the EXT-X-MEDIA tag having a value of CLOSED-CAPTIONS. This allows for explicit declaration of closed caption content in the playlist file. See the HTTP Live Streaming Draft Protocol Specification for the details.
iOS 7 also supports a CLOSED-CAPTIONS attribute for the EXT-X-STREAM-INF tag. You can declare the absence of closed caption content in the playlist by specifying CLOSED-CAPTIONS=NONE on the EXT-X-STREAM-INF tag. The "Unknown CC" option as shown in Figure 1 will not be displayed in the selection when the absence of closed caption content is declared.
Also, if the media does not contain alternate audio or subtitles, and the absence of closed caption content is declared, the alternate track button will not be displayed.

jessopb:
Restore button when above transcoder issue is closed.

PR #6315, opened 2021-06-22
Don't display CC button via videojs settings

mayeaux:
## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: #6233 

Errant CC button shows up on iOS Safari

## What is the new behavior?
Errant CC button no longer shows up

## Other information
Address issue:
https://github.com/lbryio/lbry-desktop/issues/6233

I can confirm the CC button began to show up only on more recent versions of master. May warrant further investigation as to what caused that to start appearing, this is a working hotfix to stop users from being confused.

### DIFF
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -200,6 +200,11 @@ export default React.memo<Props>(function VideoJs(props: Props) {
       eventTracking: true,
       overlay: OVERLAY.OVERLAY_DATA,
     },
+    // fixes problem of errant CC button showing up on iOS
+    // TODO: find the source of why the button recently began to show up on iOS
+    controlBar: {
+      subsCapsButton: false,
+    },
   };
 
   const tapToUnmuteRef = useRef();


DispatchCommit:
This will disable the control bar button for closed captioning, so as a hotfix it's okay, but doesn't seem like the "correct" way.

DispatchCommit:
Does the CC button only appear on files that are transcoded?

DispatchCommit:
To me this feels like we might end up just hiding an underlying issue and never fixing the real reason.

DispatchCommit:
This is an issue with metadata on transcoded files. 

From apple's dev docs: https://developer.apple.com/library/archive/qa/qa1801/_index.html
> iOS 7 now supports the TYPE attribute of the EXT-X-MEDIA tag having a value of CLOSED-CAPTIONS. This allows for explicit declaration of closed caption content in the playlist file. See the HTTP Live Streaming Draft Protocol Specification for the details.
> iOS 7 also supports a CLOSED-CAPTIONS attribute for the EXT-X-STREAM-INF tag. You can declare the absence of closed caption content in the playlist by specifying CLOSED-CAPTIONS=NONE on the EXT-X-STREAM-INF tag. The "Unknown CC" option as shown in Figure 1 will not be displayed in the selection when the absence of closed caption content is declared.
> Also, if the media does not contain alternate audio or subtitles, and the absence of closed caption content is declared, the alternate track button will not be displayed.

PR #6329, opened 2021-06-24
FIX video.js event firing issues for RecsysPlugin

jbn:
- The `rateChange` event now logs the updated speed,
   not just the    time at which it occurred.
- The `scrub` now (more) accurately logs the position
   it came from before the destination.
- The recsys events get consolidated for logical consistency.


jessopb (approved):
Makes sense.
Curious about the comment on 242 about // The problem.

PR #6342, opened 2021-06-28
RecommendedContent: prevent excessive render

infinite-persistence:
## Issue
When seeking on a video, RecommendedContent gets re-rendered, which causes the 20 ClaimPreviews to be re-rendered, several times. ClaimPreview has lots of children, plus it's constantly checking against the blacklist, among other things.

## Fix
The culprit seems to be the `recommendedContent` array -- the contents are actually the same.
Do a deep compare instead.

## Questions
- Is there a more elegant way to deep-compare?
- For `claim`, is comparing the `claim_id` sufficient?
- Not really sure why `SET_CONTENT_POSITION` would end up marking `makeSelectRecommendedContentForUri`.

## Results
Circled item indicates React activity when seeking a video. Reduced from (24ms * N) to (8ms).
- Before:
  - 
- After:
  -

### DIFF
--- a/ui/component/recommendedContent/view.jsx
+++ b/ui/component/recommendedContent/view.jsx
@@ -23,7 +23,7 @@ type Props = {
   claim: ?StreamClaim,
 };
 
-export default function RecommendedContent(props: Props) {
+export default React.memo<Props>(function RecommendedContent(props: Props) {
   const {
     uri,
     doFetchRecommendedContent,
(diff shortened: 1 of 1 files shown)


jessopb:
I think comparing claim_id should be sufficient.

jessopb (approved):
Comparing claim_id should be sufficient, because I think that's all recommended cares about at the moment. Later with johnny's work we'll have to compare the list given?

DispatchCommit:
> When seeking on a video, RecommendedContent gets re-rendered

I feel like there should be no connection here... What variable is seeking updating that is triggering the re-render?

infinite-persistence:
@DispatchCommit , I'm puzzled too.  Here's my guess:

- `SET_CONTENT_POSITION` causes a `storeState` change:
   - 
- Selectors then re-evaluate
  - But `makeSelectRecommendedContentForUri` shouldn't depend on that, though.  Does it get re-calculated regardless?
- Down the chart, props with nested objects will be considered changed due to shallow compare.  All children also get marked unnecessarily.
  - 
  - Comments is another thing to tackle.  Lots of re-renders, not just due to seeking.
  - 

Are we doing too much in selectors to make the references different each time?  Was thinking of simplifying `makeSelectRecommendedContentForUri` to just return the search results, and do the adjustments as a component state, but given upcoming changes from Johnny (not sure how things will change), doing a deep-compare seems like the cleanest quick-fix.

infinite-persistence:
## Close
Xander has better fix

infinite-persistence:
## Re-open
Took a look at Xander's fix.  While it's probaqbly the right way to sort things, it didn't addressed the problem of the selector returning a new reference each time.  Given that Johnny's update might change things in the near future, I think the deep-compare fix is a reasonable one to apply for now.

PR #6348, opened 2021-06-29
Additions to Odysee rebase

infinite-persistence:
- Hardcoding colors for now.  Open to suggestions on how to get it programmatically.  I tried `getComputedStyle`, but that seems inefficient for something simple.
- Fixed compile error (but not sure why it worked on Alice).


infinite-persistence:
`UPDATE`
- Removed the hardcoded change to avoid conflict.  We'll get the colors from env automatically when rebased to master
- [ ] Add `LOADING_BAR_COLOR=#e50054` to Odysee env.

PR #6358, opened 2021-07-01
Tech Debt: selectors/search improvements

DispatchCommit:
What if... we tried to reduce some tech debt and add more type declarations to functions and variables?

### help wanted 😢

(specifically with resolving issues in related files due to missing types and values, this is just a first pass improvement)


jessopb:
I'm tempted to put types in flow-typed as their own files as we have been doing, but I can see the value of having the types in front of your face - otoh, that's what a smart IDE should show you.
I'm Not doing anything with this this week.

kauffj:
I heard we're moving to typescript :imp:

Issue #5161, opened 2020-12-10
Links to comments don't always work


infinite-persistence:
- Normally, links to comments will bring you directly to that comment.  
- However, for this one, https://odysee.com/@lbry:3f/odysee:7?lc=29f439137dbdbc662453fe8cfd27969ed716627ddf08da6c0ebeb50ba34a86f9, it does not.  It always go to the pinned comment instead. Scroll down to find it (it is still highlighted correctly).  
- The link to the comment right above this one works, though.  And other deeper replies in the upper pages works as well.
- So, I'm not sure what the faulty pattern is.  Maybe something like "if the comment is on the 3rd reply level onwards, _and_ is on "page 2" (infinite scroll) onwards in the comments section, the link does work"?

tzarebczan:
I've also noticed that videos autoplay sometimes when clicking from notifications - should disable that too. 

Thanks for filing!

infinite-persistence:
## Close
This has been addressed while doing Comment Pagination.

If we end up reverting again in the future, the fix for this in the "fetch all" world is just to find the comment beyond the second level of replies.

PR #6390, opened 2021-07-05
Comments Pagination

### LINKED ISSUE #6158: Support Comment Pagination
Currently we have numerous claims with more comments than the max supported page size (200) of the comment server. The UI when the comments were added avoided implementing the pagination. As a work around the UI team had made the page size 99,000 comments. Once we added live streams this became a massive load as thousands of people made the call with thousands of comments. We had to limit the server temporarily. 

We continue to try to work around this but what we really need is for the API to be used correctly for proper scalability. 

The `comment.List` API has the following paramters:

```golang
// ListArgs arguments for the comment.List rpc call
type ListArgs struct {
	ClaimID       *string `json:"claim_id"`
	AuthorClaimID *string `json:"author_claim_id"`
	ParentID      *string `json:"parent_id"`
	Page          int     `json:"page"`
	PageSize      int     `json:"page_size"`
	TopLevel      bool    `json:"top_level"`
	Hidden        bool    `json:"hidden"`
}
```

`Page` and `PageSize` can be used for pagination. The default order of comments is by timestamp, descending. We can add alternative sorting options to the API as well for taking care of different lists of comments. For example, there is a section for "Controversial". This is a good example of an extension to the API.

infinite-persistence:
## Status
- Test instance: https://kp.odysee.com
- Good to go when we get Tom's seal of approval.
  - While I've tested a lot, it's still a big change (affected many files), so a second look by someone else will be good before we deploy

## Known issues
- [ ] Livestream remains unpaginated.  Wait for Cursor Pagination in v3.
- [x] <strike>Pinned comments is broken.</strike> Fixed
- [x] <strike>Linked comments doesn't work right if nested or yet to be fetched.</strike> Fixed
  - The existing version suffers the same problem too (see #5161).

## Rough summary of changes
- Top level comments are paginated using infinite-scroll style.
- Replies are paginated using a "Show more" button.
- The existing "expand all replies" behavior is possible, but disabled for now, except for linked-comments.
- Reactions:  now paginated as well, instead of always re-fetching for the entire list of comments.
  - When switching active channels, you might see like-counts temporarily reset if the fetch is slow.

## Issue
Closes #6158 - Support Comment Pagination


tzarebczan:
https://kp.odysee.com/@Odysee:8?view=discussion is a good large discussion to test on. Some findings:
1) When on large monitor, it shows a couple comments and the loading icon, but not actually loading anything: 
image

2) this shows 2 replies, but there's only 1. When you expand, you see the 1 + loading icon:
image

Other feedback
It's annoying to scroll in large lists like above, especially when you do it quickly - it seems like it only loads a few things, and then keeps showing loading icon for more. Scrolling slower seems to fair better.  

Seems pretty solid otherwise, even on mobile!

infinite-persistence:
## Update
6c80acd to 4731588
- Tell user when linked-comment is not available.
- Load comments if spinner is visible from the start (it was previously only responding to scroll events)
- Load comments when approaching viewport
  - This approach works, but currently, the issue in 6473 negates the effect at times.
- Bumped comment and reply page size to 10 (was 5 and 3, respectively)

Test instance updated as well.

----

> _this shows 2 replies, but there's only 1. When you expand, you see the 1 + loading icon:_

Yeah, saw this too.  Asking Mark if it's possible for `replies` to reflect the final listing.

PR #6392, opened 2021-07-05
i18n - fiat

infinite-persistence:
## Changes
- Put strings in localization macro
- I think we were moving away from "camel case" (e.g. "Connect A Bank Account" --> "Connect a bank account"), so I have that changed.
- Fix concatenated strings (hard to translate)

## ??
- There are strings hardcoded to "Odysee". Leaving as is for now, should probably be a variable.

----

@kauffj, assigning to you for review since you were gonna update the strings.  If it ends up with lots of delta, just close this one and I'll re-open a new one with your changes.


kauffj:
I don't want to merge this and cause extra work for @mayeaux to reconcile. Anthony, please bring this into your ongoing work.

mayeaux (approved):

mayeaux:
We can merge this and I can pull it into my current workflow without any issue.

PR #6396, opened 2021-07-05
Fix RSS icon not having a gap

infinite-persistence:
@jessopb, is this the right fix?  What puzzles me is that if `buildIcon` is needed for props to be forwarded, then shouldn't all icons use it to avoid surprises?

image


DispatchCommit (commented):
Does this not remove the ability to resize the icon?

jessopb:
It looks like buildIcon hsould take care of that. I didn't notice buildIcon.

PR #6402, opened 2021-07-06
upgrade video.js to 7.13.3

DispatchCommit:
Video.js has updates


tzarebczan:
Package.json should also be updated, and yarn file?

jessopb:
yarn.lock is there, but package.json is not. also rebase pls.

PR #6411, opened 2021-07-07
Handle stale/missing chunks

infinite-persistence:
@jessopb, if the solution is acceptable, I'll rework the entire to repo to change from `React.lazy` to `lazyImport`  
Currently, I only applied it to `Settings` and `TipModal` to test the concept.

Solutions that I know of:
1. Reload on behalf of user (this PR). 
2. Similar to 1, but instead of reloading directly, we display a message somewhere saying "A new version is available. Please refresh".
3. Run a `serviceWorker` to listen for changes, and somehow update the current page (this is beyond my understanding to implement at the moment).

It seems like either approach would require a reload, so number 1 seems the easiest.

There are a few caveats, though:
1. It works seamlessly for Pages (other than the user seeing a brief flash from the reload), but it's a bit odd for smaller lazy-loaded components like the Tip Modal -- it doesn't open the modal after the reload.
2. Reloading directly can result in data-loss if the user in the middle of something.

### DIFF
--- a/static/index-web.html
+++ b/static/index-web.html
@@ -3,6 +3,8 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="Pragma" content="no-cache" />
+    <meta http-equiv="Cache-Control" content="no-cache, no-store, must-revalidate" />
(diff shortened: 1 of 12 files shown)


jessopb:
I think this path sounds good.

PR #6460, opened 2021-07-13
Fix "exact match" being applied to Recommended

### LINKED ISSUE #6414: Don't use "exact match" setting outside of main search - cuases no results in related
exact match setting from search is passed along to related queries, which will produce no results in most cases.

infinite-persistence:
## Issue
Closes #6414 <Don't use "exact match" setting outside of main search - cuases no results in related>

`doSearch` was used in Wunderbar and Recommended, but it was adding in Wunderbar options all the time.

## Changes
- Fix `doSearch` to not use hardcoded options (take from arguments).
- Refactors the searchOption-creation code so that we don't duplicate the logic in both .js and .jsx.

### DIFF
--- a/ui/page/search/index.js
+++ b/ui/page/search/index.js
@@ -1,43 +1,42 @@
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router';
 import { doSearch } from 'redux/actions/search';
-import { SIMPLE_SITE } from 'config';
 import {
   selectIsSearching,
   makeSelectSearchUris,
-  makeSelectQueryWithOptions,
   selectSearchOptions,
   makeSelectHasReachedMaxResultsLength,
 } from 'redux/selectors/search';
 import { selectShowMatureContent } from 'redux/selectors/settings';
 import { selectUserVerifiedEmail } from 'redux/selectors/user';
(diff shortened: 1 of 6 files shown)


ui/page/search/index.js
```diff
@@ -5,39 +5,39 @@ import { SIMPLE_SITE } from 'config';
...
-      ? { nsfw: false, isBackgroundSearch: false }
-      : showMature === false
-      ? { nsfw: false, isBackgroundSearch: false }
-      : { isBackgroundSearch: false }
-  )(state);
 
+  const searchOptions = {
+    ...selectSearchOptions(state),
+    isBackgroundSearch: false,
+    nsfw: SIMPLE_SITE ? false : showMature,
```
jessopb, line 28: the selectShowMature selector already accounts for SIMPLE_SITE. :)
```
export const selectShowMatureContent = !ENABLE_MATURE
  ? createSelector(() => false)
  : makeSelectClientSetting(SETTINGS.SHOW_MATURE);
  ```

jessopb (approved):

infinite-persistence:
### Update (1c5f99d to b822ae3)
- Removed redundant SIMPLE_SITE check.  Thanks, Jessop!

PR #6504, opened 2021-07-16
Optimize channel banner

infinite-persistence:
_Channel banners can hit several MBs._

## Target
Optimize the banner filesize, but don't compromise on the quality (i.e. never show blurry banner)

## Outcome
1. Visually, the banner is as sharp as the version before this PR, with the exception of the following scenarios:
    - If you load the page in a smaller zoom and then zoomed out, it will be potentially blur if you go from one extreme to the other. A reload is required to get a sharp image in the new zoom setting.
      - It's the same in YT's banner.  When cache is disabled, YT seems to listen to window-resizing and grabs an updated image.  I don't think we need to do that.
    - Depending on the format (mostly on JPG), images with sharp edges will be slightly aliased due to the compression. But this is usually not noticeable unless you look really close.
      - Same as YT.  General compression behavior, I guess.
2. On a 1080p screen in various zoom settings, on average we save about 10-40% in image size.

### DIFF
--- a/ui/component/optimizedImage/index.js
+++ b/ui/component/optimizedImage/index.js
@@ -1,0 +1,2 @@
+import OptimizedImage from './view';
+export default OptimizedImage;
--- a/ui/component/channelThumbnail/view.jsx
+++ b/ui/component/channelThumbnail/view.jsx
@@ -5,12 +5,7 @@ import classnames from 'classnames';
 import Gerbil from './gerbil.png';
 import FreezeframeWrapper from 'component/fileThumbnail/FreezeframeWrapper';
 import ChannelStakedIndicator from 'component/channelStakedIndicator';
-import { getThumbnailCdnUrl } from 'util/thumbnail';
-
-const FONT_PX = 16.0;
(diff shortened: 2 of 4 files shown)


infinite-persistence:
Set to `Draft`
- Found a bug when applying it on channel thumbnails.

infinite-persistence:
`Ready for review`
- Skip optimization for files hosted in `/public`, like Gerbil. If we want to optimize those, we'll need to provide the full URL in the code, otherwise the CDN lookup will fail.

ui/component/optimizedImage/index.js
```diff
@@ -0,0 +1,6 @@
+import { connect } from 'react-redux';
+import OptimizedImage from './view';
+
+const perform = (dispatch) => ({});
+
+export default connect(null, perform)(OptimizedImage);
```
jessopb, line 6: connect has nothing to do here.
index.js  can just import and export the default from ./view
Though, I sometimes do this if I expect to add some selectors or actions later.

jessopb (approved):

PR #6507, opened 2021-07-16
RSS: Add enclosure for media

### LINKED ISSUE #6369: RSS feed issues and improvements
- [x] figure out what the total number of items we want to return. Right now it's hardcoded to the last 50. Maybe have a lastest 50 and latest 1K feed? 
- [x] received complaints that RSS readers can't download the content. We can try adding a .mp4 and/or ?download=1 to the media URL : https://cdn.lbryplayer.xyz/api/v4/streams/free/this-hardcore-fps-is-actually-awesome/0325d87d94c858b26089a3ff0dc48a1843d5fcef/5c0f5d.mp4?download=1 
- [x] shorten urls? https://odysee.com/$/rss/@GivesAMinute/2e4c63a39e28e8e2bcac74debece9f7893fc3fe5

infinite-persistence:
RSS: Add enclosure for media

## Issue
Closes #6369

> _received complaints that RSS readers can't download the content. We can try adding a .mp4 and/or ?download=1 to the media URL_

## Comments
Not really sure if I'm addressing the issue, as I couldn't find a reader that downloads meida within a feed item.  It sounded like they want a feed that links to the direct download, rather than the page?

Anyway, comparing with feeds from other sites, it seems like putting the media into the 'enclosure' tag should be sufficient.  The enclosure provides the direct link to the CDN.

## Code Changes
- Moved the thumbnail from 'enclosure' to 'description' by appending the thumbnail in front.
- Generate enclosure for media. It's up to readers on how they will handle it. Some will display the inline video player, some will just provide a download link.

## Sample
- Feedly
    - 
- QuiteRss
    -

### DIFF
--- a/web/src/rss.js
+++ b/web/src/rss.js
@@ -1,3 +1,4 @@
+const { generateDownloadUrl } = require('../../ui/util/web');
 const { URL, SITE_NAME, LBRY_WEB_API } = require('../../config.js');
 const { Lbry } = require('lbry-redux');
 const Feed = require('feed').Feed;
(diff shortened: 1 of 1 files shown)


kauffj:
https://lbryians.slack.com/archives/C1R07H02X/p1626446034243500

^ That doesn't need to block merging this, but should be addressed so we don't lose view counts

PR #6557, opened 2021-07-19
Fix comment create redirection

### LINKED ISSUE #5321: Comments section redirects to create channel
When the Odysee web application uploads a post from server resources and clicks the comment section, the channel creation section opens. This happens when you are still loading libraries and resources. I took a screenshot to represent the error. This error has not appeared on lbry.tv at the moment, only on Odysee. I can provide more details if necessary.

image

saltrafael:
## PR Checklist

- [X] I added a line describing my change to CHANGELOG.md

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Documentation changes

## Fixes

Issue Number: Closes #5321


tzarebczan:
Will test this out, but wanted to confirm it fixes the issue where the channel list may still be loading (looks like some of that code was removed). 

The main bug was that it would tell you to create a channel because the call wasn't done loading.

infinite-persistence (changes requested):
> _Will test this out, but wanted to confirm it fixes the issue where the channel list may still be loading (looks like some of that code was removed)._
>
> _The main bug was that it would tell you to create a channel because the call wasn't done loading._

This scenario isn't covered.  If `channel_list` stalls, the code incorrectly thinks the user has no channels.  It should probably wait until `isFetchingChannels` is done before enabling the section.

saltrafael:
I reverted it to the current way but waiting for `isFetchingChannels ` now

PR #6581, opened 2021-07-20
Change API server address to odysee-api

andybeletsky:


jessopb:
Need to update some explicit .envs on production

Issue #6068, opened 2021-05-14
Fix "Cumulative Layout Shift" for Core Web Vitals


infinite-persistence:
<sub>Related: [#6057 Investigate low Core Web Vitals score from google](https://github.com/lbryio/lbry-desktop/issues/6057)</sub>

I just realized that while CLS has a lesser impact in Lighthouse scores, it's the top 3 metric for Core Web Vitals.  The results in Lighthouse showed green for CLS (so I thought we were ok in this department), but Core Web Vitals says we are failing.

![image](https://user-images.githubusercontent.com/64950861/122735443-9ec9c600-d2b1-11eb-8f73-cb6434545444.png)

## Items
- [x] Main Page
  - Fixed during initial code-split work.
- [ ] Post page (0.5++)
- [ ] Video page (0.3++)

PR #6612, opened 2021-07-21
Improve CLS on Posts

infinite-persistence:
## Issue
Closes [#6068 Fix "Cumulative Layout Shift" for Core Web Vitals](https://github.com/lbryio/lbry-desktop/issues/6068)

In Posts, the Comments section appears first while we fetch the MD file. When the MD is fetched, Comments get pushed to the bottom (or shifted up for short posts), hence the red CLS scores.

## Approach
There are too many layers between `<FilePage>` and `<DocumentViewer>` to pass the `loading` state around to hide the Comments section, so just make Comments fade in after a 2s delay.

## Changes
- Posts: Add 2s fade-in delay to Comments.
- Posts: remove the gray placeholder. This improves the score a bit more, and reduces flicker as well. There's already a spinner from `FileRenderInline` to tell the user to be patient.
- Posts: add a minimum 30vh height so that short posts don't get collapsed too much, causing the `FileDetails` and Comments to shift. Small shifts are fine as long as CLS is below 0.1.

## Result
CLS back to green.

### DIFF
--- a/ui/component/postViewer/view.jsx
+++ b/ui/component/postViewer/view.jsx
@@ -104,8 +104,10 @@ function PostViewer(props: Props) {
 
       <ClaimAuthor uri={uri} />
 
-      <FileRenderInitiator uri={uri} />
-      <FileRenderInline uri={uri} />
+      <div className="file-render--post-container">
+        <FileRenderInitiator uri={uri} />
+        <FileRenderInline uri={uri} />
+      </div>
       <FileActions uri={uri} />
     </div>
   );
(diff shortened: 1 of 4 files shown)


ui/component/postViewer/view.jsx
```diff
@@ -104,8 +104,10 @@ function PostViewer(props: Props) {
 
       <ClaimAuthor uri={uri} />
 
-      <FileRenderInitiator uri={uri} />
-      <FileRenderInline uri={uri} />
+      <div className="file-render--post-container">
+        {<FileRenderInitiator uri={uri} />}
```
jessopb, line 108: whatis this enclosure doing?

infinite-persistence: Enforcing a minimum height.  The 2 childs have sub-childs that come and go, so added an enclosure instead of tweaking every child's css

jessopb (approved):

PR #6692, opened 2021-07-26
Logo component

jessopb:


jessopb:
This handles header, embed, embed-ended cases for logos. on lbrytv and odysee.

infinite-persistence (approved):
Tested on mobile and desktop. Looks good.

- [ ] `OptimizeImg` doesn't work -- file size not optimized (my bad).  We can use plain `` for now until I figure it out.  Sorry.

Non blockers:
- The wunderbar shifts after the Odysee logo is fetched.  The CLS penalty is surprisingly low (0.004) despite so obvious visually, while things shifting behind the scenes have high penalty (weird).  Anyway, it would be nice if we can allocate the space for the logo.  I tried it this morning, but couldn't get it work without blowing up the image size.
- When you enter http://localhost:9090/$/embed/matrix-vs.-xmpp-which-is-better-for/9f0caf336db6332084574a17b5df6392edce36d0?r=Ajpwc7LAwte8Zj5D59ogTm1sYVPUgWGn and while it is loading, the logo is super huge.

PR #6694, opened 2021-07-26
Option to change comments-server (desktop)

### LINKED ISSUE #5459: Interact directly with commenting servers
- [x] Drop SDK calls (SDK is dropping `comment_xxx` commands)
- [x] Replace with direct calls to server matching `commentron` APIs
- [ ] Add setting for changing your comment server. Visible on desktop (and possibly defaulting to Odysee URL), hidden on odysee.
- [ ] Revise comment onboarding on desktop.

infinite-persistence:
## Issue
Closes #5459
> _Add setting for changing your comment server. Visible on desktop (and possibly defaulting to Odysee URL), hidden on odysee._

## Changes
- Added Desktop-only options to define custom server. [Settings > Advanced Settings > "Comment server" section].

## Comments
Not sure how this would actually work properly without the user recompiling the app to handle server differences. For example, even when we use our own server but switch between v1 and v2, some code changes are need to handle the differences. At that point, it seems easier for the user to just change the .env file? 

But I guess it is still useful as a quick way to switch servers when there are no API interface issues.


infinite-persistence:
fabbcee to dce101c
- Rebased to `master` to resolve string conflicts

ui/comments.js
```diff
@@ -31,8 +38,10 @@ const Comments = {
 };
 
 function fetchCommentsApi(method: string, params: {}) {
-  if (!Comments.enabled) {
-    return Promise.reject('Comments are not currently enabled'); // eslint-disable-line
+  if (!Comments.url) {
+    return Promise.reject(new Error('Comments-server URL not defined.'));
```
kauffj, line 42: Commenting server is not set.

ui/component/settingCommentsServer/view.jsx
```diff
@@ -0,0 +1,69 @@
...
+    return () => clearTimeout(timer);
+  }, [customUrl, setCustomServerUrl]);
+
+  return (
+    <React.Fragment>
+      <fieldset-section>
+        <FormField
+          type="radio"
+          name="use_default_comments_server"
+          label={__('Use official Odysee comments server')}
```
kauffj, line 34: `__('Default comments server (%DEFAULT_COMMENT_SERVER_NAME%)', { DEFAULT_COMMENT_SERVER_NAME })`

with the default server name moved to a config

kauffj: alternatively, add an array config value for all known servers, and make this a select with custom as the last option

ui/component/settingCommentsServer/view.jsx
```diff
@@ -0,0 +1,69 @@
...
+          onChange={(e) => {
+            if (e.target.checked) {
+              setCustomServerEnabled(false);
+            }
+          }}
+        />
+        <FormField
+          type="radio"
+          name="use_custom_comments_server"
+          label={__('Use custom comments server')}
```
kauffj, line 45: Custom comments server

(drop use)

ui/redux/actions/comments.js
```diff
@@ -96,20 +98,32 @@ export function doCommentList(
...
-            },
-          });
-        } else {
-          devToast(dispatch, `doCommentList: ${error.message}`);
-          dispatch({
-            type: ACTIONS.COMMENT_LIST_FAILED,
-            data: error,
-          });
+        switch (error.message) {
+          case 'comments are disabled by the creator':
```
kauffj, line 102: where is this coming from? can/should it be a const?

kauffj (changes requested):
I do not need to review this after these changes, just ship it.

ui/redux/actions/comments.js
infinite-persistence: :+1: Yup, recently cleaned up to be constants in `master`.

PR #6711, opened 2021-07-27
Notifications improve links

### LINKED ISSUE #5062: React/reply to comments from notifications
Would be nice to react (fire/slim/creator heart) and reply to comments directly from the notifications page

### LINKED ISSUE #5095: Allow right click menu on notifications page
So you can open content in new tabs

saltrafael:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Documentation changes

## Fixes

Issue Number: Closes #5095

## What is the new behavior?

- Notification links improved
- Now can right click on notification items, open in new tabs, open other users channel, etc.
- Added reactions to comments:
[](https://postimg.cc/D8BxG13w)

### DIFF
--- a/ui/component/notification/view.jsx
+++ b/ui/component/notification/view.jsx
@@ -16,6 +16,8 @@ import FileThumbnail from 'component/fileThumbnail';
 import { Menu, MenuList, MenuButton, MenuItem } from '@reach/menu-button';
 import NotificationContentChannelMenu from 'component/notificationContentChannelMenu';
 import LbcMessage from 'component/common/lbc-message';
+import UriIndicator from 'component/uriIndicator';
+import { NavLink } from 'react-router-dom';
 
 type Props = {
   notification: WebNotification,
@@ -29,13 +31,12 @@ export default function Notification(props: Props) {
   const { notification, menuButton = false, doReadNotifications, doDeleteNotification } = props;
   const { push } = useHistory();
   const { notification_rule, notification_parameters, is_read, id } = notification;
+
   const isCommentNotification =
     notification_rule === RULE.COMMENT ||
     notification_rule === RULE.COMMENT_REPLY ||
     notification_rule === RULE.CREATOR_COMMENT;
   const commentText = isCommentNotification && notification_parameters.dynamic.comment;
-  const channelUrl =
-    (notification_rule === RULE.NEW_CONTENT && notification.notification_parameters.dynamic.channel_url) || '';
 
   let notificationTarget;
   switch (notification_rule) {
@@ -51,21 +52,14 @@ export default function Notification(props: Props) {
       notificationTarget = notification_parameters.device.target;
   }
 
-  let notificationLink = formatLbryUrlForWeb(notificationTarget);
-  let urlParams = new URLSearchParams();
-  if (isCommentNotification && notification_parameters.dynamic.hash) {
-    urlParams.append('lc', notification_parameters.dynamic.hash);
-  }
-
-  try {
-    const { isChannel } = parseURI(notificationTarget);
-    if (isChannel) {
-      urlParams.append(PAGE_VIEW_QUERY, DISCUSSION_PAGE);
-    }
-  } catch (e) {}
-
-  notificationLink += `?${urlParams.toString()}`;
-
+  const creatorIcon = (channelUrl) => {
+    return (
+      <UriIndicator uri={channelUrl} link>
+        <ChannelThumbnail small uri={channelUrl} />
+      </UriIndicator>
+    );
+  };
+  let channelUrl;
   let icon;
   switch (notification_rule) {
     case RULE.CREATOR_SUBSCRIBER:
@@ -73,16 +67,20 @@ export default function Notification(props: Props) {
       break;
     case RULE.COMMENT:
     case RULE.CREATOR_COMMENT:
-      icon = <ChannelThumbnail small uri={notification_parameters.dynamic.comment_author} />;
+      channelUrl = notification_parameters.dynamic.comment_author;
+      icon = creatorIcon(channelUrl);
       break;
     case RULE.COMMENT_REPLY:
-      icon = <ChannelThumbnail small uri={notification_parameters.dynamic.reply_author} />;
+      channelUrl = notification_parameters.dynamic.reply_author;
+      icon = creatorIcon(channelUrl);
       break;
     case RULE.NEW_CONTENT:
-      icon = <ChannelThumbnail small uri={notification_parameters.dynamic.channel_url} />;
+      channelUrl = notification_parameters.dynamic.channel_url;
+      icon = creatorIcon(channelUrl);
       break;
     case RULE.NEW_LIVESTREAM:
-      icon = <ChannelThumbnail small uri={notification_parameters.dynamic.channel_url} />;
+      channelUrl = notification_parameters.dynamic.channel_url;
+      icon = creatorIcon(channelUrl);
       break;
     case RULE.DAILY_WATCH_AVAILABLE:
     case RULE.DAILY_WATCH_REMIND:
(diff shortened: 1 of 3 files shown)


ui/component/notification/view.jsx
```diff
@@ -140,35 +175,61 @@ export default function Notification(props: Props) {
...
+
+              {isCommentNotification && (
+                <div className="notification__reactions">
+                  <Button
+                    label={__('Reply')}
+                    className="comment__action"
+                    onClick={handleCommentReply}
+                    icon={ICONS.REPLY}
+                  />
+                  <CommentReactions uri={notificationTarget} commentId={notification_parameters.dynamic.hash} />
```
infinite-persistence, line 217: `CommentReactions` will show 0 reactions each time unless you've visited the comment before.
One suggestion is to collect a list of comment IDs at one level up and fetch all reactions in one shot.

ui/component/notification/view.jsx
```diff
@@ -57,6 +62,24 @@ export default function Notification(props: Props) {
...
+  });
+  const title = titleSplit.map((message) => {
+    if (creatorName === message) {
+      return (
+      <UriIndicator uri={notification_parameters.dynamic.comment_author} link />
+      );
+    } else {
+      return <LbcMessage>{__(' %message% ', { message })}</LbcMessage>;
+    }
+  });
```
infinite-persistence, line 81: @jessopb, I believe this block is trying replace the channel name with something that's clickable.  I recall this issue also affects livestream chat and maybe other places.  Do you have a recommendation on how to address this issue in general?

The above breaks up the string.  This will break any hopes of localizing the notifications page in the future.
  - 
  - 

Maybe something like how `LbcMessage` does the replacement will be a cleaner approach?

But anyway, since Commentron doesn't support a i18n-friendly format yet, this solution is still better than current state.

infinite-persistence (changes requested):
I noticed late that this is probably still work-in-progress, but left a few comments anyway.
Nice work!

ui/component/notification/view.jsx
infinite-persistence: Or maybe something like this, which also fixes the i18n problem:
```
  const commentedOnClaimMsg = (
    <I18nMessage
      tokens={{
        comment_author: <UriIndicator uri={notification_parameters.dynamic.comment_author} link />,
        claim_title: notification_parameters.dynamic.claim_title,
      }}
    >
      %comment_author% commented on %claim_title%
    </I18nMessage>
  );
```

sapioit:
Thanks for adding this feature! It makes it possible to open all the new-video notifications in new tabs, and binge-watch them all.

kauffj:
Is this ready or are you still working on it @saltrafael?

saltrafael:
> Is this ready or are you still working on it @saltrafael?

I had messaged infinite-persistence to see the latest changes, then it just needs rebase and fix Reply !

ui/component/notification/view.jsx
```diff
@@ -97,6 +107,52 @@ export default function Notification(props: Props) {
...
+    } else {
+      fullTitle.push(message);
+
+      if (index === (titleSplit.length - 1)) {
+        return (
+          <LbcMessage>{fullTitle.join(' ')}</LbcMessage>
+        );
+      }
+    }
+  });
```
infinite-persistence, line 141: Feels like can be improved, but will pass for now so as not to stall the PR further.

With a little more work, I think we can avoid the unnecessary loop and also enable localization at the same time using something like the following (incomplete example).  It'll also prevent the need to put a blanket `LbcMessage` for notifications that don't need it. We can also customize the phrase like adding quotes to the title for readability.

```
function generateTitle(rule, params) {
    switch (rule) {
      case RULE.COMMENT:
        return (
          <I18nMessage
            tokens={{
              commenter: <UriIndicator uri={notification_parameters.dynamic.comment_author} link />,
              title: notification_parameters.dynamic.claim_title,
            }}
          >
            %commenter% commented on "%title%"
          </I18nMessage>
        );

      case RULE.CREATOR_SUBSCRIBER:
        // TODO
        break;
      case RULE.CREATOR_COMMENT:
        // TODO
        break;
      case RULE.COMMENT_REPLY:
        // TODO
        break;
      case RULE.NEW_CONTENT:
      case RULE.NEW_LIVESTREAM:
      case RULE.DAILY_WATCH_AVAILABLE:
      case RULE.DAILY_WATCH_REMIND:
      case RULE.MISSED_OUT:
      case RULE.REWARDS_APPROVAL_PROMPT:
      case RULE.FIAT_TIP:
      default:
        // Use Commentron default
        return params.device.title;
    }
  }

  const title = generateTitle(notification_rule, notification_parameters);

```

ui/component/notification/index.js
```diff
@@ -1,8 +1,19 @@
 import { connect } from 'react-redux';
```
infinite-persistence, line 1: image

- [ ] When I click on this kind of notification, I'll go to the intended page, but it immediately sends me back to the Notification page.

infinite-persistence, line 1: image

The navlink only works when clicked on text.  If you middle-click on empty areas, it doesn't work.
`ClaimPreview` solves this by adding a wrapper to the blank areas. 

Not super critical -- can postpone.

infinite-persistence (changes requested):

kauffj:
@saltrafael if you've adjusted a PR in response to feedback, please click the icon to re-request a review (I just did this, so you won't see it)

ui/component/notification/index.js
```diff
@@ -1,8 +1,10 @@
...
```
infinite-persistence, line 1: When a "comment notification" is clicked, it navigates to the claim page but also reloads the site, causing the cache to be cleared and re-authenticated. Should avoid a reload if possible.

infinite-persistence, line 1: After leaving a comment reply directly from the notification component, the reply is not immediately visible. This will be confusing, and I think a toast wouldn't cut it either.  I understand it is troublesome to place the reply component there with current code.

infinite-persistence (changes requested):

ui/component/notification/index.js
saltrafael: [](https://postimg.cc/FfXSnRwh)

like this?

infinite-persistence (changes requested):
Only did some minor testing -- there are still issues.  Also, as noted in the comments, I wonder if there's a better solution than to place "stop propagation" all over the place, and tweaking `form-field` needs more testing to ensure it doesn't break all clients.

Handling the comment direct-reply and reaction seems tricky.  Also, with an upcoming push "min-tips" push, we'll have to take more things into account.
   - **Suggestion**:  perhaps we can limit this PR to just the "channel link/allow navigation", and move the "comment"-related parts into a new PR.  It'll be easier to test/track/revert (not lumping multiple items).

saltrafael:
> Only did some minor testing -- there are still issues. Also, as noted in the comments, I wonder if there's a better solution than to place "stop propagation" all over the place, and tweaking `form-field` needs more testing to ensure it doesn't break all clients.

✔️ Ended  "stop propagation" hell

> Handling the comment direct-reply and reaction seems tricky. Also, with an upcoming push "min-tips" push, we'll have to take more things into account.
> * **Suggestion**:  perhaps we can limit this PR to just the "channel link/allow navigation", and move the "comment"-related parts into a new PR.  It'll be easier to test/track/revert (not lumping multiple items).

I hid the support option from direct-reply which was causing the most issues, (also looked confusing, since replying to a comment sends a HyperChat to the content and not the reply). See if it's still needed to take it out

tzarebczan:
@saltrafael can you rebase and fix the conflict and we'll take another look soon.

ui/component/notification/index.js
```diff
@@ -1,8 +1,14 @@
...
```
infinite-persistence, line 1: When changing the active channel via "Set replying as", we'll need to re-fetch the reactions to get the right numbers (mine vs others).  Currently, it resets to zero.

infinite-persistence, line 1: Second level reply is a bit weird. The GUI remains stuck at the following, with SuperChat appearing.  When I pressed Cancel, I noticed a ton of comment was sent.

image

infinite-persistence, line 1: When I press the Creator Like (Heart), a medium-sized avatar gets stuck in the screen.

image

infinite-persistence (changes requested):
Few more issues unfix/uncovered.

I'm still in favor of splitting out [#5062 React/reply to comments from notifications ](https://github.com/lbryio/lbry-desktop/issues/5062) into another PR so that we can ship your navigation link changes first. 

The comment-replies doesn't seem straightforward to implement. YT does it by moving into a virtual "sub-page" that is basically a `CommentsList` -- I think that totally avoids all the issues and hacks that we've been trying to do in embedding the `Comment` component within `Notification`.

ui/component/notification/index.js
tzarebczan: Noticed this too, probably minor but should be fixed.

tzarebczan: I ran into this but wasn't sure what happened - but now this makes sense.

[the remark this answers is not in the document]
tzarebczan: That's a known bug on master - we should not be sending that notification to already approved users.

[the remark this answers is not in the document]
saltrafael: > That's a known bug on master - we should not be sending that notification to already approved users.

So that explains it 😅
Peek 2021-08-18 14-37

PR #6722, opened 2021-07-29
Fill in remaining Recsys fields

### LINKED ISSUE #6366: Recsys Evaluation Telemetry
See: https://github.com/lbryio/clickstream/issues/2

infinite-persistence:
## Test
- Available for testing in kp.odysee.com
    - 

## Known issues
- **Right-click events are not captured.**
    - I can detect the right-click event but I cannot get the selected value.  The only solution I know is that we'll need to roll our own context-menu, which I don't think we want to do it that way.  
    - The other alternative is to just consider right-clicks as left-clicks, but would that pollute the telemetry?
    - Status:
        - ✓ Capture left-click
        - ✓ Capture ctrl-click (open in new tab)
        - ✓ Capture shift-click (open in new window)
        - ✓ Capture middle/wheel-click (open in new tab)
        - ❌ Capture right-click --> open in new tab
 
## Ticket
Closes #6366 Recsys Evaluation Telemetry

The recommended list from lighthouse is obtained from `makeSelectRecommendedContentForUri`. This list is further tweaked by the GUI (e.g. move autoplay next item to top, remove blocked content, etc.). Recsys wants the final recommendation list and the clicked index (in exact order), so we need pass these info to the videojs recsys plugin somehow. Also, Recsys wants a recommendation list ID as well as the parent (referrer) ID, we so need to track the clicks and navigation.

## General Approach
- It seems easiest to just spew the final (displayed) list and all the required info back to Redux; then,the recsys plugin (or anyone else in the future) can grab it.
- Try to touch few files as possible. The dirty work should all reside in `<RecommendedContent>` only.

## Changes
- `ClaimPreview`: add optional parameters to store an ID of the container that it is in (for this case, it is `ClaimList`) as well as the index within the container.
  - When clicked, we store the container ID in the navigation history `state` object.
    - For general cases, anyone can check this state from  `history.location.state` to know which container referred/navigated to the current page. For the recsys use case, we can use this as the `parentUUID`.
- `ClaimList`: just relay `onClick` and set IDs.
- `RecommendedContent`: now handles the uuid generation (for both parent and child) and stores the data in Redux.


infinite-persistence:
#### 6ca42a6 to b79f66b
- Fixed `parentUUID` not populated when the empty space on `ClaimPreview` is clicked instead of at the text.
- Rebased to `master`

jessopb:
We have some desktop-side checks for things that are 'analytics-y'
https://github.com/lbryio/lbry-desktop/blob/9dd065a59548951c167dd96be6ce59468f03c6ac/ui/analytics.js#L72
Need to make sure we have a way to respect these?

infinite-persistence:
b79f66b to 8a68ec0 

#### Previous:
~~Although it was collecting info for all platforms, the data is only sent for Odysee.~~ 
_Correction_: it became enabled for all platform after `master` became Odysee.

#### New:
Only send out the telemetry for `IS_WEB` or if there's user consent.

jbn:
I don't know enough react to be able to review it competently, but as far as verifying that it does want I want when looking at the network traffic console, it LGTM.

jessopb (approved):

PR #6757, opened 2021-08-02
Fallback image for <FileThumbnail>

### LINKED ISSUE #5457: Create file thumbnail fallback image for odysee
We just display a grey box for thumbnails that error currently. We should have some placeholder thumbnial.

Screen Shot 2021-02-01 at 1 49 57 PM

infinite-persistence:
## Ticket
Closes #5457 Create file thumbnail fallback image for odysee

## Approach
<strike>Since `background-image` does not invoke an `onerror` event, use the stacking attribute of `background-image` to display the fallback.</strike>

## Todo
- [x] Jessop, please help to add this in Odysee env
```
MISSING_THUMB_DEFAULT=https://spee.ch/missing-thumb-png:3.png
```

## Screenshot
- e.g. Blocked a specific image from dev tools:
  -


infinite-persistence:
### Back to draft
Hm ... the fallback will often appear first when scrolling due to caching.  It's pretty annoying.
Finding another solution.

jessopb:
I did add the .env fallback to prod.

infinite-persistence:
d97a26f 
@jessopb, how about this one.  Too expensive?

jessopb:
https://www.npmjs.com/package/lozad should we try this?

infinite-persistence:
> _https://www.npmjs.com/package/lozad should we try this?_

@jessopb, I took a look at that, and it doesn't provide any on-error fallback capability (there's a request for that). They are pretty much discussing the same issue we are facing ... in 2018 lol.

Seems like we can't avoid moving from `background-image` to `` (to get the `onError` functionality) and mess with all the css to make it behave the same as existing?  Not sure if the new `object-fit` behaves exactly the same.

infinite-persistence:
### d97a26f to 9481777 (version 3 - ready for review)
After looking at various plugins, seems like we can't avoid using a test `Image`.  However, we can avoid the silly timer.

jessopb (approved):

PR #6488, opened 2021-07-15
Mobile view: Make comments expandable

saltrafael:
## PR Checklist

Please check all that apply to this PR using "x":

- [X] I have checked that this PR is not a duplicate of an existing PR (open, closed or merged)
- [X] I have checked that this PR does not introduce a breaking change
- [ ] This PR introduces breaking changes and I have provided a detailed explanation below

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting)
- [ ] Refactoring (no functional changes)
- [ ] Documentation changes
- [ ] Other - Please describe:

## Fixes

Issue Number: Closes #6230

## What is the current behavior?

Big comment sections makes it hard to reach Recommended/Playlist tabs

## What is the new behavior?

At first comments show contracted, and new pages can be loaded or the whole thread can be collapsed:

[](https://postimg.cc/YGKW6bSV) [](https://postimg.cc/Y4npkS5x)

PR #6778, opened 2021-08-04
Revert "Mobile view: Make comments expandable"

jessopb:
Reverts lbryio/lbry-desktop#6488


jessopb:
Reverting because of missing view. Will add again later.

Issue #6420, opened 2021-07-07
Pinned Updates in SuperChats


tiger5226:
Commentron now sends a message over websocket when a creator pins a comment. This can be used to pin it in the super chat.

PR #6822, opened 2021-08-09
Livestream: implement Pinned Comments

infinite-persistence:
## Issue
Fixes [#6420 Pinned Updates in SuperChats ](https://github.com/lbryio/lbry-desktop/issues/6420)

## Behavior
The top (latest) pinned comment will be displayed at the top of the livestream chat.  Only 1 pinned message will be displayed (same as other platforms).  When a comment is unpinned, the next pinned comment (if exists) will replace it's spot.

## Approach
- Livestream comments are retrieved from 2 ways:
    - From initial entry, `comment.List` is used to retrieved the chat.  Due to the recent pagination changes, the pinned comments will always be placed at the "top" (or in the chat GUI, the "bottom").
    - Subsequent messages will come via websocket. 
- We handle both cases by storing the pinned comment IDs in the reducer stage.

## Test
kp.odysee.com

### DIFF
--- a/flow-typed/Comment.js
+++ b/flow-typed/Comment.js
@@ -38,6 +38,7 @@ declare type CommentsState = {
   topLevelTotalCommentsById: { [string]: number }, // ClaimID -> total top level comments in commentron.
   commentById: { [string]: Comment },
   linkedCommentAncestors: { [string]: Array<string> }, // {"linkedCommentId": ["parentId", "grandParentId", ...]}
+  pinnedCommentsById: {}, // ClaimId -> array of pinned comment IDs
   isLoading: boolean,
   isLoadingByParentId: { [string]: boolean },
   myComments: ?Set<string>,
--- a/ui/component/livestreamComment/view.jsx
+++ b/ui/component/livestreamComment/view.jsx
@@ -21,10 +21,22 @@ type Props = {
   stakedLevel: number,
   supportAmount: number,
   isFiat: boolean,
+  isPinned: boolean,
 };
 
 function LivestreamComment(props: Props) {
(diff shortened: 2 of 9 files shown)


tzarebczan:
This is working great, nice job KP! @jessopb feel free to ship after code review. 

In the future, it may be nice to allow them to edit the pinned comment - then they don't have to retype/pin a new one.

infinite-persistence:
### 2da7815 to aaf37be
- Corrected "Pinned" being inside the author name highligted bubble
- Fixed padding so that the message, if not scrolling, looks centered.

tzarebczan:
Just saw the comment about latest pinned being displayed if something is unpinned...that may be a bit strange in some cases for creators. 

I think updating the pin would mostly solve it - as they'd keep a single pin. We can rebroadcast the message into chat when a pin update is done (UI side or needs commentron?). Will file these ideas.

infinite-persistence:
> _Just saw the comment about latest pinned being displayed if something is unpinned...that may be a bit strange in some cases for creators.
I think updating the pin would mostly solve it - as they'd keep a single pin. We can rebroadcast the message into chat when a pin update is done (UI side or needs commentron?). Will file these ideas._

- Just a note:  YouTube's help page tells the user to delete and re-create. But I agree it's not user-friendly.
- #6858 - Can just copy/paste for now instead of refactor if needed quickly.
- Will also have to add something in the GUI to enforce a single pinned message per livestream.  
    - Unlikely event:  if creator decides to pin a comment via dev-tools Network, I think we would still end up with multiple pins.

jessopb (approved):

infinite-persistence:
### 67c8b83 to 4731786 
- Rebased to resolve conflict

PR #6824, opened 2021-08-09
Creator: Enable "min tips" and "min hyperchat"

### LINKED ISSUE #6119: Creator Settings: beyond "Muted Words"
Continuation of [#5833 Support creator controlled commenting settings (mainly muted word list for now) ](https://github.com/lbryio/lbry-desktop/issues/5833);  mostly just reminders of things to take care of.

- [ ] (Tom) for the comment tip setting, I think the intention was to force all comments to have tips in order to be posted - which would be an anti spam mechanism. That should be made clearer and ensure tiger5226 enforces. That would apply to hyperchats as well. So that setting is more global and forces tips, whereas the other is a min when a tip is used.
- [x] (Tom) anyone with `comments-disabled` channel tag should have setting defaulted to 0 comments enabled
  - In my previous discussion with Mark, we think this should just be removed.  Anyway, we can re-decide this when `setting.CommentsEnabled` is implemented.
- [ ] (KP) Handle LBC vs. satoshi changes
  - There was back-and-forth about where the satoshi-conversion should take place.  We've decided to do this in the app.
- [ ] (KP) Debounce numerical controls
  - Currently, only the "muted words" tag component is being debounced. Any numerical control-change would result in many `setting.Update` calls.

infinite-persistence:
## Issue
Closes #6119 (min tips)

## Test
jimbo.lbry.tv (⚠️ different test instance from usual -- troublesome to merge with existing PRs multiple times).

## General description of changes
- "Creator Settings" page now include a "Min Tip" and "Min Hyper Chat" amount setting.
- An advisory text will be displayed beside the comment-create submission button to indicate the minimum amount.  For livestreams, and additional bubble will appear to describe the min hyper chat amount.

## Concerns
- How does fiat-comments come into play?
- There is no mechanism to avoid mismatched payment.  I tried to handle it as much as possible, as describe in the Notion post.

### DIFF
--- a/flow-typed/Comment.js
+++ b/flow-typed/Comment.js
@@ -58,7 +58,6 @@ declare type CommentsState = {
   blockingByUri: {},
   unBlockingByUri: {},
   togglingForDelegatorMap: {[string]: Array<string>}, // {"blockedUri": ["delegatorUri1", ""delegatorUri2", ...]}
-  commentsDisabledChannelIds: Array<string>,
   settingsByChannelId: { [string]: PerChannelSettings }, // ChannelID -> settings
   fetchingSettings: boolean,
   fetchingBlockedWords: boolean,
@@ -222,10 +221,20 @@ declare type ModerationAmIParams = {
 };
 
 declare type SettingsParams = {
-  channel_name: string,
+  channel_name?: string,
   channel_id: string,
-  signature: string,
-  signing_ts: string,
+  signature?: string,
+  signing_ts?: string,
+};
+
+declare type SettingsResponse = {
+  words?: string,
+  comments_enabled: boolean,
+  min_tip_amount_comment: number,
+  min_tip_amount_super_chat: number,
+  slow_mode_min_gap: number,
+  curse_jar_amount: number,
+  filters_enabled?: boolean,
 };
 
 declare type UpdateSettingsParams = {
--- a/ui/comments.js
+++ b/ui/comments.js
@@ -34,6 +34,7 @@ const Comments = {
   setting_unblock_word: (params: BlockWordParams) => fetchCommentsApi('setting.UnBlockWord', params),
   setting_list_blocked_words: (params: SettingsParams) => fetchCommentsApi('setting.ListBlockedWords', params),
   setting_update: (params: UpdateSettingsParams) => fetchCommentsApi('setting.Update', params),
+  setting_get: (params: SettingsParams) => fetchCommentsApi('setting.Get', params),
   super_list: (params: SuperListParams) => fetchCommentsApi('comment.SuperChatList', params),
 };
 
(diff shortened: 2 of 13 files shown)


infinite-persistence:
Found few items to improve while testing.  Setting back to `DRAFT`.

infinite-persistence:
### 4a4f6d9 to 6dbd24b (ready for review at Jimbo)
- Use a less-verbose notice message on amount. Push the verbose version as a tooltip instead.
    - Before-and-after
        -  --> 
- Fetch `setting.Get` when clicking the LBC icon as well (try to get the latest value as frequent as possible, but not too frequent).

flow-typed/Comment.js
```diff
@@ -221,10 +220,20 @@ declare type ModerationAmIParams = {
...
-  signing_ts: string,
+  signature?: string,
+  signing_ts?: string,
+};
+
+declare type SettingsResponse = {
+  words?: string,
+  comments_enabled: boolean,
+  min_tip_amount_comment: number,
+  min_tip_amount_super_chat: number,
```
infinite-persistence, line 233: Note that Commentron seem to enforce a minimum amount of `0.01`.  Not sure if this is good or bad.
Setting to anything smaller like `0.001` will automatically reset it to `0`.  It would be nicer if it clamped to `0.01` instead.

tzarebczan: I think a min of 0.01 is fine.

tzarebczan:
Looking great, I know some creators that are waiting to get their hands on this. 

Reiterating feedback from Slack:
- [ ] grey out min hyperchat if min comment is used. If hyperchat is filled, and then comment amount is added, reset hyperchat amount. Or maybe grey out in both cases (i.e. if hyperchat min is used, grey out comments)? Add note: "This setting not available if all comments require a tip" (and vice versa for other if we go that route). 
- [ ] fix regular comments when min hyperchat is used (missing the notification)
- [ ] improve error message if comment is less than min comment/hyperchat amount

- [ ] I adjusted commentron to not apply these settings for creators. The UI can still show the message, but should allow creator to send messages regarldess. I.e. I may want to pin something.

infinite-persistence:
### 6dbd24b to c8f41a8 
Ready for re-review, and I'll test more tomorrow.
- Covered all 4 items mentioned above.
- Improved the amount checking. As long as the setting changed after user entered the "Review" stage, we prompt a toast asking the user to double-confirm. This also covers the case of creator lowering the setting (prevent overpay).

Commentron Issues:
- [ ] Commentron should provide better error when out of range:
    - 
- [ ] The scenario is prevented in the GUI now, but Commentron produces error in satoshis
    -

tzarebczan:
This is working great! @jessopb please merge after your review.

infinite-persistence:
c8f41a8 to 799a194
- Just rebased to resolve code comment conflict with Stripe merge
- Also re-tested for a few minutes.

infinite-persistence:
### 799a194 to 7817c57
Rebased again to resolve conflict with "pinned comments".
Commenting is a hot area, lol.

PR #6843, opened 2021-08-10
fix collection preview links

jessopb:


jessopb:
https://www.gitmemory.com/issue/ReactTraining/react-router/7582/734436092

PR #6875, opened 2021-08-12
send recsys powered-by

jessopb:


jessopb:
Had to change urisByQuery to resultsByQuery so that I could store the recsys id with the query since it changes by load balancer.

ui/redux/selectors/search.js
```diff
@@ -42,9 +43,9 @@ export const makeSelectSearchUris = (query: string): ((state: State) => Array<st
     if (query) {
       query = query.replace(/^lbry:\/\//i, '').replace(/\//, ' ');
       const normalizedQuery = createNormalizedSearchKey(query);
-      return byQuery[normalizedQuery];
+      return byQuery[normalizedQuery] && byQuery[normalizedQuery]['uris'];
     }
-    return byQuery[query];
+    return byQuery[query] && byQuery[query]['uris'];
   });
```
infinite-persistence, line 49: Wunderbar broke a bit (not sure if it's specifically due to this line or not)

Steps to reproduce:
1. Homepage
2. Search something in Wunderbar and press enter (I used "manchester").  We'll end up in the Search Page.
3. Clear Wunderbar and type "manchester" again but don't press enter.  
    - In the drop down, `Most Supported` appears, but `Search results` is blank.  Reverted the change and `Search results` will appear.

ui/redux/selectors/search.js
```diff
@@ -28,7 +29,7 @@ export const selectIsSearching: (state: State) => boolean = createSelector(selec
 
 export const selectSearchUrisByQuery: (state: State) => { [string]: Array<string> } = createSelector(
```
infinite-persistence, line 30: Maybe a rename would be good to reflect the actual returned object.

infinite-persistence (commented):
:+1: Good to go with the following issues left:

infinite-persistence (approved):

Issue #6107, opened 2021-05-21
muted/blocked not applying to related content


tzarebczan:
Mute a related content channel and then refresh - it still shows

infinite-persistence:
`Mute` seems to work, but `Block` doesn't.

tzarebczan:
It's also not applying to commenters you block outside of your own content. Not sure if there's a better way to handle this outside of front end blocking...

infinite-persistence:
> _Mute a related content channel and then refresh - it still shows_

- `ClaimPreview` is already filtering based on muted and blocked list, so it's working from that perspective.
- The scenario above only happens if you refresh _immediately_ after muting.  The wallet/settings sync reverted the change, and the channel actually didn't get muted. #5974 is the root-cause?

tzarebczan:
"Block and mute" may be better if they expect not to see these channels come up on their content also.

PR #6878, opened 2021-08-13
Hide comments based on own blocklist

infinite-persistence:
- Tom, this is the changes we talked about in Slack.
- Available in the usual dev instance.
- Given how close this is to Saturday's stream, we should only merge in next week (just in case).

----

<details>
  <summary>Comments: include Commentron blocklists when filtering</summary>

## Issue
- When you block Channel-X, Channel-X's comments will still be visible on someone else's content. This feels odd.
- https://github.com/lbryio/lbry-desktop/issues/6107#issuecomment-897884114

## Change
In addition to the blacklist, filter-list and muted-list, we now include the Commentron blocklists (personal, admin, moderator) when filtering out comments.

## Specifics
`makeSelectCommentsForUri`, `makeSelectTopLevelCommentsForUri` and `makeSelectRepliesForParentId` all perform the same filtering code. Factor that out to `makeSelectFilteredComments` and add the Commentron lists into the mix.

## Downsides
This probably adds to the already-high CPU usage in rendering comments.
</details>

<details>
  <summary>Comments: simplify blocked replies display</summary>

## Issue
Previously, we decide when to display "Show More" based on the current fetched reply count vs. total replies in Commentron. It was already troublesome as `comment.List` and `comment.replies` give different values (one includes blocked content, while another does not).

Now, we are further filtering the list with Commentron blocklists (personal, admin, moderator), so it is not feasible to run the logic based on reply count.

## Solution
- Keep track of number of remaining pages instead and use that to determine when to display "Show More".
  - While it doesn't solve the "Show N replies" mismatch (YT has this problem too), it prevents the button from lingering.
- In the event that all replies are blocked, just show an empty space (same as YT). I didn't like the previous version that cluttered the space with "comment(s) blocked".
</details>

### DIFF
--- a/flow-typed/Comment.js
+++ b/flow-typed/Comment.js
@@ -35,7 +35,7 @@ declare type CommentsState = {
   byId: { [string]: Array<string> }, // ClaimID -> list of fetched comment IDs.
   totalCommentsById: {}, // ClaimId -> ultimate total (including replies) in commentron.
   repliesByParentId: { [string]: Array<string> }, // ParentCommentID -> list of fetched replies.
-  totalRepliesByParentId: {}, // ParentCommentID -> total replies in commentron.
+  repliesTotalPagesByParentId: {}, // ParentCommentID -> total number of reply pages for a parentId in commentron.
   topLevelCommentsById: { [string]: Array<string> }, // ClaimID -> list of fetched top level comments.
   topLevelTotalPagesById: { [string]: number }, // ClaimID -> total number of top-level pages in commentron. Based on COMMENT_PAGE_SIZE_TOP_LEVEL.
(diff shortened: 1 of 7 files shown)


tzarebczan:
Thanks! Will give this a test, but agreed, let's not merge til next week.

tzarebczan:
This was super buggy on kp.odysee.com

infinite-persistence:
> _This was super buggy on kp.odysee.com_

We found out it was due to the sidebar resolving a large Following -- also present in current production._

infinite-persistence:
### 53a9d34 to e33e8d1
- Rebased to resolve conflict

Issue #6879, opened 2021-08-13
Previously pinned livestream comments show as latest 


tzarebczan:
![image](https://user-images.githubusercontent.com/8120721/129381273-ba802c7c-dccb-4ba3-9b5f-503cae567d71.png)
circled test 4 comment was made recently, but the 3 comments below that were old pinned comments. Happens after refresh if you're on a page already .

PR #6888, opened 2021-08-16
Livestream: stop pinned comments from appearing as latest

infinite-persistence:
## Ticket
Closes [#6879: Previously pinned livestream comments show as latest](https://github.com/lbryio/lbry-desktop/issues/6879)

## Issue
`comment.List` will always display the pinned comment first, hence the problem when the chat is refreshed.

## Approach
Completely split pinned comments from top-level comments in the Reducer, and the let the GUI (e.g. regular comments, livestream comments) decide how they want to display it.

For the case of livestream, there is no need to repeat the pinned comments in the regular chat area, since there is a dedicated area on top.

### DIFF
--- a/ui/component/commentsList/index.js
+++ b/ui/component/commentsList/index.js
@@ -15,6 +15,7 @@ import {
   selectMyReactionsByCommentId,
   makeSelectCommentIdsForUri,
   selectSettingsByChannelId,
+  makeSelectPinnedCommentsForUri,
 } from 'redux/selectors/comments';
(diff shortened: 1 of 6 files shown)


tzarebczan:
LGTM! 
It is kind of strange that the other pinned comments disappear from the list, but I think that's okay / kind of expected.

infinite-persistence:
Yeah, it's kind of all-or-nothing for now until `comment.List` has the ability to not float pinned comments to the top, so that it'll be in the correct sort order.  Briefly discussed in v3 pagination.

Issue #6886, opened 2021-08-16
Livestream auto-scroll problems


infinite-persistence:
## Issues
- [ ] When sending a new comment in a livestream, your comment sometimes is not visible until you manually scroll down, even when you are at the bottom before sending the message.  On other times, it is partially visible with it being obscured by the "Recent Comments" button.
- [ ] When receiving a new comment in livestream, sometimes the message isn't visible until manually scrolled down.

## Steps:
Couldn't find a direct way to reproduce this, as it seems to depend on zoom level and exact scroll position.  However, scrolling up the chat, then pause, then scrolling back down before sending/receiving the message seems to increase the chances of it happening.

PR #6889, opened 2021-08-16
Fix livestream autoscroll

infinite-persistence:
## Ticket
Closes [#6886 Livestream auto-scroll problems](https://github.com/lbryio/lbry-desktop/issues/6886)

## Issue
- `performedInitialScroll` was problematic as it won't allow auto-scroll even when user scrolled back to the bottom.
- The dependence on `commentElement` seems to assume a certain comment height? Not sure.

## Approach
- Add a scroll listener and stash the last scroll position.
- When a message is received, check if it's at the bottom (>=0). If yes, maintain that position after the new comment is added. If not, leave as is.
- When submitting a comment, always reset to the bottom.


jessopb (approved):
Nice

PR #6897, opened 2021-08-18
Reload when auth token mismatch is detected

### LINKED ISSUE #5504: Signing out of account causes page to break in other tabs
Open up multiple tabs of odysee.com.

In tab A open a video, then go to tab B and sign out of odysee. Now return to tab A and an error will popup.

### LINKED ISSUE #6829: merged accounts - force log out / fail sync when x-auth-token and cookie auth token are different
There's a scenario where two accounts can get merged if multiple tabs are used and accounts are logged out. This has to happen quickly enough where the background process doesn't log out the account. 

The newly logged in token is saved and run against a different x-auth-token.

infinite-persistence:
## Tickets
- Closes [#5504 Signing out of account causes page to break in other tabs](https://github.com/lbryio/lbry-desktop/issues/5504)
- Closes [#6829 merged accounts - force log out / fail sync when x-auth-token and cookie auth token are different](https://github.com/lbryio/lbry-desktop/issues/6829)

## Steps to replicate
1. Login to odysee with account-A.
2. Open another tab, and split both tabs on the screen.
3. Logout from the 1st tab. Do not activate (focus) the 2nd tab.
4. On the 1st tab, login with account-B.
5. Activate (focus) the 2nd tab. The wallet would have been merged, and we are still logged in as account-A.

## Approach
Reload when the LBRY API token no longer matches the auth token.

## Concerns
- Will there be valid scenarios where the API token can be different?

### DIFF
--- a/ui/redux/actions/sync.js
+++ b/ui/redux/actions/sync.js
@@ -3,9 +3,10 @@ import { Lbryio } from 'lbryinc';
 import { SETTINGS, Lbry, doWalletEncrypt, doWalletDecrypt } from 'lbry-redux';
 import { selectGetSyncIsPending, selectSetSyncIsPending, selectSyncIsLocked } from 'redux/selectors/sync';
 import { makeSelectClientSetting } from 'redux/selectors/settings';
-import { getSavedPassword } from 'util/saved-passwords';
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks @infinite-persistence. There should be no scenario where these two are different.

PR #6916, opened 2021-08-19
watchman send dsk for desktop

jessopb:


mayeaux:
Does Andrey's backend account for this? It may throw an error if he hasn't added it to the enum values

PR #6947, opened 2021-08-24
Use solid heart for "subscribed"

infinite-persistence:
## Objective
Make it easier to know the subscription state in a glance, without actually having to reading the text, especially for certain languages where it's just one character difference.

image
image

image

## Aside
Noticed that mobile forces the "unsubscribe" (broken heart) in place of "subscribe", I believe due to the fact that we can't mouse-hover.  So, the solid heart doesn't appear in Mobile.


jessopb (commented):
Love it

PR #6975, opened 2021-08-27
Add play button on pause ios

mayeaux:
Adds a play button when content is paused on iOS. Useful because the current play button is very small and in an annoying place for righthanded users, and because iOS Safari is requiring a manual play intervention on each video upload.

### DIFF
--- a/ui/component/viewers/videoViewer/internal/videojs.jsx
+++ b/ui/component/viewers/videoViewer/internal/videojs.jsx
@@ -19,8 +19,6 @@ import isUserTyping from 'util/detect-typing';
 // @endif
 const isDev = process.env.NODE_ENV !== 'production';
 
-process.on('unhandledRejection', console.log)
-
 export type Player = {
   on: (string, (any) => void) => void,
   one: (string, (any) => void) => void,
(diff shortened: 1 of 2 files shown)


jessopb (approved):

jessopb (approved):
This may be a sideways change if @infinite-persistence is right. Explore it?

mayeaux:
> We are currently using `videojs-mobile-ui` for Android for the mobile style. It was disabled for IOS back then due to something (I forgot what) which I think was recently fixed.
> 
> * Update: I recall a bit now ... I think it was "use native on iOS for some reason", but we recently switched to just use the regular videojs version on iOS, so at that point we can use `videojs-mobile-us` again.
> 
> The plugin was never tested on IOS, but if it works, maybe we should just use common code rather than introduce new one for IOS? It'll provide other functionality like "tap-to-seek-5s"
> 
> ```
> const mobileUi = function(options) {
>   // if (videojs.browser.IS_ANDROID || videojs.browser.IS_IOS) {
>   if (videojs.browser.IS_ANDROID) {
>     this.ready(() => {
>       onPlayerReady(this, videojs.mergeOptions(defaults, options));
>     });
>   }
> };
> ```

Thanks for the info I will try that out right now!

PR #6977, opened 2021-08-27
recsys v0.2

jessopb:


ui/component/claimPreview/view.jsx
```diff
@@ -223,7 +223,7 @@ const ClaimPreview = forwardRef<any, {}>((props: Props, ref: any) => {
 
   const handleNavLinkClick = (e) => {
     if (onClick) {
-      onClick(e, indexInContainer);
+      onClick(e, claim, indexInContainer); // not sure indexInContainer is used for anything.
```
infinite-persistence, line 226: The index was previously used in `handleRecommendationClicked`, but seems like the index is determined differently now.
Can retain for other future clients, or remove entirely (I can clean up later)

infinite-persistence (commented):
Some questions here: https://www.notion.so/lbry/Recsys-v2-testing-b6618e578d2b4922a47869ad0ac14c95

Ignore if it doesn't make sense :D

jessopb:
I was looking for a reliable way to send when a page is left, but I decided to just listen to history changes and send every time as long as the claimId isn't currently playing. Currently playing will happen either on dispose or on tab close/hide.

PR #6990, opened 2021-08-31
Pressing escape in wunderbar should do something useful

ddifiore:
## Fixes

Issue Number: 2116

Pressing escape would cause the wunderbar to lose focus.

## What is the new behavior?

Pressing escape now clears out the wunderbar and brings it in focus so the user can continue searching. Also Ctrl+K will now highlight anything in the wunderbar and bring it into focus.

## PR Checklist

<details><summary>Pressing escape in wunderbar should do something useful.</summary>

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

</details>

### DIFF
--- a/ui/component/wunderbarSuggestions/view.jsx
+++ b/ui/component/wunderbarSuggestions/view.jsx
@@ -28,6 +28,7 @@ const SEARCH_PREFIX = `$/${PAGES.SEARCH}q=`;
 const INVALID_URL_ERROR = "Invalid LBRY URL entered. Only A-Z, a-z, 0-9, and '-' allowed.";
 const TAG_SEARCH_PREFIX = 'tag:';
 
+const K_KEY_CODE = 75;
 const L_KEY_CODE = 76;
 const ESC_KEY_CODE = 27;
 
--- a/ui/component/wunderbarSuggestions/view.jsx
+++ b/ui/component/wunderbarSuggestions/view.jsx
@@ -228,7 +228,7 @@ export default function WunderBarSuggestions(props: Props) {
   React.useEffect(() => {
     function handleKeyDown(event) {
       const { ctrlKey, metaKey, keyCode } = event;
-      
+
       if (!inputRef.current) {
         return;
       }
--- a/ui/component/wunderbarSuggestions/view.jsx
+++ b/ui/component/wunderbarSuggestions/view.jsx
@@ -228,12 +228,19 @@ export default function WunderBarSuggestions(props: Props) {
     function handleKeyDown(event) {
       const { ctrlKey, metaKey, keyCode } = event;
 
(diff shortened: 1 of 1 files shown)


ui/component/wunderbarSuggestions/view.jsx
```diff
@@ -228,12 +229,19 @@ export default function WunderBarSuggestions(props: Props) {
     function handleKeyDown(event) {
       const { ctrlKey, metaKey, keyCode } = event;
 
+      if (keyCode === K_KEY_CODE && ctrlKey) {
```
jessopb, line 232: I'm not certain that the next block `if (!inputRef.current) return;` is necessary, but if it is, I imagine it should be before this one. (Bail early?)

ui/component/wunderbarSuggestions/view.jsx
```diff
@@ -228,12 +229,19 @@ export default function WunderBarSuggestions(props: Props) {
...
+      }
+      
       if (!inputRef.current) {
         return;
       }
 
       if (inputRef.current === document.activeElement && keyCode === ESC_KEY_CODE) {
-        inputRef.current.blur();
+        inputRef.current.value = "";
+        inputRef.current.focus();
```
jessopb, line 244: Previously, pressing escape would blur the search focus but leave the text.
Now we're clearing the text, but keeping the focus. 

Youtube seems to clear the suggestion box on first ESC, and then blur focus on second ESC press. 
Does this seem like a good idea?

jessopb (commented):
This makes sense. But, since users are used to esc clearing focus, I think we need to keep that behavior on second press.

ui/component/wunderbarSuggestions/view.jsx
ddifiore: You are right!  I thought that line was guarding against when the wunderbar is not selected but it is actually just checking that element exists.  I will make this change.

jessopb:
When you push up the other change, the linter complained about the double-quotes that should be single-quotes, and a trailing space somewhere.  you should be able to run "yarn precommit" to check those

tzarebczan:
Hey @ddifiore , appreciate you taking a look at this issue and welcome to our repo! We'll get another review on this once you let us know it's good to go. 

After that, can we show you some [appreciation](https://lbry.com/faq/appreciation)?

ui/component/wunderbarSuggestions/view.jsx
ddifiore: I implemented a hybrid of the two where escape will clear the text and keep the wunderbar in focus if there was any text in the box.  If there is no text in the box then the box will lose focus.  I think this is a good compromise.  It keeps the old behavior if you press escape twice and also keeps the clearing of text for faster searching.

Issue #6985, opened 2021-08-30
fromentries app crash - fix or add polyfill 


tzarebczan:
Happens on older browsers and many tv/xbox based browsers 

https://github.com/ungap/from-entries 

![image](https://user-images.githubusercontent.com/8120721/131372843-35cd6c9e-72db-4a0d-850b-07d541dd5314.png)

PR #6998, opened 2021-09-01
Fix Object.fromEntries crash on some browsers

infinite-persistence:
❗Requires [lbry-redux PR 428](https://github.com/lbryio/lbry-redux/pull/428)

## Issue
Closes [#6985 fromentries app crash - fix or add polyfill](https://github.com/lbryio/lbry-desktop/issues/6985)

## Test
- Tested on Chrome 65 on Windows 10 (via BrowserStack)
- Tested on Samsung Browser on Galaxy S10 (via BrowserStack)
- Tested on UC Browser on obscure phones (via BrowserStack)
- Roughly compared the output of the function before and after the polyfill.
- Did some regression-testing over popular browsers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "imagesloaded": "^4.1.4",
     "json-loader": "^0.5.4",
     "lbry-format": "https://github.com/lbryio/lbry-format.git",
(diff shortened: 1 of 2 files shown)


tzarebczan:
Will give this another test too, thank you! @keikari can you also try ? I think Icecat was the one we had complaints.

tzarebczan:
@JYamihud is going to be excited for this one.

Issue #7008, opened 2021-09-02
persist tip vs support last used setting


tzarebczan:
Could have sworn there was an issue for this, but can't find it. I know we discussed recently. 

We're defaulting to boost rn, which is ugly. We had persistence and lost it during stripe stuff.

PR #7038, opened 2021-09-07
Fiat tip improvements

mayeaux:
Addresses: https://github.com/lbryio/lbry-desktop/issues/7008


tzarebczan:
Comes up without any option selected. Works from that point, but will need to do some more testing with fiat too. 
image

Issue #6834, opened 2021-08-09
Support pagination + search in moderation sections


tzarebczan:
For channels or delegated mods with really long lists, loading the page of blocked channels is super slow and may not render properly. It's also really hard to unblock someone you're looking for. 

- [ ] support pagination UI side (i think commentron side is still fast enough)
- [ ] support search by channel name

PR #7055, opened 2021-09-10
Blocklist: paginate + search

infinite-persistence:
## Test
kp.odysee.com

## Ticket
Closes [#6834 Support pagination + search in moderation sections](https://github.com/lbryio/lbry-desktop/issues/6834)

## Main changes
- Added app-side pagination.
- Improved the "moderator-block" visuals a bit
- Added search bar for blocklists.
    - Fuzzy search is supported, but is slow when the list is huge. Ended up using "CONTAINS" search.
    - Can play around with other types if the search is not good enough:

        > CASE SENSITIVE EQUALS: Case-sensitive equality trumps all. These will be first. (ex. France would match France, but not france)
EQUALS: Case-insensitive equality (ex. France would match france)
STARTS WITH: If the item starts with the given value (ex. Sou would match South Korea or South Africa)
WORD STARTS WITH: If the item has multiple words, then if one of those words starts with the given value (ex. Repub would match Dominican Republic)
CONTAINS: If the item contains the given value (ex. ham would match Bahamas)
ACRONYM: If the item's acronym is the given value (ex. us would match United States)
SIMPLE MATCH: If the item has letters in the same order as the letters of the given value (ex. iw would match Zimbabwe, but not Kuwait because it must be in the same order). Furthermore, if the item is a closer match, it will rank higher (ex. ua matches Uruguay more closely than United States of America, therefore Uruguay will be ordered before United States of America)

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "express": "^4.17.1",
     "humanize-duration": "^3.27.0",
     "if-env": "^1.0.4",
+    "match-sorter": "^6.3.0",
     "parse-duration": "^1.0.0",
     "react-datetime-picker": "^3.2.1",
     "react-plastic": "^1.1.1",
--- a/ui/component/blockList/index.js
+++ b/ui/component/blockList/index.js
@@ -1,0 +1,2 @@
+import BlockList from './view';
+export default BlockList;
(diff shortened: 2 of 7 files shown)


tzarebczan:
This looks great, I know a creator that will be super happy!

Issue #6776, opened 2021-08-03
Missing follow button on channel reposts


tzarebczan:
![image](https://user-images.githubusercontent.com/8120721/128072850-dffd5af0-cd6c-443e-bf63-783a322a7224.png)

tzarebczan:
Also on a channel page, we show a follow button but it doesn't reflect your following data (i.e. if you follow a channel, it still shows follow)

PR #7056, opened 2021-09-10
Fix missing and broken "Follow" button on reposts

infinite-persistence:
Closes [#6776 Missing follow button on channel reposts](https://github.com/lbryio/lbry-desktop/issues/6776)

| Issue | Fix |
| ------------- | ------------- |
| Missing "follow" button  | For reposts, `parseURI` will fail to determine if it's a channel because the URL doesn't have enough info. Determine from the claim object instead. |
| "follow" button not working in Channel Page  | The repost uri doesn't link back to the source channel, so we'll need to handle it by obtaining it from the claim object.  |


tzarebczan:
Awesome, thank you!

PR #7058, opened 2021-09-10
Changes to improve videojs playing experience on iOS

mayeaux:
Re-enables mobile-ui for videojs on iOS and plays videos by default as muted with the 'Tap to unmute' button

### DIFF
--- a/ui/component/viewers/videoViewer/view.jsx
+++ b/ui/component/viewers/videoViewer/view.jsx
@@ -341,13 +341,31 @@ function VideoViewer(props: Props) {
             // player.muted(true);
             // another version had player.play()
           }
+          console.log('running here!')
         } else {
+          console.log('running here123')
           const isPaused = player.paused();
+
+          console.log('is paused');
+          console.log(isPaused)
           if (IS_IOS && isPaused) {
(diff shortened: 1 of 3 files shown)


infinite-persistence (changes requested):

ui/component/viewers/videoViewer/view.jsx
```diff
@@ -335,6 +341,31 @@ function VideoViewer(props: Props) {
...
+            const isPaused = player.paused();
+            console.log('is paused');
+            console.log(isPaused)
+            if (IS_IOS && isPaused) {
+              document.getElementsByClassName('video-js--tap-to-unmute')[0].style.visibility = 'visible';
+              player.muted(true);
+              const iosResponse = player.play();
+              console.log(iosResponse)
+            }
+          }, 1000)
```
infinite-persistence, line 368: - Is there an alternative to a hardcoded timeout?
- If it's really needed, do the `IS_IOS` check first (instead of checking inside the timeout), so that it only runs for IOS.

infinite-persistence (changes requested):

PR #7062, opened 2021-09-11
Test langs

jessopb:


jessopb:
noticed that every collection update was doubling the language setting if set. [en] -> [en, en].

jessopb:
fixes another ui inconsistency with not deleting second language if first is set to that language

jessopb:
https://github.com/lbryio/lbry-redux/pull/430 dependency

Issue #7059, opened 2021-09-10
Add option to sort oldest first on channel page


tzarebczan:
put a ^ in front of the order_by param to reverse

PR #7069, opened 2021-09-13
Channel Page: enable filters; add "sort by" filter

infinite-persistence:
https://kp.odysee.com/@Odysee:8?order=new&sort=old

## Issue
Closes [#7059 Add option to sort oldest first on channel page](https://github.com/lbryio/lbry-desktop/issues/7059)

## Changes
- Enabled filters in Channel Page for Odysee.
    - Since we are filtering now, I thought we might as well re-enable this.  Let me know if we are not ready for this.
- Added a new "Sort By" filter, which will only appear for "New" ordering. It didn't make sense for "Trending" or "Top".


tzarebczan:
Yes, been wanting to bring to bring this filter back! We can definitely add some of the other options like language, maybe tags, etc.

PR #7112, opened 2021-09-16
Auth refactor

jessopb:
This makes "Auth" easier to understand (I think) and begins to support dual token types.


jessopb:
closing for main pr

PR #7114, opened 2021-09-16
support embed preference

jessopb:
Disable link buttons and title links on embeds during and after playback.
Still some details to work out - what to do at the end?


jessopb:
I think it may be better to interrupt/prevent the ended behavior in the video player so people can play again..

tzarebczan:
We should show a replay / rewatch button that starts the video over.

tzarebczan:
others:
- hide follow button with this tag
- top search result (unrelated to tag) - hide filtered list items.

PR #7115, opened 2021-09-17
Livestream category improvements

### LINKED ISSUE #6015: live stream categories fall out
- [x] remove timestamp from card for live streams
- [x] instead of passing all claim ids from a category to the no-source CS, find the intersection between the odysee live API channels and the category channels. Do this for all sections outside of wild west (leave as is for now). Eventually, this will need to be adjusted to a larger claim search or calling CS with 50 claims multiple times for latter pages. 
- [x] Pass release time of 3 days (let's try this for now) 
- [x] at some interval, re-run the claim searches against live stream + has no source, and refresh only those. If easier, refresh all. 
- [ ] add view count with eye

I know there were a couple more, will add them when they come to me.

### LINKED ISSUE #7123: infinite scroll on categories / discover stops working randomly
Not sure when this happens but we've seen a few complaints and have  had it  happen to me. Sometimes happens when you go into a file, then back, and then scroll further. 

Othertimes it may be related to how many tiles are shown at the bottom.

infinite-persistence:
## Status
_Sept 23rd_:  Pending reviewing on WildWest handling

## Code review
- I think it would be easiest to go through the commits one-by-one, as the deltas will be more readable, and I've added notes for each of them.
- If approved, assigned back to me, and I'll clean up the commits before merging.

## Test
https://kp.odysee.com

## Objectives
- Closes part of #6015 
    - The eye count blocked by a backend bug.
- Closes #7123 
- Reduce excessive rendering on `ClaimTileDiscover`

## Overview of changes
- Reverted the way I added livestreams to the list.
- Fetch livestream claims separately via redux, and pass to the list components using the existing `prefixUris` prop.  No more double `claim_search` calls.

## Small differences
- Previously, the total number of tiles will not be greater than requested, as the livestream tiles replaces some of the normal tiles.  Now, it just appends at the top, so we could see more than requested.  Can change back if this is not desired, but I opted for lesser code this round to keep the `Claim**Discover` components from getting out of hand.


jessopb (approved):
A thing of beauty.

tzarebczan:
"Previously, the total number of tiles will not be greater than requested," - the old method worked this way too - the only concern I have for this is how it makes the tiles uneven and this breaks infinite scroll in some cases. Doesn't need to be addressed here, I believe there's a separate issue or we can file one. 

Will give this a test soon.

tzarebczan:
This is looking great so far! we'll need to see how to handle the wild west page with the larger number of streams now showing up there since this way is more correct. 

I am seeing new live + claim search for live streams when hitting each category - did we want to avoid that? Maybe only call if we haven't refreshed streams in last X minutes? Each time they'll return the same thing in a short period of time.

infinite-persistence:
> _Is this only for claim_tiles_discover?_

It's for both `ClaimTilesDiscover` and `ClaimListDiscover`

infinite-persistence:
> _I am seeing new live + claim search for live streams when hitting each category - did we want to avoid that? Maybe only call if we haven't refreshed streams in last X minutes? Each time they'll return the same thing in a short period of time._

Ah, right. Currently I fetch Livestreams "on mount" for any pages that need it. Will think of a way to check for "last fetched".

infinite-persistence:
### ee41dcf to 4cb83fc
- Added a default minimum of 5 minutes between fetches. Clients can bypass this through `forceFetch` if needed.
- Rebased.

tzarebczan:
@infinite-persistence I spoke to Julian about the wild west section + lives - we decided it would be good to show a toggle on top to show/hide live streams. Default would be hide, and would be persisted. If show is clicked, then it pops them all on top, followed by non lives.

tzarebczan:
Jeremy actually didn't like the idea of a toggle at the top, defaulted to off. 

Maybe instead we'd show a single top line, ordered by trending, and then have a view all livestreams button below it? 

Any other ideas on how to show some at first, and then view all?

infinite-persistence:
That single-line reduces in tile count when window width is reduced (always just 1 line)? Or ok to stick with 8 tiles and reflow when necessary?

infinite-persistence:
> _Jeremy actually didn't like the idea of a toggle at the top, defaulted to off._
>
> _Maybe instead we'd show a single top line, ordered by trending, and then have a view all livestreams button below it?_
>
> _Any other ideas on how to show some at first, and then view all?_

### 4cb83fc to 796d435
- WildWest: limit to 8 livestreams initially, and user-expandable to show the rest.  Applied to both Tile and List layout.
- Rebased to latest `master`.

tzarebczan:
I like this - but can it respect the screen sizes we have? This looks strange on larger monitors: 
image

infinite-persistence:
> _but can it respect the screen sizes we have? This looks strange on larger monitors:_

#### e7f840a to 1a8f82b
- Handle tile-count on large screens the same way the homepage does it.
- Rebased.

tzarebczan:
Thank you, looking great!

PR #7132, opened 2021-09-20
ClaimPreviewTile: render optimization

infinite-persistence:
The `Date` object and blocklist arrays causes unnecessary render since their references are always different in the shallow-compare.

Here the result on one of the many extra renders, with just the Cheese section alone (50% savings per cycle):

image


jessopb (approved):
:+1:

Issue #7121, opened 2021-09-17
Missing mod block option (large number of moderated channels?)


tzarebczan:
Seeing this on an internal account, it worked about a week ago. It just worked recently on an older cached version of the site, so it must be related to the pagaination or the last commit here: https://github.com/lbryio/lbry-desktop/commit/af4ff29b238d7dade350426cf972931372af37c3

tzarebczan:
Went back to a previous commit (95278f1da15a5dd64288260d4ee9ad296f52fec4) and still missing there.

PR #7139, opened 2021-09-21
Fix moderator data misalignment.

infinite-persistence:
https://kp.odysee.com

## Issue
Closes [#7121 Missing mod block option (large number of moderated channels?)](https://github.com/lbryio/lbry-desktop/issues/7121)

## Notes
It was bad to assume `channelSignatures` would be the same length as the promise result -- any failure in signing would cause a misalignment.

For my case, an accidentally-merged channel couldn't be signed due to a missing private key. I think it's the same for Drew.


infinite-persistence:
This still does not surface any channel-signing issues. Do we want to do that, or is that rare enough to ignore?

tzarebczan:
Rare enough, they'll find out other ways.

Issue #6891, opened 2021-08-16
Add video bitrate/user bandwidth to Watchman


mayeaux:
Two optional but supported (and very valuable) fields for Watchman are the bitrate of the video and also the user bandwidth. This should be added to the integration to allow for greater understanding of buffering. 

Should be added here: ui/analytics.js:133

tzarebczan:
I opened https://github.com/lbryio/lbry-desktop/issues/6881 with this and some other suggestions.

tzarebczan:
Others to consider to see if they are helpful: 
- cumulative buffering stats?
- send location instead of using our internal apis vs IP for analytics?

jessopb:
Blocked by decision about how often to check bandwidth?

PR #7145, opened 2021-09-21
Send video bitrate and user bandwidth to Watchman

mayeaux:
Fixes issue #6891 , 

Calculates user bandwidth on analytics load, and recalculates every 30s, send both values in bits per second

### DIFF
--- a/ui/analytics.js
+++ b/ui/analytics.js
@@ -10,6 +10,13 @@ import ElectronCookies from '@exponent/electron-cookies';
 import { generateInitialUrl } from 'util/url';
 // @endif
 import { MATOMO_ID, MATOMO_URL } from 'config';
+import getConnectionSpeed from 'util/detect-user-bandwidth';
+
+let downloadSpeed;
+getConnectionSpeed(function(speedInMbps){
+  downloadSpeed = speedInMbps;
+  console.log(downloadSpeed);
+});
 
 const isProduction = process.env.NODE_ENV === 'production';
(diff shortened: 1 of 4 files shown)


infinite-persistence:
Don't forget to run lint and the auto-formatter, else it will cause unnecessary deltas for the next person :)

tzarebczan:
Seems a bit overkill to re-download a file every 30 seconds to grab bandwidth. This itself can slow down their stream if their bandwidth is limited. If the file is small, it shouldn't be an issue, but it also may not be a good bandwidth test.

mayeaux:
> Seems a bit overkill to re-download a file every 30 seconds to grab bandwidth. This itself can slow down their stream if their bandwidth is limited. If the file is small, it shouldn't be an issue, but it also may not be a good bandwidth test.

The file is 2MB, the smaller the file the less accurate the value, ideally we would serve a file from one of our own CDNs so it would more accurately match to their bandwidth, I couldn't find a file though that would be suitable (was large enough), but hopefully if we can find a large enough file (1MB+ for sure), then we can maybe change the interval speed to one minute or let me know what you think would be suitable.

tzarebczan:
@nikooo777 thoughts on how to deal with the bandwidth test here? 
 
I'd say let's push with bitrate for now until we figure that out. Please fix linting and comment out the bandwidth test.

jessopb:
Blocked by deciding how often to check bandwidth?

PR #7154, opened 2021-09-22
Fix issue where channel upload viewcounts were creating a new line

### LINKED ISSUE #7078: improve wrapping for views + date on channel pages

mayeaux:
Fixes #7078

This address specifically 'large' view of over 1600px, where the worst offending was happening, but should also fix issues on other sizes as well, if anyone sees the view counts creating a third line on channel overview let me know and I'll take a look.

### DIFF
--- a/ui/component/fileViewCountInline/view.jsx
+++ b/ui/component/fileViewCountInline/view.jsx
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import '../../scss/component/_view_count.scss';
+import 'scss/component/_view_count.scss';
 
 type Props = {
   uri: string,
--- a/ui/component/fileViewCountInline/view.jsx
+++ b/ui/component/fileViewCountInline/view.jsx
@@ -15,6 +15,7 @@ export default function FileViewCountInline(props: Props) {
   let formattedViewCount;
 
   try {
+    // SI notation that changes 1234 to 1.2K, look up Intl.NumberFormat() for docs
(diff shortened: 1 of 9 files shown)


infinite-persistence (commented):
- The difference in margin feels incoherent in a glance with sibling components (title).  Makes it too cramped to the left side?  
    - image
    - image
- The channel-stake-indicator gets cut off.
    - image
- Adding the tweak for specific layouts does cover most cases in English (nice), but even in other layouts it will still reflow:
    - image
- Other languages still would reflow after the tweak (although I understand we are just trying to cover the most common cases):
    - image

YT does it by tweaking tile count, and ultimately just reflow.

ui/component/fileViewCountInline/view.jsx
```diff
@@ -30,13 +30,16 @@ export default function FileViewCountInline(props: Props) {
   // clean up (only one place edit/remove).
   const isChannelPage = window.location.pathname.startsWith('/@');
 
+  // dont show if no view count, if it's a repost, a livestream or isn't a channel page
   if (!viewCount || (claim && claim.repost_url) || isLivestream || !isChannelPage) {
```
infinite-persistence, line 34: Feels like the variables names are already self-explanatory?  There's no additional info here.

tzarebczan:
Let Anythony know linting was failing, otherwise should be good to go.

jessopb:
I think lint is failing because it's a new pattern, and we may not need a new pattern. @infinite-persistence can say whether this optimization is necessary for code splitting?

Issue #5146, opened 2020-12-07
Show all own comments


tzarebczan:
Would require https://github.com/lbryio/commentron/issues/14

tzarebczan:
- allow easy moderation from this area

tzarebczan:
This should be restricted to a signed channel request for the author only (so we don't allow others to see someone else's comments)

PR #7171, opened 2021-09-27
List own comments

infinite-persistence:
## Test
KP » Settings » Account » Comments

## Issue
Closes [#5146 Show all own comments](https://github.com/lbryio/lbry-desktop/issues/5146)

## Known issues
<strike>The total number of comments that Commentron reports vs. what it returns, can be different.  Checking with Mark/Tom.</strike> Fixed


infinite-persistence:
### cd2f3d2 to bac26bc
Commentron now requires authorization for this action.

tzarebczan:
Don't love the placement of where this is avaialble, but probably the best bet for now. 

Checking with Beamer on a sorting issue, otherwise it looks good!

infinite-persistence:
#### bac26bc to 05f2aed
- Add i18n strings
- Rebased

tzarebczan:
There's a new sorting method (5 I think - check DMs) to remove pins. 

Can we make the comment clickable or linkable?

infinite-persistence:
05f2aed to 48247cb
- Mobile: removed indent; reduced claim thumbnail size.
- Remove automatic pinning when fetching.
- Clicking on the claim (left side) now links to the comment directly.  The comment (right side) remains unclickable due to:
    - I find hard to copy/paste parts of the comment if the entire section is clickable.
    - But I did try and couldn't get `NavLink` to work with `Comment` -- it causes lots of renders and blinking.  Maybe re-visit this next time, but for now use the left-side to click?
        - Also considered adding a "View comment" button at the lower-right of each row, but that made the page busier than it is already.

PR #7180, opened 2021-09-29
Memoize 'mutedAndBlockedChannelIds'

infinite-persistence:
@jessopb, can you help review this, particular the usage of `selectCommentsState` (whether it is anti-pattern or not). If it is good, I'll apply the same method to other beefy selectors.

## Issue
Muted and blocked channel IDs are constantly being re-calculate on every update cycle. This PR tries to memoize it through `createSelector`

## Notes
The input selectors to `createSelectors` must be plain, i.e. no transformations, otherwise it breaks memoization 
<sub>(aside: I believe most of our `makeSelect**` selectors break this rule, which explains why we've been getting new references in each cycle, causing extra renders).</sub>

That was why I used `selectCommentsState` instead of importing `selectModerationBlockList` from the other file. Even if I created a new `selectModerationBlockListRaw` (which doesn't transform) and then import that, it didn't work (memoization didn't happen). It could just be a coder error, though.

## How I tested it
- I added logs for the selector call. Most of the time, the selector doesn't re-calculate and just returns past value.  However, it does re-calculate occasionally, and I believe it is due to the preference sync invalidating the entire store.
    - Screenshot from 2021-09-29 17-03-04
    - Screenshot from 2021-09-29 17-03-17
- Result: it doesn't re-calculate on each update if there's no change.  The one instance shown is an exception -- from a tab-focus preference sync.
    - Screenshot from 2021-09-29 17-04-08

### DIFF
--- a/ui/redux/selectors/blocked.js
+++ b/ui/redux/selectors/blocked.js
@@ -1,5 +1,6 @@
 // @flow
 import { createSelector } from 'reselect';
+import { splitBySeparator } from 'lbry-redux';
 
 const selectState = (state: { blocked: BlocklistState }) => state.blocked || {};
 
@@ -11,3 +12,15 @@ export const makeSelectChannelIsMuted = (uri: string) =>
   createSelector(selectMutedChannels, (state: Array<string>) => {
     return state.includes(uri);
   });
+
+export const selectMutedAndBlockedChannelIds = createSelector(
+  selectState,
+  (state) => state.comments,
+  (state, commentsState) => {
+    const mutedUris = state.blockedChannels;
+    const blockedUris = commentsState.moderationBlockList;
+    return Array.from(
+      new Set((mutedUris || []).concat(blockedUris || []).map((uri) => splitBySeparator(uri)[1]))
+    ).sort();
+  }
+);
--- a/ui/component/claimTilesDiscover/index.js
+++ b/ui/component/claimTilesDiscover/index.js
@@ -6,14 +6,12 @@ import {
   selectFetchingClaimSearchByQuery,
   SETTINGS,
   selectClaimsByUri,
-  splitBySeparator,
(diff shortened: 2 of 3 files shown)


ui/redux/selectors/blocked.js
```diff
@@ -11,3 +12,17 @@ export const makeSelectChannelIsMuted = (uri: string) =>
   createSelector(selectMutedChannels, (state: Array<string>) => {
     return state.includes(uri);
   });
+
+const selectCommentsState = (state) => state.comments || {};
```
jessopb, line 16: you could just inline:
`state => state.comments,`
But I think it's fine.

jessopb (approved):

jessopb:
Looks fine.
What's the necessity of the sort()?
There was the possibility of storing the Set rather than Array in redux store.

infinite-persistence:
```    
    const mutedUris = state.blockedChannels;
    const blockedUris = commentsState.moderationBlockList;
    return Array.from(
      new Set((mutedUris || []).concat(blockedUris || []).map((uri) => splitBySeparator(uri)[1]))
    ).sort();
```
The need for the `sort` is super weird.  I stepped through the code to confirm the order of `blockedUris` before running the `return` statement (I had 2 blocked entries).  The return statement occasionally changes the order of the final 2 uris, so the GUI ended up blinking because the "array changed". 

I've even tried changing the return statement to a simple for-loop and pushing it one by one, and still the order changed -- doesn't make sense since it should follow the index.  I temporarily gave up, sorted it, and memoized to negate the wasted cpu cycles for sorting.

infinite-persistence:
### 5247b59 to a8149fe
- Switched to inline style.  Thanks for the tip!

jessopb:
Ah, right. It's because sets do not guarantee order.

PR #7226, opened 2021-10-04
fix notifications page on unauthed app

jessopb:
Notifications page was broken on app if unauthed


infinite-persistence (approved):
The call runs without errors even if `isAuthenticated` is false. Was the broken part just "it doesn't make sense to call it"? (just for future reference when I'm touching settings again)

PR #7227, opened 2021-10-05
Comments: fix memoization of selectors

infinite-persistence:
Jessop, this serves as a one-time review for all upcoming PRs that are similar, so I don't need to bug you every time.

Pretty much what we've discussed in Slack:  general idea is to ensure we don't re-calculate heavy items and return new references:
1. Ensure selectors uses stashed value from `createSelector`.
2. For selectors where the input differs between instances (e.g. `props.uri` passed in), use the suggested factory method to create a unique selector per instance.
3. At the React side, we can remove any deep-compare work done in `React.memo` or `shouldComponentUpdate` to skip renders, since the default shallow compare would suffice.

----

_See commit messages, especially the last one, regarding a quick solution to avoid repetition in `CommentReactions`._ 

For this specific example, the `Comment` component renders 4-5 times because it thinks the `othersReacts` object is different, when it actually only updated once. This is just 1 comment alone.

After the changes, the number of renders is cut down to 2 -- one for initial render (can't avoid), and the next one when reactions are fetched:

### DIFF
--- a/ui/component/comment/index.js
+++ b/ui/component/comment/index.js
@@ -19,24 +19,30 @@ import { selectActiveChannelClaim } from 'redux/selectors/app';
 import { selectPlayingUri } from 'redux/selectors/content';
 import Comment from './view';
 
-const select = (state, props) => {
-  const activeChannelClaim = selectActiveChannelClaim(state);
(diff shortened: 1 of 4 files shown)


jessopb (approved):
I think this looks good

PR #7228, opened 2021-10-05
Skip muted list update if no change

infinite-persistence:
`selectMutedChannels` run and return a new reference each time the app gains focus, causing components like `ClaimPreview` and several others to render unnecessarily.

`createSelector` will not help in this case, because we are indeed invalidating the data in the store.

Not sure if it's a valid pattern to return existing state, but it helps in not invalidating cache.


jessopb (approved):
I think it's ok, given the way sync populate works 
How big of a hit is filtering the blocklist every doPopulate?

infinite-persistence:
Will combine this in the Ody-repo PR instead.  Desktop can use it if interested.

PR #7387, opened 2022-01-03
Unused environment vars/secrets

danger89:
I think the following 4 environment variables/secrets aren't used at all during the build.

Am I right?

Regards,
Melroy vd Berg


jessopb:
They are used to upload the build to aws.

jessopb:
Sorry, you're right.

danger89:
Indeed.. AWS is another step.

PR #7418, opened 2022-01-17
Fix shared prefs all values

jessopb:
Turns out this was a quick fix.


jessopb:
This isn't right yet. ...

Issue #3139, opened 2019-10-29
Ability to search through publishes


kauffj:
It should be possible to filter your own publishes via a search box, similar to the way that one can search their library or within a channel.

These PRs should the addition of this behavior for the Library area and may be a useful reference:
https://github.com/lbryio/lbry-desktop/pull/2969
https://github.com/lbryio/lbry-redux/pull/214

jessopb:
![image](https://user-images.githubusercontent.com/36554050/160070751-1563f9ea-ee8a-4179-9844-a892bc92829e.png)
Probably lose the upload button, and add the search like on the collectoins page:

jessopb:
![image](https://user-images.githubusercontent.com/36554050/160070889-b0edc1df-06b8-4fdf-9300-3eb1680d43c5.png)

PR #7535, opened 2022-04-06
Add ability to search through publishes.

Ruk33:
## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/3139

Users can't search through their publishes.

## What is the new behavior?

Users can search through their publishes:

image

image

image

## PR Checklist

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

</details>


ui/page/fileListPublished/view.jsx
```diff
@@ -42,12 +58,58 @@ function FileListPublished(props: Props) {
...
-    if (paramsString && fetchClaimListMine) {
-      const params = JSON.parse(paramsString);
-      fetchClaimListMine(params.page, params.page_size, true, filterBy.split(','));
+    if (!disableHistory) {
+      const params = new URLSearchParams(search);
+      params.set(PAGINATE_PARAM, '1');
+      history.push('?' + params.toString());
     }
-  }, [paramsString, filterBy, fetchClaimListMine]);
+  }, [filteredClaims]);
```
Ruk33, line 108: @jessopb I'm not entirely happy with this but I couldn't think of a better way to work around this issue.

jessopb: Seems to work really well. I don't think you need "disableHistory" - that mostly doesn't exist except if it's passed to <Paginate as a prop, and only on one page.

jessopb:
I think this works well. 2 things, 
1) During loading, let's not say "No results".
2) It would be good to debounce the filtering while typing?
3) Maybe make Reposts searchable or disabled?

jessopb:
I think we should put the search term in the params too, so when you navigate forward, then back again, you have both page and search term.

ui/page/fileListPublished/view.jsx
Ruk33: Updated. Thanks @jessopb for the comment.

PR #7563, opened 2022-04-26
Revert "Updated important instances of LBRY.tv to LBRY.com "

jessopb:
Reverts lbryio/lbry-desktop#7558

I thought this might be ok, but it wasn't.


kcseb:
So, do we not really have a way of getting rid of all the irrelevant LBRY.tv text without breaking things?

PR #7594, opened 2022-06-01
Recsys: determine the clicked index based on raw results instead of rendered order

### LINKED ISSUE #1606: Fix Linux upgrade process
## The Issue
Since we started building the LBRY app for 32 and 64 bit, the extension on the file downloaded by the linux upgrade process is different: 
What the LBRY app tries to open: /tmp/ar1h20/LBRY_0.21.6_amd64.deb
The file downloaded is LBRY_0.21.6.deb

Instead of launching the deb automatically which never installs correctly because the Ubuntu software center is awful, we should give the user a button which opens the temp folder to the deb file, so they can run it themselves if it works, or run dpkg -i in terminal to install. Give this hint on the upgrade screen too. 

## System Configuration

- LBRY Daemon version:
- LBRY App version:
- LBRY Installation ID:
- Operating system:

infinite-persistence:
## Ticket
Closes #1606

## Issue
The displayed order of the recs are often re-adjusted -- viewed content are pushed to the bottom to avoid auto-play-next from picking it up again. This was causing recsys to be confused, because the click order does not correspond the results.

## Change
Refactored the selector we can grab the raw results and use that as the index reference.


infinite-persistence:
Wrong repo

PR #7634, opened 2022-07-01
separate out advanced textarea, fix comment channel selector width, a…

jessopb:
separate out advanced textarea
fix comment channel selector width
add advanced text / simple text icons


jessopb:
channel edit description is broken...

PR #7650, opened 2022-07-14
add wallet sync export feature

jessopb:


jessopb:
screenshots accidentally in issue instead of pr

Issue #7365, opened 2021-12-20
Swap comment servers without going to settings page


jessopb:
Provide settings to add preferred comment servers to a list.
Allow selecting them.
Requested by @kodxana

jessopb:
![image](https://user-images.githubusercontent.com/36554050/177879602-36e7bc15-fb13-4936-9393-58c0871a7f57.png)
Place a dropdown similar to the "Comment as" Next to the refresh button *if* they have added other comment (under Comments Server in Settings)

Also a checkbox setting to show the comment server selector. (Follow "Viewed Hosting" setting style)

jessopb:
@ByronEricPerez Still working on this?

ByronEricPerez:
Hi @jessopb yes, i'm still working on this

kodxana:
Just letting know that my comment server is back online :)
https://comments.madiator.com/api/v2

jessopb:
@ByronEricPerez How is this going? Do you have any questions?

ByronEricPerez:
Hi @jessopb , this one was a bit more difficult than the others but it's going well. I just need to synchronize the input with the setting and that's it.

![image](https://user-images.githubusercontent.com/89370060/183733772-7eb07cb0-404b-48b7-8992-8e07d90fa475.png)

PR #7670, opened 2022-08-09
Swap comment servers without going to settings page #7365

ByronEricPerez:
## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/7365

## PR Checklist

- [ ] Bugfix
- [x] Feature
- [ ] Documentation changes

</details>

### DIFF
--- a/ui/component/commentsList/view.jsx
+++ b/ui/component/commentsList/view.jsx
@@ -1,6 +1,6 @@
 // @flow
 import { COMMENT_PAGE_SIZE_TOP_LEVEL, SORT_BY } from 'constants/comment';
-import { ENABLE_COMMENT_REACTIONS } from 'config';
+import { ENABLE_COMMENT_REACTIONS, COMMENT_SERVER_API, COMMENT_SERVER_NAME } from 'config';
 import { useIsMobile, useIsMediumScreen } from 'effects/use-screensize';
 import { getCommentsListTitle } from 'util/comments';
 import * as ICONS from 'constants/icons';
@@ -15,6 +15,8 @@ import Empty from 'component/common/empty';
 import React, { useEffect } from 'react';
 import Spinner from 'component/spinner';
 import usePersistedState from 'effects/use-persisted-state';
+import { FormField } from 'component/common/form';
+import Comments from 'comments';
 
 const DEBOUNCE_SCROLL_HANDLER_MS = 200;
 
@@ -52,6 +54,9 @@ type Props = {
   fetchReacts: (commentIds: Array<string>) => Promise<any>,
   resetComments: (claimId: string) => void,
   doResolveUris: (uris: Array<string>, returnCachedClaims: boolean) => void,
+  customCommentServers: Array<CommentServerDetails>,
+  setCommentServer: (string) => void,
+  commentServer: string,
 };
 
 export default function CommentList(props: Props) {
@@ -80,10 +85,16 @@ export default function CommentList(props: Props) {
     fetchReacts,
     resetComments,
     doResolveUris,
+    customCommentServers,
+    setCommentServer,
+    commentServer,
   } = props;
 
   const isMobile = useIsMobile();
   const isMediumScreen = useIsMediumScreen();
+
+  const defaultServer = { name: COMMENT_SERVER_NAME, url: COMMENT_SERVER_API };
+  const allServers = [defaultServer, ...(customCommentServers || [])];
 
   const spinnerRef = React.useRef();
   const DEFAULT_SORT = ENABLE_COMMENT_REACTIONS ? SORT_BY.POPULARITY : SORT_BY.NEWEST;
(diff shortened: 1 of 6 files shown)


jessopb:
Good progress. Besides fixing styling, this should be at the top next to the refresh button. 

The refresh button reloads all comments. The server selector changes the server for all comments.

Below, the comment field adds one comment. 

Also, don't forget to make the selector only show up when the number of servers is > 1.

kodxana:
@ByronEricPerez What do you think if we could have option in app to integrate using multiple comments servers into single feed.
Aka at the same time user could see Odysee comments and also other comments server, or user can decide to limit to selected comments servers or just use single one.

ByronEricPerez:
Hi @jessopb how are you, I have already corrected the location, in addition to the fact that it only shows if there are 2 or more commentServer, I just need to add that it shows the comments. And finally some css.

Hi @kodxana , I think the idea is good, it would also be more practical to have the option of being able to show them all in one.


image
image

ByronEricPerez:
Hi @jessopb , I'm having trouble seeing the comments on all the videos. 
The btn moved it to the correct component, it already synchronizes, I need to align it a little more. What do you think?

image

Although I comment I get an error asking me to restart, but I have that comment in my history.
I also don't see comments on any video. This error started appearing yesterday.

image

image

ByronEricPerez:
Fixed why it didn't show comments  , the error was in index.js

image

ui/scss/component/_card.scss
```diff
@@ -274,6 +274,23 @@
   display: flex;
   flex-direction: column;
   overflow: hidden;
+
+  .select--slim {
+    //margin-bottom: var(--spacing-xxs);
+    display: flex;
+    //justify-content: center;
+    margin: 0px 0px 10px 10px;
```
jessopb, line 282: We should prefer spacing vars instead of px values wherever possible.

jessopb (changes requested):
Good progress!
I think we keep refresh button all the way to the right.
And the alignment issue you mentioned.

ui/scss/component/_card.scss
ByronEricPerez: Hi @jessopb , Alignment fixed. Create a new class for the element.

image

image

What do you think?

jessopb:
Put the refresh button last, and it's a bit too wide, I think.

ui/scss/component/_card.scss
jessopb: Ideally, it's only as wide as the longest server name, but also up to a limit.

jessopb:
@kodxana would you like to test this branch?

ui/scss/component/_card.scss
jessopb: I think merging comments sounds really interesting, but sounds like a heavy lift, given how it currently works.

ui/scss/component/_card.scss
```diff
@@ -642,3 +658,17 @@
     }
   }
 }
+
+.button_selectedServer {
+  display: inline;
+  float: right;
+  min-width: var(--side-nav-width--micro);
```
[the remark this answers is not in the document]
jessopb: Also, did you try a `text-overflow: ellipsis;`

jessopb (commented):
Almost there!

ui/scss/component/_card.scss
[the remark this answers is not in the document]
ByronEricPerez: Hi @jessopb ,  what do you think?

image

I limit it to 10 characters, how much do you think is optimal?

ui/scss/component/_card.scss
```diff
@@ -642,3 +658,17 @@
...
+  max-width: var(--side-nav-width);
+  margin-left: -10px;
```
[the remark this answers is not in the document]
jessopb: I think the second one, but it's good to check how other places in the app look so that it's consistent.

ui/component/commentsList/view.jsx
```diff
@@ -398,9 +398,14 @@ const CommentActionButtons = (actionButtonsProps: ActionButtonsProps) => {
             value={commentServer}
           >
             {allServers.map(function (server) {
+              let name = server.name;
+              if (name.length > 10) {
+                name = name.substring(0, 10);
+                name = name + '...';
+              }
```
jessopb, line 405: Interesting technique. Try using CSS overflow: and text-overflow: ellipsis

ui/scss/component/_card.scss
```diff
@@ -662,9 +662,8 @@
 .button_selectedServer {
   display: inline;
   float: right;
-  min-width: var(--side-nav-width--micro);
-  max-width: var(--side-nav-width);
-  margin-left: -10px;
+  min-width: var(--input-select-server-min-width);
```
jessopb, line 665: It's good to use variables when the variable is going to be reused multiple places, but this is pretty specific here. 
In this code, probably something like: max-width: 12rem - pick some rem number that is reasonable.
Then shrink down to "mobile" size and see if you need a 
`@media (max-width: $breakpoint-small) { ... }` to adjust.

jessopb (changes requested):

jessopb (commented):
I tested it out and it looks pretty good. Two more things. 
 - Rebase it on master
 - call refresh when it switches

jessopb (approved):

Issue #7646, opened 2022-07-13
In the cover upload form, trying to upload anything other than an image doesn't reset the form


Ruk33:
**Bug**
In the cover upload form, trying to upload anything other than 
an image doesn't reset the form, thus, the user is unable to
try with another upload.

**To Reproduce**
1. Go to the Edit channel view
2. Click on the Upload cover button
3. Select a nonimage file
4. Click upload
5. See the error, the upload field remains disabled

**Expected behavior**
The form should be reset so the user can upload a new file.

**Screenshots**
![image](https://user-images.githubusercontent.com/1719111/178767251-1d8d415c-45d3-457c-8f94-4304505c5362.png)

**System Configuration**

- LBRY Daemon version: 0.107.2
- LBRY App version: 0.53.4
- LBRY Installation ID: 5rtRpW1S82tCGZZaTvkVrY4L52xZVpsPv8QAVTKRz63TDFzD7y1BVBycrF85GLSANW
- Operating system: Windows (Windows-10-10.0.19041-SP0)

**Additional context**

PR #7672, opened 2022-08-11
Allow only images in modal image uploader.

Ruk33:
## Fixes

Issue Number: https://github.com/lbryio/lbry-desktop/issues/7646

Even though a file filter is being used, the "All Files"
option allows users to select any file when uploading 
an image.

## What is the new behavior?

The electron filter is being used so the "All Files" 
option is no longer displayed.

image

## PR Checklist

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

</details>

### DIFF
--- a/ui/component/common/file-selector.jsx
+++ b/ui/component/common/file-selector.jsx
@@ -83,7 +83,16 @@ class FileSelector extends React.PureComponent<Props> {
         if (!result) {
           return;
         }
-        const file = new File([result.buffer], result.name);
+        const file = new File([result.buffer], result.name, {
+          type: result.mime,
+        });
+        // "path" is a read only property so we have to use this
+        // hack to overcome the limitation.
+        // $FlowFixMe
+        Object.defineProperty(file, 'path', {
+          value: result.path,
+          writable: false,
+        });
         this.props.onFileChosen(file);
       });
   };
--- a/ui/component/common/file-selector.jsx
+++ b/ui/component/common/file-selector.jsx
@@ -77,7 +77,6 @@ class FileSelector extends React.PureComponent<Props> {
         if (path) {
           return ipcRenderer.invoke('get-file-from-path', path);
         }
-        return undefined;
       })
       .then((result) => {
         if (!result) {
--- a/ui/component/common/file-selector.jsx
+++ b/ui/component/common/file-selector.jsx
@@ -8,7 +8,7 @@ import { FormField } from 'component/common/form';
 type Props = {
   type: string,
   currentPath?: ?string,
-  onFileChosen: (WebFile) => void,
+  onFileChosen: (FileWithPath) => void,
   label?: string,
   placeholder?: string,
   accept?: string,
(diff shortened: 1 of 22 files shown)


ui/component/common/file-selector.jsx
```diff
@@ -64,13 +66,26 @@ class FileSelector extends React.PureComponent<Props> {
...
+          return ipcRenderer.invoke('get-file-from-path', path);
+        }
+        return undefined;
+      })
+      .then((result) => {
+        if (!result) {
+          return;
+        }
+        const file = new File([result.buffer], result.name);
+        this.props.onFileChosen(file);
```
jessopb, line 87: I'm concerned that changing this from { path } to file will break publishing. Did you try?
https://github.com/lbryio/lbry-desktop/blob/b9be8d9f3a80d06c3e620ad12fc692f907c46223/ui/component/publishFile/view.jsx#L273 expects file.path on electron because the sdk needs the path.

It may be good to send { file: file , path: path } and adjust all the functions assigned to onFileChosen to take that.

Ruk33: Sounds good @jessopb, nice catch. Perhaps as a small note for the future, we should avoid using these union types. Notice how I'm still respecting the interface (by sending a `File` type) but still, it manages to break parts of the app...

jessopb: You're probably right. The union types were originally added when the codebase was supporting both web and app. 

Maybe the type should be ChosenFile = { file: WebFile, path: string }

Ruk33: @jessopb I have updated the code so the `File` now contains a `path` and a `mime` type. With this, the file upload for posts works fine. Please let me know what you think. Thanks.

ui/component/fileDrop/view.jsx
```diff
@@ -11,10 +11,10 @@ import Icon from 'component/common/icon';
 
 type Props = {
   modal: { id: string, modalProps: {} },
-  filePath: string | WebFile,
+  filePath: File,
```
jessopb, line 14: Also breaks here on file drop. This should be a path.

jessopb (changes requested):
Sorry it took a while to get to this. Tested publish and it's broken.

jessopb (approved):

ui/modal/modalImageUpload/view.jsx
```diff
@@ -14,10 +14,13 @@ type Props = {
 
 function ModalImageUpload(props: Props) {
   const { closeModal, currentValue, title, assetName, helpText, onUpdate } = props;
+  const filters = React.useMemo(() => [{ name: 'Images', extensions: ['jpg', 'png', 'gif'] }]);
```
jessopb, line 17: this blocks `jpeg` - possibly some other valid types. svg has been possible and allowed so far.

jessopb (changes requested):
One small snag.

PR #7677, opened 2022-08-22
Fix viewer position and shape in mobile mode

jm-morani:
## Fixes

Issue Number:

When the screen is narrow enough, the video is not render at the same location than the cover.
Because it's rather on the top of the window, the top of the video is hidden by the header.

## What is the new behavior?
The video is played at the exact same location than the cover, and is fully visible.
As well, because that area has then no margin with the border of the window, I removed the rounded border (seems to already be a pattern).

## Other information

This is a very minor change.

(On narrow window, it may look even better with an overlay scrollbar.)
Anyway, it may not be the best fix, but it solves my problem. Thank you for making lbry.

## PR Checklist

- [X] Bugfix
- [ ] Feature
- [ ] Documentation changes

</details>

### DIFF
--- a/ui/component/fileRenderFloating/view.jsx
+++ b/ui/component/fileRenderFloating/view.jsx
@@ -21,7 +21,6 @@ import AutoplayCountdown from 'component/autoplayCountdown';
 // scss/init/vars.scss
 // --header-height
 const HEADER_HEIGHT = 60;
-const HEADER_HEIGHT_MOBILE = 60;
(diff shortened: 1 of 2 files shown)


jm-morani:
I wrote "mobile" because that what is in the code. But, I actually use lbry on a laptop running debian bullseye and the i3 window manager.

jessopb (approved):
Nice

Issue #7687, opened 2022-09-10
53.5 upload form bugs


keikari:
* [ ] "name" doesn't auto-fill after selecting a file to upload (53.4 auto-fills with the file name)
* [ ] Need to select file twice for file path to show on the "file" field
* [ ] https://github.com/lbryio/lbry-desktop/issues/7685 (got filed before, just mentioning here)

Ruk33:
Hello @keikari I wasn't able to reproduce the second issue. Did you try the latest changes in `master`?

keikari:
Oh, actually all of the above seems to work in master like expected. Guess this can be closed.
(Only checked 53.5 at the  time)

Issue #7685, opened 2022-09-07
Transcode on upload broke on 53.5


keikari:
Reported in reddit
https://www.reddit.com/r/lbry/comments/x7eylc/why_is_my_optimize_and_transcode_video_box/

Seeing same here on Linux. The checkbox to optimize and transcode video stays grayed out.
Fine on 53.4

PR #7688, opened 2022-09-13
Better handling of uploaded files.

Ruk33:
## Fixes

Issue Number:
- https://github.com/lbryio/lbry-desktop/issues/7687
- https://github.com/lbryio/lbry-desktop/issues/7685

The behavior when selecting a file is different
than the behavior when it's dragged and dropped.

## What is the new behavior?

The behavior when selecting a file and 
dropped no longer differs.

## PR Checklist

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

</details>

### DIFF
--- a/ui/component/fileDrop/view.jsx
+++ b/ui/component/fileDrop/view.jsx
@@ -80,9 +80,6 @@ function FileDrop(props: Props) {
           navigateToPublish();
           updatePublishForm({
             filePath: selectedFile.path || selectedFile.name,
-            fileDur: 0,
-            fileSize: 0,
-            fileVid: false,
           });
         }, NAVIGATE_TIME_OUT);
       }, HIDE_TIME_OUT);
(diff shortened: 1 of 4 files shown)


ui/component/fileDrop/view.jsx
```diff
@@ -80,9 +80,6 @@ function FileDrop(props: Props) {
           navigateToPublish();
           updatePublishForm({
             filePath: selectedFile.path || selectedFile.name,
-            fileDur: 0,
-            fileSize: 0,
-            fileVid: false,
```
jessopb, line 85: The purpose of these lines was to do the math to guess Mbps and judge whether file needs to be transcoded (is a video, and size is large compared to duration). Is that done somewhere else now?

jessopb (commented):
I tried this out - after selecting a transcoded video, it still says "your video may not be the best format.

ui/component/fileDrop/view.jsx
Ruk33: @jessopb yes, it was removed so this logic happens only in one place (publishFile/view.jsx)

jessopb: Ok. The messaging seems wrong. `Your video may not be the best format.` is always the message which is the message for (isVid && !duration) so something isn't working.

jessopb (approved):

PR #7709, opened 2022-10-25
Minor layout fixes depending on window size

jm-morani:
## Disable theatre mode styles for medium screen (or lower)
The theatre mode icon only shows up when window is greater than 1150px ($breakpoint-medium).
Users had to increase the window width to toggle it, then shrink the window back.
Now, the use of theatre mode styles is consistent with the display of the icon.
Moreover, those styles didn't compound well with small screen (900px or less).

## Fix the width of the cover in theatre mode
The CSS "100vw" is the width of the window not reduced of the scroll bar.
Therefor, the right-hand side of the cover was hidden under the scroll bar.
As well, the slightly bigger width used to trigger a slightly bigger height.
This height is used as a the reference for the player's height.
So the extra height could lead to black margins on top and bottom of the video.
(Reproducible in theatre mode, when window width slightly more than 1150px and height big enough)
Now, I removed it, so it rather inherits "100%".

## Fix oddities when screen width is exactly 1150px
According to useIsMediumScreen(), the screen is medium when width is up to 1150px.
This media query is equivalent `(max-width: $breakpoint-medium)`.
But, `(min-width: $breakpoint-medium)` is a wrong opposite as it includes 1150px.
When the width is exactly at the limit, medium and large styles overlap.
(Eg. Check the width of covers in the list of the main page, when window width is 1150px)
Same problem for other $breakpoint, but I haven't fixed those.

## Question: Is it possible to lower the minWidth of the window?
I regularly use the application at width 670px (thank to a tiling window manager).
(Including this PR), the small screen mode seems fine. So, why not to make it available to all users?
(It may even possible to shrink further, but I haven't test enough to recommend that).

## PR Checklist

- [x] Bugfix
- [ ] Feature
- [ ] Documentation changes

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
  - Show downloads newest first ([#7684](https://github.com/lbryio/lbry-desktop/pull/7684))
  - Only allow images in image uploader ([#7672](https://github.com/lbryio/lbry-desktop/pull/7672))
  - Fixed bug with csv exports ([#7697](https://github.com/lbryio/lbry-desktop/pull/7697))
- - Fixed small screen viewer position ([#7677](https://github.com/lbryio/lbry-desktop/pull/7677))
  - Fixed various upload bugs including transcoding ([#7688](https://github.com/lbryio/lbry-desktop/pull/7688))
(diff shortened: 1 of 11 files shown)


jessopb:
Thanks for this - can we show you some [appreciation](https://lbry.com/faq/appreciation)?

jessopb:
Thanks for this @jm-morani , can we show you some [appreciation](https://lbry.com/faq/appreciation)?

